PR #28, opened 2020-05-20
Update strings.xml

Konradrundfunk:
Hi I have seen the german transation and some strings where missleading and not common in the language.


geeksville:
thank you @Konradrundfunk !

PR #40, opened 2020-06-11
Add tooltip to status image icon

lgoix:


app/src/main/java/com/geeksville/mesh/MainActivity.kt
```diff
@@ -406,6 +400,26 @@ class MainActivity : AppCompatActivity(), Logging,
...
+                R.drawable.cloud_off,
+                R.string.disconnected
+            )
+            else -> Pair(R.drawable.cloud_off, R.string.disconnected)
+        }
+
+        connectStatusImage.setImageResource(image)
+        TooltipCompat.setTooltipText(connectStatusImage, getString(tooltip))
+    }
+
```
geeksville, line 421: Your code is super clean and nice.  I especially like how you hoisted the two bits of state out of the when.  A very 'kotlin' approach!

app/src/main/res/values/strings.xml
```diff
@@ -55,6 +55,7 @@
     <string name="connected_count">Connected: %s of %s online</string>
     <string name="list_of_nodes">A list of nodes in the network</string>
     <string name="update_firmware">Update Firmware</string>
+    <string name="connected">Connected to radio (%s)</string>
```
geeksville: This whole pull request is great (and I'll reply to your kind email tomorrow - currently drinking beer and watching a movie ;-) ).  Can you make one change though?  change this resource name back to connected_to, because I don't want to lose the various language translations others have donated in values_XX.

lgoix: My bad, I've forgot to remove (%s) from this new label. no changes for **connected_to**. Did not see it when I test it because the device was not connected.

I will fix it and add missing french translation

app/src/main/java/com/geeksville/mesh/MainActivity.kt
lgoix: Thanks

I tried something easy to get familiar with the project and github process. In fact, I did not like kotlin when I followed a tutorial for the Jetpack menu navigation. This project is perfect to change my mind I hope !

app/src/main/res/values/strings.xml
geeksville: heh!  last night when I was sleeping I briefly woke and realized oh!  it probably was a new string and the %s was probably just an oops.  No worries thanks.  And OMG a pull request for french would be awesome.

PR #42, opened 2020-06-12
missing translation in french

lgoix:


geeksville:
Looks great @lgoix.  Is it okay if I send you an invite to join this org?  It would allow you to review my commits (if you have time and interest) and allow either of us to approve pushes into the master tree.

lgoix:
Sure

I saw your post on the forum about what we discussed , I will read in
details later this pm after work. I guess we are not on the same timezone.

Ludovic


Le ven. 12 juin 2020 12:18, Kevin Hester <notifications@github.com> a
écrit :

Issue #64, opened 2020-06-24
Display battery level of remote nodes


lgoix:
The node page will display the battery level

The work in progress: 
![image](https://user-images.githubusercontent.com/16083263/85608998-a06f9800-b623-11ea-8ae0-bbacca5035f4.png)

PR #75, opened 2020-07-01
Display battery level of remote nodes #64

lgoix:


geeksville:
ooh looks awesome to me!

PR #85, opened 2020-07-03
added option to clean message history

chepa92:
nice feature that I needed, its my first commit :)


geeksville:
@chepa92 yeah - the service 'owns' the message DB and all persistent state (which I think is a pretty good abstraction).  If you want to make a 'delete old messages' feature (a good idea) I'd recommend:

* edit IMeshService.aidl and add 'void deleteOldMessages()'
* build, compiler will now know enough to complain about missing method in the MeshService.kt stub.  Right click to implement that method.  In that method you can clear 'recentDataPackets'
* In the gui, call that method and then either also clear the gui livedata or (cleaner I think) reread service.getOldMessages().

PR #93, opened 2020-07-07
fix a couple of subtle autobugs (see commits for details) and we were leaking macaddrs (PII) into logs

geeksville:


geeksville:

https://meshtastic.discourse.group/t/dear-person-with-a-redmi-5-plus-phone-who-clicked-report-bug/657/1

Issue #99, opened 2020-07-15
battery level 0 in protobuf should render as "battery level unknown"


geeksville:
more info here: https://meshtastic.discourse.group/t/alpha-tester-thread-no-new-tests-needed-at-the-moment/298/189?u=geeksville

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/alpha-tester-thread-no-new-tests-needed-at-the-moment/298/189

lgoix:
A charging icon could be displayed

PR #102, opened 2020-07-16
Display power plug icon when no battery #99

lgoix:


geeksville:
awesome!  thanks for all you do @lgoix.

PR #103, opened 2020-07-16
Set crash report text background

zjwhitehead:
(Disclaimer: still very new to Android dev)
On Pixel 3XL when resizing to do split window the text for the crash report checkbox overlaps the text behind it. This just adds a background to that text box


geeksville:
thanks!  a good improvement but just fyi (to help you learn android - bravo to you btw): I think a better fix would be to somehow change the constraints in the contraintlayout so that as the window gets smaller that warning text (you haven't yet paired...) gets assigned less vertical height.  I don't know exactly how to do this ;-) but I bet that's the solution.

Issue #84, opened 2020-07-03
Connection status is the opposite of the status in settings fragment 


lgoix:
There are some instabilities when the activity pause/resume that are caused by observer

fix some issues the user are having like:

![image](https://user-images.githubusercontent.com/16083263/86469827-35e7e780-bd08-11ea-91ae-7b4ff0c76377.png)
The icon display disconnect when it is not

**Scope of Activity vs ViewModel:**

![image](https://user-images.githubusercontent.com/16083263/86469993-80696400-bd08-11ea-8bc2-4ea0eed9a156.png)

geeksville:
oh - smart!

lgoix:
I was wrong, I am trying to fix this issue but the title is wrong, this is not an issue

PR #105, opened 2020-07-17
Fix bad connection status #84 

lgoix:
When the activity recreates while connected to the radio, the menu was not inflated, default Disconnect value was displayed


geeksville:
yay thanks!

PR #109, opened 2020-07-21
Update strings.xml

eriktheV-king:
Just modified french language ressource files with some more added strings !
Left 5 untranslatables at EOF.
Trying to send these in as a pull request (replacing the existing files)
eriktheV-king github username


geeksville:
Awesome!  Thanks for the PR

PR #112, opened 2020-07-24
Fix time issue and Toast not showing when notification are disabled

lgoix:


app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -942,22 +943,22 @@ class MeshService : Service(), Logging {
         val p = packet.decoded
 
         // If the rxTime was not set by the device (because device software was old), guess at a time
-        val rxTime = if (packet.rxTime == 0) packet.rxTime else currentSecond()
+        val rxTime = if (packet.rxTime != 0) packet.rxTime else currentSecond()
```
lgoix, line 946: @geeksville : I think this was not good if rxTime was 0 if not present in the packet

geeksville: @lgoix hmm - I don't quite understand.  if rxTime is not specified by the device, why would we not want to pick the local phone time?

app/src/main/java/com/geeksville/mesh/MainActivity.kt
```diff
@@ -869,7 +869,7 @@ class MainActivity : AppCompatActivity(), Logging,
         try {
             val packageInfo: PackageInfo = packageManager.getPackageInfo(packageName, 0)
             val versionName = packageInfo.versionName
-            Toast.makeText(this, versionName, Toast.LENGTH_LONG).show()
+            Toast.makeText(applicationContext, versionName, Toast.LENGTH_LONG).show()
```
geeksville, line 872: smart!

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
lgoix: Yes that is the fix

It was: if rxtime is set then use the alocal phone time

app/src/main/java/com/geeksville/mesh/MainActivity.kt
lgoix: rxTime=0 when packet.rxTime==0

lgoix: The code was:
```
val rxTime = if (packet.rxTime == 0) packet.rxTime else currentSecond()
```

If device has rxTime, we were using currentSecond()

```
if (packet.rxTime == 0)
   val rxTime = 0
else
   val rxTime = currentSecond()
```

geeksville: @lgoix oh I see!  I'm a moron and was not thinking carefully!  Thank you for fixing this.  Sorry for being silly!

PR #127, opened 2020-08-11
Put channel options into URL fragments

### LINKED ISSUE #126: QR sharing URL format potentially leaks channel keys to remote server
Right now the URL format for the QR-shared channel details looks like this: `https://www.meshtastic.org/c/CBciENTxuzogKQdZ8Lz_q89Oab8qB0RlZmF1bHQ=`

Which encodes the channel name, parameters, and key into the URL path. When shared, there's a non-zero chance that this URL gets fetched. Someone taps the link (and admittedly gets a 404) but the URL _path_ is being sent to `meshtastic.org`. Anyone with access to the hosting logs can potentially see the `GET /c/CBciENTxuzogKQdZ8Lz_q89Oab8qB0RlZmF1bHQ=` request and take the channel configuration including the key.

I believe there's a fairly easy solution here: placing the channel configuration in the URL fragment instead of the path. The URL fragment identifier is _not_ sent to the server when browsing to the URL. So even if the URL is clicked/tapped/cached by a browser fetch the channel options never get sent to a remote machine.

gkelly:
Put the channel options into the URL fragment instead of the URL path.
This ensures that the channel options (including the channel key) is
never accidentally leaked to a remote party. This URL format appears to
be backwards compatible, URLs generated in the new form are properly
parsed by the old version. URLs generated by the old version are _not_
parsed by the new version.

Closes #126.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/model/Channel.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/Channel.kt
@@ -32,7 +32,7 @@ data class Channel(
                 .setModemConfig(MeshProtos.ChannelSettings.ModemConfig.Bw125Cr45Sf128).build()
         )
 
-        const val prefix = "https://www.meshtastic.org/c/"
+        const val prefix = "https://www.meshtastic.org/c/#"
 
         private const val base64Flags = Base64.URL_SAFE + Base64.NO_WRAP
 


geeksville:
I guess we could make old urls  be properly parsed by the new version, but since the app is still young I think the hassle is not worth it.

geeksville:
cc @lgoix because clever problem and clever fix.

PR #163, opened 2020-09-17
Simplified calls to system services

KenVanHoeylandt:
Rather than calling the Android interfaces, I made some read-only extension properties to easily (and consistently) call the system services.
I also fixed some code that assumed BluetoothManager to never be NULL.

The consuming code of the `UsbManager` assumes that the `UsbManager` is always there, so I enforced that too.


geeksville:
Looks great!

PR #174, opened 2020-09-27
Always keep first character of long name when making short name

timgunter:
This should ensure that the first character of the long name is retained when forming short names even if it is a vowel.


geeksville:
ooh thanks!

PR #198, opened 2020-10-21
Add Exit button in dropdown menu

### LINKED ISSUE #186: Add stop app button/reboot start option
Once the app starts, it will stay on until the phone is rebooted, then it stays off.

If this is for securing that messages are receieced, it should restart at reboot, so you don't miss any messages. 

Also, a stop button that cleanly exits the app so when things go wonky you don't have to go through android/settings/app/force termination.

hestela:
This mostly fixes #186 .
After the app "closes", a few seconds later the bluetooth notification goes away.
User still needs to close app from open apps but no longer needs to force close
the app from the system settings in order to get rid of the bluetooth
notification.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/MainActivity.kt
+++ b/app/src/main/java/com/geeksville/mesh/MainActivity.kt
@@ -60,6 +60,7 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancel
 import java.nio.charset.Charset
+import kotlin.system.exitProcess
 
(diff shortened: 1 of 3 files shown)


geeksville:
Hi @hestela thank you for the very kind pull request.  I'm kinda disinclined to add this feature though.  Can you take a look at the comments I just added on bug #186 and add your thoughts there?  (Sorry for the late reply - I've been moving and only just now have a desk ;-) )

geeksville:
ok closing for now - because app seems pretty stable and android doesn't need exit buttons on apps (usually)

PR #220, opened 2021-01-26
Material design for Message UI

vfurman-gh:
This is how it looks: 
Screenshot_20210125-173539

Mostly UI changes to the layout, but there are also code changes:
* Update myNode LiveData when messages are loaded from the storage, otherwise the UI will not classify messages correctly until BT connection is established
* Manually set colors/margins in the code


geeksville:
looks great @vfurman-gh!  I'll push out an alpha later this week with your changes!

PR #223, opened 2021-01-31
Fix background color

vfurman-gh:
Semi-transparent background color in combination with rounded corners creates a dark border on Android >= 9.0, probably due to a way hardware renders it.


geeksville:
thanks for the quick fix!

PR #225, opened 2021-02-01
Moved try to catch all RemoteExceptions inside connection coroutine

vfurman-gh:
When an uncaught RemoteException is thrown inside mainScope.handledLaunch, it is silently ignored, but leaves the mainScope in "cancelled" state and all subsequent coroutine calls using mainScope never actually execute the body of coroutine.
Proposed minimal fix is to enclose all remote invocations inside this particular coroutine with try/catch block. But perhaps we should think of a better way to handle exceptions, as handledLaunch is used in many other places.


geeksville:

https://meshtastic.discourse.group/t/new-android-alpha-release-1-1-37/2326/11

geeksville:
a great find - thanks @vfurman-gh.  And yeah - I'm still pretty new on kotlin coroutines, so if you have ideas for fixes on the root problem they are wecome!

PR #233, opened 2021-02-06
Added ability to edit broadcast_position and ls_sleep settings

vfurman-gh:
* I exposed broadcast_position and ls_sleep properties in the UIState model
* When  broadcast_position == 0, I also do location_share = LocDisabled, which feels correct, but I maybe wrong. Let me know.
* Adds corresponding views to the resources and bind them to the model.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -121,7 +121,7 @@
             android:name="com.geeksville.mesh.MainActivity"
             android:label="@string/app_name"
             android:screenOrientation="portrait"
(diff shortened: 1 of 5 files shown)


vfurman-gh:
Here is the [screenshot](https://drive.google.com/file/d/1-31BI8LyDGa2kDKQcm-SKz31wVxRpRSo/view?usp=sharing)

geeksville:
looks great!  note for the future: eventually we might want to move these fields to some other 'advanced settings' screen.  To discourage novice users from playing with them.

geeksville:
(also I really need to fix the CI build system)

PR #236, opened 2021-02-07
Map UI update fixes

vfurman-gh:
Fixes for issues discovered during testing map updates
* The isViewVisible workaround for mapbox crashes was not working as expected, as isViewVisible was false all the time (and not updating the map). Rather than fixing the expression for it, I tested on my devices without the workaround and the crash does not happen. I suggest we try to release without the workaround (this PR)
* Follow up to recent change to allow increasing frequency of position broadcast:  
** update gps wakeup frequency, otherwise we were sending the same position all the time.
** shorten period of owner updates according to broadcast frequency (I wish we had an independent period for owner updates..)
* Do not ignore position updates for our device
* Zoom map camera to user location only once, afterwards if user scrolls elsewhere, it becomes annoying.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/model/UIState.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/UIState.kt
@@ -21,6 +21,7 @@ import com.geeksville.mesh.database.entity.Packet
 import com.geeksville.mesh.service.MeshService
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
+import kotlin.math.max
 
 /// Given a human name, strip out the first letter of the first three words and return that as the initials for
 /// that user. If the original name is only one word, strip vowels from the original name and if the result is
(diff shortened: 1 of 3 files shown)


geeksville:
great fixes!  Though I think I'd like to keep the isViewVisible for a bit longer.  Though with my (fairly recent ;-)) breakage of the test polarity fixed ;-)

This nasty hack only manifests in very occasional crashes of the mapbox native code.  I only received them via analytics from a 1%ish of usage.  So hard to reproduce on a particular machine.  But with the (original - working) version of this test the problem was fixed.

Btw: do you want access to see the crashlytics reports (no pressure)?  If so, what email should I send the invite to?

vfurman-gh:
OK, i'll send you another PR with isViewVisible re-resurrected, did not realize it was 1% case. I actually got the app frozen on the map view after a day of testing, perhaps it is related.
My email is vfurman@gmail.com

geeksville:
re: another change
no need - but please review my recent checkin and see if you think it looks
okay?

PR #238, opened 2021-02-08
Add a separate notification channel for messages

vfurman-gh:
* Split notification channel into 'service' and 'messages' notifications. With a separate channel, we can reasonably enable sounds for incoming messages, otherwise it will be too annoying, as service notifications are very frequent.
* Added deduping for service notifications to reduce their frequency, but in the end it proved not sufficient, so went for the split channel option.
* Provided separate "notification" methods for messages and service, to make things look cleaner.
* Removed recentReceivedTextPacket member, seemed a bit hacky.


geeksville:

https://meshtastic.discourse.group/t/message-notifications-channel/2443/1

PR #239, opened 2021-02-08
misc fixes

geeksville:


vfurman-gh (commented):
LGTM
Is this sufficient to resolve the crash? (Previously it was also checking isResumed)
If so, maybe a better name (isDestroyed)?

geeksville:
@vfurman-gh yes I think so, isResumed I think was unneeded.  thanks!

PR #259, opened 2021-03-15
Display time ago and coords for users

Crotery:
The lastSeen time is buggy, it shows only time without day. It would be more informative, if android app show real age of packet.
Also, there is no way to know, where the station is, using only android phone - let's show coordinates of station on Users List, because map cannot work offline and even there you cannot get the station coordinates :(
image


randybb:
Please add url for these coordinates in _geo:latitude,longitude_ format - this will open these coordinates in mapping app (OruxMaps, Locus,...).
+ a calculated distance from your node would be helpful

Crotery:
Hi there,
distance is already shown near coords. The url is a good idea, agree!

Crotery:
dear devs,
Had to make changes to hwModelDeprecated due to compile problems. Seems like protobufs+master were broken.
Also, I could not setup testing data to show the coords, seems like protocol problem or so. Tested on real phone&1.2.10 firmware, works. There is a screenshot from emulator
image

Crotery:
here it is from phone
image

randybb:
I think distance and bearing between nodes is visible only on device itself, not in the app.

Crotery:
There is a code for distance and it's in the desingn...

But, seems it's broken in alpha, could not see it today while testing on latest master version. Hope it will be fixed, it is not about this PR
image

geeksville:

https://meshtastic.discourse.group/t/nodes-out-of-range-still-listed/2616/14

geeksville:
looks great!  And a good discussion in this bug.  Merging

geeksville:
> Had to make changes to hwModelDeprecated due to compile problems. Seems like protobufs+master were broken.

This I think is by intent, the submodule version in this project is locked to the version of the protobufs it was built with.  Updating to master for protobufs may pull in not yet ready for android changes.  I'll submit my devbranch with these changes.

> Also, I could not setup testing data to show the coords, seems like protocol problem or so. 

This isn't really documented but you can add this info to MockInterface as fake from device packets.  Hopefully this app will be more testable in the future.

Crotery:
> This isn't really documented but you can add this info to MockInterface as fake from device packets. Hopefully this app will be more testable in the future.

Yeah, there were some data (with position) and even I could change the name of node and it changed in emulator. But lat&long from the mockInterface somehow did not get to the UI, so I tested only on the phone.


Thanks for merging, waiting for the 1.2 release =)

luxonn:
That's great.  I have offline maps installed, it would be great to choose the application that you switch to after clicking coordinates or to be able to copy the coordinates.

PR #260, opened 2021-03-18
Save messages in CSV file

vfurman-gh:
* Add menu option to store  (From, SNR, Time, Distance) from the last 500 messages in a CSV file. We only use POSITION_APP messages for that.
* Fixed location request interval for GPS-less devices - align that with positionBroadcastSec in radio preferences and ensure that when interval changes (usually when preferences are downloaded from radio), the location request is properly restarted.


geeksville:

https://meshtastic.discourse.group/t/save-position-messages-in-csv/2900/1

geeksville:
looks great!

PR #264, opened 2021-03-22
Fix expired link

danielnemenyi:
The 'Firmware update required' modal contains an expired link to the
Wiki. This commit points the user towards the Firmware Installation
section on Github instead.

* app/src/main/res/values/strings.xml


geeksville:
thanks!  good idea!

PR #276, opened 2021-04-14
Two fixes

vfurman-gh:
* Removed checking for the device name to contain its own MAC address. We still verify that the device name starts with "Mesh".
* Added "utf8" var to DataPacket class. This is to make IMeshService.aidl (and its dependency DataPacket) compile outside of geeksville package (to be used as a remote service).


geeksville:
yay thanks for this!

PR #282, opened 2021-06-20
POC: Show nodes in Locus Maps. 

ray0711:
On each update of the nodeDb Broadcast the first 1k nodes so that a running LocusMaps instance can display them as Map Points. The current state is neither in the perfect place nor perfect but might be a good start.

Things to take into consideration
- Should we litter the MeshService with that or is there a better place where optional code should be located and still called?
- We call the method quite often, any idea of intelligent filtering? To keep this from becoming a battery hog.
- Cleanup on exit? We could send an empty points list
- Configuration? If Locus isn't installed it doesn't hurt anybody but is hust useless
- Alternative import methods (starting locus etc.)

Extra changes:  jcenter replaced with maven central since jcenter is deprecated

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -36,9 +36,17 @@ import com.google.protobuf.ByteString
 import com.google.protobuf.InvalidProtocolBufferException
 import kotlinx.coroutines.*
 import kotlinx.serialization.json.Json
+import locus.api.android.ActionDisplayPoints
+import locus.api.android.objects.PackPoints
+import locus.api.objects.extra.Location
+import locus.api.objects.geoData.Point
+import java.text.SimpleDateFormat
+
+import java.time.format.DateTimeFormatter
 import java.util.*
 import kotlin.math.absoluteValue
 import kotlin.math.max
+import kotlin.math.min
 
 /**
  * Handles all the communication with android apps.  Also keeps an internal model
@@ -131,6 +139,9 @@ class MeshService : Service(), Logging {
             connect()
         }
     }
+
+    private val isoLocalDateTimeFormatter: DateTimeFormatter =
+        DateTimeFormatter.ISO_LOCAL_DATE_TIME;
 
     private val locationCallback = MeshServiceLocationCallback(
         ::perhapsSendPosition,
@@ -1015,6 +1026,7 @@ class MeshService : Service(), Logging {
     /// If we just changed our nodedb, we might want to do somethings
     private fun onNodeDBChanged() {
         maybeUpdateServiceStatusNotification()
+        exportVisibleNodesToLocusMaps()
     }
 
     private fun setupLocationRequest() {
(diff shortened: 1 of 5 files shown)


geeksville:

https://meshtastic.discourse.group/t/offline-maps-on-meshtastic/137/34

geeksville:
super cool!  sorry I was away from github for a while.  but I'm back now and will look and reply tomorrow!

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -1880,6 +1892,47 @@ class MeshService : Service(), Logging {
...
+        val packPoints = PackPoints("Meshtastic Nodes")
+        packPoints.bitmap = serviceNotifications.getBitmapFromVectorDrawable(R.mipmap.ic_launcher2_round);
+
+        exportPoints.subList(
+            0, min(
+                exportPoints.size, 999
+            ) // more than 999 points  requires file based transport
+        ).forEach { packPoints.addPoint(it) }
+
+        val send = ActionDisplayPoints.sendPackSilent(this, packPoints, false)
```
menion, line 1909: [Suggestion]
In case you worry about 1MB limit for intent `Bundle` object, I really suggest to use system over `FileProvider`. Sample is already in [Locus API](https://github.com/asamm/locus-api/blob/a1f278b5669841124ccb825753b4029c780f199a/locus-api-android-sample/src/main/java/com/asamm/locus/api/sample/utils/SampleCalls.kt#L288) and it should work perfectly fine even with modern 30+ API level.

The main reason is that the current limit to 1000 points is hidden from the user. Missing data on the map without obvious reason is really bad UX.

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -1880,6 +1892,47 @@ class MeshService : Service(), Logging {
...
+            val loc = Location()
+            val point = Point(nodeInfo.user?.longName ?: nodeInfo.num.toString() , loc)
+            val lastHeardString =
+                SimpleDateFormat("yyyy-MM-dd HH:mm:ss").format(java.sql.Timestamp(nodeInfo.lastHeard.toLong() * 1000))
+
+            point.parameterDescription =
+                "User: ${nodeInfo.user?.longName}\nLast heard: ${lastHeardString}\n" +
+                        "Battery: ${nodeInfo.batteryPctLevel}\n:Online ${nodeInfo.isOnline}\n"
+
+            loc.latitude = nodeInfo.position?.latitude ?: Double.NaN
```
menion, line 1926: [Suggestion]
If there is a risk for invalid location, highly suggest just omitting these points. They are useless.

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -1880,6 +1892,47 @@ class MeshService : Service(), Logging {
...
+            loc.longitude = nodeInfo.position?.longitude ?: Double.NaN
+            nodeInfo.position?.altitude?.let { loc.altitude = it.toDouble() }
+            nodeInfo.position?.time?.let { loc.time = 1000 * it.toLong() }
+            loc.provider = "Meshtastic"
```
menion, line 1930: [Suggestion]
To save a few bytes, remove this String. It is not used in the app at all (only rarely to detect if `Location` is GPS or Network based).

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -1880,6 +1892,47 @@ class MeshService : Service(), Logging {
             r.toString()
         }
     }
+
+    fun exportVisibleNodesToLocusMaps() {
```
menion, line 1896: [Question]
I see direct implementation in the `MeshService` object, but I see nowhere any check if Locus Map is installed and exists valid version and is running! Really suggest doing it because the current implementation seems to me like a waste of resources if previous conditions are not met!

You may use `LocusUtils.getActiveVersion` to get current active app version and if any exists, use `ActionBasics.getLocusInfo` to get metadata like if app currently runs.

PR #284, opened 2021-07-20
Russian translation update

rkuksa:


geeksville:
great!  I'll include in next weeks release.  Thanks so much!

PR #291, opened 2021-08-31
Feature: Added Send button

andrekir:
Missed having a "Send button" to send messages, so I made one.

Tried to keep it simple and follow current design aesthetic.

Screenshot_20210831-171208 Screenshot_20210831-172952

added:

- **sendButton** (ImageButton -> messages_fragment.xml + setOnClickListener -> MessagesFragment.kt)
- **multi-line text input** (removed android:singleLine="true" -> messages_fragment.xml)
- **200 character max text message limit** (android:maxLength="200" -> messages_fragment.xml) (real limit is 255 per packet, any way to calculate exact max text message size?)

Feedback is welcome!


geeksville:
looks great! thanks andre!

PR #304, opened 2021-11-06
Added Exit button to kill app and related services

jacky4566:
Added Exit button to main menu.
Added _android:stopWithTask="true"_ which kills related services when user "closes" app from task manager.

Do the devs want me to add all the language variants for the "exit" string?

Secondarily, it seems weird committing to main branch. It would be good to have a DEV and RELEASE branches for this git.


jacky4566:
Got too excited.
Requested changes need more debug. 
Ill be back!

PR #307, opened 2021-11-17
Feature: Camera QR Code scanning

andrekir:
Basic QR Code scanning feature using ZXing library IntentIntegrator.

Requires:
- android.permission.CAMERA

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -14,6 +14,9 @@ jobs:
         with:
           submodules: true
       - name: Loading secrets
+        env:
+          gservices: ${{ secrets.gservices }}
+          mapboxtoken: ${{ secrets.mapboxtoken }}
         run: |
           # not yet needed echo $gservices > ./app/google-services.json
           rm ./app/google-services.json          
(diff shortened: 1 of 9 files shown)


geeksville:
ooh sounds promising.

andrekir:
not sure how to deal with the camera permission. treating CAMERA as minimal requirement (current logic) seems excessive (persistent warnings, no camera = quit).

can we just let Android deal with the Camera permission? or could this break early Android versions 6.0+?)

feedback appreciated!

andrekir:
Added camera permissions following best practices: https://developer.android.com/guide/topics/permissions/overview

- Ask for permissions in context, when the user starts to interact with the feature that requires it.
- Don't block the user. Always provide the option to cancel an educational UI flow related to permissions.
- If the user denies or revokes a permission that a feature needs, gracefully degrade your app so that the user can continue using your app, possibly by disabling the feature that requires the permission.
- Don't assume any system behavior. For example, don't assume that permissions appear in the same permission group. A permission group merely helps the system minimize the number of system dialogs that are presented to the user when an app requests closely-related permissions.

mc-hamster:
Merged. Error is with map box.

geeksville:

PR #313, opened 2021-11-28
increase message max length to 228 chars

andrekir:
increase message maxLength to 228 characters (from 200).

in theory 237 should be max, but 234 appears to be the max for Android App and 228 for iOS App.
using 228 to match iOS.


geeksville:

PR #323, opened 2021-12-14
update permissions

andrekir:
- allow user to deny permissions (`warn` instead of `error`);
- improve and better explain permission dialogs;
- comply with Google's requirements for background location access.

references:
https://developer.android.com/training/location/permissions#request-background-location
https://support.google.com/googleplay/android-developer/answer/9799150


andrekir:
- added GPS sensor check;
- added 'location only' logic to make sure we have it before asking for background location;
- disabled WRITE_EXTERNAL_STORAGE (only used in debug).

geeksville:

PR #326, opened 2021-12-22
include firmware pre-releases in android release build 

andrekir:
previous "latest release" filter excluded firmware pre-releases. this change will use the latest device firmware, including pre-releases, for the OTA firmware update feature and keep device vs. Android versions closer and more compatible.


geeksville:

https://meshtastic.discourse.group/t/meshtastic-android-1-2-52-alpha/4613/1

PR #332, opened 2021-12-27
Remove duplicate permission

caveman99:
During commandline/CI builds with Gradle chokes on this duplicate declaration.


caveman99:
duplicated with a change from PR #331

Issue #275, opened 2021-04-13
App crash if changing tab quickly after BLE connection


MeshM-M:
I just sent a crash report from the app in the last hour with a similar title. OnePlus phone under android 8.

Scenario 1 -> crash
Launch meshtastic app from closed.
Connect by selecting proper device.
Once connection is done (name and region appear), switch to another tab (any tab).
-> crash

Link to movie:
https://www.dropbox.com/s/uk2hgmwq1fzf3fs/App%20crash%20after%20quick%20change%20of%20tab.MP4?dl=0

Scenario 2 -> no crash
Launch meshtastic app from closed.
Connect by selecting proper device.
Once connection is done (name and region appear), wait for say 5 secs before switching to another tab (any tab).
-> no crash

Link to movie:
https://www.dropbox.com/s/n24ey4rh6mjyp0f/No%20app%20crash%20if%20delay%20before%20changing%20tab.MP4?dl=0

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/android-app-1-2-22-crash-when-switching-tab-after-ble-pairing/3142/1

geeksville:
fixed!

PR #335, opened 2022-01-02
fix tabLayout smoothScroll crash

andrekir:
Issue #275

switching across all tabs with `smoothScroll` animations could create a race condition (with mapbox throwing exceptions) while trying to render all fragments at the same time (plus Bluetooth threads).

fix: `smoothScroll: false`


geeksville:

https://meshtastic.discourse.group/t/meshtastic-android-1-2-53-alpha/4778/1

PR #336, opened 2022-01-04
add provideLocation service start/stop

andrekir:
- start MeshService `setupLocationRequests` (to use phone GPS) when preference is enabled;
- attach Settings tab `provideLocationCheckbox` to call MeshService `setup/stopLocationRequests`;
- disable "Provide location to mesh" checkbox if Google Play Services is not available (prerequisite).


geeksville:

https://meshtastic.discourse.group/t/meshtastic-android-1-2-53-alpha/4778/1

PR #347, opened 2022-01-19
fix permissions for android >= 31

andrekir:
some Android 12 devices crash when legacy Bluetooth permissions are used.


mc-hamster:
thanks!

Issue #201, opened 2020-11-07
Add the ability to clear/delete messages 


nasprouse:
Currently there is no ability to delete individual  messages from the thread or clear all messages from the thread.

Adding this would allow messages to be cleared so a user could start fresh for the day/ event or to correct erroneous messages that clog the thread.

geeksville:
good idea

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/1-2-31-feedback/3370/1

PR #356, opened 2022-01-29
!WIP Merge feature/delete-messages into Master

PWRxPSYCHO:
This PR is targeted at Issue: #201 

The current attempt in this PR is just to delete messages locally. 

Looking into "delete all messages" method & UI integration

Further discussions/enhancements can be made about having the option to delete messages from the entire mesh, similar to Signal's delete functionality (DELETE FOR ME, DELETE FOR EVERYONE).

### DIFF
--- a/app/src/main/aidl/com/geeksville/mesh/IMeshService.aidl
+++ b/app/src/main/aidl/com/geeksville/mesh/IMeshService.aidl
@@ -69,6 +69,8 @@ interface IMeshService {
 
     void delete(int position);
 
+    void deleteAllMessages();
+
     /**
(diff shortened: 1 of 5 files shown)


MaximKrasikov:
Hi! What about "DELETE FOR EVERYONE" function?

garthvh:
> Hi! What about "DELETE FOR EVERYONE" function? 

It would create a lot of mesh activity and you can't really ensure that every node that received the message gets the delete message packet.

MaximKrasikov:
Is there any way to solve this problem or alternative ways?

Issue #303, opened 2021-11-06
Closing app and service


jacky4566:
Is there a way to kill the app service without going through the setting panel? Seems weird we don't have an "exit" button in the app to kill the service. 

It would also be nice if `android:stopWithTask="true"` was set so when you "close" the app from the task manager it destroys the service.

andrekir:
Duplicate of #186

PR #358, opened 2022-02-01
Close Application and Stop Services

PWRxPSYCHO:
Targeted at issue #303 

- Added "Exit" option from main menu which closes application and running services
- Added `android:stopWithTask="true" ` to AndroidManifest to end running services when user "swipes" application to exit.

Looking into why "Disconnected" notification still shows up on screen. (Seems to appear a few minutes after application is closed and BT is disabled)

Also looking into something better than `exitProcess(0)` when killing the app.


PWRxPSYCHO:
After discussion with @andrekir it was determined that the services need to continue running in the background after the application has been closed in order for specific plugins to work. As well as the mesh service. Persistence is required.

PR #360, opened 2022-02-02
Introduce Hilt dependency injection

mcumings:
Uses [Hilt](https://dagger.dev/hilt/) to get the database initialization off of the main thread.

The initial introduction always has a disproportionate fan-out of boilerplate.  In this case, all entry points which
were using `UIViewModel` needed to be annotated in order to let the code gen know that they needed to support
it.

The `PacketRepository` is injected into things via the main thread (e.g., the `MeshService`) but due to the
lazy declaration, the database isn't hydrated until the DAO is access while on an IO thread.


mcumings:
First in the installments for #369.  A toe in the water for Hilt dependency injection.

PR #365, opened 2022-02-04
Fixes #340 Improve CSV file export

### LINKED ISSUE #340: Export messages as CSV generate an empty file
When saving messages as CSV - it generates an empty file. Once it generated a file with headers only , but no messages.

csv

Android 11,  App version 1.2.51

mcumings:
Fixes  #340 and expands upon the written fields such that it (mostly) matches the range test plugin's exported file content.

Restructured this a bit to get more business logic out of the activity layer.  I moved it down to the viewmodel for now but this would really be moved into a repository.


andrekir:
thanks!

mc-hamster:
Thank you!

Issue #359, opened 2022-02-01
add offline map options to mapbox


andrekir:
Mapview
---

- [x] migrate to mapbox API v10 ([Installation](https://docs.mapbox.com/android/maps/guides/install/), [Migrate to v10](https://docs.mapbox.com/android/maps/guides/migrate-to-v10/))
- [x] add option to import map tiles / offline maps

garthvh:
@joshpirihi there is some interest in doing the same thing you did for the iOS app in android app, can you help @andrekir with any tile questions he runs into?  @thebentern has volunteered to do some android testing.


QGIS is the application we have been using to make the maps, the built in OpenStreetMap layers are a good place to start.

joshpirihi:
Happy to help! From what I could tell, MapBox doesn't have native support for local mbtiles map files. 

But I have seen a hack where you run a local http server on some random port (on the device), and tell the mapbox sdk that is a network source of mbtiles data.

PWRxPSYCHO:
> ## Mapview
> 
>     * [ ]  migrate to mapbox API v10 ([Installation](https://docs.mapbox.com/android/maps/guides/install/), [Migrate to v10](https://docs.mapbox.com/android/maps/guides/migrate-to-v10/))
> 
>     * [ ]  add option to import map tiles / offline maps

I've gotten some of the map migrated over to V10.

Still need to fix the following:
 - Gesture Controls
 - Adding Image,Source & layers
 - Updating Nodes
 - Camera Movement
 - Remove old dependency
 
 
I can open a PR with the changes so far.

andrekir:
> I can open a PR with the changes so far.

@PWRxPSYCHO great! please do.

PWRxPSYCHO:
@joshpirihi I'll have to look into .mbtiles. I'm not sure what the properties are on that file and if we can convert them over to what the offline manager needs to load the map (I would assume it is possible).

[MapBox V10 Offline Android SDK](https://docs.mapbox.com/android/maps/guides/offline/)

PWRxPSYCHO:
I will investigate this option and see what I can do! It looks like MapBox has an example one I can mess around with which is perfect

PWRxPSYCHO:
I've been chugging along at this and I think I'm about 75% of the way through getting something working. (There are a few menus/quality of life improvements I need to make still).

[...]

PR #368, opened 2022-02-05
Migrating to Mapbox API V10

PWRxPSYCHO:
Part 1 of issue #359

Working on the migration to V10 of MapBox

**Key things to note in this PR:**
- Min SDK version has been bumped from `15` to `21`
- `com.github.mik3y:usb-serial-for-android` has been updated from `v3.0.0` to `v3.4.3` (This might cause issues so we can attempt to revert if necessary)   
   
 **Work that has been completed:**
  - Upgraded MapBox dependency
  - Updated map_view.xml to reference to new version
  - Updated the rendering of MapBox in the MapFragment.kt file
  - Updated adding Layers, Source, & Image to Layer View
  - Updated Gesture Control
  - Updated Nodes Position
  - Updated Camera Movement (and initial load zoom level)
  - Removed old dependency

  
I do not have 2 GPS capable devices to test multiple nodes. So that should be tested before this gets merged. If there are any issues I would recommend whoever has the devices makes the changes so that it can be tested more efficiently.


app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
```diff
@@ -42,22 +40,20 @@ class MapFragment : ScreenFragment("Map"), Logging {
...
-        textColor(Color.RED),
-        textVariableAnchor(arrayOf(TEXT_ANCHOR_TOP)),
-        textJustify(TEXT_JUSTIFY_AUTO),
-        textAllowOverlap(true)
-    )
+    private val labelLayer = SymbolLayer(labelLayerId, nodeSourceId)
+        .textField(Expression.get("name"))
+        .textSize() // TODO Set text size
+        .textColor(Color.RED)
+        .textVariableAnchor(arrayListOf(TextAnchor.TOP.toString()))
```
PWRxPSYCHO, line 54: This is not correct.

I am looking into updating this

andrekir:
thank you!

PR #391, opened 2022-03-12
Feature/offline maps

PWRxPSYCHO:
Issue: #359

This my first attempt to get some of the offline maps feature working inside of Meshtastic's MapBox.
This feature is limited at it current state. In order to use the feature you must set the `offlineEnabled` flag to true on line 67 inside of MapFragment.kt. 

After the flag is enabled, the flow for getting an offline region is as follows:

- Long click on area on the map that you wish to download. A bounding box will then appear with a "location icon" at the center.
(the bounding box is 2.5miles x 2.5 miles)
- Then a FAB will appear with a download icon.
- Clicking the FAB will open up a menu with the following options
     - Save
     - Cancel
     - View Region
     - Add Custom URI?
- Saving without adding the custom URI will just download the current region to your current MapBox instance. If you close the application, the downloaded region will be removed.
- Once the region has finished downloading, open the Save Region dialog again and click "View Region". The camera will then fly over to the region and center the camera over top of, leaving the bounding box in place to show what you have downloaded.
- Clicking cancel will delete the offline region, and remove the bounding box from your screen and hide the FAB.

Custom URI:
MapBox allows you to create and publish your own styles onto MapBox studio. You are then able to import those styles into an existing MapBox instance. You can change the layers, text, add routes, icons, overlays, etc in these styles. Once you have the styles created you can then take the url (`mapbox://styles/...`) and add it to the "Custom URI" box to import it into MapBox when you download a region.

This feature is not perfect and I expect to make some enhancements with it in the future.


andrekir:
thanks!

Issue #13, opened 2020-04-11
add an option to force language X


geeksville:
Supposedly possible with code like this.

https://stackoverflow.com/questions/11013904/disable-localization-in-android-application/11014290

requested here:
https://meshtastic.discourse.group/t/call-for-help-please-localize-the-app-for-your-language/107/12?u=geeksville

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/call-for-help-please-localize-the-app-for-your-language/107/13

PR #401, opened 2022-03-30
add an option to force language X

wkawecki:
Enhancement https://github.com/meshtastic/Meshtastic-Android/issues/13 add an option to force language X


andrekir:
looks great! one minor detail I noticed, languages with regional variations `ko-rKR` and `pt-rBR` are not taking the change after restart. is that something you could fix before merging?

andrekir:
thank you

PR #409, opened 2022-04-08
MapFragment crashes if analytics are not allowed

mcumings:
`binding` is only set in `onCreateView` when analytics are
enabled, leading to a crash in `onViewCreated` when
`binding` - a lateinit var - is attempted to be accessed.


andrekir:
thanks!

PR #410, opened 2022-04-08
Initial step in refactoring RadioInterfaceService for dependency injection

mcumings:
Extracts USB device management into a `UsbRepository`.

In order for `SerialInterface to gain access to this prior to
the `RadioInterfaceService` being fully natively dependency
injected, all `InterfaceFactory` implementations needed
to be modified to accept the `UsbRepository` via argument.  This
will go away in a future PR.

Changed `assumePermission` constant to `false` as it was preventing
the request for permission from occurring, breaking serial connectivity.

Minor improvement: `SerialInterface` re-bonding by device name is
now supported.

Another minor improvement:  Adds USB serial connectivity support for TBeam and RAK 4631 devices. 😁

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -88,6 +88,7 @@ android {
 
     kotlinOptions {
         jvmTarget = "1.8"
+        freeCompilerArgs += [ '-Xopt-in=kotlin.RequiresOptIn' ]
     }
     lint {
         abortOnError false
--- a/app/src/main/java/com/geeksville/mesh/repository/usb/UsbRepository.kt
+++ b/app/src/main/java/com/geeksville/mesh/repository/usb/UsbRepository.kt
@@ -1,0 +1,77 @@
+package com.geeksville.mesh.repository.usb
+
+import android.app.Application
+import android.hardware.usb.UsbDevice
+import android.hardware.usb.UsbManager
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.coroutineScope
+import com.geeksville.android.Logging
+import com.geeksville.mesh.CoroutineDispatchers
+import com.hoho.android.usbserial.driver.UsbSerialProber
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.*
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Repository responsible for maintaining and updating the state of USB connectivity.
+ */
+@OptIn(ExperimentalCoroutinesApi::class)
+@Singleton
+class UsbRepository @Inject constructor(
+    private val application: Application,
+    private val dispatchers: CoroutineDispatchers,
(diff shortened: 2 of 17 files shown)


app/build.gradle
```diff
@@ -88,6 +88,7 @@ android {
 
     kotlinOptions {
         jvmTarget = "1.8"
+        freeCompilerArgs += [ '-Xopt-in=kotlin.RequiresOptIn' ]
```
mcumings, line 91: Adding this option tells the Kotlin compiler that we want to support the `@OptIn` feature.  This basically allows code to use unstable APIs (e.g., coroutines) without requiring consumers of that code to also declare it.  i.e., it prevents propagation into the consumer code.

app/src/main/java/com/geeksville/mesh/repository/usb/UsbRepository.kt
```diff
@@ -0,0 +1,77 @@
...
+    private val usbManagerLazy: dagger.Lazy<UsbManager?>,
+    private val usbSerialProberLazy: dagger.Lazy<UsbSerialProber>
+) : Logging {
+    private val _serialDevices = MutableStateFlow(emptyMap<String, UsbDevice>())
+
+    @Suppress("unused") // Retained as public API
+    val serialDevices = _serialDevices
+        .asStateFlow()
+
+    val serialDevicesWithDrivers = _serialDevices
```
mcumings, line 37: Eventually it would be cool if the UI responded to this state flow reactively.  For now we just use it to hold the most recent state.

app/src/main/java/com/geeksville/mesh/repository/usb/UsbRepository.kt
```diff
@@ -0,0 +1,77 @@
...
+import kotlinx.coroutines.withContext
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Repository responsible for maintaining and updating the state of USB connectivity.
+ */
+@OptIn(ExperimentalCoroutinesApi::class)
+@Singleton
+class UsbRepository @Inject constructor(
```
mcumings, line 23: I want to add one more piece of functionality to this repository as a subsequent step: management of the serial communication itself, as currently handled by `SerialInterface`.  Doing this work on top of this PR's worth of code change was getting a little unwieldy so I stepped back and stabilized this portion of the change in isolation, first.

app/src/main/java/com/geeksville/mesh/service/SerialInterface.kt
```diff
@@ -36,77 +35,44 @@ class SerialInterface(service: RadioInterfaceService, private val address: Strin
...
             return false
         }
 
-        private fun findSerial(context: Context, rest: String): UsbSerialDriver? {
-            val drivers = findDrivers(context)
-
-            return if (drivers.isEmpty())
-                null
-            else  // Open a connection to the first available driver.
-                drivers[0] // FIXME, instead we should find by name
```
mcumings, line 65: This FIXME has been fixed. 😄

PR #412, opened 2022-04-12
 meshtastic#73 'Resend Button' on message selection

wkawecki:
Resend Button' on message selection

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/MessagesFragment.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/MessagesFragment.kt
@@ -307,11 +307,14 @@ class MessagesFragment : Fragment(), Logging {
                                 }
                                 R.id.resendButton -> {
                                     debug("User clicked resendButton")
-                                    var resendText:String = ""
+                                    var resendText = ""
                                     selectedList.forEach {
                                         resendText = resendText + it.text + System.lineSeparator()
-                                        binding.messageInputText.setText(resendText)
                                     }
+                                    if (resendText!="")
+                                        resendText = resendText.substring(0, resendText.length - 1)
+                                    binding.messageInputText.setText(resendText)
+                                    mode.finish()
                                 }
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/geeksville/mesh/ui/MessagesFragment.kt
```diff
@@ -305,6 +305,14 @@ class MessagesFragment : Fragment(), Logging {
                                     actionMode?.title = selectedList.size.toString()
                                     notifyDataSetChanged()
                                 }
+                                R.id.resendButton -> {
+                                    debug("User clicked resendButton")
+                                    var resendText:String = ""
+                                    selectedList.forEach {
+                                        resendText = resendText + it.text + System.lineSeparator()
+                                        binding.messageInputText.setText(resendText)
```
andrekir: line 313 should go outside the forEach loop

app/src/main/java/com/geeksville/mesh/ui/MessagesFragment.kt
```diff
@@ -305,6 +305,14 @@ class MessagesFragment : Fragment(), Logging {
...
+                                    }
+                                }
```
andrekir, line 318: consider calling `mode.finish()` at the end

app/src/main/res/menu/menu_messages.xml
```diff
@@ -1,6 +1,11 @@
 <?xml version="1.0" encoding="utf-8"?>
 <menu xmlns:android="http://schemas.android.com/apk/res/android"
     xmlns:app="http://schemas.android.com/apk/res-auto">
+    <item
+        android:id="@+id/resendButton"
+        android:icon="@drawable/ic_retry_24"
```
andrekir: please use a `twotone` icon so it matches the color/style of the other icons.

andrekir:
thanks

Issue #369, opened 2022-02-08
Application Architecture Modernization (Modern Android Development / MAD)


mcumings:
The current application suffers from a lot of artificial complexity resulting from not having a well defined architecture.  Additionally, there are some bad practices which have become prevalent and are impacting app performance.  Some examples of this include:
- Blocking I/O on the main thread.  Examples:
  - Database is being initialized on the main thread
  - SharedPreferences updates are using `commit` versus `apply`
- Lots of business logic in the view layer (which happens to primarily run on the main thread)
- Poor encapsulation. Examples:
  - mesh onConnected calback being used to set `UiViewModel.meshService` field
  - `MeshService.kt` has LOTS of different responsibilities (and is nearly 1900 lines of code!)
  - Individual files defining multiple/many classes
- State is being managed in a variety of locations:
  - Different node databases in UiViewModel and MeshService

This tends to make the app difficult to understand and more fragile when it comes time to make changes.  By applying the architectural principles in Google's recommended application architecture (https://developer.android.com/topic/architecture) many of these complexities will be naturally eliminated.  For example:
- [Using dependency injection](https://developer.android.com/training/dependency-injection) can allow portions of the app to remain decoupled from each other, allowing them to be more freely changed, independently of one another while hiding their implementation details.
- [Driving the UI directly from data models](https://developer.android.com/jetpack/guide#drive-ui-from-model) can create a unidirectional data flow (e.g., radio -> database -> view model -> view) which can make it easier to debug the application as well as put all application state into a place which supports transactional change.
- Moving business logic out of the view layer and into [domain and/or data layers](https://developer.android.com/jetpack/guide/data-layer) allows for easier combination of discrete portions of the application and makes each component much more testable in isolation.

This issue is a proposal to shift the app closer to this recommended architecture in small, incremental steps.

andrekir:
considering your review I would say database needs the most attention. MeshService is the heart of the project and basically shares code between all different platforms. so we might leave that for last.

if there are tasks we can split or anything I can help, let me know.

mcumings:
Ok, I think that starting with preferences is sub-optimal for one main reason: when we change the database schema we either need to blow away the existing data or we need to define migrations to mutate the database schema over time, properly.

Preferences are one of those things which should remain persistent, indicating that the database would need to be leveraging migrations.

[...]

PR #417, opened 2022-04-22
Prep work for radio service repository

mcumings:
Continuing work on #369 

This pair of commits simply creates the `repository/radio` dir and moves the radio service code into that new location.

Doing this in a dedicated PR to make the next steps more clear/transparent.

Future/next steps:
- Converting the AIDL `IRadioInterfaceService` / `RadioInterfaceService` from a `Service` into a vanilla object instance.  Using services implies RPC calls - an expensive choice for something not exported outside of the application.
- Native dependency injection of the `InterfaceFactory` instances and `RadioInterfaceService` itself.
- Conversion of `RadioInterfaceService` into a proper repository form.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/RadioNotConnectedException.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/RadioNotConnectedException.kt
@@ -1,0 +1,4 @@
+package com.geeksville.mesh.service
+
+open class RadioNotConnectedException(message: String = "Not connected to radio") :
+    BLEException(message)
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/geeksville/mesh/service/RadioNotConnectedException.kt
```diff
@@ -0,0 +1,4 @@
+package com.geeksville.mesh.service
+
+open class RadioNotConnectedException(message: String = "Not connected to radio") :
```
mcumings, line 3: This exception is used outside of the radio service code.  Keeping it in its original location...

PR #420, opened 2022-04-23
`RadioInterfaceService` is no longer an Android `Service`

mcumings:
More work on #369 :

Removes the AIDL and associated support for the `IRadioInterfaceService`.
This should give some performance benefit since radio data no longer
has to round-trip through the platform.


andrekir:
this is great, thanks!

PR #454, opened 2022-07-09
add CH9102F USB serial (used on some tlora-v2-1-1.6 variants)

majbthrd:
The LILYGO LoRa32 modules that I recently purchased off Amazon use the CH9102F instead of a CP2102.  (The PCB markings are T3_V1.6.1 20210104, the same as one of the images on meshtastic.org)

felHR95/UsbSerial already supports devices using the CDC protocol (which the Qinheng CH9102F adopts), so adding support is as straightforward as adding the VID:PID to device_filter.xml.


andrekir:
thanks

PR #458, opened 2022-07-29
Explicitly enable DTR/RTS to fix RAK serial comms

mcumings:
Found the issue behind RAK nodes not connecting properly over serial/USB.  It looks like we have to explicitly raise data-terminal-ready and ready-to-send in order to start receiving data.

I found that gem in the source code published for the Serial USB Terminal app.


andrekir:
excellent, thanks!

PR #460, opened 2022-08-10
Added App Introduction Screens

### LINKED ISSUE #333: Feature Request: App first time launch walk-through
When I first launched the app, I had a background in how LoRa works and I had a rough idea of what to expect. I recently watched a Ham radio YouTuber demo Meshtastic and it seems even he really didn't understand some of the fundamentals of how the Meshtastic app works. He did not seem to get that the QR code stores the frequency, spreading factor, channel name, and encryption key.

The Meshtastic app could really use a "first time launch walk-through" where a few features are explained from a high level for first time users.

A few ideas on my mind:

- Link to the Meshtastic documentation somehow.
- Link to the buying opportunity for Meshtastic compatible devices if there is no device found after a few minutes. (Amazon searches can be affiliate links! Circle it back to Geeksville Industries LLC to help grow.) Example: https://amzn.to/314m7yZ Not sure if there is already a link in the documentation, I didn't see one. (Obviously selling devices to users carries a different set of risks than just linking to vendors or even search terms of compatible devices.) 
- Explain that the QR code carries the frequency/spreading factor/channel name/encryption key info somehow in a friendly way.
- Wireless sharing is very nice, is there a way to promote that to be the primary option? Link sharing wouldn't work without internet.
- Somehow remind users that they might have to wake up their nodes if they are battery powered. (Especially if settings are trying to be pushed and there is no node connected and the app appears to freeze.)
- Somehow educate users that this is a really low bandwidth system, and pushing video over it is not feasible.
- Discourage use of "Long range / Slow". I think most people might confuse this with "Of course I want long range!" (and not understand that LoRa has a HUGE link budget and SF7 works for many applications, especially if you have mesh nodes all over.) Maybe present an "are you sure?" popup explaining that even "Short range / Fast" can go miles with line of sight. (Allow the user to do it anyway, I'm just saying there should be education there.)
- Explain that Meshtastic is much better than goTenna because it's open source and you can modify it however you like! (Also, goTenna has much worse signal range because it's not LoRa.)

legendgroupv2:
Added three screens explaining what Meshtastic is, how to use the app and how to set up encryption. 
This is only shown on the first launch of the app - after completing the introduction the app returns to normal functionality.

Closes #333 

<details><summary>Screenshots</summary>

Screenshot_20220810-171356_Meshtastic
Screenshot_20220810-171400_Meshtastic
Screenshot_20220810-171404_Meshtastic

</details>


app/src/main/res/values/strings.xml
```diff
@@ -144,4 +144,13 @@
...
     <string name="shutdown">Shutdown</string>
     <string name="reboot">Reboot</string>
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="hello_blank_fragment">Hello blank fragment</string>
+    <string name="show_intro">Show Introduction</string>
+    <string name="intro_welcome_title">Welcome to Meshtastic</string>
+    <string name="intro_meshtastic_desc">Meshtastic is an open-source, off-grid, encrypted communication platform. The meshtastic radios form a mesh network and communicate using the LoRa protocol to send text messages</string>
+    <string name="intro_get_started">...Let\'s get started!</string>
+    <string name="intro_started_text">Connect your meshtastic device by using either bluetooth or serial. \n\nYou may need to wake your node if they are battery powered. \n\nYou can see which devices are compatible at www.meshtastic.org/docs/hardware</string>
+    <string name="intro_ecryption_title">"Setting up encryption"</string>
```
andrekir: typo

```suggestion
    <string name="intro_encryption_title">"Setting up encryption"</string>
```

app/src/main/java/com/geeksville/mesh/AppIntroduction.kt
```diff
@@ -0,0 +1,68 @@
...
+        ))
+        addSlide(AppIntroFragment.createInstance(
+            title = resources.getString(R.string.intro_get_started),
+            description = resources.getString(R.string.intro_started_text),
+            imageDrawable = R.drawable.icon_meanings,
+            backgroundColorRes = R.color.colourGrey,
+            descriptionColorRes = R.color.colorOnPrimary
+        ))
+        addSlide(AppIntroFragment.createInstance(
+            title = resources.getString(R.string.intro_ecryption_title),
```
andrekir: typo
```suggestion
            title = resources.getString(R.string.intro_encryption_title),
```

app/src/main/AndroidManifest.xml
```diff
@@ -1,196 +1,186 @@
...
-                <action android:name="android.intent.action.MY_PACKAGE_REPLACED" />
-                <!-- I was using PACKAGE_REPLACED, but MY_PACKAGE_REPLACED is newer and seems cleaner
-                <data
-                    android:scheme="package"
-                    android:path="com.geeksville.mesh" /> -->
-            </intent-filter>
-        </receiver>
-    </application>
-
-</manifest>
```
andrekir, line 196: please only change lines relevant to the PR

app/src/main/res/values/strings.xml
```diff
@@ -144,4 +144,13 @@
...
+    <string name="intro_encryption_text">To enable encryption, go to the channel tab and change the channel name. This will set a random key and enable AES256 encryption. \n\nTo communicate with other devices they will need to scan your QR code or follow the shared link to configure the channel settings.</string>
```
andrekir: the default channel has a public AES256 key, but encryption is enabled. maybe "to create your own channel and encryption key" ...

app/src/main/res/values/strings.xml
```diff
@@ -144,4 +144,13 @@
     <string name="resend">Resend</string>
     <string name="shutdown">Shutdown</string>
     <string name="reboot">Reboot</string>
+    <!-- TODO: Remove or change this placeholder text -->
+    <string name="hello_blank_fragment">Hello blank fragment</string>
+    <string name="show_intro">Show Introduction</string>
+    <string name="intro_welcome_title">Welcome to Meshtastic</string>
+    <string name="intro_meshtastic_desc">Meshtastic is an open-source, off-grid, encrypted communication platform. The meshtastic radios form a mesh network and communicate using the LoRa protocol to send text messages</string>
+    <string name="intro_get_started">...Let\'s get started!</string>
+    <string name="intro_started_text">Connect your meshtastic device by using either bluetooth or serial. \n\nYou may need to wake your node if they are battery powered. \n\nYou can see which devices are compatible at www.meshtastic.org/docs/hardware</string>
```
andrekir: - app now supports BLE, Serial and HTTP.
- "You may need to wake your node if they are battery powered." no longer applies.

```suggestion
    <string name="intro_started_text">Connect your meshtastic device by using either Bluetooth, Serial or WiFi. \n\nYou can see which devices are compatible at www.meshtastic.org/docs/hardware</string>
```

andrekir (changes requested):
@legendgroupv2 Thank you for your contribution!
Please check my reviews 👍

app/src/main/res/values/strings.xml
legendgroupv2: Have reworded it to reflect your suggestions. Thanks :)

app/src/main/AndroidManifest.xml
legendgroupv2: Apologies it seems like my IDE reformatted the file so my changes aren't clear. I've reverted the file back to the original, committed it, and then added my code to the latest commit. It's the activity relating to AppIntroduction. Hopefully, this resolves the issue.

legendgroupv2:
Thank you for your reviews. I've addressed them all and committed the changes. In regards to the AndroidManifest file, it seems like my IDE reformatted the file so my changes aren't very clear. I've reverted the file back to the original, committed it, and then added my code to the latest commit. Hopefully, this resolves the issue.

andrekir:
I took care of some leftovers in the manifest. if you run into this again, a tip is to revert at the file level on your local copy. you can review the diffs in the IDE after. 👍

andrekir (approved):
all good! thank you!

PR #468, opened 2022-08-23
Add factory reset button

Douile:
Add a button to factory reset the device in advanced settings.

I cannot actually confirm whether this works due to https://github.com/meshtastic/Meshtastic-device/issues/1597 but it does brick the device :joy:


andrekir:
thanks!

PR #469, opened 2022-08-23
Shows a prompt when Bluetooth is off and trying to add a device

legendgroupv2:
When trying to add a Meshtastic device, the app will check if Bluetooth is turned on, if it isn't then it will display a message.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
@@ -506,10 +506,7 @@ class SettingsFragment : ScreenFragment("Settings"), Logging {
     private fun checkBTEnabled(
         warningReason: String = getString(R.string.requires_bluetooth)
     ) {
-
-        var btAdapter = BluetoothAdapter.getDefaultAdapter();
-
-        if (!(btAdapter.isEnabled())) {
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
```diff
@@ -501,6 +503,18 @@ class SettingsFragment : ScreenFragment("Settings"), Logging {
         }
     }
 
+    private fun checkBTEnabled(
+        warningReason: String = getString(R.string.requires_bluetooth)
+    ) {
+
+        var btAdapter = BluetoothAdapter.getDefaultAdapter();
+
+        if (!(btAdapter.isEnabled())) {
```
andrekir: Bluetooth state is handled via `bluetoothViewModel.enabled.value`, then `BluetoothAdapter` import can also be removed.

```suggestion
        if (bluetoothViewModel.enabled.value == false) {
```

andrekir (changes requested):
@legendgroupv2 Thanks for your contribution!
Please check my reviews 👍

legendgroupv2:
Thanks for your reviews. All done :)

andrekir (approved):
thanks

Issue #465, opened 2022-08-16
migrate to maplibre 


andrekir:
(this is a draft, please comment if you'd like to contribute)
* [ ] create free MapTiler Cloud account & api key
* [x] refactor gradle/MainActivity code to use maplibre
* [ ] recreate MapFragment
* [ ] cleanup mapbox code ([guide](https://docs.mapbox.com/android/maps/guides/install/))

Tutorial: https://github.com/maptiler/mobile-code-samples/blob/main/samples/android/TUTORIAL.md

recreating MapFragment from scratch is probably more efficient considering current maplibre SDK 9.5.2 is missing too many mapbox features in use.

## Requirements:
- show on map nodes positions with pins and labels (name, lastHeard, etc);
- consider/include logic for waypoint packets ([waypoint_app portnum](https://github.com/meshtastic/Meshtastic-protobufs/blob/2fadf011e574c180afb395bc2a751b3f7e56839c/mesh.proto#L600));
- bonus points for offline maps.

## Considerations:
- MapTiler free account capacity ([plans](https://www.maptiler.com/cloud/pricing/))
- other map source alternatives?

if I'm missing anything, suggestions, please comment.

Issue #19, opened 2020-04-22
Support off-line maps


geeksville:

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/offline-maps-on-meshtastic/137/3

aHVzY2g:
I would suggest the usage of OsmAnd. The only job to be done would be to integrate the "aidl interface", of Osmand. https://github.com/osmandapp/Osmand/blob/master/OsmAnd/src/net/osmand/aidl/IOsmAndAidlInterface.aidl#L445

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/offline-maps-on-meshtastic/137/21

aHVzY2g:
@geeksville did you already had a thought, which "OfflineMap integration" would fit the scope of the project best?

geeksville:
@aHVzY2g Mapbox as a really nice and easy looking API to turn on off-line maps.  I just didn't call it ;-).  So I'd say the "low hanging fruit" solution is for someone to just use the mapbox version.  

Porting to some other API would be fine if someone wants to do that and it provides the same level of polish.  But the docs for mapbox are super good and it was super easy to for me add maps with their api.  the last time I looked at OsmAndroid it was much more rough - but that was for a project like 6 yrs ago (andropilot and droneshare) so it has probably gotten better.

jetatomic:
Please consider non-urban areas when looking at map examples. Many maps which look very cool in cities end up being completely blank in areas with no roads. OpenStreetMap, which is the source OsmAnd primarily draws from, only has 1 layer (of 5) which shows useful details when there are no roads. If OSM sources are selected, please consider using the _Cycle Map_ layer if no other robust data sources are layered over the basemap. Otherwise, people hiking might find this is all the map shows, which is the 4km surrounding my home:

![Screenshot_2020-07-11 OpenStreetMap(1)](https://user-images.githubusercontent.com/3219559/87235328-7fc77200-c387-11ea-8923-b548dedfbca0.png)


[Example in the Meshtastic forum](https://meshtastic.discourse.group/t/offline-maps-on-meshtastic/137/24?u=jetatomic).

chron0:
[...]

PR #470, opened 2022-08-24
Feature/osmdroid migration

PWRxPSYCHO:
#465  (This option is an alternative to Maplibre)

#19 This should hopefully close out this issue

Phase 1: MVP - get application working as close to if not the same as it was with MapBox

TODO
- [x] Verify removal of all MapBox configurations
- [x] Change TileSource while map is open/app is open
- [x] Render Nodes on map
- [x] Zoom to node/nodes when first coming into application
- [x] Set zoom level as far back as possible
- [x] Show node name & timestamp
- [x] Copyright shows on each map

**_Offline Maps will be moved out to phase 2_**

Library Being used: https://github.com/osmdroid/osmdroid 

- [x] Updating single node
**Given**: I have multiple nodes displaying on the map
**When**: The location for one node changes
**Then**: The old node location will not display and the new node location will display

- [x] Updating multiple nodes
**Given**: I have multiple nodes displaying on the map
**When**: The location for multiple node changes
**Then**: The old nodes location will not display and the new nodes location will display

- [x] Updating multiple nodes and changing map layers
**Given**: I have multiple nodes displaying on the map
**When**: The location for all the node changes and I change the map source
**Then**: The old node locations will not display and the new node location will display on the new map source


app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
```diff
@@ -1,680 +1,316 @@
...
-
-        // We might not have a real mapview if running with analytics
-        if ((requireContext().applicationContext as GeeksvilleApplication).isAnalyticsAllowed) {
-//            binding.fabStyleToggle.setOnClickListener {
-//                //TODO: Setup Style menu for satellite view, street view, & outdoor view
-//            }
-            binding.downloadRegion.setOnClickListener {
-                // Display menu for download region
-                this.downloadRegionDialogFragment()
+    private fun setESRITileSource() {
```
PWRxPSYCHO: I would like to move this out to its own class where we can add extra tile sources/layers more easily in the future.

app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
```diff
@@ -1,680 +1,316 @@
...
-                style.addSource(geoJsonSource)
-                style.addPersistentLayer(lineLayer)
-                style.addLayer(userTouchLayer)
-            } else {
-                style.removeStyleLayer(lineLayerId)
-                style.removeStyleSource(boundingBoxId)
-                style.addSource(geoJsonSource)
-                style.addLayer(lineLayer)
-            }
+    private fun matchOnlineTileSourceBase(name: String): OnlineTileSourceBase {
```
PWRxPSYCHO: I would like to try and make this more dynamic, Otherwise every time we add a new source it must be added here

app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
```diff
@@ -1,680 +1,316 @@
...
-            lineWidth(1.5)
-            lineColor("#888")
-        }
-
-        if (point != null) {
-            binding.downloadRegion.visibility = View.VISIBLE
+    override fun onResume() {
+        super.onResume()
+        map.invalidate()
+        val tileSourceName = mPrefs.getString(
```
PWRxPSYCHO: I need to determine if this is actually needed after adding the setDestroyMode(false)

app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
```diff
@@ -1,680 +1,316 @@
...
-            // and zoom ranges of the tiles within the region.
-
-            // The tileset descriptor encapsulates the tile-specific data, such as which tilesets, zoom ranges,
-            // pixel ratio etc. the cached tile packs should have. It is passed to the Tile Store along with
-            // the region area geometry to load a new Tile Region.
-
-            // The OfflineManager is responsible for creating tileset descriptors for the given style and zoom range.
+    override fun onPause() {
+        val edit = mPrefs.edit()
+        edit.putString(prefsTileSource, map.tileProvider.tileSource.name())
```
PWRxPSYCHO: I need to determine if this is actually needed after setting setDestroyMode(false)

app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
```diff
@@ -1,680 +1,316 @@
...
-
-        val style = styleURI.ifEmpty {
-            mapView?.getMapboxMap()
-                ?.getStyle()?.styleURI.toString()
+    private fun loadOnlineTileSourceBase(): OnlineTileSourceBase {
+        val prefs = context?.getSharedPreferences(uiPrefs, Context.MODE_PRIVATE)
+        val mapSourceId = prefs?.getInt(mapStyleId, 1)
+        debug("mapStyleId from prefs: $mapSourceId")
+        val mapSource = when (mapSourceId) {
+            0 -> TileSourceFactory.MAPNIK
```
PWRxPSYCHO: Potentially make this more dynamic

app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
```diff
@@ -1,680 +1,316 @@
...
-                f
+                marker
             }
+            return mrkr
+        }
+        nodePositions = getCurrentNodes()
+    }
+
+    private fun drawOverlays() {
+        map.overlayManager.overlays().clear()
```
PWRxPSYCHO, line 145: Temporary solution. Ideally we should only update the locations of each node, rather than re-draw each layer anytime there is an update

app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
PWRxPSYCHO: Determined this wasn't needed

app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
PWRxPSYCHO: Determined that this is not needed

app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
PWRxPSYCHO: Not needed after removing onPause() functionality

app/src/main/java/com/geeksville/mesh/ui/MapFragment.kt
PWRxPSYCHO: Not sure if I will be able to change this. It does seem lightweight enough to re-draw the top layers on update.

andrekir: > It does seem lightweight enough to re-draw the top layers on update.

agreed.

Issue #478, opened 2022-09-08
Waypoint message support


MaximKrasikov:
would it be possible to implement sending a location on the map, as a named marker displayed on the map, without having a device at that specific location? It would be great to be able to e.g define a common meeting point that appears on everyone’s map without having to be at that point already.

andrekir:
this can be done adding support for Waypoint messages: https://meshtastic.org/docs/developers/protobufs/api#waypoint

MaximKrasikov:
will i be able to send list of markers to other users via waypoint ? Do I need to fix the firmware of the device and proto for this?

andrekir:
> will i be able to send list of markers to other users via waypoint ?

yes. same as other message types.

> Do I need to fix the firmware of the device and proto for this?

no. just the app side.

MaximKrasikov:
I apologize in advance for the stupid question. I just started to deal with meshtastic
I made changes to NodeInfo like EnvironmentMetrics in your last commits and add handleReceivedWaypoint in MeshService. Also create my custom marker in MapFragment via long click on osm view and save it in NodeInfo my new Waypoint, but nothing is sent. 
Are these all the changes I need?
my device-firmware 1.3.39 and 1.3.40 application in android
Thanks for adding osm)

andrekir:
feel free to ask questions. make sure firmware and app are on 1.3.40, there are lots of breaking changes with previous versions.

waypoints are handled as DataPacket instead of NodeInfo. quick draft:

- add WAYPOINT_APP_VALUE to recentDataPackets in MeshService (same as TEXT_MESSAGE_APP_VALUE)
- add waypoint object in model/MessageState and filters by type == TEXT_MESSAGE_APP_VALUE / WAYPOINT_APP_VALUE for each list
- from MapFragment use MessageState.waypoints.observe to build markers
- add logic to send waypoint messages similar to MessageState.sendMessage (initially just send to the primary channel).

also need to check how the waypoint payload looks like and how to get the data from DataPacket.bytes

MaximKrasikov:
Do I need to change something in MainActivity?
Need to make edits only in MeshService, MessageState, MapFragment and NodeInfo?

andrekir:
[...]

PR #485, opened 2022-09-22
Display waypoints on map

PWRxPSYCHO:
#478 - Waypoints 
This PR setups the MapFragment to display waypoints as an overlay.


andrekir:
thank you!

Issue #488, opened 2022-10-03
Phase 2: Clear map cache


PWRxPSYCHO:
Sometimes the maps will not correctly update, they will require the user to download tiles from the tile server, however there can be issues when old tiles are cached and new tiles are being downloaded for different zoom levels. Having an option to clear the tile cache should allow brand new tiles to be downloaded for the user.

 - Look into clearing cache for specific tiles
 - Otherwise clear all cache.

PR #489, opened 2022-10-08
Osmdroid Phase 2

PWRxPSYCHO:
### **Change Log**
- New Map sources (Open Topo)
- ESRI tile source concurrent download threads have increased from 2->4 (this should allow quicker rendering of tiles)
- Download Region functionality (#19 - Limited in use but functional for this first iteration.)
  - **NOT ALL MAP SOURCES CAN BE DOWNLOADED FOR OFFLINE USE**
  - A download button will appear at the bottom right corner of the screen when the map can be downloaded
Screenshot_20221007-203104
  - Offline Manager Window will appear with the following options:
Screenshot_20221007-203341
  - Current Cache Size - calculates  current cache usage and displays information to screen
  - Download Region - opens up download region dialog and bounding box
  - Clear Download Cache #488  - Map SQL database cache is now clear-able
  - Cancel - exits menu
  
  
Download Region:
1. A Bounding Box will appear on the screen surrounding an area that you can download.
2. You can select from 3 different zoom levels:
    - 5 miles (approx 12.5 sq miles)
    - 10 miles (approx 50sq miles)
    - 15 miles (approx 112.5 sq miles)
The distance given in each button is measured by the the total height of the bounding box. Which is an approximation.
3. You can pan the map in any direction to fine tune the region you wish to download. (Zoom will not work on this screen)
4. An approximate tile count will appear below the box options and above the "Start Download" & "Cancel" button
5. The region downloaded will cover the current zoom level all the way down to the furthest in zoom. (Furthest back: 11.5, closest: 15)
Screenshot_20221007-203349


andrekir (approved):
excellent, thank you!

thebentern:
Wow! This is incredible

PR #505, opened 2022-10-27
Map View Updates

PWRxPSYCHO:
### Change Log

- New Map Source ([ESRI World Topo](https://www.arcgis.com/apps/mapviewer/index.html)) 
https://www.arcgis.com/home/item.html?id=6e850093c837475e8c23d905ac43b7d0#!
- Updated `Clear Downloaded Tiles` dialog
    - You can now select which tile source you wish to clear cache for
- Updated min zoom level for ESRI Clarity  and URL
  
Screenshot_20221027-092821

Other code added is for future use.

Most upcoming updates will be halted until the map is migrated over to JetPack Compose. This will require a redesign of the architecture.


andrekir (approved):
thanks!

PR #532, opened 2022-12-03
Account for negative metric values in telemetry

thebentern:
Fixes issue reported here:
https://meshtastic.discourse.group/t/eviromental-sensor-wont-report-temperatures-below-0-c/6871/3


andrekir:
Thanks, Ben!

Issue #580, opened 2023-02-03
Use "min" instead of "m" for Minutes


micheljung:
It took me a while to understand that "8m" doesn't mean 8 meters but 8 Minutes in the Peers or Map view.

This should be changed, because according to the [International Symbols of Units](https://en.wikipedia.org/wiki/International_System_of_Units), `m` stands for Meter while `min` stands for Minute.

micheljung:
Wait, does it stand for Meter after all? I'm so confused :D

Issue #592, opened 2023-02-21
allow copying the gps coordinates of the peers


krasi-georgiev:
on the map when clicking a peer it shows its distance and GPS coordinates, but to be able to navigate to it need to also copy the coordinates and use these in google maps for navigation.

so long press-to-copy feature would be a great addition.

RightHandMan1:
In the peers tab(click the people), it shows they're position as a link. Clicking that link, will let you open it in maps,etc.

krasi-georgiev:
Thanks I just tried it, and it does open google maps but doesn't populate the coordinates in the search box of maps

PR #609, opened 2023-03-27
Fix issue #580 with formatting

GMart:
### Two fixes:
#580 fixed. Replaces m with min, and added space to the format. I also changed the threshold for switching the units around to something that makes more sense to me. (So we won't see "99 hours ago" any more)
#592 Fixed. When Google Maps now opens, it drops a pin instead of just going to the location. Google Maps has a non-standard geo URI scheme.
You can copy the coords from Maps.
Future improvement could be to add a hold action to copy the coords directly from the app.


andrekir:
a bit stuck here. #592 is about coords from the app's map (not Google Maps or 3rd party) and not sure if the Geo scheme change is Google Maps specific (we don't want to break compat with other map apps).

I'll need more time to review, sorry. if I could offer a tip, it's usually better to address issues in their own separate PRs.

GMart:
@andrekir I understand.
For #592 I was actually focusing on the second and third comments made in the issue chain. I have reverted that change and want to make this change just about #580. I'll open another PR for the other change.

PR #650, opened 2023-07-04
Update strings.xml

adoeller:
lange Reichweite -> hohe reichweite

german translatin wron in these three lines. "hohe" was already used elsewhere in the file.


andrekir:
translations are done using Crowdin, please submit your changes at https://crowdin.meshtastic.org/

PR #651, opened 2023-07-04
Add option for Miles and Feet for distance

### LINKED ISSUE #621: [Bug]:  Android app vs 2.1.4 , peers show distance in km instead of miles
Version 2.1.4 of the android, (2.1.17 firmware), in device settings I have Imperial selected, but the peers list keeps showing their distance in km. this node IS connected to wifi and mqtt, but assume it should still translate the distance.

jp-bennett:
closes #621

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
+++ b/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
@@ -218,11 +218,14 @@ data class NodeInfo(
     }
 
     /// @return a nice human readable string for the distance, or null for unknown
-    fun distanceStr(o: NodeInfo?) = distance(o)?.let { dist ->
+    fun distanceStr(o: NodeInfo?, prefUnits: Int = 0) = distance(o)?.let { dist ->
         when {
             dist == 0 -> null // same point
-            dist < 1000 -> "%.0f m".format(dist.toDouble())
-            else -> "%.1f km".format(dist / 1000.0)
+            prefUnits == ConfigProtos.Config.DisplayConfig.DisplayUnits.METRIC_VALUE && dist < 1000 -> "%.0f m".format(dist.toDouble())
+            prefUnits == ConfigProtos.Config.DisplayConfig.DisplayUnits.METRIC_VALUE && dist >= 1000 -> "%.1f km".format(dist / 1000.0)
+            prefUnits == ConfigProtos.Config.DisplayConfig.DisplayUnits.IMPERIAL_VALUE && dist < 1609 -> "%.0f ft".format(dist.toDouble()*3.281)
+            prefUnits == ConfigProtos.Config.DisplayConfig.DisplayUnits.IMPERIAL_VALUE && dist >= 1609 -> "%.1f mi".format(dist / 1609.34)
+            else -> null
         }
     }
 }
--- a/app/src/main/java/com/geeksville/mesh/ui/UsersFragment.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/UsersFragment.kt
@@ -194,7 +194,7 @@ class UsersFragment : ScreenFragment("Users"), Logging {
             }
 
             val ourNodeInfo = model.ourNodeInfo.value
-            val distance = ourNodeInfo?.distanceStr(n)
+            val distance = ourNodeInfo?.distanceStr(n,  model.config.display.units.number)
             if (distance != null) {
                 holder.distanceView.text = distance
                 holder.distanceView.visibility = View.VISIBLE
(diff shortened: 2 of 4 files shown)


jp-bennett:
There is one problem, that the current view doesn't update when the units are switched. I'm not sure the best way to trigger a refresh.

app/src/main/java/com/geeksville/mesh/MainActivity.kt
```diff
@@ -170,12 +170,16 @@ class MainActivity : AppCompatActivity(), Logging {
         (application as GeeksvilleApplication).isInTestLab
     }
 
+    companion object {
+        public var prefUnits = 0
```
andrekir: storing public variables in an Activity should be avoided

app/src/main/java/com/geeksville/mesh/MainActivity.kt
```diff
@@ -858,4 +866,32 @@ class MainActivity : AppCompatActivity(), Logging {
...
+            prefUnits = unitTags.values.elementAt(position)
+            debug("Set Units pref to $prefUnits")
+            prefs.edit().putInt("unit", prefUnits).apply()
+            //AppCompatDelegate.setDefaultNightMode(selectedTheme)
+            dialog.dismiss()
+        }
+
+        val dialog = builder.create()
+        dialog.show()
+    }
```
andrekir: this isn't needed, instead use the radio config stored in `viewModel.config.display.units`
--> DisplayConfig.DisplayUnits.METRIC_VALUE == 0
and DisplayConfig.DisplayUnits.IMPERIAL_VALUE == 1

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -221,8 +222,11 @@ data class NodeInfo(
     fun distanceStr(o: NodeInfo?) = distance(o)?.let { dist ->
         when {
             dist == 0 -> null // same point
-            dist < 1000 -> "%.0f m".format(dist.toDouble())
-            else -> "%.1f km".format(dist / 1000.0)
+            prefUnits == 0 && dist < 1000 -> "%.0f m".format(dist.toDouble())
+            prefUnits == 0 && dist > 1000 -> "%.1f km".format(dist / 1000.0)
```
andrekir: missing dist == 1000, use >=

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -221,8 +222,11 @@ data class NodeInfo(
...
+            prefUnits == 1 && dist < 1609 -> "%.0f feet".format(dist.toDouble()*3.281)
```
andrekir: convention is to use the abbreviated form 'ft' here

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -221,8 +222,11 @@ data class NodeInfo(
...
+            prefUnits == 1 && dist >= 1609 -> "%.1f miles".format(dist / 1609.34)
```
andrekir: same, use 'mi' here

andrekir (changes requested):

andrekir:
simplest way:

1. in fragments that call distanceStr() --> add `val prefUnits = model.config.display.units.number `
2. add the prefUnits argument in the fragments distanceStr() calls: distanceStr(x) --> distanceStr(x, prefUnits)
3. in NodeInfo.kt you pass the value adding a parameter to distanceStr() --> like: distanceStr(o: Nodeinfo, unit: Int = 0) ...

let me know if you need any help

app/src/main/java/com/geeksville/mesh/MainActivity.kt
jp-bennett: I really struggled to find another way to store global configuration data. What is the better way to do it?

andrekir: in Android always: UI -> ViewModel -> data sources

metric/imperial units is already stored in the UIState viewmodel. see my other comments. nothing should be added in MainActivity.

jp-bennett: Am I understanding correctly, that `model.config.display.units.number` should already contain an int value corresponding to either `ConfigProtos.Config.DisplayConfig.DisplayUnits.METRIC_VALUE` or `ConfigProtos.Config.DisplayConfig.DisplayUnits.IMPERIAL_VALUE`? And I shouldn't have to add a menu option at all?

andrekir: > Am I understanding correctly, that `model.config.display.units.number` should already contain an int value corresponding to either `ConfigProtos.Config.DisplayConfig.DisplayUnits.METRIC_VALUE` or `ConfigProtos.Config.DisplayConfig.DisplayUnits.IMPERIAL_VALUE`?

yes, `METRIC_VALUE` is 0 and `IMPERIAL_VALUE` is 1.

> And I shouldn't have to add a menu option at all?

correct

jp-bennett:
@andrekir I think that should put this in better shape. Thanks for the hand-holding.

andrekir:
looks good! thanks!

Issue #649, opened 2023-06-30
[FEATURE] Custom connection options


tb0hdan:
Hi,

Please add the following:

- Custom IP:Port for Meshtastic devices behind router (and possibly HTTP basic or similar)
- Keepalive / Timeout / Reconnect options for WiFi + request above

Much appreciated.

ut3uku:
I consider it's important!

G2Dolphin:
Yeah, I already [https://github.com/meshtastic/Meshtastic-Android/issues/643](posted) similar FR, which is similar to [https://github.com/meshtastic/Meshtastic-Android/issues/472](other FR) before.

jp-bennett:
It doesn't sound like a good idea to expose a Meshtastic device to the Internet, even on a random port. A VPN like Wireguard is a much better approach. PR #663 does allow that use case.

PR #663, opened 2023-07-17
Adds Field to Specify Device IP

### LINKED ISSUE #472: Allow adding nodes via IP connection by manually providing an IP/host
In the latest 1.3 release the android application lists meshtastic nodes within the same network connected via WiFi. It would also be interesting to allow a connection to a remote node by providing an IP / host that is not within the same network / ip range.

jp-bennett:
Closes #472

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
@@ -14,6 +14,7 @@ import android.os.Bundle
 import android.os.Handler
 import android.os.Looper
 import android.os.RemoteException
+import android.util.Patterns
 import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
@@ -24,6 +25,7 @@ import android.widget.RadioButton
 import android.widget.TextView
 import android.widget.Toast
 import androidx.activity.result.contract.ActivityResultContracts
+import androidx.core.widget.doAfterTextChanged
 import androidx.fragment.app.activityViewModels
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.asLiveData
@@ -455,6 +457,27 @@ class SettingsFragment : ScreenFragment("Settings"), Logging {
         }
     }
 
+    private fun addManualDeviceButton() {
+        val b = binding.radioButtonManual
+        val e = binding.editManualAddress
+
+        b.isEnabled = false
+
+        binding.deviceRadioGroup.addView(b)
+
+
+        b.setOnClickListener {
+
+
+            b.isChecked = onSelected(BTScanModel.DeviceListEntry("", "t" + e.text, true))
+
+        }
+        binding.deviceRadioGroup.addView(e)
+        e.doAfterTextChanged {
+            b.isEnabled = Patterns.IP_ADDRESS.matcher(e.text).matches()
+        }
+    }
+
     private fun updateDevicesButtons(devices: MutableMap<String, BTScanModel.DeviceListEntry>?) {
         // Remove the old radio buttons and repopulate
         binding.deviceRadioGroup.removeAllViews()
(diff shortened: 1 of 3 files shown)


jp-bennett:
Also partially address #649 .

app/src/main/res/layout/settings_fragment.xml
```diff
@@ -109,6 +109,22 @@
...
+                android:layout_height="wrap_content"
+                android:text="@string/IP_Address" />
+
+            <EditText
+                android:id="@+id/editManualAddress"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:ems="10"
+                android:hint="@string/IP_Address"
+                android:inputType="text"
```
andrekir: I suggest using `android:inputType="number|text"` for IP address. it opens the numeric keyboard and allows IP input.

app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
```diff
@@ -455,42 +457,70 @@ class SettingsFragment : ScreenFragment("Settings"), Logging {
...
+                // Note: we pull this into a tempvar, because otherwise some other thread can change selectedAddress after our null check
+                // and before use
+                val curAddr = scanModel.selectedAddress
+                if (curAddr != null) {
+                    val curDevice = scanModel.getDeviceListEntry(curAddr)
+                    addDeviceButton(curDevice, model.isConnected())
+                }
+            }
+        }
+        addManualDeviceButton()
```
andrekir, line 507: this is the only line you need to add in updateDevicesButtons(), the rest shouldn't change so it only runs when devices != null.

app/src/main/res/values/strings.xml
```diff
@@ -50,6 +50,7 @@
     <string name="connected_count">Connected: %1$s of %2$s online</string>
     <string name="list_of_nodes">A list of nodes in the network</string>
     <string name="update_firmware">Update Firmware</string>
+    <string name="IP_Address">IP Address:</string>
```
andrekir: please use snake_case here for the string name (all lowercase)

andrekir (changes requested):

app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
jp-bennett: Is Devices == null a special case beyond just an empty list of devices?

jp-bennett:
@andrekir Requested changes made. Tested and it's still working as expected. Thanks!

app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
andrekir: devices should rarely be null except before being initialized, mostly because it is MutableLiveData.

app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
```diff
@@ -481,16 +504,24 @@ class SettingsFragment : ScreenFragment("Settings"), Logging {
...
         // If we are running on an emulator, always leave this message showing so we can test the worst case layout
         val curRadio = scanModel.selectedAddress
 
-        if (curRadio != null && !MockInterface.addressValid(requireContext(), usbRepository, "")) {
+        if (curRadio != null && !MockInterface.addressValid(
+                requireContext(),
+                usbRepository,
+                ""
+            )
+        ) {
```
andrekir: also please do not reformat lines that you didn't have to change the code on

andrekir (changes requested):
looks really good. added one small request.

app/src/main/java/com/geeksville/mesh/ui/SettingsFragment.kt
jp-bennett: I think Android Studio decided to reformat that for me, and I didn't notice. How helpful. I'll get it fixed.

jp-bennett:
> looks really good. added one small request.

Cleaned that up. And some extra bits in the fragment XML that didn't need changed.

andrekir (approved):

andrekir:
looks great, thank you!

PR #687, opened 2023-08-21
Treat Detection Sensor messages like text messages

thebentern:
I had to remove Android Studio from my Macbook due to disk space, but I think this should work. I can pull it down later on my Windows PC.

Update: Tested and working on my Samsung Galaxy Note 10+. I'm seeing push notifications and the messages show up as texts in the normal chat view as well now.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/DataPacket.kt
+++ b/app/src/main/java/com/geeksville/mesh/DataPacket.kt
@@ -48,10 +48,10 @@ data class DataPacket(
     )
 
     /**
-     * If this is a text message, return the string, otherwise null
+     * If this is a text or detection sensor message, return the string, otherwise null
      */
     val text: String?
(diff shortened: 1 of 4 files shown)


andrekir:
are the text messages with `DETECTION_SENSOR_APP` portnum one-way? ie. receive only?

if so we can save them as `TEXT_MESSAGE_APP` [here](https://github.com/meshtastic/Meshtastic-Android/blob/1e03b0b5e8899af06c6cb82b5d64b6e3cd1730fa/app/src/main/java/com/geeksville/mesh/service/MeshService.kt#L555) and avoid the extra handling.

andrekir:
do we need to do the same for `RANGE_TEST_APP` ? would be a good fit for both.

thebentern:
> do we need to do the same for `RANGE_TEST_APP` ? would be a good fit for both.

Yes, but that one is trickier. We only have interest in those packets if the connected node has RangeTest module enabled. I may need some help hunting down how to do that.

PR #734, opened 2023-09-19
Add hop limits to gui

wrewolf:


garthvh:
What is the purpose of having the hop limit on the node? It is set per packet.

wrewolf:
> What is the purpose of having the hop limit on the node? It is set per packet.

Only showing in UI. It help to understand how far this node.

andrekir:
we don't know nodes hop_limit, packet hop_limit is different. this was discussed before: meshtastic/firmware#2003

there is packet hop history in 'Export rangetest.csv' or for in depth network survey/planning check: https://github.com/meshtastic/network-management-client

G2Dolphin:
> we don't know nodes hop_limit, packet hop_limit is different. this was discussed before: [meshtastic/firmware#2003](https://github.com/meshtastic/firmware/issues/2003)

It's not about nodes hop_limit setting, it's about TTL of packets we receive from this node. Of course it makes little sense in networks where nodes use different settings, but if they don't — you could easily guess this node's route without running traceroutes or basically anything that would make some load on the network.
Software the link you provided to may be great, but obviously usage is very different from having the value directly in the phone application.

osmanovv:
In a dense city wide mesh network it would be great to see the latest hop limit from the node, to know exactly which nodes you link direct. It also help to direct antennas. Also there is an agreement on hop_limit value in the network.

garthvh:
It really means nothing for end users, could be from a node 10 feet away that just sent it, or a message that came in over mqtt, or a lora connected node a hundred miles away.  You can trace route to see how stuff is routing, but the hop limit on a packet (which was then being saved incorrectly on the node in this case) is not useful data.

G2Dolphin:
> It really means nothing for end users
Does RSSI and SNR coming from MQTT-connected node makes sense? There's basically no such thing if packet came through Internet, yet I still see them and wonder. 
Are those metrics being almost the same in case where many nodes behind only one is very useful data? It simply doesn't show the truth. 
Are RSSI and SNR really mean anything for those you call "end users"? Or maybe Air Utilisation showing on countries where's no such thing at all?

There's enough people whom tiny four symbol length field would be useful, yet you still decide what's good by yourself.

kri164:
Can you create "stop closing the trace results dialog when tap on background" as separated PR?
I found it very usefull, as current UX behaviour sucks.

PR #830, opened 2024-02-01
Annotated debug panel of to/from fields with hex form

mcumings:
What do you think about doing something like this?

This PR dynamically annotates the text in the debug panel, appending the hex form of the from and to fields.  This makes it easier to associate the integer forms of the node IDs to the hex/string forms.

The textual portion of the annotation is applied as the data is being stored int he database.  There is then a detection and styling step applied at render-time.

Before:
before

After (Light):
annotation-light

After (Dark):
annotation-dark


app/src/main/java/com/geeksville/mesh/ui/DebugAdapter.kt
```diff
@@ -35,11 +37,63 @@ class DebugAdapter internal constructor(
...
+    /**
+     * Annotate the message with the from/to addresses in hex if they are present.
+     *
+     * This makes it easier to correlate messages with to/from in integer form and
+     * node IDs in hex form.
+     */
+    @OptIn(ExperimentalStdlibApi::class) // toHexString
+    private fun annotateMessage(current: MeshLog): CharSequence {
+        val meshPacket = current.meshPacket ?: return current.raw_message
+        val from = meshPacket.from
```
andrekir: looks like `from` is missing the same `toUInt()` conversion used below in `to`?

app/src/main/java/com/geeksville/mesh/ui/DebugAdapter.kt
```diff
@@ -35,11 +37,63 @@ class DebugAdapter internal constructor(
...
+        val to = meshPacket.to.let {
+            if (it == -1) UInt.MAX_VALUE else it.toUInt()
+        }
+        val fromString = "from: $from\n"
+        val fromHex = "(!${from.toHexString()})"
+        val toString = "to: $to\n"
+        val toHex = "(!${to.toHexString()})"
```
andrekir: we've been using `"!%08x".format(n)` to [convert from node number to node ID](https://github.com/meshtastic/Meshtastic-Android/blob/abb1eff4260220b48c4b7ecc2118a8fb6f606f83/app/src/main/java/com/geeksville/mesh/DataPacket.kt#L170).

so we could also use `"(!%08x)".format(meshPacket.to/from)` here unless `toHexString()` is preferred.

andrekir:
I like it! added a couple notes for review.

app/src/main/java/com/geeksville/mesh/ui/DebugAdapter.kt
mcumings: Ah, nice!  I'll take another stab at this when I get some time.  I'm also going to try to get the work to be done upon database ingress rather than at render time, or at least see if I can lighten the load a bit.

It may take me a few days but I'll get it done. :)

mcumings:
Okay, I took a different approach and I like this a LOT better now.

I had to add a color to the XML values.  I just picked something arbitrary.  I'm not sure if you have a style guide specification that we'd want to adhere to or if this is ok for now?

andrekir:
no worries about the color! 

but I'm afraid this breaks parsing the String back to protobuf ([MeshPacket](https://github.com/meshtastic/Meshtastic-Android/blob/d75188f03c80614e33762212fa4df3b10eddc6a6/app/src/main/java/com/geeksville/mesh/database/entity/MeshLog.kt#L24) and [NodeInfo](https://github.com/meshtastic/Meshtastic-Android/blob/d75188f03c80614e33762212fa4df3b10eddc6a6/app/src/main/java/com/geeksville/mesh/database/entity/MeshLog.kt#L37))?

mcumings:
Argh..  I must have missed that in my Find Usages.  Ok, I'll take care of that somehow as well...

mcumings:
Ok, that should do it.  Probably a little cleaner anyhow.

andrekir:
looks good, thanks!

Issue #778, opened 2023-11-11
[Feature request] link user from chatlist direct to node


universal7:
Could be useful if the username in the chat is directly linked to the node in nodelist.
example: I've read a message in the chat tap on the username an that node is opend in the nodelist. So i can directly see: node is online/offline, position and so the language i can answer or even to ignore.

PR #844, opened 2024-02-10
Feature: Jump to node info from message

daviscodesbugs:
I believe this fully addresses issue #778

From within a message, tapping on the node short name will navigate to the node list tab and scroll to that node. Added bonus: the node will flash a couple times to draw the users attention. Figured this is the best solution at the moment since there isn't a more in-depth node details view.

Also did a bit of refactoring around the tabs of the main menu to make simplify future changes to icons or strings in the main menu, as well as translations. Important to note that the order that the values of the MainTab enum are defined will be the order they appear in the UI.

https://github.com/meshtastic/Meshtastic-Android/assets/22112533/96410f68-a1dc-427d-8148-f21dcbe3656c


app/src/main/res/values/strings.xml
```diff
@@ -11,6 +11,13 @@
     <string name="sample_coords" translatable="false">55.332244 34.442211</string>
     <string name="sample_message" translatable="false">hey I found the cache, it is over here next to the big tiger. I\'m kinda scared.</string>
 
+    <!-- Main Tabs -->
+    <string name="main_tab_lbl_messages">Messages</string>
+    <string name="main_tab_lbl_users">Users</string>
+    <string name="main_tab_lbl_map">Map</string>
+    <string name="main_tab_lbl_channel">Channel</string>
+    <string name="main_tab_lbl_settings">Settings</string>
```
andrekir: can you add `translatable="false"` into these? I don't want to send them for translation just yet.

andrekir (changes requested):
great work! added a small note for review.

app/src/main/res/values/strings.xml
daviscodesbugs: Sure thing! Made those changes.
Is there some automated process to translate those? Just wondering why they need to be marked not translatable

andrekir: the plugin can be a pain, so just to be safe. but this is an exception, these names are ancient and you'll notice some variation over time.

PR #847, opened 2024-02-11
Select active device via automation Intent

kvj:
Add new broadcast receiver to change active node (or disconnect when address is empty).

This allows opens possibilities to automate node switching based on external events and automation platforms. 
For example I now can change active node from Home Assistant using following service call:
```yaml
service: notify.mobile_app_my_device
data:
  message: command_broadcast_intent
  data:
    intent_package_name: com.geeksville.mesh
    intent_action: com.geeksville.mesh.CHANGE_DEVICE
    intent_extras: com.geeksville.mesh.Address:t192.168.0.99
```

A call without `intent_extras` disconnects from active node


andrekir:
there's already an API for all service methods including `setDeviceAddress()`

https://github.com/meshtastic/Meshtastic-Android/blob/master/app/src/main/aidl/com/geeksville/mesh/IMeshService.aidl#L12-L41

kvj:
hmm, that's quite different from what was in the PR. In order to bind AIDL service, you actually need a custom application. But automation using Intents doesn't require it, as many automation app support this feature.

PR #850, opened 2024-02-13
Refactor and cleanup

daviscodesbugs:
Just some small changes and simplifications in preparation for moving the settings fragment to Compose. I'll probably be doing it in chunks, so that the PRs aren't so huge


andrekir:
looks good.

in SettingsFragment, there is no need to keep the firmware update logic and SoftwareUpdateService. those are outdated and unused.

daviscodesbugs:
> looks good.
> 
> in SettingsFragment, there is no need to keep the firmware update logic and SoftwareUpdateService. those are outdated and unused.

Okay, sweet! I'll see if I can't get those cleaned out as well

Issue #853, opened 2024-02-14
Remove firmware update logic and SoftwareUpdateService


daviscodesbugs:
As mentioned in #850 
The firmware update logic and SoftwareUpdateService is outdated and unused. It can be refactored out
Remove the "Update Firmware" button on the settings page

andrekir:
> Is the "Update Firmware" button on the settings page also okay to remove?

yes.

> Also, is there another mechanism for updating radio firmware from the app, or is it just no longer supported?

there is Bluetooth OTA support in the firmware but it was replaced in 2.0, so the app side will have to start from scratch.

PR #870, opened 2024-02-24
Remove unused firmware update UI and logic

daviscodesbugs:
This addresses #853 

Due to changes in the firmware for updates, the app needs to re-implement updates to conform to the new method. This is just cleanup prior to that effort.

Also included is an extension function to use non-deprecated method for checking IP address format


andrekir:
thanks!

Issue #817, opened 2024-01-19
Black cursor on node name field


crypto-smoke:
Hard to see in dark mode
![image](https://github.com/meshtastic/Meshtastic-Android/assets/86024507/14a8fbe8-fb8b-458b-a866-4dfdb2e34c50)

daviscodesbugs:
Yes, I mentioned something similar in #616 
I'll look at this soon

PR #871, opened 2024-02-25
Fix UI contrast and theming

### LINKED ISSUE #616: Buttons illegible in dark mode
In dark mode, white text on green buttons are illegible. 

Ver: FDroid-2.1.4
Device: Sony Xperia 5ii, Android 11

daviscodesbugs:
This should address #817 where the cursor and label were hard to read. This happens because of how the primary color was being used. In dark mode it was set to the black, and then used like in the light theme. This creates some difficulties with the default coloring of most components, so this PR sets the primary back to the Meshtastic green and then rectifies the contrast issues in other spots where custom theming is already being used.

Here is that input field after the changes

image

This should also close #616

image

I did my best to ensure that all other parts of the app looked properly colored with these changes (both in dark and light mode) but would love some additional testing/verification before merging


andrekir:
nice work

PR #874, opened 2024-02-26
Improved battery/voltage info in node list

### LINKED ISSUE #826: [Feature Request] - More info on node page
I posted this on discord and it seemed to have positive feedback. My suggestion is to add a single extra line to the node page for each individual node so the following is shown.

BAT% BATVoltage
Last Heard
LAT LON ALT #SATS
RSSI SNR Channel (and/or freq.)
TEMP HUM PRESS AIR-RESISTANCE (future sensors)

daviscodesbugs:
Doesn't completely resolve #826, but addresses the voltage info complaint from @HikerScott 

Primarily, moves the battery info into a composable and uses that in the adapter
I added in a couple more battery icons to get a quick idea of the level. I set the ranges arbitrarily, so we could tune these later to match user expectation better
I also cleaned up the adapter item layout for nodes to avoid manually declaring margins on everything. Additionally, moved the sample text to a tools attribute so it still shows up in the preview, without actually setting the fields in the app

image

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/BatteryInfo.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/BatteryInfo.kt
@@ -1,11 +1,8 @@
 package com.geeksville.mesh.ui
 
 import android.content.res.Configuration
-import androidx.compose.foundation.layout.Arrangement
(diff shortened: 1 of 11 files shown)


andrekir:
thanks

PR #878, opened 2024-02-27
Move last heard in node info to compose

daviscodesbugs:
Depends on #877


andrekir:
@daviscodesbugs a few conflicts came up after merging #877, can you have a look?

daviscodesbugs:
> @daviscodesbugs a few conflicts came up after merging #877, can you have a look?

Okay, I think that should do it

PR #879, opened 2024-02-27
Move signal info to compose

daviscodesbugs:
Depends on #878


andrekir:
@daviscodesbugs a few conflicts here as well, can you check?

daviscodesbugs:
> @daviscodesbugs a few conflicts here as well, can you check?

That should be cleared up as well. Sorry; I must have mangled my branches a bit 🙃

PR #883, opened 2024-02-28
Add number of GPS satellites in view to DB for later display

daviscodesbugs:
First steps in fulfilling #826 

The app currently doesn't hang on to the number of satellites info that is on the position message. This adds that field to the DB so we can later display it in the node info

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
+++ b/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
@@ -68,7 +68,10 @@ data class Position(
     val longitude: Double,
     val altitude: Int,
     val time: Int = currentTime(), // default to current time in secs (NOT MILLISECONDS!)
-    val satellitesInView: Int = 0
+    val satellitesInView: Int = 0,
+    val groundSpeed: Int = 0,
+    val groundTrack: Int = 0, // "heading"
+    val precisionBits: Int = 0,
 ) : Parcelable {
 
     companion object {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/geeksville/mesh/database/MeshtasticDatabase.kt
```diff
@@ -27,7 +29,7 @@ import com.geeksville.mesh.database.entity.QuickChatAction
     autoMigrations = [
         AutoMigration (from = 3, to = 4),
```
andrekir, line 28: here you can simply add "_AutoMigration (from = 4, to = 5),_" instead of using the Migration object.

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -67,8 +67,10 @@ data class Position(
     val latitude: Double,
     val longitude: Double,
     val altitude: Int,
-    val time: Int = currentTime() // default to current time in secs (NOT MILLISECONDS!)
+    val time: Int = currentTime(), // default to current time in secs (NOT MILLISECONDS!)
+    val satellitesInView: Int = 0
```
andrekir: can you also add the fields below?

```kotlin
    val groundSpeed: Int = 0,
    val groundTrack: Int = 0, // "heading"
    val precisionBits: Int = 0,
```

andrekir (changes requested):

app/src/main/java/com/geeksville/mesh/database/MeshtasticDatabase.kt
daviscodesbugs: I tried that originally, but it didn't seem to work. I was thinking it didn't support adding columns, but maybe I did something wrong

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
daviscodesbugs: Yeah, totally. I'll add that

app/src/main/java/com/geeksville/mesh/database/MeshtasticDatabase.kt
daviscodesbugs: Now that I think about it, I might not have had the new schema file added yet.. I'll try it again

davis-gotenna:
@andrekir Yup, turns out it was a caching issue. Automigration works great. Also added those additional fields

andrekir (approved):

PR #886, opened 2024-02-29
Move Node info completely to Compose

daviscodesbugs:
Node info is now completely in Compose

Moved over to use constraints for more dynamic layout
Added a string composing method for environment metrics that doesn't have the space baked in to each value.
Note: the feature that navigates from a chat to the node list and highlights the node will be broken. The navigation aspect works, but the blink does not. I'll have to follow up on that

- Layout adjust to accommodate wide coordinate formats
- Intelligently update node list only where information changed

Coordinates are aligned to top of signal info, but wrapped with last heard. This leaves some blank space to the side, but isn't too significant considering this should only happen on node long names that wrap
image

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
+++ b/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
@@ -170,6 +170,32 @@ data class EnvironmentMetrics(
     override fun toString(): String {
         return "EnvironmentMetrics(time=${time}, temperature=${temperature}, humidity=${relativeHumidity}, pressure=${barometricPressure}), resistance=${gasResistance}, voltage=${voltage}, current=${current}"
     }
+
+    fun getDisplayString(inFahrenheit: Boolean = false): String {
+        val temp = if (temperature != 0f) {
+            if (inFahrenheit) {
+                val fahrenheit = temperature * 1.8F + 32
+                String.format("%.1f°F", fahrenheit)
+            } else {
+                String.format("%.1f°C", temperature)
+            }
+        } else null
+        val humidity = if (relativeHumidity != 0f) String.format("%.0f%%", relativeHumidity) else null
+        val pressure = if (barometricPressure != 0f) String.format("%.1fhPa", barometricPressure) else null
+        val gas = if (gasResistance != 0f) String.format("%.0fMΩ", gasResistance) else null
+        val voltage = if (voltage != 0f) String.format("%.2fV", voltage) else null
+        val current = if (current != 0f) String.format("%.0fmA", current) else null
+
+        return listOfNotNull(
+            temp,
+            humidity,
+            pressure,
+            gas,
+            voltage,
+            current
+        ).joinToString(" ")
+    }
+
 }
 
 @Parcelize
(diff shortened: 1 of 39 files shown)


app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
```diff
@@ -0,0 +1,238 @@
...
+    thisNodeInfo: NodeInfo,
+    thatNodeInfo: NodeInfo,
+    gpsFormat: Int,
+    distanceUnits: Int,
+    tempInFahrenheit: Boolean,
+    isIgnored: Boolean = false,
+    onClicked: () -> Unit = {}
+) {
+    val unknownShortName = stringResource(id = R.string.unknown_node_short_name)
+    val unknownLongName = stringResource(id = R.string.unknown_node_long_name)
```
andrekir: there is a resource string for "unknown name" --> _R.string.unknown_username_

app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
```diff
@@ -0,0 +1,238 @@
...
+
+    val nodeName = thatNodeInfo.user?.longName ?: unknownLongName
```
andrekir, line 51: also suggest using `getInitials(nodeName)` instead of a resource string for unknownShortName:

```kotlin
val unknownShortName = getInitials(nodeName)
```
or directly:

```kotlin
val shortName = thatNodeInfo.user?.shortName ?: getInitials(nodeName)
```

(then using `shortName` in line 90 below)

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -179,6 +179,32 @@ data class EnvironmentMetrics(
...
+
+        return listOfNotNull(
+            temp,
+            humidity,
+            pressure,
+            gas,
+            voltage,
+            current
+        ).joinToString(" ")
+    }
```
andrekir, line 206: can we remove `envMetricStr` and `envFormat` after this change?

andrekir (changes requested):
super! added some small notes for review.

app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
daviscodesbugs: Yes! I did see that, but that value is "Unknown Username" and the one I added is just "Unknown" (like in the examples above) Do we want it to say "Unknown Username"?

app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
daviscodesbugs: Oh cool! Didn't know about this method. I'll mention you in another comment on this

app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
andrekir: yeah, "Unknown Username" / _R.string.unknown_username_ is what is use elsewhere specific for unknown node longNames.

(you'll notice "node" and "user" are often used interchangeably)

app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
andrekir: just checking in for an update on this one. if you add `unknown_node_short_name` as "???", please tag it translatable=false.

andrekir (approved):

PR #895, opened 2024-03-04
Add elevation and number of GPS satellites to node info

daviscodesbugs:
Depends on #886 

Displays number of satellites and elevation above sea level

image


andrekir:
looks good! thanks!

Issue #683, opened 2023-08-13
[Feature Request] Add User Search in Units Tab


tttmmmsss:
Add Search in Units Tab to search for specific users, querying short and long name.  Possibly with live filtering for each additional character added in the search string.

PR #920, opened 2024-03-19
Feature: Filter node list

daviscodesbugs:
Adds a text field to the node list that filters against node short name and long name. Filter updates with each character

Fulfills #683

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/model/UIState.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/UIState.kt
@@ -141,6 +141,18 @@ class UIViewModel @Inject constructor(
     private val _focusedNode = MutableStateFlow<NodeInfo?>(null)
     val focusedNode: StateFlow<NodeInfo?> = _focusedNode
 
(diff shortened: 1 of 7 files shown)


prokrypt:
would it be easy to have it search nodeid too, for times when we need to look up a mystery node?

andrekir:
looks good!

since the filter can remove `node[0]`, one small change is needed at: https://github.com/meshtastic/Meshtastic-Android/blob/44aa28554bce0796574c7af94944752b0ead00a5/app/src/main/java/com/geeksville/mesh/ui/UsersFragment.kt#L182-L183

 we can use `model.ourNodeInfo.value` here and update `thisNodeInfo` param type to `NodeInfo?`.

davis-gotenna:
@andrekir 
Finally getting back to this; should be good now!

daviscodesbugs:
@andrekir
Finally getting back to this; should be good now!

andrekir:
looks great, thanks!

PR #923, opened 2024-03-20
Improvement: Increase max zoom level for USGS tile sources

danwelch3:
The default osmdroid USGS Topo and Sat tile source definitions limit the max zoom to 15.  When using the app, this makes it tough to differentiate node positions when they are close together.  This PR defines these sources and changes the max zoom to 20, to match the other tile sources defined in CustomTileSource.kt


andrekir:
tile source definitions are not arbitrary. USGS doesn't support max zoom > 15, increasing this value will cause errors downloading tiles.

danwelch3:
Interesting.  I'd still like to work out a way to zoom further on these maps.  I will open an issue to discuss other ways to accomplish this.

PR #939, opened 2024-03-28
Fixed 59 minute issue

RCGV1:
Fixed the issue where 59 minutes would show up as "0 hours"
Ignore my misnamed commit message, lol.
image


andrekir:
good catch!

Issue #909, opened 2024-03-13
Long-press of node pin on map should jump to nodelist


prokrypt:
Instead of jumping to direct message, I think highlighting the node in the nodelist would be more useful, and we can then initiate a direct message from there after reviewing the node stats.

PR #955, opened 2024-04-05
Long press node in map jumps to node in node list

daviscodesbugs:
Previously, long pressing a node in the map would jump to a DM with that node. Now, it will jump to that node in the node list.

Fulfills #909


madeofstown:
Could we also change it so that clicking on GPS coordinates in the node list take you to the map in the app instead of Google maps or whatever is set to default?

Issue #922, opened 2024-03-19
Hops Away


asjmcguire:
The Android app currently lacks the display of how many hops away nodes are.
The thing that would would make the most sense is to only display RSSI and SNR when hops away is zero, or missing. Because obviously since when it is 1 or more hops away, then the signal strength no longer makes sense, people think they have an excellent connection to a node, but really they only have an excellent connection to the 1st hop, the connection between the 1st hop and the destination node, or just the next hop, might be terrible, but that information is never communicated (or available) to the sender.

It makes more sense to show the number of hops in place of the RSSI and SNR when the hop is higher than zero.

prokrypt:
I think the RSSI and SNR readings in those cases still can be useful--they just need to be somehow linked to the neighbor that repeated the packet.

asjmcguire:
> I think the RSSI and SNR readings in those cases still can be useful--they just need to be somehow linked to the neighbor that repeated the packet.

Which if I understand correctly from the firmware channel on discord, is not information that is available. The only node information is the original sender, nothing from any node the packet has passed through. Maybe V3 will change this, but there is nothing in V2.

francisuk1989:
So many features we are missing out compared the iOS version
- Emojis
- root topic recommends
- Sending a GPS location

and more

![image](https://github.com/meshtastic/Meshtastic-Android/assets/11011688/059eb3d0-5486-4701-8f04-b09f73220f91)

DayleDrinkwater:
After fiddling around for the past few days, I've finally worked out how to get the data. I'll tidy everything up and make UI elements for it. 
![image](https://github.com/meshtastic/Meshtastic-Android/assets/7864598/f6c4c57e-ff96-4b63-89e9-1ece7c1ff3fe)

joneyes:
Good result.
One of the other devs has also added the info to the python cli --nodes table

PR #966, opened 2024-04-11
Hops Away implementation

DayleDrinkwater:
A basic implementation of Hops Away, closing #922 

Modifies SignalInfo, if hopsAway is more than 0 it replaces the RSSI with the number of hops.

hops

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -5,7 +5,19 @@
       <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
     </inspection_tool>
+    <inspection_tool class="PreviewApiLevelMustBeValid" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
+    </inspection_tool>
     <inspection_tool class="PreviewDimensionRespectsLimit" enabled="true" level="WARNING" enabled_by_default="true">
+      <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
+    </inspection_tool>
+    <inspection_tool class="PreviewFontScaleMustBeGreaterThanZero" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
+    </inspection_tool>
+    <inspection_tool class="PreviewMultipleParameterProviders" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
     </inspection_tool>
@@ -14,6 +26,14 @@
       <option name="previewFile" value="true" />
     </inspection_tool>
     <inspection_tool class="PreviewNeedsComposableAnnotation" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
+    </inspection_tool>
+    <inspection_tool class="PreviewNotSupportedInUnitTestFiles" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
+    </inspection_tool>
+    <inspection_tool class="PreviewPickerAnnotation" enabled="true" level="ERROR" enabled_by_default="true">
       <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
     </inspection_tool>
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -2,39 +2,15 @@
   <profile version="1.0">
     <option name="myName" value="Project Default" />
     <inspection_tool class="PreviewAnnotationInFunctionWithParameters" enabled="true" level="ERROR" enabled_by_default="true">
-      <option name="composableFile" value="true" />
-      <option name="previewFile" value="true" />
-    </inspection_tool>
-    <inspection_tool class="PreviewApiLevelMustBeValid" enabled="true" level="ERROR" enabled_by_default="true">
-      <option name="composableFile" value="true" />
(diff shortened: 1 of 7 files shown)


prokrypt:
Nice! But it leaves an ambiguity: a node with RSSI/SNR displayed is either a neighbor or a packet from an older firmware node/repeater/router. Perhaps try to a way to differentiate between the two...

DayleDrinkwater:
Hi @prokrypt , good point. I've now made some changes. It now marks RSSI on nodes with a (!) if `hopStart` is 0, but `hopLimit `isn't. this will differentiate between nodes which are direct (no (!)) and nodes which are running old FW (not sending `hopStart`).

by default it marks all nodes with (!) until it knows for certain, at which point it will remove the ! if the are direct, or show Hops Away. if they are running old FW or it doesn't know, it'll continue to mark them with a (!)
hops2

prokrypt (commented):
Possible lifehack:
`if (packet.hopStart < packet.hopLimit) ...` (just a random musing. totally untested!)

DayleDrinkwater:
I did think that after I commited, I'll try it later on

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -219,6 +220,8 @@ data class NodeInfo(
     var snr: Float = Float.MAX_VALUE,
     var rssi: Int = Int.MAX_VALUE,
     var lastHeard: Int = 0, // the last time we've seen this node in secs since 1970
+    @ColumnInfo(name="hopsAway", defaultValue = "-1")
+    var hopsAway: Int = -1,
```
andrekir: 1. this need to be added at the end of the object, or it creates a breaking change for the API
2. there is no need to add **@ColumnInfo**

.idea/inspectionProfiles/Project_Default.xml
```diff
@@ -2,15 +2,39 @@
...
+    </inspection_tool>
+    <inspection_tool class="PreviewNotSupportedInUnitTestFiles" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
+      <option name="previewFile" value="true" />
+    </inspection_tool>
+    <inspection_tool class="PreviewPickerAnnotation" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
     </inspection_tool>
   </profile>
```
andrekir: please do not check in files that are unrelated to this pull request

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -1020,11 +1020,20 @@ class MeshService : Service(), Logging {
                 it.snr = packet.rxSnr
                 it.rssi = packet.rxRssi
             }
-
+            //generate our own hopsAway, comparing hopStart to hopLimit.
+            when {
```
andrekir: setting `hopsAway` needs to move inside the `updateNodeInfo()` block [above](https://github.com/meshtastic/Meshtastic-Android/blob/b69466f669ce360577267e9adc41a970abae0421/app/src/main/java/com/geeksville/mesh/service/MeshService.kt#L1014). extra calls will trigger unnecessary node change broadcasts for every packet received.

this should match how `hopsAway` is calculated in the firmware.

andrekir (changes requested):

.idea/inspectionProfiles/Project_Default.xml
DayleDrinkwater: Apologies, I must've missed that one, I've reverted it now

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
DayleDrinkwater: I've now moved `hopsAway `in to the `updateNodeInfo()` block as required.

Regarding `hopsAway `calculation, this essentially does follow the Firmware implementation, it checks if `hopStart `is not 0, then `hopStart `- `hopLimit`.

The one thing it does extra from the firmware is it checks if `hopStart` is less than `hopLimit`, if it is, set `hopsAway `to -1. It then uses this to append `(!)` to the RSSI data, to signify that it's probably from a device with old Firmware. 

The firmware version of this implementation does not differentiate between a packet which is received direct, and a packet from a node with old firmware (both return 0). This can be confusing to the user, and may make them think all nodes are local to them.

Is this OK? If not I have no problem removing it.

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
DayleDrinkwater: I've now moved hopsAway to the end.

I'm having issues removing `@ColumnInfo` though, when I remove that line it will not build, I receive the error `"New NOT NULL column'hopsAway' added with no default value specified. Please specify the default value using @ColumnInfo."`. I'm not exactly sure why it is required compared to other columns, but looking online it's related to versioning of the Room database.

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
andrekir: keep it simple and align with the code in the firmware. `hopsAway` from the radio (removed in [`b69466f`](https://github.com/meshtastic/Meshtastic-Android/pull/966/commits/b69466f669ce360577267e9adc41a970abae0421#diff-3a6bc031c4bd5b2872f201d2b4b98ed404f169f548d64821749bd6144efaf854L1322)) must be used and any extra handling must be compatible.

keep in mind all this added complexity will be unnecessary in a few weeks once everyone upgrades to 2.3.0 or later. if there is room for improvement here or on the firmware implementation, address them in separate PRs.

DayleDrinkwater: no problem, I've now removed the extras and `hopsAway` now follows the firmware implementation.

andrekir: thanks

andrekir (approved):

PR #970, opened 2024-04-16
Show nodes we have not yet received a NodeInfo from.

### LINKED ISSUE #912: [Improvement] Show nodes that have been seen but for which there is no NodeInfo
If I've gotten an ack or other packet from a node but haven't yet gotten a NodeInfo packet, it seems that it won't show at all in the tab. This matches what the CLI does (though I'll be putting in a feature request there too), but does not match the web client (which shows "UNK: <integer node id>") nor iOS (which shows "Meshtastic XXXX"). I think it's valuable to see those nodes in the list so it's less mysterious who's acking things or to identify nodes that you have a poor connection to (hence no NodeInfo) but see sometimes, as those nodes are the exact sort that smaller improvements like mast height or antenna upgrades can help be seen more reliably.

armooo:
We will receive random packets from nodes with a marginal connection but often will never get a NodeInfo. The connection notification will then show more nodes online that are visible in the node list. This fixes #912.

I decided to give the unknown nodes the name "UNK: !<hex id>". This matches what was already being done in the chat view where only the hex id is used. It was also somewhat simpler because the `to` and `from` fields on `DataPacket` already had this format.

Screenshot_20240415-200635
Screenshot_20240415-203556
Screenshot_20240415-203937
Screenshot_20240415-204009
Screenshot_20240415-223109
Screenshot_20240415-223123

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
+++ b/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
@@ -2,10 +2,11 @@ package com.geeksville.mesh
 
 import android.graphics.Color
 import android.os.Parcelable
-import androidx.room.ColumnInfo
+import androidx.annotation.NonNull
 import androidx.room.Embedded
 import androidx.room.Entity
 import androidx.room.PrimaryKey
+import androidx.room.ColumnInfo
 import com.geeksville.mesh.util.GPSFormat
 import com.geeksville.mesh.util.bearing
 import com.geeksville.mesh.util.latLongToMeter
@@ -213,8 +214,9 @@ data class EnvironmentMetrics(
 data class NodeInfo(
     @PrimaryKey(autoGenerate = false)
     val num: Int, // This is immutable, and used as a key
+    @NonNull()
     @Embedded(prefix = "user_")
-    var user: MeshUser? = null,
+    var user: MeshUser,
     @Embedded(prefix = "position_")
     var position: Position? = null,
     var snr: Float = Float.MAX_VALUE,
--- a/app/src/main/java/com/geeksville/mesh/model/NodeDB.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/NodeDB.kt
@@ -1,10 +1,15 @@
 package com.geeksville.mesh.model
 
+import android.app.Application
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.coroutineScope
+import com.geeksville.mesh.R
 import com.geeksville.mesh.MyNodeInfo
 import com.geeksville.mesh.NodeInfo
+import com.geeksville.mesh.MeshUser
+import com.geeksville.mesh.DataPacket
 import com.geeksville.mesh.database.dao.NodeInfoDao
+import com.geeksville.mesh.MeshProtos.HardwareModel
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
@@ -18,6 +23,7 @@ import javax.inject.Singleton
 @Singleton
 class NodeDB @Inject constructor(
     processLifecycle: Lifecycle,
+    private val app: Application,
     private val nodeInfoDao: NodeInfoDao,
 ) {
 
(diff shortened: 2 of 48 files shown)


prokrypt:
Nice! Does this also make position packets from those nodes appear on the map?

armooo:
I expect it should. The map and the contacts list both use `model.nodeDB.nodes` which is populated now due to calling `installNodeDB` and `upsert` with userless nodes in `MeshService`. I will see if I can get my node to broadcast a location before I get a NodeInfo from it to confirm.

armooo:
Good call on testing the map view, there was a `node.user!!` hiding in it.

prokrypt:
Possible bug somewhere: only one UNK shows on the map (possibly the latest one). Or maybe it's something I did. hmm.

andrekir (changes requested):
your approach looks good overall, but instead of doing work on the UI side, consider adding a default `user` when a new `NodeInfo` DB entry is created in `MeshService`:

```kotlin
    // given a nodeNum, return a db entry - creating if necessary
    private fun getOrCreateNodeInfo(n: Int): NodeInfo = nodeDBbyNodeNum.getOrPut(n) {
        val defaultUser = MeshUser(
            id = ...,
            longName = ...,
            shortName = getString(R.string.unknown_node_short_name),
            hwModel = MeshProtos.HardwareModel.UNSET,
        )
        NodeInfo(n, defaultUser)
    }
```

also `unknown_username_node_id` seems a bit redundant as the unknown shortname/longname res strings can be used.

armooo:
Sure I can give synthesizing a `MeshUser` a shot and see how that works out.

> also unknown_username_node_id seems a bit redundant as the unknown shortname/longname res strings can be used.

I was thinking we would want to keep the string concatenation in the template for RTL languages if the translator wanted to localize and UNK and move in the string.

nagumii:
Thanks for thinking of RTL languages.

armooo:
> your approach looks good overall, but instead of doing work on the UI side, consider adding a default user when a new NodeInfo DB entry is created in MeshService.

This mostly worked but in cases like a resetting the NodeDB you can still have messages referencing nodes which will not exist. I added an infallible getNode method to the model.NodeDB as well which will make a NodeInfo if once can not be found.

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -213,8 +214,9 @@ data class EnvironmentMetrics(
 data class NodeInfo(
     @PrimaryKey(autoGenerate = false)
     val num: Int, // This is immutable, and used as a key
+    @NonNull()
     @Embedded(prefix = "user_")
-    var user: MeshUser? = null,
+    var user: MeshUser,
```
andrekir, line 219: In Android, when working with AIDL and Parcelable objects, it is a common standard to use nullable types for properties, especially if the underlying data can be optional or can change over time. Most properties in `NodeInfo` are nullable and its more consistent to keep `user` nullable as well.

This would also be a breaking change for our API.

app/src/main/java/com/geeksville/mesh/model/NodeDB.kt
```diff
@@ -60,6 +66,22 @@ class NodeDB @Inject constructor(
...
+                DataPacket.idToDefaultNodeNum(num),
+                MeshUser(
+                    id = num,
+                    longName = app.getString(R.string.unknown_username, num),
+                    shortName = app.getString(R.string.unknown_node_short_name),
+                    hwModel = HardwareModel.UNSET,
+                )
+            );
+        }
+    }
```
andrekir, line 84: this should be avoided. overriding database results can lead to consistency issues.

also in Kotlin semicolons are not used.

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -327,10 +326,7 @@ class MeshService : Service(), Logging {
 
         // put our node array into our two different map representations
         nodeDBbyNodeNum.putAll(nodes.map { it.num to it })
-        nodeDBbyID.putAll(nodes.mapNotNull {
-            // ignore records that don't have a valid user
-            it.user?.let { user -> user.id to it }
-        })
+        nodeDBbyID.putAll(nodes.map { it.user.id to it })
```
andrekir, line 329: this shouldn't be necessary?

app/src/main/res/values/strings.xml
```diff
@@ -11,7 +11,7 @@
     <string name="sample_coords" translatable="false">55.332244 34.442211</string>
     <string name="sample_message" translatable="false">hey I found the cache, it is over here next to the big tiger. I\'m kinda scared.</string>
 
-    <string name="unknown_node_short_name" translatable="false">\???</string>
+    <string name="unknown_node_short_name">UNK</string>
```
andrekir, line 14: the short name has a limit of 4 bytes. that is why it is using `?` and is non-translatable.

app/src/main/res/values/strings.xml
```diff
@@ -23,7 +23,7 @@
     <string name="unset">Unset</string>
     <string name="connection_status">Connection status</string>
     <string name="application_icon">application icon</string>
-    <string name="unknown_username">Unknown Username</string>
+    <string name="unknown_username">UNK: %1$s</string>
```
andrekir, line 26: because we use [Crowdin](https://crowdin.meshtastic.org/android) for translations, source strings can't be changed or replaced like this. instead, add a new string.

there is also no need to change the translated strings (ie. `values-fr/strings.xml`, etc)

andrekir (changes requested):

PR #993, opened 2024-04-24
chore: Add the difference between Traditional Chinese and Simplified Chinese

wayne2002tw:


andrekir:
translations are done using Crowdin, please submit your changes at https://crowdin.meshtastic.org/

Issue #635, opened 2023-05-16
feature request: let us select which channels to generate URL/QR for


prokrypt:
maybe have primary channel be required, but let us choose which secondary channels to share in case we don't want someone in our admin channel...

Ferretdog:
This, but also the ability to import just that channel - and not to replace your entire channel set with the one you are importing. 

I want to share a channel with friends without them having to manually add it (requiring me to send the encryption key using a clunky cut and paste), or remove all their own channels to accept it. It would ideally check that the modem preset and channel were the same before importing.

SenorFusion:
> This, but also the ability to import just that channel - and not to replace your entire channel set with the one you are importing.

This is the really important part.

Now that this is available on iOS and there is a working method, would it be easier to implement for Android?

This is a real hold-up for on-boarding other families and groups who are not tech savvy enough to use the CLI or web flasher for channel management

Issue #953, opened 2024-04-05
Honor add=true query string value when scanning QR codes


garthvh:
When scanning a QR code honor the new add=true query string value and append the channels to the list if the query string value is present.

PR #1038, opened 2024-05-14
Qr channel select

Robert-0410:
#### Addressing Issue #635
These changes allow the user to select what channels they want included in the URL/QR.


Robert-0410:
> Does this PR do anything for #953 or is it only for selecting which channels to share?
I'm not aware of what "add=true" is. Is it something meant to help with the feature?

Robert-0410:
> > > Does this PR do anything for #953 or is it only for selecting which channels to share?
> > 
> > 
> > I'm not aware of what "add=true" is. Is it something meant to help with the feature?
> 
> It would be so when you add selected channels from a QR code, they’ll be added versus overwriting all channels. It was recently added in the Apple apps: [meshtastic/Meshtastic-Apple@2a38dbb](https://github.com/meshtastic/Meshtastic-Apple/commit/2a38dbbb1cfab3b25e9bd1ae9b6b62c10ed5d0bd)

Understood, These commits don't provide said feature.
But I wouldn't mind working on it after this.

andrekir (changes requested):
while the effort is appreciated, this implementation is needlessly complex. the changes to `Channel`, `ChannelSet`, and viewmodel are unnecessary. instead, simply filter the checked channels to build a `ChannelSet` and generate the QR code/URL from that.

also consider using a separate UI screen or modal for channel selection and QR code/URL generation, for example, similar to how it’s done in the web app:

ezgif-1-bcf3779d20

Robert-0410:
Thanks for the feedback at @andrekir , and I appreciate the hard work that has gone into the app.

Ill take another crack at it 🏗️

Robert-0410:
Took a fresh approach with #1051

PR #1051, opened 2024-05-22
Qr channel select att2

Robert-0410:
#### Addressing Issue #635
These changes allow the user to select what channels they want included in the URL/QR.
- Made the QR/URL generation UI similar to the web app.


andrekir:
sorry for the late response. I'm working on this area and made a few changes to the code.

I like the new channel selection, but it just looks confusing with everything in the same screen.
- primary channel is shown twice
- strings are out of context (title and description work for a separate screen/modal)

we can show only the selected channels (removing what is above it) and figure out a way to switch between "channel selection & qr code" from the "channel editor" before we can merge this.

Screenshot_20240630_094900_Meshtastic

PR #1065, opened 2024-06-02
feat: add nodelist sort options

andrekir:
- adds nodelist sort options;
- moves nodelist filter into database query;
- adds option to include unknown names.


prokrypt:
Where is Hops 😆

Issue #1042, opened 2024-05-15
[Feature Request]: Clear display of IQA.


memo-567:
It would be nice if there was also an understandable display of the IQA for the Android app.
Currently, the value is displayed in `MΩ` only.

This is already available for the iPhone app: https://github.com/meshtastic/Meshtastic-Apple/commit/52dd086e09b7b7e6d1e5e693e90e6db38fe98501

At the moment it looks like:

![20240515-1030](https://github.com/meshtastic/Meshtastic-Android/assets/58190287/c848b981-d4f6-430f-9b02-78990fe5306f)

grenius:
That's power telemetry isn't it? Though i agree a weather telemetry page like iOS would be nice.

memo-567:
> That's power telemetry isn't it? 

This value the BME680 sends as IAQ (Indoor Air Quality) https://en.wikipedia.org/wiki/Indoor_air_quality

grenius:
bme680 only has pressure/temp/humidity.

memo-567:
Nope: https://www.bosch-sensortec.com/products/environmental-sensors/gas-sensors/bme680/

grenius:
Oh my bad. Must have been thinking of the 280. Either way i'm onboard with what you're asking.

grenius:
FYI:

MΩ
A unit of resistance in electricity that is equal to one million ohms, also known as a meg

memo-567:
Thanks for the tip, but I'm very familiar with the basics of electrical engineering 😉

andrekir:
IAQ values were added in 2.3.12

PR #1078, opened 2024-06-08
feat: support for switching between devices

andrekir:
This update separates the messages database using `myNodeNum` as an index key to support switching between devices.

- New messages will be stored separately per node.
- Old messages (received before this feature) will remain visible for any connected node. Users can manually delete them to ensure only messages for the connected node are displayed.


caveman99:
this also fixes https://github.com/meshtastic/Meshtastic-Android/issues/1042

Issue #1086, opened 2024-06-11
Android/Apple Client Disparity, Cannot Share PSKs of 192bits


AddisonTustin:
## Summary
It does not seem possible to share PSKs between Apple/Android clients when the PSK is created/generated by the Apple client with a key size of 192 bits.

## Steps to reproduce
- In Apple client, create a new channel with an auto-generated key of 192bits.
- Share that PSK to an Android client/user, and have them attempt to create a channel by pasting the PSK. 

There's no warnings or error messages for why this is an invalid channel key for Android, while it's supported on the Apple client.

## Apple Image
![apple-meshtastic-192-bits](https://github.com/meshtastic/Meshtastic-Android/assets/1168530/beea486e-f8d3-4533-b3cd-9d0f63bbab41)
## Android Image
![android-meshtastic-192-bits](https://github.com/meshtastic/Meshtastic-Android/assets/1168530/d83d60da-8d9c-4d19-96e9-f00eccc8435f)

AddisonTustin:
Trying to spin-up on some of the Android package, but is there any reason we shouldn't be able to accept an additional [key size here](https://github.com/meshtastic/Meshtastic-Android/blob/5b3c78316b8accc789471f7ebdcd36e295c76682/app/src/main/java/com/geeksville/mesh/ui/components/config/EditChannelDialog.kt#L102) of 24 bytes?

andrekir:
firmware doesn't support 192 bit keys

https://github.com/meshtastic/protobufs/blob/master/meshtastic/channel.proto#L36
https://github.com/meshtastic/firmware/blob/master/src/mesh/CryptoEngine.h#L38

AddisonTustin:
I'd take a look at the [comment @ianmcorvidae](https://github.com/meshtastic/Meshtastic-Android/pull/1087#issuecomment-2162340971) added to #1087. 

The firmware pads the key to be 256-bits and the [Meshtastic-Apple client](https://github.com/meshtastic/Meshtastic-Apple/blob/a4e26de6e5423ddeb70d40388d80690006f21bb1/Meshtastic/Views/Settings/Channels/ChannelForm.swift#L60) 'accepts' a 192-bit key.

PR #1087, opened 2024-06-12
fix: allow 192-bit PSKs

AddisonTustin:
Adds support for 192-bit PSKs. Solves #1086 

Android client currently considers 192-bit PSKs invalid, which causes usability issues when trying to share some PSKs generated on iOS clients (or any 192-bit PSK).

I've validated that a 192-bit PSK generated with an iOS client can be copy/pasted when setting up a new channel on a node using the Android client, and can successfully communicate.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/components/config/EditChannelDialog.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/components/config/EditChannelDialog.kt
@@ -46,6 +46,9 @@ import com.google.protobuf.ByteString
 import com.google.protobuf.kotlin.toByteString
 import java.security.SecureRandom
 
+// Allows 0, 128, 192, and 256-bit PSKs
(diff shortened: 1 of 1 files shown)


ianmcorvidae:
Just as a point of information, I believe the firmware doesn't directly support aes192 and just pads the key out to aes256 with zeros: https://github.com/meshtastic/firmware/blob/master/src/mesh/Channels.cpp#L159 -- I guess it's not strictly invalid and it should be possible to share between clients, of course

AddisonTustin:
That makes sense. 

Yeah, actually running a test w/ two Heltec V3 dev-boards was successful when sharing a 192-bit psk.

andrekir:
confirmed only 128 or 256 bit keys are valid. looks like the crypto engine can use 192 bit keys, if it gets added we can revisit this. closing for now.

Issue #1109, opened 2024-06-26
Add interface for BLE 'logradio' logging 


thebentern:
If you set `security.debug_log_api_enabled` to true, you will receive `LogRecord`s over the API (BLE, TCP, or Serial). I believe this should come up over the
LOG_RADIO characteristic UUID: `5a3d6e49-06e6-4423-9944-e9de8cdf9547`.

I've only implemented this on Apple clients so far:
![image](https://github.com/meshtastic/Meshtastic-Android/assets/9000580/241e9c3c-6188-41b0-a94e-a785960459a2)

Example code on iOS:
https://github.com/meshtastic/Meshtastic-Apple/pull/749/files

I think it would be a neat feature to expose on Android as well, but I'm not sure the most idiomatic way of representing the data. Definitely interested in hearing what UI could work best from the Android platform experts.

jamesarich:
@thebentern #3691 is the implementation for bluetooth -  does this also work over serial/tcp?

thebentern:
> [@thebentern](https://github.com/thebentern) [#3691](https://github.com/meshtastic/Meshtastic-Android/pull/3691) is the implementation for bluetooth - does this also work over serial/tcp?

Should work over both. Thanks for calling that one out, I thought you had implemented something but figured I was just dreaming that up 😅

jamesarich:
> > [@thebentern](https://github.com/thebentern) [#3691](https://github.com/meshtastic/Meshtastic-Android/pull/3691) is the implementation for bluetooth - does this also work over serial/tcp?
> 
> Should work over both. Thanks for calling that one out, I thought you had implemented something but figured I was just dreaming that up 😅

fwiw i haven't tested tcp/serial (or seen working) - so we may want to break off a specific one for those transport types

PR #1110, opened 2024-06-27
Add logradio BLE characteristic

thebentern:
Start of #1109.
They're coming in at least
image


andrekir:
nice! inline with https://github.com/meshtastic/firmware/pull/4187 we could try to route this using the `LogRecord` FromRadio so it doesnt need separate plumbing all the way to the UI

PR #1111, opened 2024-06-28
Adds hardware model string to NodeInfo card ui if available

jamesarich:
Adds hardware model string to NodeInfo cards if available

image

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
@@ -90,7 +90,7 @@ fun NodeInfo(
                     .background(bgColor)
                     .padding(8.dp)
             ) {
-                val (chip, dist, name, pos, alt, sats, batt, heard, sig, env) = createRefs()
+                val (chip, dist, name, hw, pos, alt, sats, batt, heard, sig, env) = createRefs()
                 val barrierBattHeard = createStartBarrier(batt, heard)
                 val sigBarrier = createBottomBarrier(pos, heard)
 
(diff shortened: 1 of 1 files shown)


jamesarich:
> rather than the little preview card

I could see a case for an expandable card as well, with action buttons.

jamesarich:
Expandable cards would need to have more consideration and discussion beyond the original intent of this PR which is to expose the node's hardware model (which fwict is not visible anywhere else). Is there any roadmap or ux  guidelines that we could reference here for future UI improvements to expose any other node info not currently visible to the user?

app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
```diff
@@ -149,20 +149,44 @@ fun NodeInfo(
...
-                        )
+                            )
                         width = Dimension.preferredWrapContent
                     },
                     text = nodeName,
                     style = style,
                     textDecoration = TextDecoration.LineThrough.takeIf { isIgnored },
                 )
 
+                val hwInfoString = thisNodeInfo?.user?.hwModelString
```
andrekir, line 160: just noticed this should be using `thatNodeInfo`, otherwise it will show our local node's hardware for everyone.

let me know if you want to follow up with a fix.

jamesarich: big oop - i can do a quick PR

jamesarich: @andrekir https://github.com/meshtastic/Meshtastic-Android/pull/1123 fix here 🙃

PR #1119, opened 2024-07-01
feat: adds detekt gradle plugin

jamesarich:
Adds [detekt](https://detekt.dev/) gradle plugin to do code smell analysis and prevent smelly code.

Includes default generated config file. Also includes baseline file that ignores already smelly code to focus on ensuring no new smells get added.


jamesarich:
We could also consider adding github a status check for this.

andrekir:
> We could also consider adding github a status check for this.

sounds good

PR #1120, opened 2024-07-01
feat: adds ktlint gradle plugin

jamesarich:
Adds [ktlint gradle plugin](https://github.com/JLLeitschuh/ktlint-gradle) to provide gradle tasks for [ktlint](https://github.com/pinterest/ktlint) style checks and autoformatting.

Also includes baseline files that ignore any existing issues to focus on ensuring cleanly styled code is added.


jamesarich:
We could also consider adding github a status check for this.

andrekir:
> We could also consider adding github a status check for this.

yes, would be nice to run detekt and ktlint early in the CI

PR #1133, opened 2024-07-13
refactor: migrate `MessagesFragment` RecyclerView to Compose

andrekir:
wip


app/src/main/java/com/geeksville/mesh/ui/MessagesFragment.kt
```diff
@@ -305,8 +120,7 @@ class MessagesFragment : Fragment(), Logging {
 
         val contactKey = arguments?.getString("contactKey").toString()
         val contactName = arguments?.getString("contactName").toString()
-        val title = contactName
-        binding.toolbar.title  = title
+        binding.toolbar.title = contactName
```
jamesarich, line 123: Does it make any sense to move to [Scaffold and TopAppBar](https://developer.android.com/develop/ui/compose/components/scaffold) here? The toolbar view is currently being housed inside the fragment xml already and could be replaced with compose component.

jamesarich: Also if there's a better place to display the dm channel info in/after this rework, I'm open.

jamesarich (commented):
Lookin' good so far

app/src/main/java/com/geeksville/mesh/ui/MessagesFragment.kt
andrekir: > Does it make any sense to move to [Scaffold and TopAppBar](https://developer.android.com/develop/ui/compose/components/scaffold) here?

soon-ish, yeah. similar to: https://github.com/meshtastic/Meshtastic-Android/blob/de180f2d3556a2e8338099450c2f9a4d8c9f5972/app/src/main/java/com/geeksville/mesh/ui/DeviceSettingsFragment.kt#L133-L135

and we'll need to extract the `TopAppBar` so that it can be used in all screens.

> Also if there's a better place to display the dm channel info in/after this rework, I'm open.

looks good to me!

PR #1141, opened 2024-07-14
Channel scan select

Robert-0410:
### I believe this is related to #953.

Implemented a new dialog that is prompted when the user scans a correct QR code. The dialog provides additional
actions they can perform such as:
- Cancel: maintains the users current channels.
- Replace: makes the scanned channels the new user channels. (What would happen before)
- Save: appends the selected channels to the users existing channels.

### Notes
- The save button will only be enabled if the sum of the current and selected channels is <= max channels.
- The user can delete channels to make space, if cancel is pressed we maintain the same channels we had before the scan.

Screenshot_20240714-164454_Meshtastic


app/src/main/java/com/geeksville/mesh/ui/ChannelFragment.kt
```diff
@@ -547,6 +565,157 @@ private fun ChannelSelection(
...
 
+/**
+ * Enables the user to select which channels to accept after scanning a QR code.
+ */
+@Composable
+fun ScannedQrCodeDialog(
+    viewModel: UIViewModel = viewModel(),
+    scannedQR: String?,
+    onDismissRequest: () -> Unit
+) {
```
andrekir, line 579: avoid passing ViewModels to components, instead you can simply use:

``` kotlin
fun ScannedQrCodeDialog(
    channels: AppOnlyProtos.ChannelSet, (currentChannelSet)
    incoming: AppOnlyProtos.ChannelSet, (incomingChannelSet)
    onDismiss: () -> Unit,
    onConfirm: (AppOnlyProtos.ChannelSet) -> Unit,
) {
```

andrekir:
needs a few small changes but nice work!

regarding the buttons:
> Replace: makes the scanned channels the new user channels. (What would happen before)

convention is the default action ("positive") is placed on the right side of the dialog.
> Save: appends the selected channels to the users existing channels.

ensure the dialog buttons provide clear feedback of what they do: "Save" should read "Add".

app/src/main/java/com/geeksville/mesh/ui/ChannelFragment.kt
```diff
@@ -547,6 +565,157 @@ private fun ChannelSelection(
...
+                                style = MaterialTheme.typography.body1,
+                                text = stringResource(id = R.string.cancel)
+                            )
+                        }
+
+                        /* Replace - Replaces the previous set with the scanned channel set */
+                        Button(
+                            onClick = {
+                                onDismissRequest.invoke()
+                                viewModel.setRequestChannelUrl(Uri.parse(scannedQR))},
```
andrekir: `setRequestChannelUrl()` calls another dialog after this one (ie. the "old" dialog we're replacing), you can use `installSettings()` instead.

andrekir (changes requested):

app/src/main/java/com/geeksville/mesh/ui/ChannelFragment.kt
Robert-0410: Thanks, good stuff!

andrekir (approved):
well done, thanks!

Robert-0410:
Thank you!!

PR #1144, opened 2024-07-19
Expanding node info cards

### LINKED ISSUE #1097: Enhancement : If unknown node - show HEX nodeID
Instead of showing Unknown Username, show node ID instead.

This would open the way to the next step to write to a certain nodeID with the app or CLI command.
This would allow to address e.g. a known node with a new device.

jamesarich:
This is a rework of the UI for NodeInfo cards to be expandable. Tapping the card now expands the view to show less pertinent details as discussed in discord, https://github.com/meshtastic/Meshtastic-Android/pull/1111#issuecomment-2197270357, and #1138. A global show details toggle is available in the filter menu.
As part of this -  the views were reworked from a Constraint view into Rows for future readability and maintainability.

Hides the coordinates by default as mentioned in @garthvh's comment here https://github.com/meshtastic/Meshtastic-Android/discussions/1138#discussioncomment-10046245

Adds the node id when available - resolves  #1097.

Before
image

After
image

Also adds a global Show Details toggle to the Filter Menu. 
image


b8b8:
Can this be locked in expanded view?

jamesarich:
> Can this be locked in expanded view?

it's currently just toggle-able per card

b8b8:
ok, can you make it lock open if we want. some of us want to see all the data all the time by just scrolling vertically instead of having to tap each node. But so far this looks great. Moving the coordinate down to expanded view but keeping all other information visible in collapsed view was the right call. now we aren't relegating important info the only the expanded view while still having lots of room for future data to not impede the look.

b8b8:
no worries, I just would like it to be a future consideration/option.
There's been a lot of talk of hiding all the info behind separate cards, and stripping away the node info page to only the most basic info. I think the way shown here is the best solution. Not everything needs to be dumbed down.

jamesarich:
@b8b8 - added a global toggle in the filter menu

b8b8:
THANK YOU

jamesarich:
fixed expand toggle logic, added longpress to copy gps coordinates
image

jamesarich:
added node info text selection
image

app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
```diff
@@ -74,222 +95,172 @@ fun NodeInfo(
...
+                        verticalAlignment = Alignment.CenterVertically,
+                    ) {
+                        Chip(
+                            modifier = Modifier
+                                .width(72.dp)
+                                .padding(end = 8.dp)
+                                .defaultMinSize(minHeight = 32.dp)
+                                .clickable {
+                                    chipClicked()
+                                },
```
andrekir: .clickable can be removed here

andrekir:
we no longer need `contraintlayout-compose` dependency, please remove:

https://github.com/meshtastic/Meshtastic-Android/blob/de180f2d3556a2e8338099450c2f9a4d8c9f5972/app/build.gradle#L196

app/src/main/java/com/geeksville/mesh/ui/SignalInfo.kt
```diff
@@ -17,11 +17,14 @@ fun signalInfo(
...
-            nodeInfo.deviceMetrics?.airUtilTx
-        )
+    var text = ""
+    text = if (isThisNode) {
+        nodeInfo.deviceMetrics?.let { metrics ->
+            "ChUtil %.1f%% AirUtilTX %.1f%%".format(
+                metrics.channelUtilization,
+                metrics.airUtilTx
+            )
+        } ?: ""
```
andrekir: I don't think this is necessary. device metrics should never be null. if it is, this makes it clear to users something is wrong.

andrekir (changes requested):

andrekir (approved):
looks good, thanks!

PR #1151, opened 2024-07-21
feat: support add=true in QR codes

### LINKED ISSUE #953: Honor add=true query string value when scanning QR codes

AddisonTustin:
Support when `?add=true` is present in the scanned QR code. 

I've been working with maintainers in the Meshtastic-Apple and web repos to relocate the `?add=true` before the URI fragment, but until those changes are generally available, this change supports determining if it should add the channel configurations if `?add=true` is either before or inside the fragment.

Screenshots below on behavior. All QR codes were generated from a Meshtastic-Apple client with "Primary" channel excluded.

**Add Single Channel**:
Single Channel Add
After Single Channel Add

**Add Multiple Channels**:
Multi Channel Add
After Multi Channel Add

Resolves #953


app/build.gradle
```diff
@@ -12,8 +12,11 @@ plugins {
 }
 
 unMock {
-    keep "android.net.Uri"
     keep "android.util.Base64"
+    keepStartingWith "libcore."
+    keep "android.net.Uri"
+
+    keepAndRename "java.nio.charset.Charsets" to "xjava.nio.charset.Charsets"
```
AddisonTustin, line 19: I'm happy to document this in the build script, but the ChannelSet tests were throwing NoClassDefFoundError for `java.nio.charset.Charsets` and `libcore.net.UriCodec`.  Documentation in [unmock-plugin](https://github.com/bjoernQ/unmock-plugin/blob/master/README.md) indicated that these might be necessary, which resolved the test failures.

config/detekt/detekt.yml
```diff
@@ -365,7 +365,7 @@ naming:
     packagePattern: '[a-z]+(\.[a-z][A-Za-z0-9]*)*'
   TopLevelPropertyNaming:
     active: true
-    constantPattern: '[A-Z][A-Za-z0-9]*'
+    constantPattern: '[A-Z][_A-Za-z0-9]*'
```
AddisonTustin, line 368: I'm not sure why underscores (`_`) weren't included in the constant pattern, but it seems incorrect to prohibit something like `MESHTASTIC_DOMAIN` in favor of `MESHTASTICDOMAIN`.

The [default Detekt configurations](https://detekt.dev/docs/rules/naming/#configuration-options-15) allow underscores, but I'm happy to change this back if it's preferable.

jamesarich: Agree, feel free to adjust - it's a new add, and may need some tuning.

app/src/main/java/com/geeksville/mesh/model/UIState.kt
```diff
@@ -386,13 +386,29 @@ class UIViewModel @Inject constructor(
...
+            newRadioSettings = channelSet.settingsList
+        }  else {
+            // To guarantee consistent ordering, using a LinkedHashSet which iterates through it's
+            // entries according to the order an item was *first* inserted.
+            // https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-linked-hash-set/
+            val uniqueChannels =
+                LinkedHashSet(channels.value.settingsList + channelSet.settingsList).toList()
+            getChannelList(uniqueChannels).forEach(::setChannel)
+            newRadioSettings = uniqueChannels
+        }
```
andrekir:
```suggestion
        val newRadioSettings = if (overwrite) {
            channelSet.settingsList
        } else {
            // To guarantee consistent ordering, using a LinkedHashSet which iterates through it's
            // entries according to the order an item was *first* inserted.
            // https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/-linked-hash-set/
            LinkedHashSet(channels.value.settingsList + channelSet.settingsList).toList()
        }

        getChannelList(newRadioSettings, channels.value.settingsList).forEach(::setChannel)
```

app/src/main/java/com/geeksville/mesh/model/UIState.kt
```diff
@@ -78,12 +78,12 @@ fun getInitials(nameIn: String): String {
...
  *
  * @param new The updated [ChannelSettings] list.
- * @param old The current [ChannelSettings] list (required to disable unused channels).
+ * @param old The current [ChannelSettings] list (required when disabling unused channels).
  * @return A [Channel] list containing only the modified channels.
  */
 internal fun getChannelList(
     new: List<ChannelSettings>,
-    old: List<ChannelSettings>,
+    old: List<ChannelSettings> = emptyList(),
```
andrekir: please don't skip this this parameter.

app/src/main/java/com/geeksville/mesh/model/ChannelSet.kt
```diff
@@ -12,22 +12,31 @@ import java.net.MalformedURLException
...
 /**
- * Return a [ChannelSet] that represents the URL
+ * Return a [ChannelSet] that represents the ChannelSet encoded by the URL and [Boolean] if the URL
+ * indicates the [ChannelSet] should be added to the existing configuration.
  * @throws MalformedURLException when not recognized as a valid Meshtastic URL
  */
 @Throws(MalformedURLException::class)
-fun Uri.toChannelSet(): ChannelSet {
-    val urlStr = this.toString()
+fun Uri.toChannelSet(): Pair<ChannelSet, Boolean> {
```
andrekir: `toChannelSet()` should remain generic and just return `ChannelSet`. please move `shouldAdd` logic to a separate function.

app/src/main/java/com/geeksville/mesh/model/ChannelSet.kt
```diff
@@ -12,22 +12,31 @@ import java.net.MalformedURLException
...
+    if (fragment.isNullOrBlank() || host != MESHTASTIC_DOMAIN || path != MESHTASTIC_CHANNEL_CONFIG_PATH) {
```
andrekir: here `host` and `path` matching needs to ignore case. ".equals(string, true)" should work.

suggestion: consider naming more in line with the Uri reference, ie. MESHTASTIC_HOST and MESHTASTIC_PATH.

andrekir (changes requested):
thanks for your contribution. looks really good overall, just a few small changes and we can merge.

config/detekt/detekt.yml
andrekir: Compose [introduced `PascalCase` convention](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md#singletons-constants-sealed-class-and-enum-class-values) but `CAPITALS_AND_UNDERSCORES` is also fine.

AddisonTustin: fwiw, I actually think the compose documentation might be incorrect or out-of-date? The [Kotlin convention](https://kotlinlang.org/docs/coding-conventions.html#property-names) is shouty snake case (CAPITALS_AND_UNDERSCORES):
> Names of constants (properties marked with const, or top-level or object val properties with no custom get function that hold deeply immutable data) should use uppercase underscore-separated ([screaming snake case](https://en.wikipedia.org/wiki/Snake_case)) names.

app/src/main/java/com/geeksville/mesh/model/UIState.kt
AddisonTustin: Ah, I understand. I had a poor assumption in how the result of `getChannelList` was being used.

Thanks!

app/src/main/java/com/geeksville/mesh/model/UIState.kt
AddisonTustin: :+1:

app/src/main/java/com/geeksville/mesh/model/ChannelSet.kt
AddisonTustin: Can do

app/src/main/java/com/geeksville/mesh/model/ChannelSet.kt
AddisonTustin: Oop totally, nice catch! I'll add some tests to verify this behavior for the future too.

AddisonTustin:
> thanks for your contribution. looks really good overall, just a few small changes and we can merge.

Thanks for the review, I'll get those updated!

config/detekt/detekt.yml
andrekir: think of Kotlin guidelines as general rules and Compose guidelines as specific rules for Compose, which is a framework within Kotlin. So basically, what is mentioned in the Compose guidelines is what is different from Kotlin guidelines.

top-level constants used here are a common Compose pattern, so that's why Detekt wants `PascalCase` but since it doesn't really apply (ie. not Compose code) both should be allowed.

AddisonTustin:
Added the changes. I tried to see if there's a preferred process for fixups (addt'l commits vs amend). If you don't plan on doing the squash as a part of the merge, let me know and I can clean up the git history 🙂.

Thanks again!

andrekir:
> Added the changes. I tried to see if there's a preferred process for fixups (addt'l commits vs amend).

yes, adding commits is preferred.

andrekir (approved):
looks good, thanks!

PR #1157, opened 2024-07-26
Unknow Username to Node ID

musznik:
Instead of showing "unknown username", display the node ID.

Why is this important? It allows for locating the node using an online map and better tuning of the antenna over long distances, recognizing distant nodes, and distinguishing them on the list.

Screenshot_1082


garthvh:
You should use the last 4 of the hex node id in the colored pull too

jamesarich:
My expanding nodeinfo card PR  #1144 _does_ display the device id, but I like the idea of it being the default fallback.

musznik:
> You should use the last 4 of the hex node id in the colored pill too

Screenshot_1083

suggestion implemented

prokrypt:
I think it may be better to have UNK or similar in the pill. That way it is easier to spot an unknown node on the map, and makes it possible to be searched/filtered. If more info on an UNK node is needed when looking at the map, a single tap is all it takes to show the whole nodeid.

thebentern:
> I think it may be better to have UNK or similar in the pill. That way it is easier to spot an unknown node on the map, and makes it possible to be searched/filtered. If more info on an UNK node is needed when looking at the map, a single tap is all it takes to show the whole nodeid.

I disagree entirely. UNK provides zero helpful information. It's an all-caps acronym that is does not translate well and is not immediately apparent to all users. Whereas the last 4 hex gives you actual relevant information about the node. We can add filters for unknown nodes without having that useless string in the pill.

thebentern (approved):
Love it! Great work!

prokrypt:
> > I think it may be better to have UNK or similar in the pill. That way it is easier to spot an unknown node on the map, and makes it possible to be searched/filtered. If more info on an UNK node is needed when looking at the map, a single tap is all it takes to show the whole nodeid.
> 
> I disagree entirely. UNK provides zero helpful information. It's an all-caps acronym that is does not translate well and is not immediately apparent to all users. Whereas the last 4 hex gives you actual relevant information about the node. We can add filters for unknown nodes without having that useless string in the pill.

But the downside with the current shortname is that there is no way on the map to differentiate an unknown node and one with nodeinfo, and there is a high chance the shortname will change when nodeinfo is received.

thebentern:
> But the downside with the current shortname is that there is no way on the map to differentiate an unknown node and one with nodeinfo, and there is a high chance the shortname will change when nodeinfo is received.

A NodeInfo (User) will either arrive eventually or it won't. Knowing whether a node is "unknown" does not give the user any actionable path.

prokrypt:
> A NodeInfo (User) will either arrive eventually or it won't. Knowing whether a node is "unknown" does not give the user any actionable path.
If an auto generated short name is displayed, a user is more likely to use it to search on an online map and likely not be able to find it. Isn't it better to make it obvious there is no nodeinfo so a user will search using the nodeid?

garthvh:
> > A NodeInfo (User) will either arrive eventually or it won't. Knowing whether a node is "unknown" does not give the user any actionable path.
> 
> If an auto generated short name is displayed, a user is more likely to use it to search on an online map and likely not be able to find it. Isn't it better to make it obvious there is no nodeinfo and force a user to search using the nodeid?
> 
> Maybe another method would be to have a ! in the short name too, if 5 characters is possible.

A year in not a single user has requested the return of ??? On iOS instead of relevant data I think you are a real edge case here.

prokrypt:
> A year in not a single user has requested the return of ??? On iOS instead of relevant data I think you are a real edge case here.

I just feel like a known (or rather, UNKnown or ???) is replaced with an unknown (a generated shortname that may change at any time). Currently, it is difficult to tell if a node in chat has nodeinfo or not if their shortname is in hex format. If you press on their shortname, it just jumps to the nodelist unless you enable "Include unknown".

garthvh:
> > A year in not a single user has requested the return of ??? On iOS instead of relevant data I think you are a real edge case here.
> 
> I just feel like a known (or rather, UNKnown or ???) is replaced with an unknown (a generated shortname that may change at any time). Currently, it is difficult to tell if a node in chat has nodeinfo or not if their shortname is in hex format. If you press on a shortname with no nodeinfo, it just jumps to the nodelist unless you enable "Include unknown".

This is a seperate issue and can be determined from the lack of hardware model data 
IMG_2964

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -414,10 +414,17 @@ class MeshService : Service(), Logging {
 
     // given a nodeNum, return a db entry - creating if necessary
     private fun getOrCreateNodeInfo(n: Int) = nodeDBbyNodeNum.getOrPut(n) {
+        val defaultId = DataPacket.nodeNumToDefaultId(n)
+        val shortName = if (defaultId.length >= 4) {
```
andrekir, line 419: nodeNumToDefaultId() returns a fixed length of 9 characters, so this is always true and can be removed

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -414,10 +414,17 @@ class MeshService : Service(), Logging {
...
+            defaultId.takeLast(4)
```
andrekir, line 420: use .takeLast(n = 4) to fix the lint error

andrekir (changes requested):

PR #1174, opened 2024-08-01
feat: Add role to NodeInfo

jamesarich:
image

As suggested by **https://github.com/meshtastic/Meshtastic-Android/discussions/1138#discussioncomment-10203963**

This adds a new field `role` to the `NodeInfo` entity in the MeshtasticDatabase. The `role` field is an integer with a default value of 0. This change allows the app to store and display the role of a user in the mesh network.

@andrekir - I found the `role` defined in the `User` proto [here](https://buf.build/meshtastic/protobufs/docs/main:meshtastic#meshtastic.User), and added it to our local model. Let me know if there's a better/more preferred way to add a field from a proto.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
+++ b/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
@@ -35,11 +35,11 @@ data class MeshUser(
 
     override fun toString(): String {
         return "MeshUser(id=${id.anonymize}, " +
-                "longName=${longName.anonymize}, " +
-                "shortName=${shortName.anonymize}, " +
-                "hwModel=${hwModelString}, " +
-                "isLicensed=${isLicensed}), " +
-                "role=${role})"
+            "longName=${longName.anonymize}, " +
+            "shortName=${shortName.anonymize}, " +
+            "hwModel=$hwModelString, " +
+            "isLicensed=$isLicensed, " +
+            "role=$role)"
     }
 
     /** Create our model object from a protobuf.
@@ -60,7 +60,7 @@ data class MeshUser(
             .setShortName(shortName)
             .setHwModel(hwModel)
             .setIsLicensed(isLicensed)
-            .setRole(ConfigProtos.Config.DeviceConfig.Role.forNumber(role))
+            .setRoleValue(role)
             .build()
 
     /** a string version of the hardware model, converted into pretty lowercase and changing _ to -, and p to dot
--- a/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
+++ b/app/src/main/java/com/geeksville/mesh/NodeInfo.kt
@@ -29,10 +29,17 @@ data class MeshUser(
     val shortName: String,
     val hwModel: MeshProtos.HardwareModel,
     val isLicensed: Boolean = false,
+    @ColumnInfo(name = "role", defaultValue = "0")
+    val role: Int = 0,
 ) : Parcelable {
 
     override fun toString(): String {
-        return "MeshUser(id=${id.anonymize}, longName=${longName.anonymize}, shortName=${shortName.anonymize}, hwModel=${hwModelString}, isLicensed=${isLicensed})"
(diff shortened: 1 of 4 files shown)


app/schemas/com.geeksville.mesh.database.MeshtasticDatabase/8.json
```diff
@@ -98,7 +98,7 @@
...
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`num` INTEGER NOT NULL, `snr` REAL NOT NULL, `rssi` INTEGER NOT NULL, `lastHeard` INTEGER NOT NULL, `channel` INTEGER NOT NULL, `hopsAway` INTEGER NOT NULL DEFAULT 0, `user_id` TEXT, `user_longName` TEXT, `user_shortName` TEXT, `user_hwModel` TEXT, `user_isLicensed` INTEGER, `user_role` TEXT, `position_latitude` REAL, `position_longitude` REAL, `position_altitude` INTEGER, `position_time` INTEGER, `position_satellitesInView` INTEGER, `position_groundSpeed` INTEGER, `position_groundTrack` INTEGER, `position_precisionBits` INTEGER, `devMetrics_time` INTEGER, `devMetrics_batteryLevel` INTEGER, `devMetrics_voltage` REAL, `devMetrics_channelUtilization` REAL, `devMetrics_airUtilTx` REAL, `devMetrics_uptimeSeconds` INTEGER, `envMetrics_time` INTEGER, `envMetrics_temperature` REAL, `envMetrics_relativeHumidity` REAL, `envMetrics_barometricPressure` REAL, `envMetrics_gasResistance` REAL, `envMetrics_voltage` REAL, `envMetrics_current` REAL, `envMetrics_iaq` INTEGER, PRIMARY KEY(`num`))",
```
andrekir: we never change an existing schema directly. this can cause data loss and crashes. use Room auto migration to add a new schema.

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -29,10 +30,16 @@ data class MeshUser(
     val shortName: String,
     val hwModel: MeshProtos.HardwareModel,
     val isLicensed: Boolean = false,
+    val role: DeviceConfig.Role
```
andrekir: using protobuf enums for types (like hwModel) is problematic, instead use `Int` to work with the "_VALUE" suffix (e.g., CLIENT -> CLIENT_VALUE) and role.number

then you can tag "= 0" as default value (CLIENT) so we don't need any extra handling elsewhere like in testing

```kotlin
data class MeshUser(
    val id: String,
    val longName: String,
    val shortName: String,
    val hwModel: MeshProtos.HardwareModel,
    val isLicensed: Boolean = false,
    val role: Int = 0, // ConfigProtos.Config.DeviceConfig.Role.CLIENT_VALUE
) : Parcelable {
```

andrekir (changes requested):

andrekir:
please follow the [official docs](https://developer.android.com/training/data-storage/room/migrating-db-versions) for details. check the database git history for examples. this can cause big problems if not done right.

jamesarich:
> please follow the [official docs](https://developer.android.com/training/data-storage/room/migrating-db-versions) for details. check the database git history for examples. this can cause big problems if not done right.

Gotcha - I'll move this to draft and do a little research. Thanks for reviewing.

jamesarich:
@andrekir - I think I got it sorted, the Automigration was failing due to expecting a nullable `role`.  Want to give it another look?

andrekir:
you can define a default value for a column by using the annotation @ColumnInfo(defaultValue = "...")
```kotlin
@ColumnInfo(defaultValue = "0")
val role: Int = 0,
```

jamesarich:
@andrekir another round? :sweat:

andrekir:
that should do it. I'll get back to it soon. 👍

andrekir (approved):

PR #1178, opened 2024-08-03
Fix: Periodically update last heard time

### LINKED ISSUE #1136: [Bug]: Times in node list do not update
When viewing the node list, the last seen time appears to be essentially static unless a remote node is seen, then the time will update to 'now'. None of the other nodes will count their time up on a refresh interval. A node may say last seen 'now' but the actual last transmission may have been 20 mins ago, or even hours. Terminating the app and reopening is the only way to get the times to update.

jamesarich:
This introduces a periodic update to the last heard time displayed in the UI. The time is now refreshed every 1 minute
 to ensure accurate and up-to-date information.

fixes #1136


andrekir:
adding a `LaunchedEffect` for every node can be quite resource-intensive, instead it should be hoisted this all the way up to `NodesScreen`.

you could wrap a System.currentTimeMillis() tick and pass that to be consumed in the `LastHeardInfo` component.

PR #1190, opened 2024-08-09
Fix: Colored transparent precision circles around nodes on map

jamesarich:
This updates precision circles around nodes on the map to better visually represent the accuracy of their positions. The circles are now colored according to the node's color and are now more transparent.

This adds precision circle to the `MarkerLabel` canvas, so it can be managed alongside markers. This fixes the redraw on zoom issue that users were experiencing.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/model/map/MarkerWithLabel.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/map/MarkerWithLabel.kt
@@ -27,8 +27,8 @@ class MarkerWithLabel(mapView: MapView?, label: String, emoji: String? = null) :
     }
 
     @Suppress("MagicNumber")
-    private val precisionMeters =
-        when (precisionBits) {
+    private fun getPrecisionMeters(): Double? {
+        return when (precisionBits) {
             10 -> 23345.484932
             11 -> 11672.7369
             12 -> 5836.36288
@@ -41,6 +41,7 @@ class MarkerWithLabel(mapView: MapView?, label: String, emoji: String? = null) :
             19 -> 45.58554
             else -> null
         }
+    }
 
 
     private var onLongClickListener: (() -> Boolean)? = null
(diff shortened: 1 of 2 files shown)


andrekir:
there are other Polygon() instances in the code, and managing them separately will be very complicated. to avoid that you can simply move the polygon into `MarkerWithLabel`.

https://github.com/meshtastic/Meshtastic-Android/blob/6d01fb4bc41b072903c4a2470bd78362b9474f3a/app/src/main/java/com/geeksville/mesh/model/map/MarkerWithLabel.kt#L60-L69

`org.osmdroid.views.overlay.Polygon` also has a draw method:

```kotlin
class MarkerWithLabel(mapView: MapView?, label: String, emoji: String? = null) : Marker(mapView) {
(...)

    private var precisionBits: Int? = null

    fun setPrecisionBits(bits: Int) {
        precisionBits = bits
    }

(...)
    override fun draw(c: Canvas, osmv: MapView?, shadow: Boolean) {
    (...)
        precisionBits?.let {
            val polygon = Polygon(osmv).apply {
                points = Polygon.pointsAsCircle(
                    position,
                    bitsToMeters(it)
                )
                fillPaint.color = Color.RED
                fillPaint.alpha = 64
                outlinePaint.color = Color.RED
            }
            polygon.draw(c, osmv, false)
        }
    }
```

jamesarich:
Good catch - don't want to nuke those other Polygons, makes it simpler to just add them to the marker canvas.

app/src/main/java/com/geeksville/mesh/model/map/MarkerWithLabel.kt
```diff
@@ -15,6 +16,33 @@ class MarkerWithLabel(mapView: MapView?, label: String, emoji: String? = null) :
...
+    }
+
+    private var precisionBits: Int? = null
+    fun setPrecisionBits(bits: Int) {
+        precisionBits = bits
+    }
+
+    @Suppress("MagicNumber")
+    private val precisionMeters =
+        when (precisionBits) {
```
andrekir: without a getter this variable is static and will always point to null

app/src/main/java/com/geeksville/mesh/model/map/MarkerWithLabel.kt
```diff
@@ -15,6 +16,33 @@ class MarkerWithLabel(mapView: MapView?, label: String, emoji: String? = null) :
         private const val LABEL_Y_OFFSET = 100F
     }
 
+    private var nodeColor: Int = Color.GRAY
+    fun setNodeColors(colors: Pair<Int, Int>) {
+        nodeColor = colors.first
```
andrekir: = colors.second

andrekir (changes requested):

andrekir (approved):

PR #1195, opened 2024-08-14
Offline map - OrganicMap integration  

alexander-smyslov:
This is a prof of concept offline map in Meshastic.
OrganicMap best offline map  (in my opinion)
Use this https://omaps.app/api to interact with  OrganicMap 

Need custom button icon.


andrekir:
Thanks for sharing! Just a heads up, there's an API built specifically for app integration. To listen to node positions you only need to register for `NODE_CHANGE` broadcasts: 
https://github.com/meshtastic/Meshtastic-Android/blob/6b4603af9d306d2acd6b3ae7a47b7a747c67475f/app/src/main/aidl/com/geeksville/mesh/IMeshService.aidl#L11-L44

Issue #559, opened 2023-01-11
[Feature request] Telemetry / Device Metrics Log


IZ1IVA:
I don't own an iOS 16 device, but I've discovered by chance in a [post](https://meshtastic.discourse.group/t/t-beam-and-device-metrics-log-only-storing-4-values/7044?u=iz1iva) a "Device Metrics Log" and graph tab, which is not present in the Android app.
Would it be possibile to add it to the Android app?
How about "syncing" features between the apps?

Cheers

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/t-beam-and-device-metrics-log-only-storing-4-values/7044/9

hitachi8:
hello, if i might please bring my comment into this.

i would also love to be able to log, save and backup the telemetry data from a node of my choosing unto my own smartphone in order to make a wall display, or graph later on. basically have the data available on that cellphone.
here would be my setup:
[distant WX station] --LoRa--> [personal node] --> store received telemetry data 

also if possible: 
-save it automatically instead of having to do it manually?
-have the data in a .txt file line by line format as shown in the IOS app (easier to use)
-being able to select what node to store said telemetry data from
-input where to save the .txt file 

Thank you all for your work.

IZ1IVA:
Cheers!

PR #1198, opened 2024-08-17
Initial Charts

Robert-0410:
Issue #559 seems relevant to this Pull Request.

I've implemented a first graph/chart for the device metrics as a start.

My plan is to address any valuable input anyone may have and based on those improvements we can plan the next steps.

### My Concerns
- Anything wrong with the way I got the data?

Note: The data used for the chart below is not real.

Screenshot_20240817-141800_Meshtastic


app/src/main/java/com/geeksville/mesh/ui/NodeDetailsFragment.kt
```diff
@@ -0,0 +1,156 @@
...
+    val data = mutableListOf<DataEntry>()
+
+    /* Retrieve only the data belonging to the selected node. */
+    for (log in logs) {
+        log.nodeInfo?.let { nodeInfo ->
+            val currentId = nodeInfo.user.id
+            if (nodeInfo.hasDeviceMetrics() && currentId == nodeId)
+                data.add(DataEntry(log.received_date, DeviceMetrics(nodeInfo.deviceMetrics)))
+        }
+    }
```
andrekir: we want to do all the heavy lifting in the ViewModel, and the UI should be just the consumer for the data.

you can add your own custom ViewModel ("NodeDetailsViewModel"?) using "DebugViewModel" as a baseline and build the `DataEntry` list there.

I added a method to the MeshLog repository to retrive Telemetry for a specific node that you can use: https://github.com/meshtastic/Meshtastic-Android/blob/7e0cfff67bd38f0eaeed05798456c2e8c430fcf2/app/src/main/java/com/geeksville/mesh/database/MeshLogRepository.kt#L36

app/src/main/java/com/geeksville/mesh/model/UIState.kt
```diff
@@ -194,10 +195,20 @@ class UIViewModel @Inject constructor(
...
     )
 
+    @OptIn(ExperimentalCoroutinesApi::class)
+    fun getNodeByNum(nodeNum: Int): StateFlow<NodeInfo?> = nodeDB.nodeDBbyNum.mapLatest { nodes ->
+        nodes[nodeNum]
+    }.stateIn(
+        scope = viewModelScope,
+        started = WhileSubscribed(STOP_TIMEOUT_MILLIS),
+        initialValue = null,
+    )
```
andrekir: then this can also move to your own ViewModel

andrekir (commented):
looking good. added a few notes to improve how the data is handled.

app/src/main/java/com/geeksville/mesh/ui/NodeDetailsFragment.kt
Robert-0410: Awesome with that new method I was able to get all the data!

Screenshot_20240818-225816_Meshtastic

app/src/main/java/com/geeksville/mesh/model/UIState.kt
Robert-0410: I ended not using that function because it was overkill for what we needed it for. Instead I only retrieve the short name of the selected node. This helped avoid unnecessary re-composition.

Robert-0410:
I'm going to move forward with the Environmental Metrics chart. I was thinking about using a Pager that allows the user to swipe between graphs. Let me know if that works or if anyone has another suggestions.

jamesarich:
> I'm going to move forward with the Environmental Metrics chart. I was thinking about using a Pager that allows the user to swipe between graphs. Let me know if that works or if anyone has another suggestions.

nice progress - Pager w/ tabs or similar controls probably makes sense here. Found some [guidelines](https://m2.material.io/design/communication/data-visualization.html#behavior:~:text=the%20previous%20day.-,Data%20controls,-Data%20can%20be) (material 2 but seems relevant.

Robert-0410:
> > I'm going to move forward with the Environmental Metrics chart. I was thinking about using a Pager that allows the user to swipe between graphs. Let me know if that works or if anyone has another suggestions.
> 
> nice progress - Pager w/ tabs or similar controls probably makes sense here. Found some [guidelines](https://m2.material.io/design/communication/data-visualization.html#behavior:~:text=the%20previous%20day.-,Data%20controls,-Data%20can%20be) (material 2 but seems relevant).

Thanks, cool guide had not seen that yet.

andrekir:
> I'm going to move forward with the Environmental Metrics chart. I was thinking about using a Pager that allows the user to swipe between graphs. Let me know if that works or if anyone has another suggestions.

sounds good.

I'd suggest working with protobufs directly instead of our POJOs (ie. `com.geeksville.mesh.TelemetryProtos.EnvironmentMetrics` instead of `com.geeksville.mesh.EnvironmentMetrics`)

as the POJOs will always behind (sometimes not available at all) for some telemetry variants.

Robert-0410:
> > I'm going to move forward with the Environmental Metrics chart. I was thinking about using a Pager that allows the user to swipe between graphs. Let me know if that works or if anyone has another suggestions.
> 
> sounds good.
> 
> I'd suggest working with protobufs directly instead of our POJOs (ie. `com.geeksville.mesh.TelemetryProtos.EnvironmentMetrics` instead of `com.geeksville.mesh.EnvironmentMetrics`)
> 
> as the POJOs will always behind (sometimes not available at all) for some telemetry variants.

Makes sense.

Robert-0410:
The first iterations of graphs are ready for review. They still have room to improve but I would say its useful in it's current state.

app/src/main/java/com/geeksville/mesh/model/NodeDetailsViewModel.kt
```diff
@@ -0,0 +1,57 @@
...
+    private val _environmentMetrics = MutableStateFlow<List<Telemetry>>(emptyList())
+    val environmentMetrics: StateFlow<List<Telemetry>> = _environmentMetrics
+
+    /**
+     * Gets the short name of the node identified by `nodeNum`.
+     */
+    @OptIn(ExperimentalCoroutinesApi::class)
+    suspend fun getNodeName(nodeNum: Int): String? {
+        return nodeDB.nodeDBbyNum.mapLatest { it[nodeNum] }.first()?.user?.shortName
+    }
```
andrekir: you can access `nodeDB.nodeDBbyNum.value` property to avoid suspending and creating a flow pipeline:

``` kotlin
fun getNodeName(nodeNum: Int): String? = nodeDB.nodeDBbyNum.value[nodeNum]?.user?.shortName
```

app/src/main/java/com/geeksville/mesh/ui/NodeDetailsFragment.kt
```diff
@@ -0,0 +1,173 @@
...
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        val nodeNum = arguments?.getInt("nodeNum")
+        if (nodeNum != null)
+            model.setSelectedNode(nodeNum)
+        /* We only need to get the nodes name once. */
+        var nodeName: String? = ""
+        lifecycleScope.launch { nodeName = model.getNodeName(nodeNum ?: 0) }
+
```
andrekir, line 72: then this can be simplified with

``` kotlin
val nodeName = model.getNodeName(nodeNum ?: 0)
```

app/src/main/java/com/geeksville/mesh/ui/NodeDetailsFragment.kt
```diff
@@ -0,0 +1,173 @@
...
+            else
+                Color.LightGray
+
+            val imageVector = if (iteration == 0)
+                ImageVector.vectorResource(R.drawable.baseline_charging_station_24)
+            else
+                ImageVector.vectorResource(R.drawable.baseline_thermostat_24)
+            Icon(
+                imageVector = imageVector,
+                contentDescription = stringResource(R.string.tab),
```
andrekir: this should point to `R.string.device_metrics` / `R.string.environment_metrics` depending on what is selected

andrekir (commented):
added a few notes but overall looks good!

app/src/main/java/com/geeksville/mesh/ui/NodeDetailsFragment.kt
Robert-0410: I wondered about that, thanks.

app/src/main/java/com/geeksville/mesh/model/NodeDetailsViewModel.kt
Robert-0410: :+1:

app/src/main/java/com/geeksville/mesh/ui/NodeDetailsFragment.kt
Robert-0410: :+1:

DOC-HOC:
Hi Robert,

How about adding the following features to the Device Metrics?
- Changeable timeframe: i.e. choose between history of one day, 3 days, one week of history
- choose what should be displayed: i.e. checkboxes for each sensor so you can unselect battery voltage and would see Channel and Air utilization with a better resolution
- add position to the history, ideally with a map and a track on it

Thanks,
Chris

PR #1219, opened 2024-08-30
feat: adds UI for unknown node details

### LINKED ISSUE #1218: Wrong role displayed on device list (client instead of Router)
### Contact Details

[mertszrdev@gmail.com](mailto:mertszrdev@gmail.com)

### What happened?

Device list shows shows all routers as clients, 
I have 2 routers and 1 old router_client.
One of the routers (R2) is near me so its more up to date v2.4.1 
I don't remember the versions of the other two.
List shows my own client_mute device correctly.

Screenshot_2024-08-30-13-17-00-112_com geeksville mesh~2
Screenshot_2024-08-30-13-25-55-323_com geeksville mesh~2
Screenshot_2024-08-30-13-22-01-792_com geeksville mesh~2

### App Version

2.4.4

### Phone

Xiaomi 13T

### Device

Rak4631

### Firmware

2.4.1

### Relevant log output

_No response_

### Code of Conduct

jamesarich:
This commit provides default device details shown in the NodeInfo cards, providing visual clarity when the information is unknown.
image

resolves #1218


app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -29,8 +29,8 @@ data class MeshUser(
     val shortName: String,
     val hwModel: MeshProtos.HardwareModel,
     val isLicensed: Boolean = false,
-    @ColumnInfo(name = "role", defaultValue = "0")
-    val role: Int = 0,
+    @ColumnInfo(name = "role", defaultValue = "-1")
```
andrekir: creating a new migration to change the `defaultValue` has no effect

andrekir (changes requested):
thanks for looking into this but looks like the issue is the role is not being set here:

https://github.com/meshtastic/Meshtastic-Android/blob/cb2ea2804a19caec29f0b8322cf26872e21904a5/app/src/main/java/com/geeksville/mesh/service/MeshService.kt#L419-L426

and also for unknown:

https://github.com/meshtastic/Meshtastic-Android/blob/cb2ea2804a19caec29f0b8322cf26872e21904a5/app/src/main/java/com/geeksville/mesh/service/MeshService.kt#L798-L810

app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt
```diff
@@ -87,7 +87,12 @@ fun NodeInfo(
     val (textColor, nodeColor) = thatNodeInfo.colors
 
     val position = thatNodeInfo.position
-    val hwInfoString = thatNodeInfo.user?.hwModelString
+    val hwInfoString = thatNodeInfo.user?.hwModelString ?: MeshProtos.HardwareModel.UNSET.name
+    val roleName = thatNodeInfo.user?.role?.let { role ->
+        DeviceConfig.Role.forNumber(role)?.name
```
andrekir: we won't be able to set `Role.UNRECOGNIZED` as default for unknown users in [MeshService](https://github.com/meshtastic/Meshtastic-Android/blob/24886994a6afc1a443b92e2630903612403158f8/app/src/main/java/com/geeksville/mesh/service/MeshService.kt#L373-L379) so let's use an if/else block to check `isUnknownUser` and assign `Role.UNRECOGNIZED.name` when its true.

andrekir (approved):
looks good, thanks!

PR #1241, opened 2024-09-12
Display PKI encryption and Public Key status

jamesarich:
This commit adds an icon to messages to indicate whether they are encrypted with PKI. Also adds icon to contacts and nodes to indicate that a public key exists.

The icon is a lock if the message is encrypted and a crossed lock if it is not.
The icon is a lock if the contact or node has a public key, crossed lock if not.
The icon is displayed in the shortname chip.
A content description is also added to the icon to provide more information to users who use screen readers.

image

image


andrekir:
thanks for the effort! there are a lot of changes in this area that I’ll be pushing soon, so let’s hold off for now and we can revisit after that.

jamesarich:
> thanks for the effort! there are a lot of changes in this area that I’ll be pushing soon, so let’s hold off for now and we can revisit after that.

Sure - It may be helpful to coordinate more on the work you're planning to do to avoid wasted effort or rework.

PR #1245, opened 2024-09-14
Mesh Packet Limit

Robert-0410:
### Problem
Users are experiencing long wait times when retrieving data for the metrics.


Robert-0410:
Also started separating the components for the respective charts.

Issue #1207, opened 2024-08-24
[Feature Request]: Rearrange traceroute result and add some differentiating colour


b8b8:
### Tell us your idea.

My suggestion is to have the traceroutes present vertically and with different colours for hops, transmit, and receive. Also including the hopping SNR would be beneficial.
![signal-2024-08-24-144701](https://github.com/user-attachments/assets/c1b0cd92-88ee-4e55-ad4b-ddf31e37d73c)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

GUVWAF:
Hey @b8b8, maybe it would be good to consider here how traceroute in 2.5 is going to look like in general. Apart from the SNR per hop, we can also get a two-way traceroute result. There is some validation needed, because it depends on the firmware of the nodes in the route whether the info is accurate or not. I opened a PR for the Python CLI (https://github.com/meshtastic/python/pull/654) to accommodate for this; basically the length of the list of SNRs should be 1 longer than the route list and nodes and SNR can be unknown (represented by `0xFFFFFFFF` and -128 respectively).

Maybe it can look something like this then (excuse me for my Paint skills):
![traceroute_android](https://github.com/user-attachments/assets/b73b37c2-8447-480f-98d9-8a3af925fcb4)

Not sure how feasible all of this is, but it would be cool to show it like this :)

b8b8:
Dude this looks even better.

Nestpebble:
I've already mentioned it in the discord, but apparently not asked for it as a feature request. Putting the results in the chat history for a node would solve some of the things people have asked for above

b8b8:
> I've already mentioned it in the discord, but apparently not asked for it as a feature request. Putting the results in the chat history for a node would solve some of the things people have asked for above

yea, how about the pop-up, but then also the pop-up results end up saved in the appropriate DM with text of all the info people are asking for. Pop just shows info GUVWAF posted. DM has all this plus any other info like timing as text.

Nestpebble:
> > I've already mentioned it in the discord, but apparently not asked for it as a feature request. Putting the results in the chat history for a node would solve some of the things people have asked for above
> 
> yea, how about the pop-up, but then also the pop-up results end up saved in the appropriate DM with text of all the info people are asking for. Pop just shows info GUVWAF posted. DM has all this plus any other info like timing as text.

Yep. They're a bit ephemeral at the minute, even if they succeed after you switch tabs they're lost.

[...]

PR #1248, opened 2024-09-15
Show results for two-way traceroute, including SNR values between hops

djholt:
This adds traceroute paths in both directions, including SNR values. This still uses a text dialog, but readability is improved as suggested in #1207. More UI improvements can be added once the functionality can be tested thoroughly with both new and old firmwares.

I've tested this with firmware 2.5 nodes and a routes with a combination of 2.5 and 2.4 nodes, but not using 2.4 nodes exclusively.


andrekir:
looks good, thanks

PR #1250, opened 2024-09-16
refactor: introduce `NodeEntity` protobuf-based database entity

andrekir:
This PR aims to decouple the `NodeInfo` API layer from the database by introducing a new protobuf-based entity.

By separating the internal database model from the API representation, this reduces tight coupling and makes it easier to update or modify the internal database without impacting the API. Using protobuf ensures changes to the project's protobuf definitions are available in the database without migrations or schema changes, improving database flexibility and reducing maintenance overhead.

Key changes:

- Introduces a protobuf-based `NodeEntity` as the new Room database entity;
- Replaces `NodeInfo` references with `NodeEntity` in the DAO layer, `NodeDB`, and `MeshService`;
- Decouples `NodeInfo` from the database schema, allowing future changes without impacting the API;
- Protobuf definitions are used to define the data format independently of the API.

Note:
- Further work is required to fully replace all instances of `NodeInfo` with `NodeEntity` exposed to the UI layer.

### DIFF
--- a/app/schemas/com.geeksville.mesh.database.MeshtasticDatabase/10.json
+++ b/app/schemas/com.geeksville.mesh.database.MeshtasticDatabase/10.json
@@ -1,0 +1,642 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 10,
+    "identityHash": "d3d6934fdce32237a4c8f2cb24455a59",
+    "entities": [
+      {
+        "tableName": "MyNodeInfo",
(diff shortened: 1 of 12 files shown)


jamesarich:
Nice. It'll be good to get this better aligned.

andrekir:
yeah. next big lift will be refactoring [/mesh/ui/NodeInfo.kt](https://github.com/meshtastic/Meshtastic-Android/blob/master/app/src/main/java/com/geeksville/mesh/ui/NodeInfo.kt) with the new `NodeEntity`.

PR #1253, opened 2024-09-17
feat: Add ability to display environment metrics in Fahrenheit

### LINKED ISSUE #1251: [Bug]: Metrics don't respect unit preferences
### Contact Details

@madeofstown

### What happened?

Loving the new metrics graphs but they only display temperature in Celsius. I believe they should respect the unit preferences from the Display settings section. 

### App Version

2.4.10

### Phone

Pixel 7a, GrapheneOS

### Device

Portduino, Heltec-Wireless-Tracker

### Firmware

2.5.0

### Relevant log output

_No response_

### Code of Conduct

jamesarich:
The temperature values in the
 environment metrics charts and cards are now displayed in Fahrenheit or Celsius based on the user's preference.
Celsius is still used as the base unit for calculations and storage.
 
resolves #1251

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/components/EnvironmentMetrics.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/components/EnvironmentMetrics.kt
@@ -36,27 +36,50 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.dp
 import com.geeksville.mesh.R
 import com.geeksville.mesh.TelemetryProtos.Telemetry
+import com.geeksville.mesh.copy
 import com.geeksville.mesh.ui.components.CommonCharts.ENVIRONMENT_METRICS_COLORS
 import com.geeksville.mesh.ui.components.CommonCharts.LEFT_CHART_SPACING
 import com.geeksville.mesh.ui.components.CommonCharts.MS_PER_SEC
 import com.geeksville.mesh.ui.components.CommonCharts.TIME_FORMAT
 
-
 @Composable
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/geeksville/mesh/ui/components/EnvironmentMetrics.kt
```diff
@@ -253,9 +274,9 @@ private fun EnvironmentMetricsCard(telemetry: Telemetry) {
                             style = TextStyle(fontWeight = FontWeight.Bold),
                             fontSize = MaterialTheme.typography.button.fontSize
                         )
-
+                        val textFormat = if (environmentDisplayFahrenheit) "%s %.1f°F" else "%s %.1f°C"
```
Robert-0410, line 357: Since we have to check if the user uses Fahrenheit here we could just do the conversion here instead of altering the telemetries.

jamesarich: The reason I was modifying the telemetries was to also pass it to the chart with the correct temperature values without branching and passing the boolean everywhere they're accessed or displayed on the chart.

Robert-0410: Got it, looks good from my perspective. :smiley:

andrekir:
got a few conflicts after merging #1252, should be good to go once they are resolved

jamesarich:
ok @andrekir - should be up to date

PR #1258, opened 2024-09-18
feat: Add IAQ to environment metrics

jamesarich:
Adds Indoor Air Quality (IAQ) to the environment metrics chart and card.

The IAQ is displayed as
 a green line on the chart and as a dot with a numerical value on the card.
The IAQ scale is available as a legend in the IAQ component.


b8b8:
This will be so useful. thanks.

PR #1260, opened 2024-09-19
Fix lat/lng/alt config validation errors when position is null

djholt:
This fixes persistent validation errors when setting fixed latitude, longitude and altitude when the Position object is null.

Root issue: state comparison always fails during error checking when passing a literal double or int value into EditTextPreference (in this case 0.0 for lat and lng, and 0 for altitude).


andrekir:
good catch

PR #1206, opened 2024-08-24
New feature: Request UserInfo

musznik:
The requestUserInfo function added to the Meshtastic Android project offers several key advantages:

https://github.com/user-attachments/assets/b3f863e8-df40-4284-95f2-11ab5fa08e1e

(preview from old dev version)

**Improved Data Completeness**: This function allows actively requesting user information when the longName and shortName are unknown. This ensures complete data for each node, especially when initial packets lack user details.

**Enhanced Reliability**: It addresses the issue of missing user information due to initial packet loss or communication challenges with distant nodes. This improves synchronization and data reliability across the network.

**Minimal Network Load**: The function is designed to minimize network traffic by limiting requests to a minimum interval of 5 minutes per node, preventing excessive network load.

The function works similarly to position requests, providing a unified approach to querying nodes and facilitating potential future expansions. Overall, this function enhances data reliability and completeness in the Meshtastic network without significantly impacting performance.

PR #1279, opened 2024-10-02
Adds Request user info

jamesarich:
Carrying on from @musznik's stale pull #1206 to request a node's info - fixed conflicts, using NodeEntity classes.


app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -1931,7 +1931,25 @@ class MeshService : Service(), Logging {
                 removeByNodenum = nodeNum
             })
         }
-
+        override fun requestUserInfo(packetId: Int, destNum: Int, sourceNodeInfo: NodeInfo) = toRemoteExceptions {
```
andrekir: we can simplify this a bit. we only need `destNum` as parameter since `packetId` is unused and the service has the nodeDB with our user info (already as MeshProtos.User).

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -1931,7 +1931,25 @@ class MeshService : Service(), Logging {
...
+            }.build()
+
+            if (destNum != myNodeNum) {
+                sendToRadio(newMeshPacketTo(destNum
+                ).buildMeshPacket(
+                    channel = nodeDBbyNodeNum[destNum]?.channel ?: 0
+                ) {
+                    portnumValue = Portnums.PortNum.NODEINFO_APP_VALUE
+                    wantResponse = true
+                    payload = protoSourceNodeInfo.toByteString()
```
andrekir:
``` kotlin
payload = nodeDBbyNodeNum[myNodeNum]!!.user.toByteString()
```

andrekir (changes requested):

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
jamesarich: Nice, I'll clean it up.

andrekir (approved):

PR #1284, opened 2024-10-03
Allow zooming past map source max

### LINKED ISSUE #926: [Improvement] View maps a zoom beyond tile max zoom
Background: When multiple nodes are close by each other their markers on the map overlap and some are not visible at some zoom levels.  I find this most often on the USGS Topo layer which only zooms to 15.  Since this is one of the few downloadable layers, it's not easily remedied.

Details:  When submitting a PR to increase the max zoom of the USGS Topo layer, I was informed that the USGS tile server doesn't work >15. When looking into this I found that their [map server spec](https://basemap.nationalmap.gov/arcgis/rest/services/USGSTopo/MapServer) specifies zoom levels up to 23, but in testing anything with a z of 16 and below seems to work just fine within the US (areas outside the US are max 12. (Side note, if anyone can give me more info on why it's limited to 15 instead of 16, I'd like to know)

Just for reference, here's an example tile at z=16
https://basemap.nationalmap.gov/arcgis/rest/services/USGSTopo/MapServer/tile/16/24567/13229

It looks like osmdroid supports zooming further than the max tile zoom. This can be seen in their demo app under the "Zoom Restricted Online Tiles" sample.  Is there interest in decoupling the max view zoom from the max layer tile zoom?  If so, I will work on a new PR.

danwelch3:
Resolves #926
When nodes are located close together, some of the map layers do not zoom in enough to view markers without overlapping.  This sets the max zoom level to at least 20.


andrekir:
the issue here is the local var for the tile provider `zoomLevelMax` and the MapView `maxZoomLevel` are tied together

https://github.com/meshtastic/Meshtastic-Android/blob/e7d675e831ee3d353367abbd8fd791d35db5c93a/app/src/main/java/com/geeksville/mesh/ui/map/MapFragment.kt#L522

setting the MapView max also sets the tile provider max to 20.0, which can throw errors downloading offline maps

to separate them you can set a fixed maxZoomLevel = 20.0 then then set `zoomLevelMax` in loadOnlineTileSourceBase() instead of `map.maxZoomLevel`

danwelch3:
@andrekir I see your point.  Thanks.  I addressed it by resetting the max map zoom when the "Download Region" is selected to the map source's max.  This has the benefit of zooming back out for the user and feels more user friendly to me.  Once the Cancel button is hit to close the download dialog, it re-applies the 20 or higher max zoom.

FYI, I'm using the coerceAtLeast to set 20 or higher because OSMDroid supports >20 on source layers.

andrekir:
- zoomLevelMax: is a local var controlling the max zoom level for tile providers used in offline downloads
- MapView.maxZoomLevel: is the maximum zoom level for the map view (that will to change to 20.0)

to separate them this line needs to be removed:

https://github.com/meshtastic/Meshtastic-Android/blob/e7d675e831ee3d353367abbd8fd791d35db5c93a/app/src/main/java/com/geeksville/mesh/ui/map/MapFragment.kt#L522

coerceAtLeast(20.0) will always return 20.0, so just set `maxZoomLevel` to 20.0 (same as [`minZoomLevel`](https://github.com/meshtastic/Meshtastic-Android/blob/master/app/src/main/java/com/geeksville/mesh/ui/map/MapFragment.kt#L643))

then set the local var `zoomLevelMax` directly from the tile provider in `loadOnlineTileSourceBase()`

danwelch3:
I made the changes you advised, but in testing found that the `CacheManager` call was now receiving 0.0 as the `zoomLevelMax` due to the it being reinitialized after `loadOnlineTileSourceBase()` runs.
https://github.com/meshtastic/Meshtastic-Android/blob/f689d772d62bf0bb7adc1ae27cb5fe77ca794136/app/src/main/java/com/geeksville/mesh/ui/map/MapFragment.kt#L534

I moved `zoomLevelMin` and `zoomLevelMax` outside the `MapView` function to avoid them being reinitialized.

Also changed this line, as it's reference to `zoomLevelDouble` could set the min level to a greater value than the max.
https://github.com/meshtastic/Meshtastic-Android/blob/f689d772d62bf0bb7adc1ae27cb5fe77ca794136/app/src/main/java/com/geeksville/mesh/ui/map/MapFragment.kt#L523
I set it to the source layers min zoom level to ensure that cached regions have smooth zooming at all levels.

andrekir:
I went ahead and made the necessary changes in https://github.com/meshtastic/Meshtastic-Android/commit/be6ea79c04e81463716d7fce52f29ea9878793f1. could you test it out and confirm it's working? thanks for the effort and let me know if you have any questions.

danwelch3:
Yes, that looks like it works.  I don't have a lot of Kotlin/Java experience, so seeing the `remember` API used in place of what I did was much cleaner.
https://github.com/meshtastic/Meshtastic-Android/blob/be6ea79c04e81463716d7fce52f29ea9878793f1/app/src/main/java/com/geeksville/mesh/ui/map/MapFragment.kt#L278-L279

I still think there some benefit to zooming out to the source max when initiating a download, so the user is shown a realistic view of what is being cached.

andrekir:
good to hear it's working. this part of the code is a unique mix for sure, and can be tricky to navigate. let's keep it simple, most users don't zoom all the way in to select areas for download.

I'm going to close this now that it's sorted out.

PR #1287, opened 2024-10-03
Implement marker clustering

danwelch3:
Implements M Kergall's [osmbonuspack](https://github.com/MKergall/osmbonuspack) Marker Clustering feature.

I was not able to implement the library as a dependency, as I could not customize the markers to match the markers currently used. I put the relevant parts of the library in with modifications to the markers.

image

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/map/MapFragment.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/map/MapFragment.kt
@@ -85,6 +85,7 @@ import org.osmdroid.views.overlay.Polygon
 import org.osmdroid.views.overlay.gridlines.LatLonGridlineOverlay2
 import org.osmdroid.views.overlay.infowindow.InfoWindow
 import org.osmdroid.views.overlay.mylocation.MyLocationNewOverlay
+import com.geeksville.mesh.model.map.clustering.RadiusMarkerClusterer
 import java.io.File
 import java.text.DateFormat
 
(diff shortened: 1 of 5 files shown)


jamesarich:
I'm unsure how copying code directly from other projects needs to be attributed, but it does look like the [license is pretty liberal](https://github.com/MKergall/osmbonuspack#licence).

danwelch3:
Agreed. I would prefer to link directly to the library if possible, so any suggestions on using the markers from com.geeksville.mesh.model.map.MarkerWithLabel instead of the default OSMDroid markers would be appreciated.

Its my impression that in terms of license, this is not a problem.  The osmbonuspack library is licensed under LGPLv3 which should be compatible with GPLv3.  [See GNU's license matrix here.](https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility)

image

app/src/main/java/com/geeksville/mesh/ui/map/MapFragment.kt
```diff
@@ -283,6 +290,9 @@ fun MapView(
     val map = rememberMapViewWithLifecycle(context)
     val state by model.mapState.collectAsStateWithLifecycle()
 
+    val nodeClusterer = RadiusMarkerClusterer(context)
+    map.overlays.add(nodeClusterer)
```
andrekir: please move this into `drawOverlays()` and should be ready to merge

danwelch3:
Moved `map.overlays.add(nodeClusterer)` to `MapView.drawOverlays()`.

I left the `nodeClusterer` definition where it was, otherwise it would break the `UpdateMarkers()` call.

DonRichie:
This change reduced the usability of my map. I can't see when I had contact with the nodes without zooming in.
If you insist on the feature, please add a button to turn it off.

danwelch3:
I'll look into adding a toggle.  The clustering shouldn't go into effect until nodes are overlapping visually though.  There might be some tuning of the cluster settings that needs done as well.  Can you attach screenshots of the same nodes with and without this feature showing the problem you're having?

andrekir:
there is no need for a toggle. the clustering was fine tuned in app version 2.5.1, please update.

PR #1292, opened 2024-10-07
Small improvement to the pi approximation.

Robert-0410:
Small pi improvement pointed out by users on Discord.

[Pi](https://en.wikipedia.org/wiki/Pi) is approximately equal to 3.14159

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/util/LocationUtils.kt
+++ b/app/src/main/java/com/geeksville/mesh/util/LocationUtils.kt
@@ -181,7 +181,7 @@ fun latLongToMeter(
     lat_b: Double,
     lng_b: Double
 ): Double {
-    val pk = (180 / 3.14169)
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/geeksville/mesh/util/LocationUtils.kt
```diff
@@ -181,7 +183,7 @@ fun latLongToMeter(
     lat_b: Double,
     lng_b: Double
 ): Double {
-    val pk = (180 / 3.14169)
+    val pk = (180 / PI)
```
andrekir, line 185: Kotlin has a built-in `Math.PI` constant you can use. let me know what you prefer.

Robert-0410: Cool, Ill use the built in one instead.

PR #1295, opened 2024-10-08
Adds support for known 0 hopsAway

jamesarich:
This change updates the database schema to allow storing null values for the 'hopsAway' field in the 'nodes' table.

This allows for better representation of unknown hop counts instead of defaulting to 0.
The UI is also updated to display signal information and node details accordingly when 'hopsAway' is null _and_ zero.
image


app/src/main/java/com/geeksville/mesh/ui/SignalInfo.kt
```diff
@@ -27,7 +27,7 @@ fun signalInfo(
     } else {
         buildList {
             if (node.channel > 0) add("ch:${node.channel}")
-            if (node.hopsAway == 0) {
+            if (node.hopsAway == null) {
```
ianmcorvidae: So I'm assuming this is the key change and the rest above this is entirely stuff to make sure the DB stores nulls separate from zeros and that makes its way here for display. Let me know if that assumption is off!

I think we still want to show SNR & RSSI for nodes with both null and 0 hopsAway -- with null, it's sort of a best-effort equivalent to the current behavior; if the node's older, they might be meaningless, but that's just the risk. For actual 0, we definitely want it shown, since that's the case where the SNR/RSSI are definitely useful.

I think the summary is probably:

* display SNR/RSSI for both null and 0
* either display "Hops Away: 0" for those with an explicit 0 or display "Hops Away: unknown" for those with a null (or some other display that conveys this)
* make sure we're doing proper detection of the hopsAway/pulling it appropriately from nodedb, which I think is missing at present. In particular, in MeshService in processReceivedMeshPacket, the logic should be: if packet.hopStart == 0 or if hopLimit < hopStart (invalid), set hopsAway to null; else, set hopsAway to hopStart - hopLimit. I'm not sure where the code that instantiates this stuff on initial wantconfig lives in android, so that might be automatic there, but if not it should be doing something like (all pseudocode, heh): if !nodeInfo.hasHopsAway(): newNode.hopsAway = null; else: newNode.hopsAway = nodeInfo.hopsAway

ianmcorvidae (commented):
I'm assuming the DB stuff is right since I'm not really up on that part. I think there's an update needed in MeshService that I've outlined, plus making sure the wantconfig NodeInfo get initialized properly, and the display needed is slightly different than this

app/src/main/java/com/geeksville/mesh/ui/SignalInfo.kt
jamesarich: Gotcha - that helps, thank you!

jamesarich: Ok @ianmcorvidae - I updated the UI logic to: display `Hops Away: Unknown`,  include rssi/snr values for `hopsAway == null || hopsAway == 0`, and am explicitly checking the proto `hasHopsAway` while converting an info packet to our local entity.

Let me know if I missed anything else - Andre will review as well for the rest of the infra/db changes.

ianmcorvidae (approved):
Looks good as far as what I know at least!

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
```diff
@@ -164,7 +164,7 @@ data class NodeInfo(
     var deviceMetrics: DeviceMetrics? = null,
     var channel: Int = 0,
     var environmentMetrics: EnvironmentMetrics? = null,
-    var hopsAway: Int = 0
+    var hopsAway: Int? = null
```
andrekir: avoid making any changes to `NodeInfo`, this would break the API

andrekir (changes requested):
also please undo the protobuf update, then should be ready to merge

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
jamesarich: @andrekir it feels kindof off to be setting the `NodeInfo.hopsAway` to 0 when it's known null.. Is there another way you were considering?

andrekir: current behavior remains the same, so there's no loss. and it's a breaking change. with the new nodeDB overhaul the API parcelables should not change at all.

one option is using -1 instead of null in both `NodeEntity` and `NodeInfo`, but whether that adds enough value to justify the added refactor is up to you.

andrekir (approved):

app/src/main/java/com/geeksville/mesh/NodeInfo.kt
jamesarich: Good to merge, we can revisit for -1 if we want.

ianmcorvidae: (not an urgent question, but for when someone has a chance to explain)

what's the practical impact of not updating `NodeInfo` but only `NodeEntity`? Does this mean that null hopsAway will only get populated from over-the-mesh packets (and not from initial wantconfig/nodedb), that external integrations using the AIDL stuff will be unable to see nulls (unless the -1 thing gets added, but ignoring that for now), something else? 

Trying to slowly become a bit more aware of how things are wired up in this app, appreciate any response :)

jamesarich: [Refactor_hopsAway.PATCH](https://github.com/user-attachments/files/17311997/Refactor_hopsAway.PATCH)

@andrekir - I had a minute to work up a patch to make the `-1` change, also reverting the db change since it's no longer necessary. Give it a look?

jamesarich: Just in case we want to not have a(nother) db migration in this pr :sweat_smile:

andrekir: @jamesarich patch looks really good! I'll wait for you to push those.

at a glance I think it's only missing revert the db version bump to 14 in `MeshtasticDatabase` and the Elvis operator (= hopsAway ?: -1,) in `NodeEntity.toNodeInfo()`.

andrekir: @ianmcorvidae only external API clients like the ATAK Plugin would not see null hopsAway.

jamesarich: > @jamesarich patch looks really good! I'll wait for you to push those.
> 
> at a glance I think it's only missing revert the db version bump to 14 in `MeshtasticDatabase` and the Elvis operator (= hopsAway ?: -1,) in `NodeEntity.toNodeInfo()`.

@andrekir - applied patch, with those 2 changes - should be good to go 🤞

PR #1340, opened 2024-10-23
Signal Metrics

Robert-0410:
### Major
Added a signal graph that plots the rx snr and rssi of a mesh packet.

### Minor
- Moved the time labels above and outside of the graph area.
- Implemented info dialogs for all existing graphs.
- Implemented a signal indicator to visualize the signal quality for each data entry.

Screenshot from 2024-10-23 00-01-55

### Concerns
Retrieving the data from the DB, do I need to filter anything out with respect to hops or from?

### TODO
Planning on completing this during the review.
- Since we don't receive mesh packets from our connected node, the nav card shouldn't be available when clicking on More details for our nodes.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/database/MeshLogRepository.kt
+++ b/app/src/main/java/com/geeksville/mesh/database/MeshLogRepository.kt
@@ -1,6 +1,7 @@
 package com.geeksville.mesh.database
 
 import com.geeksville.mesh.Portnums
+import com.geeksville.mesh.MeshProtos.MeshPacket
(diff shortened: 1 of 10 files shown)


andrekir:
Looks good.

> Retrieving the data from the DB, do I need to filter anything out with respect to hops or from?

I'm working around the area and will update the DB query and filter soon.

> Since we don't receive mesh packets from our connected node, the nav card shouldn't be available when clicking on More details for our nodes.

The filter should take care of this case, except when users switch between radios. Should be fine for now.

PR #1350, opened 2024-10-25
fix: Display for self-sent messages

### LINKED ISSUE #1349: [Bug]: local name wrong and no cloud icon on sent messages
### Contact Details

valzzu_

### What happened?

local name shows as _ocal_
and messages I send no longer has the cloud icon showing the status of the message
image0

### App Version

2.5.4

### Phone

galaxy A55

### Device

T1000-E

### Firmware

2.5.6

### Relevant log output

_No response_

### Code of Conduct

jamesarich:
fixes #1349

This commit updates the `getUser` function in `NodeDB` to set the ID of the user when creating a new `User` object, ensuring that the user ID is always populated.

Additionally, it refactors the `Message` class to use the proto `User` object instead of the `MeshUser` class. This change simplifies the code and improves consistency by utilizing the proto definition for user information.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/model/Message.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/Message.kt
@@ -1,7 +1,7 @@
 package com.geeksville.mesh.model
 
+import com.geeksville.mesh.MeshProtos
 import com.geeksville.mesh.MeshProtos.Routing
-import com.geeksville.mesh.MeshUser
 import com.geeksville.mesh.MessageStatus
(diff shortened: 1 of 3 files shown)


andrekir:
Thanks for the effort but the issue here is just a missing `setId()` in `nodeDB.geUser()`. If you’re up for it, please update the PR. If not, no worries, I should get to it tomorrow.

jamesarich:
@andrekir - thank you, i missed that.

PR #1381, opened 2024-11-04
feat: show unique messaging notifications per contact

jamesarich:
Instead of a single notification for all messages, each contact now has its own, unique notification. This uses the `NotificationCompat.MessagingStyle` and the contact's name to create distinct notifications, enhancing message organization.

This is a move towards richer messaging notifications.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshServiceNotifications.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshServiceNotifications.kt
@@ -12,6 +12,7 @@ import android.media.RingtoneManager
 import android.os.Build
 import androidx.annotation.RequiresApi
 import androidx.core.app.NotificationCompat
+import androidx.core.app.Person
 import com.geeksville.mesh.MainActivity
 import com.geeksville.mesh.R
 import com.geeksville.mesh.TelemetryProtos.LocalStats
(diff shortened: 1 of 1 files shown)


jamesarich:
@andrekir I think this is ready for your perusal - I _did_ include a commit to include navigation to the conversation via observed `contact_key` in the UIViewModel. Easy enough to back out those changes if we find a better way to do that.

jamesarich:
actually hold up - i think i can access the fragmentmanager directly from MainActivity..

app/src/main/java/com/geeksville/mesh/service/MeshServiceNotifications.kt
```diff
@@ -165,10 +167,10 @@ class MeshServiceNotifications(
         )
     }
 
-    fun updateMessageNotification(name: String, message: String) =
+    fun updateMessageNotification(contactKey: String, name: String, message: String) =
         notificationManager.notify(
-            messageNotifyId,
-            createMessageNotification(name, message)
+            name.hashCode(), // show unique notifications,
```
andrekir: we could use `contactKey` instead of `name` here

jamesarich: added that change in as well

jamesarich:
[Screen_recording_20241104_155552.webm](https://github.com/user-attachments/assets/628b3d02-5441-4a35-b59c-d8abc52cb539)

PR #1389, opened 2024-11-08
feat: add text/plain intent to send message 

remi-espie:
This PR adds a text/plain intent that allow user to share a piece of text to a Meshtastic node through the application.

This PR is currently a WIP. As of right now, only the intent was added. 

The PR idea would be 2 fold:
1. Add a way to easily share the text to a user-selected contact. For this, I thought of adding an optional `txtToFill` variable to the [`navigateToMessages`](https://github.com/meshtastic/Meshtastic-Android/blob/227c65f1914cd5b4b379e94070bafc0463b08eb5/app/src/main/java/com/geeksville/mesh/ui/MessagesFragment.kt#L40) function (or create a new function altogether). If not empty, this variable would autofill the text input of the selected contact.
2. Create some kind of JSON struct that would be auto parsed, like 
```json
{
"contactKey": string,
"contactName": string,
"message": string,
"autosSend": bool
}
```

Upon completion, this PR would solve #1062 .

### DIFF
--- a/.idea/inspectionProfiles/Project_Default.xml
+++ b/.idea/inspectionProfiles/Project_Default.xml
@@ -2,6 +2,7 @@
   <profile version="1.0">
     <option name="myName" value="Project Default" />
     <inspection_tool class="PreviewAnnotationInFunctionWithParameters" enabled="true" level="ERROR" enabled_by_default="true">
+      <option name="composableFile" value="true" />
       <option name="previewFile" value="true" />
     </inspection_tool>
     <inspection_tool class="PreviewDimensionRespectsLimit" enabled="true" level="WARNING" enabled_by_default="true">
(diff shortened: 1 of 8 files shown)


jamesarich:
I have been thinking about this in conjunction with direct reply from the notification - I'm hoping that these would use similar methods/intents to collect and deliver to the appropriate places.

remi-espie:
I decided to scale down this PR to solely be able to share a string of text, removing the json parser auto sending messages.
I had trouble with the auto sending of messages. It may be a part of a future PR.

This PR adds:
- the ability to share text content to a contact
- a new screen to share to
- solve #1062

Neo-Oli:
@remi-espie Do you still intend to create this auto send feature? I would very much like a way to fully automate the sending of messages from another app.

remi-espie:
Hello,
While working on that, I saw that Meshtastic has a broadcast service. I didn't find in the the documentation, but the code is pretty straightforward (see [MeshServiceBroadcast](app/src/main/java/com/geeksville/mesh/service/MeshServiceBroadcasts.kt)). 
I don't plan to work on this feature right now, but I think it would be better to expand the broadcast service instead of an intent to send message automatically (as it would make sending messages programmatically easier).

PR #1449, opened 2024-12-06
feat: add support for Device Hardware json and svg graphics

jamesarich:
Adds device_hardware.json, DeviceHardware model, svg device graphics, and convenience methods to parse the json and map it to DeviceHardware model and image resource.

Allows for better hardware device display names, graphics, and indication of support.

image


andrekir:
Please move the images from the detail section to a header at the top of the screen. The header should place the image in focus without overwhelming the rest of the screen, so it also needs to work well with both horizontal and vertical images. Maybe use a complementary background with the hardware name under the image.

jamesarich:
@andrekir - i moved the relevant device info up to a new `Device` section above the `Details`. I also restricted the device image size and gave it a circular background with the node color (alpha 50%).

app/src/main/java/com/geeksville/mesh/model/DeviceHardware.kt
```diff
@@ -0,0 +1,45 @@
...
+        ): DeviceHardware? {
+            val json =
+                context.assets.open("device_hardware.json").bufferedReader().use { it.readText() }
+            val deviceHardware = Json.decodeFromString<List<DeviceHardware>>(json)
+            return deviceHardware.find { it.hwModel == hwModel.number }
+        }
+    }
+
+    fun getDeviceVectorImage(context: Context): Int {
+        val resourceId = context.resources.getIdentifier(
```
andrekir: this isn't working, the JSON image field uses dash (-) and the filenames have underscores (_).

getIdentifier() reflection is very unreliable, use `when` instead.

app/src/main/java/com/geeksville/mesh/ui/NodeDetail.kt
```diff
@@ -196,7 +214,51 @@ private fun NodeDetailRow(label: String, icon: ImageVector, value: String) {
 }
 
 @Composable
-private fun NodeDetailsContent(node: NodeEntity) {
+private fun DeviceDetailsContent(
+    node: NodeEntity,
+) {
+    val context = LocalContext.current
+    val deviceHardware =
+        DeviceHardware.getDeviceHardwareFromHardwareModel(context, node.user.hwModel)
```
andrekir: this can move to the ViewModel and use app context. also needs some exception handling.

andrekir (changes requested):

jamesarich:
@andrekir - I moved the DeviceHardware and image res functions to MetricsViewModel, and added some State variables.

The DeviceHardware json parsing now has exception handling and is cached in the model to avoid expensive rereads.

I also updated the image res function to use `when` mapping from HardwareModel enum to Drawable resource id. This function will need to be updated when new images are added.

andrekir (approved):

PR #1456, opened 2024-12-11
Time breaks

Robert-0410:
### Changes
- `Feature`: Drawing the battery line only when the data is continuous with out a large time separation.
- `Refactor`: Wrote helper functions to help plot data points and draw lines.

### Demo

https://github.com/user-attachments/assets/43e72c89-61a7-4028-ad50-f6b8be602382


Kealper:
I like this. On the one hand, I feel like it would be nice to have it draw lines in between those gaps so it kind of guesses dead-reckoning-style, but on the other hand I actually think the gaps there makes it much more obvious that there wasn't just linear data points and that there actually isn't data for those times. 👍

PR #1460, opened 2024-12-11
feat(config): implement excluded modules validation

andrekir:
- https://github.com/meshtastic/protobufs/pull/614


jamesarich:
Nice! This should get us closer to displaying the firmware version for administered nodes via the device metadata on Admin messages, yeah?

andrekir:
For sure. Still working out the final design, but it's a good start.

PR #1463, opened 2024-12-13
Bluetooth reconnect fix android15

tavdog:
This pull fixes two issues.
1.  The necessity to click None before manually selecting a reconnect from the BT device popup.
2.  The prolonged disconnection non-reconnect behaviour is fixed by trapping BT disconnect code 147 and calling lostConnection() function.  This sets up a reconnect attempt loop every 30 seconds. This might be power draining. Could add a user preference somewhere to do this or not. Built apk is available here for testing : https://wildc.net/tmp/meshtastic-fdroid-debug.apk

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt
+++ b/app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt
@@ -259,7 +259,7 @@ class RadioInterfaceService @Inject constructor(
      * @return true if the device changed, false if no change
      */
     private fun setBondedDeviceAddress(address: String?): Boolean {
-        return if (getBondedDeviceAddress() == address && isStarted) {
+        return if (getBondedDeviceAddress() == address && isStarted && isConnected) {
             warn("Ignoring setBondedDevice ${address.anonymize}, because we are already using that device")
             false
         } else {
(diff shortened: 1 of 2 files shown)


Franselbaer:
Thank you. Perhaps a on/off button and/or reconnect timeout set (30 sec - 300 sec) would be a good idea. Great work.

tavdog:
I just tested these changes on  a Pixel 3 - Android 12 - API 31 and it doesn't effect its ability to reconnect.
I'll be doing a battery test tonight to see if the new changes drain the battery noticeably.

app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt
```diff
@@ -259,7 +259,7 @@ class RadioInterfaceService @Inject constructor(
      * @return true if the device changed, false if no change
      */
     private fun setBondedDeviceAddress(address: String?): Boolean {
-        return if (getBondedDeviceAddress() == address && isStarted) {
+        return if (getBondedDeviceAddress() == address && isStarted && isConnected) {
```
andrekir: This is intended behavior. We specifically track `isStarted` just for this call. Please revert this change, and it should be good to merge.

andrekir (changes requested):

andrekir (approved):

PR #1490, opened 2024-12-19
refactor: Improve Reaction UI 

jamesarich:
Moves the ReactionRow below the MessageItem within a Column using negative padding to provide slight overlap.

- Updates ReactionItem UI to a more modern style with borders and rounded corners.
- Implements an ellipsis indicator for overflowing reactions, allowing users to expand and view more.
- Changes the reaction button to an icon that includes a `plus`, indicating the ability to add a reaction.

https://github.com/user-attachments/assets/f4f82e7d-610c-49ad-acee-accebf5962b8


app/src/main/java/com/geeksville/mesh/ui/message/components/MessageList.kt
```diff
@@ -95,35 +99,40 @@ internal fun MessageList(
...
-                NodeMenu(
-                    node = msg.node,
-                    showFullMenu = true,
-                    onDismissRequest = { expandedNodeMenu = false },
-                    expanded = expandedNodeMenu,
-                    onAction = onNodeMenuAction
-                )
+            Column(
+                modifier = Modifier.padding(vertical = 8.dp),
+                verticalArrangement = Arrangement.spacedBy((-12).dp)
```
andrekir: Avoid adding Columns inside LazyColumns, use spacing modifiers like padding or offset directly in `ReactionRow` where it's needed.

andrekir (changes requested):
One small change and should be ready to merge.

app/src/main/java/com/geeksville/mesh/ui/message/components/MessageList.kt
jamesarich: I couldn't find much documentation on why this should be avoided - but it was easy enough to swap out for offset and zindex.

andrekir (changes requested):
the reaction placement doesn't look right.

https://github.com/user-attachments/assets/01f885b2-70d6-4df1-995b-457d7627f7a3

jamesarich:
> the reaction placement doesn't look right.
> 
>  Screen_Recording_20250104_105825_Meshtastic.mp4

wow i didn't even catch that it was the row above... 😮‍💨 
image

jamesarich:
@andrekir  - after a lot of trial and error I _think_ I found a solution that avoids using a nested Column... it does however need to calculate the padding based on the message height, which feels terribly hacky to me.

I never did find a very great argument against nesting a Column within a LazyColumn item. When I asked Gemini: 

> Nesting a Column inside a LazyColumn like that is generally acceptable and is a common pattern in Jetpack Compose.
The LazyColumn composable is designed for efficiently displaying a large number of items by only composing and laying out the visible items. Each item within the LazyColumn's items block is individually composed, and it's perfectly fine to use a Column or other layout composables to structure the content of each item.
In your example, the Column is used to arrange the MessageItem and the FlowRow vertically within each item of the LazyColumn. This is a good way to structure your UI and it won't lead to the issues associated with nesting two lazy lists.

I'm inclined to revert to the Column implementation unless you have a different solution in mind or you're ok with this padding calculation.

PR #1498, opened 2024-12-21
fix: Adds label to alert bell quick action

jamesarich:
Use a string resource for the "Alert" text next to the bell icon in the quick action bar. Also align items vertically in the LazyRow


andrekir:
Any context for this change? Currently quick chat labels are limited to 3 characters.

jamesarich:
This was a request from @garthvh

jamesarich:
@garthvh is this still wanted?

PR #1500, opened 2024-12-22
Refactor: 24H and 48H intervals

Robert-0410:
From both using 1 hour intervals to 6 and 12 hours respectively.

### Demo

https://github.com/user-attachments/assets/f0c4dbf7-5dd9-4cdb-a87b-06aef7de1563


b8b8:
Looking good. Much more compact layout is easier to digest the data.

PR #1515, opened 2024-12-31
feat: Show ALERT_APP notifications and override silent mode

jamesarich:
This adds support for `ALERT_APP` messages.  

Includes a new Alert notification channel, and requests permissions to override Do Not Disturb for the Alert notification channel.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,5 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
-<!--
+<?xml version="1.0" encoding="utf-8"?><!--
      Copyright (c) 2024 Meshtastic LLC
 
      This program is free software: you can redistribute it and/or modify
@@ -29,14 +28,17 @@
         android:required="false" />
 
     <!-- Request legacy Bluetooth permissions on older devices -->
-    <uses-permission android:name="android.permission.BLUETOOTH"
+    <uses-permission
+        android:name="android.permission.BLUETOOTH"
         android:maxSdkVersion="30" />
-    <uses-permission android:name="android.permission.BLUETOOTH_ADMIN"
+    <uses-permission
+        android:name="android.permission.BLUETOOTH_ADMIN"
         android:maxSdkVersion="30" />
 
     <!-- API 31+ Bluetooth permissions -->
     <uses-permission android:name="android.permission.BLUETOOTH_CONNECT" />
-    <uses-permission android:name="android.permission.BLUETOOTH_SCAN"
+    <uses-permission
+        android:name="android.permission.BLUETOOTH_SCAN"
         android:usesPermissionFlags="neverForLocation"
         tools:targetApi="s" />
 
@@ -73,6 +75,9 @@
     <!-- zxing library for QR Code scanning using camera -->
     <uses-permission android:name="android.permission.CAMERA" />
 
+    <!-- Needed for AlertApp to override Do Not Disturb -->
+    <uses-permission android:name="android.permission.ACCESS_NOTIFICATION_POLICY" />
+
     <uses-feature
         android:name="android.hardware.camera"
         android:required="false" />
@@ -92,13 +97,13 @@
         android:allowBackup="false"
         android:dataExtractionRules="@xml/data_extraction_rules"
         android:fullBackupContent="false"
+        android:hardwareAccelerated="true"
         android:icon="@mipmap/ic_launcher2"
         android:label="@string/app_name"
+        android:localeConfig="@xml/locales_config"
         android:roundIcon="@mipmap/ic_launcher2_round"
         android:supportsRtl="true"
-        android:hardwareAccelerated="true"
-        android:theme="@style/AppTheme"
-        android:localeConfig="@xml/locales_config">
+        android:theme="@style/AppTheme">
 
         <!-- Default crash collection and analytics off until we (possibly) turn it on in application.onCreate -->
         <meta-data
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/geeksville/mesh/service/MeshServiceNotifications.kt
```diff
@@ -42,10 +44,19 @@ class MeshServiceNotifications(
     private val context: Context
 ) {
 
+    val notificationLightColor by lazy {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+            context.getColor(R.color.colorPrimary)
+        } else {
+            Color.GREEN
```
jsoberg: Just verifying we want this green instead of blue since it was blue before 👍

app/src/main/java/com/geeksville/mesh/MainActivity.kt
```diff
@@ -426,16 +441,45 @@ class MainActivity : AppCompatActivity(), Logging {
                     service.startProvideLocation()
                 }
             }
+            checkNotificationPermissions()
```
jsoberg, line 435: Side-note: I noticed that the base "Meshtastic needs permissions for service and message notifications" dialog shows above this if you're installed and have the notifications permissions off - trying to repro but it acted strangely and didn't go away after I accepted permissions 🤔

app/src/main/res/values/strings.xml
```diff
@@ -313,6 +314,10 @@
     <string name="unknown_age">Unknown Age</string>
     <string name="copy">Copy</string>
     <string name="alert_bell_text">Alert Bell Character!</string>
+    <string name="dnd_required">Do Not Disturb Permission Required</string>
+    <string name="why_dnd_required">Critical Alerts requires Mode Control permission to work.\n To grant Modes Control permission, tap on \'Open settings\' > Allow Modes access to \'Meshtastic\' app.</string>
```
jsoberg: I'm not entirely sure where this is supposed to go tbh - on my S23 I don't have modes control in my settings 🤔 is there an Intent we could fire to take them to this particular setting depending on their device manufacturer (though I'd imagine that would get dicey real quick)?

Would this use the same flow that something like PagerDuty does to override DND settings? It's been awhile since I've set it up fresh but I know PagerDuty has a decent setup flow that directs the user to the settings they need to enable - might be biting off more than we can reasonably chew but I wonder if we could replicate that.

jsoberg: OK I found the flow that lets you override DND on my S23:

<video src="https://github.com/user-attachments/assets/d786da9a-5d1b-4673-9312-ee8277b99685" width="300"></video>

I'm going to disable PagerDuty (pray for me 🙏 ) and see if it automatically redirects me

jsoberg: Eh even PagerDuty doesn't directly open to this, it just opens to the App's own notifications which don't have a DND override:

<video src="https://github.com/user-attachments/assets/0ed09e5d-d785-482f-8c3f-52feb68a0e41" width="300"></video>

app/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!--
-     Copyright (c) 2025 Meshtastic LLC
+     Copyright (c) 2024 Meshtastic LLC
```
todd2982: This change should be removed

app/src/main/java/com/geeksville/mesh/MainActivity.kt
jamesarich: i'll play with it a bit to see if i can delay the dnd dialog after accepted general notification perms.

app/src/main/res/values/strings.xml
jamesarich: There wasn't a ton to find out there on overriding dnd on just vanilla android.. it does not surprise me that samsung would be a headache here. Lmk if you have any better ideas to handle this.

app/src/main/java/com/geeksville/mesh/service/MeshServiceNotifications.kt
jamesarich: Set it back to blue for now. Meshtastic green LED could be nice to align with theme colors though.

app/src/main/java/com/geeksville/mesh/MainActivity.kt
jamesarich: I reworked this to fire after general notif perms are accepted - seems a little less clunky. Mind giving it another whirl @jsoberg ?

app/src/main/res/values/strings.xml
jamesarich: I'm currently having second thoughts about the UX here. We can avoid the permissions (manifest and request) and make DnD entirely opt in via app channel notification settings.

I found the intent to direct them to the system app channel settings screen 'ACTION_CHANNEL_NOTIFICATION_SETTINGS', though Samsung may still be a headache... I don't have a Samsung device for testing.

Reworking this to just inform the user and attempt to direct them to the channel settings.

jamesarich:
Reworked this a bit, no longer needs extra permission - just prompts user to change channel dnd setting, and gives them the option to open directly to channel setting screen. Curious to see how this works on Samsung devices.


https://github.com/user-attachments/assets/b03be5b2-108d-4d6d-9f9b-1fba74cc9403

jsoberg:
Looks great! I'm trying to find the best bet for branching on Samsung - the Intent that you have now opens the expected notifications channel page:




But Samsung being Samsung this doesn't include DND override settings. I did some grokking and saw that `ACTION_NOTIFICATION_POLICY_ACCESS_SETTINGS ` opens to a DND override flow, and it works on my Samsung, but Meshtastic doesn't show in the list of available apps to override 🤔

jamesarich:
Added additional dialog text and an action to open Samsung help webpage when device is detected as a Samsung.

https://github.com/user-attachments/assets/40bf00eb-d36a-4958-8378-add7c095bdd3

jsoberg:
LGTM! In the future if we find more manufacturers have differing settings, we could try to detect manufacturers and present different dialog/options per type - for now this looks great

jamesarich:
I decided to follow my own lead on #1611 and use clickable links in the dialog text. I also decided not to use device type branching.

shortwavesurfer2009:
Could someone briefly summarize exactly what this is for? I am unable to access the Discord discussion because I do not have an account there and do not wish to have one.

My first thought upon hearing this is something like if somebody were to change their name to SOS, like they needed help, you would get a critical alert or something. But I'm not totally sure.

Issue #1432, opened 2024-11-26
[Enhancement]: migrate `SettingsFragment` to Compose


materoy:
### Contact Details

materoy@proton.me

### Tell us your idea.

This is a great app with good stability but the UI is really lacking, especially the home screen with the tabs at the to and settings screen could use some bit of user experience improvement. 
I have started working on improving them and the PR will be available soon. I just wanted to put it here first so that maintainers would know. 

![image](https://github.com/user-attachments/assets/b0ac57e7-1bc8-402d-ba47-fdacd5dd0387)

This screenshot shows my first improvements but more to com. Please let me know if I'm going in the right direction. 

Cheers 
~ Matero

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

andrekir:
Thanks for your interest in improving the app! We’re almost done migrating all Fragments to Compose, with only `SettingsFragment` and parts of `ContactsFragment` (ActionMenu) and `MessagesFragment` (ActionMenu and a few components) left.

We'll likely need to finish migrating `MessagesFragment` and `ContactsFragment` before we can replace the ViewPager. How do you plan to integrate your changes with the existing Compose navigation?

materoy:
Well then if we are taking the bottom up uproach, I can work on the `SettingsFragment` and migrate it to compose, unless someone else is working on it

andrekir:
Sounds good! The "Report Bug" button will be removed, so feel free to take it out during the refactor.

Let me know if you have any questions.

b8b8:
My thumbs thank you for wanting to move the tabs to the bottom.

PR #1516, opened 2025-01-01
refactor: SettingsFragment to compose

materoy:
- #1432 

This pr is a refactor of `SettingsFragment` to `SettingsScreen` in compose

Still in draft phase, almost complete, feel free to review, suggest edits and bug fixes


andrekir:
Thanks for your contribution! As a general guideline, we want to follow the design of a traditional Settings/Preference screen, making use of our existing "Preference" components (like [SwitchPreference](https://github.com/meshtastic/Meshtastic-Android/blob/master/app/src/main/java/com/geeksville/mesh/ui/components/SwitchPreference.kt) and [DropdownPreference](https://github.com/meshtastic/Meshtastic-Android/blob/master/app/src/main/java/com/geeksville/mesh/ui/components/DropDownPreference.kt)).

The BLE/TCP/Serial device scan/selection could probably move to an item in the Settings screen to maintain a cleaner layout, but open to ideas and suggestions on how best to organize this.

Items currently in the overflow menu, like Theme and Language, will also be moved into this screen in the future.

app/src/main/java/com/geeksville/mesh/model/BTScanModel.kt
```diff
@@ -76,32 +100,34 @@ class BTScanModel @Inject constructor(
             usbRepository.serialDevicesWithDrivers,
             showMockInterface,
         ) { ble, tcp, usb, showMockInterface ->
-            devices.value = mutableMapOf<String, DeviceListEntry>().apply {
-                fun addDevice(entry: DeviceListEntry) { this[entry.fullAddress] = entry }
+            suspend fun addDevice(entry: DeviceListEntry) {
+                _uiState.emit(uiState.value.copy(devices = uiState.value.devices + (entry.fullAddress to entry)))
```
andrekir: use MutableStateFlow `.update {}` instead of `.emit`

PR #1517, opened 2025-01-02
feat: add request metadata option

andrekir:
continued work on #1460


jamesarich:
closed in favor of #1896

PR #1520, opened 2025-01-03
feat: Initial implementation of adding nodes to favorites

### LINKED ISSUE #1011: [Feature Request] Favorite Nodes in Android
I believe in iOS there is the ability to "favorite" a node, so that it does not fall off of your NodeDB. It would be nice to have something similar in Android :)

Kealper:
Initial implementation of support for favorite nodes. Adds support for and closes #1011

I've tested this in the simulator while it was connected to a real node and it seems to work as you'd expect, but it'll need a second pair of eyes on it because I'm not familiar with Kotlin and this is my first time working with it.


andrekir:
Thanks for your contribution! We're in the process of moving some of the `NodeMenu` items to the node details screen, as it is getting too crowded. Once #1517 is merged we'll revisit this.

Kealper:
> Thanks for your contribution! We're in the process of moving some of the `NodeMenu` items to the node details screen, as it is getting too crowded. Once #1517 is merged we'll revisit this.

Sounds good! Once that commit is in, I'll migrate this to the new spot.

jamesarich (commented):
Looks great for a first time Android/Kotlin first-timer! Thanks for dipping your toes in, always happy to have more help.

PR #1542, opened 2025-01-11
refactor: Lower the bluetooth timeout to 15 seconds

jamesarich:
Reduces bt timeout from 20s to 15s

reequested by @garthvh for testing against https://github.com/meshtastic/firmware/issues/5491

👀  @b8b8


jamesarich:
@thebentern closing as per our chat about firmware solution

PR #1543, opened 2025-01-11
feat: Add RAK2560 hardware model and graphic

jamesarich:
Adds the RAK2560 to the list of supported devices and associates it with the corresponding vector drawable.
re: @thebentern


thebentern (approved):

jamesarich:
@andrekir should I add this flavor to the detekt and lint runs as well?

PR #1575, opened 2025-02-08
Introduction of stable Compose UI State and some simple animations in Debug Panel

jsoberg:
Hey all!

This PR introduces stable UI state in the debug panel, as well as some simple animations. 

Using the debug panel screen as a starting point, this PR adds the conversion of the database model (`MeshLog` in this instance) to an explicitly stable UI state. There are more details on why this is important [here](https://developer.android.com/develop/ui/compose/performance/stability), but a stable UI state allows Compose to better control recompositions/etc, and remove a direct dependency of the UI to the database.

Additionally, this introduces some simple animations to the debug panel (animated scroll/placement). The difference is subtle and doesn't come through well in video, so running on device will be a better comparison - however, here are recordings for posterity:

| Before | After |
| ------ | ------ |
| <video src="https://github.com/user-attachments/assets/fe880dd3-4909-4589-8d62-e5406dcd556d" width="300"></video> | <video src="https://github.com/user-attachments/assets/bca72b2b-0363-401a-a281-c026fc4f2995" width="300"></video> |

I'll open a discussion topic about introducing this further in the project - this is just an example. Thanks!


jsoberg:
Added a discussion topic about this [here](https://github.com/orgs/meshtastic/discussions/228)

jamesarich (approved):
I like where this is headed. We'll have to tackle this in quite a few places, I think. There are some UI models  already defined that could benefit from immutability / more stable compose state.

PR #1597, opened 2025-02-19
Graphs Blue Color

Robert-0410:
Using the same blue color in both the power and environment metrics. This blue variation works better with dark themes.


jamesarich (commented):
would it make sense to hoist these various colors into the theme to keep them centralized?  Something like https://stackoverflow.com/a/75557863 ?

jamesarich (approved):

PR #1607, opened 2025-02-23
feat: Introduce Version Catalog for Dependency Management

jamesarich:
This introduces a version catalog via `libs.versions.toml` to centralize and manage project dependencies. It replaces the previous hardcoded dependency versions in `.gradle` files with references to the version catalog. This change enhances maintainability and organization by centralizing dependency management.

- Added `gradle/libs.versions.toml` to define dependencies, plugins, and their versions.
- Updated `build.gradle` to utilize references from the version catalog,
- Updated `app/build.gradle` to utilize references from the version catalog.
- Removed hardcoded version strings in gradle files.
- Added note and suppressed warning to avoid updating `zxing-core` dependency
- Consolidated dependencies into `[bundles]`
- Refactored to ensure consistency in version catalog definititons and gradle dependencies

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -141,9 +141,9 @@ androidComponents {
 
 dependencies {
 
-    implementation fileTree(dir: 'libs', include: ['*.jar'])
+    implementation(fileTree(dir: 'libs', include: ['*.jar']))
 
-    implementation libs.appcompat
+    implementation(libs.appcompat)
     // For loading and tinting drawables on older versions of the platform
     implementation(libs.appcompat.resources)
     implementation(libs.emoji2.emojipicker)
@@ -157,16 +157,16 @@ dependencies {
     implementation(libs.datastore)
 
     // Lifecycle
-    implementation libs.lifecycle.viewmodel.ktx
-    implementation libs.lifecycle.viewmodel.compose
-    implementation libs.lifecycle.livedata.ktx
-    implementation libs.lifecycle.runtime.ktx
-    implementation libs.lifecycle.runtime.compose
-    implementation libs.lifecycle.common.java8
-    implementation libs.lifecycle.process
+    implementation(libs.lifecycle.viewmodel.ktx)
+    implementation(libs.lifecycle.viewmodel.compose)
+    implementation(libs.lifecycle.livedata.ktx)
+    implementation(libs.lifecycle.runtime.ktx)
+    implementation(libs.lifecycle.runtime.compose)
+    implementation(libs.lifecycle.common.java8)
+    implementation(libs.lifecycle.process)
 
     // Room
-    implementation libs.room.runtime
+    implementation(libs.room.runtime)
     ksp libs.room.compiler
     // optional - Kotlin Extensions and Coroutines support for Room
     implementation libs.room.ktx
@@ -174,8 +174,8 @@ dependencies {
     androidTestImplementation libs.room.testing
 
     // Hilt
-    implementation libs.hilt.android
-    implementation libs.hilt.navigation.compose
+    implementation(libs.hilt.android)
+    implementation(libs.hilt.navigation.compose)
     ksp libs.hilt.compiler
     androidTestImplementation libs.hilt.android.testing
     kspAndroidTest libs.hilt.compiler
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -142,132 +142,125 @@ androidComponents {
 dependencies {
 
     implementation fileTree(dir: 'libs', include: ['*.jar'])
-    def appcompat_version = '1.7.0'
-    implementation "androidx.appcompat:appcompat:$appcompat_version"
+
+    implementation libs.appcompat
```
jsoberg: Minor nit: using parenthesis for most other deps here:
```suggestion
    implementation(libs.appcompat)
```

jsoberg: eh actually further down it's more of a norm to not use them in this file - I like the idea of using them throughout just because it's forced if we move to kts vs Groovy, but nbd either way

gradle/libs.versions.toml
```diff
@@ -0,0 +1,129 @@
...
+emoji2-emojipicker = { group = "androidx.emoji2", name = "emoji2-emojipicker", version.ref = "emoji2" }
+espresso-core = { module = "androidx.test.espresso:espresso-core", version.ref = "espresso-core" }
+ext-junit = { module = "androidx.test.ext:junit", version.ref = "junit-version" }
+firebase-analytics = { group = "com.google.firebase", name = "firebase-analytics" }
+firebase-bom = { group = "com.google.firebase", name = "firebase-bom", version.ref = "firebase-bom" }
+firebase-crashlytics = { group = "com.google.firebase", name = "firebase-crashlytics"}
+firebase-crashlytics-gradle = { module = "com.google.firebase:firebase-crashlytics-gradle" }
+fragment-ktx = { group = "androidx.fragment", name = "fragment-ktx", version.ref = "fragment-ktx" }
+google-services = { module = "com.google.gms:google-services", version.ref = "google-services" }
+gradle = { module = "com.android.tools.build:gradle", version.ref = "gradle" }
```
jsoberg: Minor nit: I might call this `agp` or `android-gradle-plugin`, was just confused at first when I saw `gradle`

jsoberg (commented):
LGTM 🥳 Happy to see this!

I pulled and tested a build to verify as well, all good (I just can't accept PRs yet). I don't think we have to change anything with Renovate for this to just work with dependency updates but I'm not 100% sure on that

jamesarich:
> LGTM 🥳 Happy to see this!
> 
> I pulled and tested a build to verify as well, all good (I just can't accept PRs yet). I don't think we have to change anything with Renovate for this to just work with dependency updates but I'm not 100% sure on that

Thank you! Also just noticed the parantheses inconsistency. Inconsistency in the version catalog definitions too: `module` vs `group` (not sure what's preferred here).

 Considering some `[bundles]` as well - gonna move this back to draft for now

gradle/libs.versions.toml
```diff
@@ -0,0 +1,170 @@
...
+
+# Data Storage
+datastore = ["datastore"]
+room = ["room-runtime", "room-ktx"]
+
+# Dependency Injection
+hilt = ["hilt-android", "hilt-navigation-compose"]
+
+# Testing
+testing = ["junit", "ext-junit"]
```
jsoberg, line 144: Unrelated: Would you be opposed to moving to Junit5? I've gone through the gamut of assertion libs (assertJ, Truth, AssertK, etc) and tend to land with AssertK being the most usable, interested to hear what you think

gradle/libs.versions.toml
```diff
@@ -0,0 +1,170 @@
...
+work-runtime-ktx = { group = "androidx.work", name = "work-runtime-ktx", version.ref = "work-runtime-ktx" }
+zxing-android-embedded = { group = "com.journeyapps", name = "zxing-android-embedded", version.ref = "zxing-android-embedded" }
+zxing-core = { group = "com.google.zxing", name = "core", version.ref = "zxing-core" }
+
+[bundles]
+# Core AndroidX
+androidx = ["core-ktx", "appcompat", "appcompat-resources", "cardview", "fragment-ktx", "activity-compose"]
+
+# UI
+ui = ["material", "constraintlayout", "viewpager2", "compose-material", "compose-material-icons-extended", "compose-ui-tooling-preview", "compose-runtime-livedata"]
```
jsoberg, line 124: We might want to move to separate `compose-ui` and `views-ui` bundles in the future if we end up modularizing out - doesn't matter right now though since we're still a monolith 👍

jsoberg (commented):
LGTM!✅

gradle/libs.versions.toml
jamesarich: I'm open. Throw in a PR when you can.

PR #1630, opened 2025-03-02
Fixes #1616: Use SP font size for marker labels on the map

### LINKED ISSUE #1616: [Bug]: Map labels seizing issue
### Contact Details

_No response_

### What happened?

the node labels on the map are incorrect size and distance from the node they are attached to. the txt is verry large taking up a lot of visual space and the label is very far away from the node they represent.
 i have pics but don't know how to ad them here.

### App Version

2.5.17

### Tablet

Samsung galaxy tab A7 lite
 one UI 6.1  android 14
screen size not listed

### Device

all

### Firmware

all

### Relevant log output

```shell

```

### Code of Conduct

jsoberg:
Fixes #1616

Marker labels were using a fixed pixel size for text (40px for text, 80px for emojis). For some devices (e.g. lower resolution tablets) this would make the labels much larger than expected.

This change fixes that by using sp font values converted to pixels. This also makes sure that the marker labels respect your device's font sizing for accessibility purposes. 

I used arbitrary values that looked about the same as they did before the change on phone (14sp for text, 20sp for emoji) that we can modify if we'd like. Note they're slightly smaller on my phone here (S23 Ultra) and I use .9f font scaling, so this is expected.

| Device | Before | After |
|------|------|-----|
| Phone (S23 Ultra) |  |  |
| Tablet (Tab S8+) |  |  |

Recordings for font scaling updates:

| Phone | Tablet |
|------|-----|
| <video src="https://github.com/user-attachments/assets/4940efaf-d994-4000-a79d-94eeacdf23f6" width="300"></video> | <video src="https://github.com/user-attachments/assets/cb71500f-fe95-4816-a993-103171bcfb53" width="300"></video> |

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/android/ContextExtensions.kt
+++ b/app/src/main/java/com/geeksville/mesh/android/ContextExtensions.kt
@@ -19,6 +19,7 @@ package com.geeksville.mesh.android
 
 import android.app.Activity
 import android.content.Context
+import android.util.TypedValue
 import android.view.inputmethod.InputMethodManager
 import android.widget.Toast
 
@@ -35,3 +36,7 @@ fun Activity.hideKeyboard() {
         imm?.hideSoftInputFromWindow(v.windowToken, 0)
     }
 }
+
+// Converts SP to pixels.
+fun Context.spToPx(sp: Float): Int =
+    TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_SP, sp, resources.displayMetrics).toInt()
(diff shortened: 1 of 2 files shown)


jamesarich (approved):

prokrypt:
There is still one small issue: the gap between marker and label appears larger when the font is smaller...

jsoberg:
oh interesting, that distance from marker to label might just need an adjustment then 🤔

prokrypt:
https://github.com/user-attachments/assets/6d2e0c78-1c38-4d68-8d2a-af63407e9890

jsoberg:
Hey @prokrypt, I added https://github.com/meshtastic/Meshtastic-Android/pull/1637 to address the variable offset issue you've described. This should make the label offset more uniform across devices

PR #1636, opened 2025-03-03
feat #1611: Add confirmation dialog when changing to router role

### LINKED ISSUE #1611: [Feature Request]: ROUTER "are you certain?"
### Contact Details

mahlon@martini.nu

### Tell us your idea.

I am noticing an uptick in new nodes in my area (awesome!), that are clearly in lower elevation areas and are set to ROUTER (not awesome!)

I'm afraid that these nodes are 'poisoning the well', and while I've tried (unsuccessfully) to reach out and request they move to CLIENT, I'm hoping the app might be able to help address this in a future update.

Perhaps when selecting the ROUTER role (each time), and upon connection (one time), if your node is in a ROUTER role, a modal appears:

"Your node is currently in a role that could negatively impact the mesh health of your area.  Please confirm that this node is at the highest possible elevation and has a 360 view of the surrounding terrain.  If not, others will not like you very much."

With the button defaults reverting things to CLIENT unless they check a box and do a dance.  Maybe some links to further reading as well.

 - https://meshtastic.org/blog/choosing-the-right-device-role/
 - https://www.youtube.com/watch?app=desktop&v=htjwtnjQkkE

I suppose this isn't Android-client specific, now that I consider it, but perhaps it's a good starting place.

### Relevant log output

```shell

```

### Code of Conduct

jamesarich:
Introduced a confirmation dialog that will be shown when changing the device role to "router". The confirmation dialog includes links to the "Device Role Documentation" and "Choosing The Right Device Role" blog post. The user must check a checkbox to confirm.

resolves #1611

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/DeviceConfigItemList.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/DeviceConfigItemList.kt
@@ -17,21 +17,34 @@
 
 package com.geeksville.mesh.ui.radioconfig.components
 
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.text.KeyboardActions
 import androidx.compose.foundation.text.KeyboardOptions
+import androidx.compose.material.AlertDialog
+import androidx.compose.material.Checkbox
 import androidx.compose.material.Divider
+import androidx.compose.material.Text
+import androidx.compose.material.TextButton
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
+import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.platform.LocalFocusManager
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.LinkAnnotation
+import androidx.compose.ui.text.SpanStyle
+import androidx.compose.ui.text.TextLinkStyles
+import androidx.compose.ui.text.buildAnnotatedString
 import androidx.compose.ui.text.input.ImeAction
 import androidx.compose.ui.text.input.KeyboardType
+import androidx.compose.ui.text.withLink
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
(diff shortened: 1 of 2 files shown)


jamesarich:
i've got a little more cookin

jamesarich:
https://github.com/user-attachments/assets/f8d59974-0fdc-44a4-a6af-2e8a2737e86a

jamesarich:
that was fun and maybe worth saving for another use case - but to keep things professional, reverted back to dialog with checkbox..

jsoberg:
Konami code is too easy to remember for this purpose anyway 😆

app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/DeviceConfigItemList.kt
```diff
@@ -97,6 +110,74 @@ fun DeviceConfigScreen(
     )
 }
 
+@Suppress("LongMethod")
+@Composable
+fun RouterRoleConfirmationDialog(
+    onDismiss: () -> Unit,
+    onConfirm: () -> Unit,
+) {
+    val dialogTitle = "Are you sure?"
```
jsoberg: Should we use string resources for these so that they can eventually be translated?

jamesarich: Oh yep - i can do that. If we just store sentence parts for these compose annotated strings - it seems like phrasing could be challenging for translators. I wonder if there's a good solve out there for that.

shortwavesurfer2009:
Is this also going to be for router client and router late?

jamesarich:
> Also, what about our iOS counterparts? Implementing this on Android will be somewhat effective, but it would be even more effective if our iOS counterparts also implement such functionality.

@RCGV1 wanted to take this on for iOS, but I'm not seeing an issue created on their repo yet.

shortwavesurfer2009:
I don't see why they would have to wait for an issue to be opened. Just submit a pull request, making the change, and explain why it's being done.

jamesarich:
> I don't see why they would have to wait for an issue to be opened. Just submit a pull request, making the change, and explain why it's being done.

All meshtastic development is done on a volunteer basis, developers donating their time and resources - feel free to do so as well. 

An issue submitted to the iOS repo will help track the request for the feature for someone to take on as they are able and want to.

shortwavesurfer2009:
Fair enough, although I can't open one because I don't own an iPhone. So for all I know, the iOS version could already have that functionality, and that would be a duplicate request. Somebody who has an iOS device and can confirm that it needs to be done will have to open an issue for it.

app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/DeviceConfigItemList.kt
jamesarich: I found [AnnotatedString.fromhtml](https://developer.android.com/reference/kotlin/androidx/compose/ui/text/AnnotatedString.Companion#(androidx.compose.ui.text.AnnotatedString.Companion).fromHtml(kotlin.String,androidx.compose.ui.text.TextLinkStyles,androidx.compose.ui.text.LinkInteractionListener)) and it's just what the doctor ordered.

jsoberg:
LGTM!

daschu117:
@jamesarich thanks for a much welcome change! Does this apply to REPEATER as well?

Kealper:
> @jamesarich thanks for a much welcome change! Does this apply to REPEATER as well?

Currently it appears to apply to ROUTER, but it could be slightly modified to catch REPEATER as well.

PR #1637, opened 2025-03-04
fix #1616: Calculate DP to pixel value for map label offset

### LINKED ISSUE #1616: [Bug]: Map labels seizing issue

jsoberg:
Similar to the change to label text [here](https://github.com/meshtastic/Meshtastic-Android/pull/1630), this PR modifies the map label offset to use DP instead of a static pixel value. This makes sure that any device will have a more uniform offset for the map labels.

I don't have my tablet available atm, but this is before and after on my S23 ultra:

| Before | After |
|------|-----|
|  |  |

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/android/ContextExtensions.kt
+++ b/app/src/main/java/com/geeksville/mesh/android/ContextExtensions.kt
@@ -40,3 +40,7 @@ fun Activity.hideKeyboard() {
 // Converts SP to pixels.
 fun Context.spToPx(sp: Float): Int =
     TypedValue.applyDimension(TypedValue.COMPLEX_UNIT_SP, sp, resources.displayMetrics).toInt()
+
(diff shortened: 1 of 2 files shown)


prokrypt:
Perfect! Works for all display size settings. Label overlaps the marker just a bit at biggest font setting but it was already like that before.
Screenshot_20250303-201901_Meshtastic

jamesarich (approved):

Issue #1626, opened 2025-03-01
[Feature Request]: Run an SMA to smooth out data


shortwavesurfer2009:
### Tell us your idea.

Run a simple moving average on say the last 10 minutes of your node data in order to smooth out lines. This screenshot shows the battery line with a 1 minute dip followed by a recovery. Though using it for ChUtil could work too.

![Image](https://github.com/user-attachments/assets/7616b8ff-38ef-47b4-b30d-48dc44bb7bdb)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1653, opened 2025-03-07
feat: Initial implementation of low battery notifications for nodes

### LINKED ISSUE #1395: [Feature Request]: Battery goes under x% notification
### Contact Details

dtrieb

### Tell us your idea.

I'd like the app to...
Get a notification on my mobile if the Akku goes under x%.

P.e. Devices like tracker you have on your backpack and didn't see if it run under 10% voltage. So a notification would be good.

### Relevant log output

_No response_

### Code of Conduct

### LINKED ISSUE #1598: [Feature Request]: Low battery notification
### Tell us your idea.

I want the app to warn users if a node has a battery percentage lower than 15%.
I think this should be configurable.

### Relevant log output

```shell

```

### Code of Conduct

Kealper:
Closes #1395 and closes #1598

This PR adds low-battery notifications for both locally-connected and remote nodes that have been added as a favorite in the device's node DB.

Currently, the notification is sent when a device goes below 20% battery remaining, but if a device reports the battery voltage as "0.00" volts, it will be exempt from alerts as that is a magic number in the firmware for no battery voltage support. In the future, this threshold will be configurable, but it is currently fixed.

For now, separate notification channels are created for locally-connected nodes and remote favorite nodes, but in the future this will be grouped together under a single notification channel that has settings to enable or disable reporting of remote favorite nodes.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshServiceNotifications.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshServiceNotifications.kt
@@ -200,7 +200,7 @@ class MeshServiceNotifications(
         return channelId
     }
 
-    // TODO: Once we get a dedicated settings page in the app, this function should be removed and
+    // FIXME, Once we get a dedicated settings page in the app, this function should be removed and
     //     the feature should be implemented in the regular low battery notification stuff
     @RequiresApi(Build.VERSION_CODES.O)
     private fun createLowBatteryRemoteNotificationChannel(): String {
@@ -272,7 +272,7 @@ class MeshServiceNotifications(
         }
     }
 
-    // TODO: Once we get a dedicated settings page in the app, this function should be removed and
+    // FIXME, Once we get a dedicated settings page in the app, this function should be removed and
     //     the feature should be implemented in the regular low battery notification stuff
     private val lowBatteryRemoteChannelId: String by lazy {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
@@ -338,7 +338,7 @@ class MeshServiceNotifications(
         )
     }
 
-    // TODO: Once we get a dedicated settings page in the app, this function should be removed and
+    // FIXME, Once we get a dedicated settings page in the app, this function should be removed and
     //     the feature should be implemented in the regular low battery notification stuff
     fun showOrUpdateLowBatteryRemoteNotification(node: NodeEntity) {
         notificationManager.notify(
@@ -531,7 +531,7 @@ class MeshServiceNotifications(
         return lowBatteryNotificationBuilder.build()
     }
 
-    // TODO: Once we get a dedicated settings page in the app, this function should be removed and
+    // FIXME, Once we get a dedicated settings page in the app, this function should be removed and
     //     the feature should be implemented in the regular low battery notification stuff
     lateinit var lowBatteryRemoteNotificationBuilder: NotificationCompat.Builder
     private fun createLowBatteryRemoteNotification(node: NodeEntity): Notification {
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/geeksville/mesh/service/MeshServiceNotifications.kt
```diff
@@ -65,6 +65,7 @@ class MeshServiceNotifications(
             createMessageNotificationChannel()
             createAlertNotificationChannel()
             createNewNodeNotificationChannel()
+            createLowBatteryNotificationChannel()
```
jamesarich, line 68: we'll want to instantiate the favorites/remote channel here as well

app/src/main/java/com/geeksville/mesh/service/MeshServiceNotifications.kt
```diff
@@ -409,4 +498,70 @@ class MeshServiceNotifications(
         }
         return newNodeSeenNotificationBuilder.build()
     }
+
+    lateinit var lowBatteryNotificationBuilder: NotificationCompat.Builder
+    private fun createLowBatteryNotification(node: NodeEntity): Notification {
```
jamesarich: if there are no visual differences between local and remote, this could be deduplicated with the use of a `isRemote` boolean param in the function - can then use `isRemote` to branch on which `channelId` to use.

Kealper: Should be good now! This has been squashed down into one function that handles both types. I'm not sure how "idiomatic Kotlin" how I did it was but it does still work and removes a lot of duplicated code.

jamesarich (approved):
LGTM thank you!

shortwavesurfer2009:
So far I have received the low battery notification like 7 or 8 separate times because of fluctuating voltages. This should fix it. https://github.com/meshtastic/Meshtastic-Android/issues/1626

shortwavesurfer2009:
Okay, so I'm still getting the notifications. And now I'm not sure if it's because of fluctuating voltages or if it's because the node reports data every single minute. And so I would continue getting the notifications until such time as it is no longer at 19% or whatever.

Kealper:
> Okay, so I'm still getting the notifications. And now I'm not sure if it's because of fluctuating voltages or if it's because the node reports data every single minute. And so I would continue getting the notifications until such time as it is no longer at 19% or whatever.

With Android's notifications, it'll keep the existing low battery notification for a particular node updated with the most recent battery percent while it's below the threshold. For a locally-connected node that will be every minute, and if you dismiss the low battery notification for it while the battery is still low, it'll cause the notification to appear again the next time the node sends the status to your phone.

The notifications are sent to a different "channel" than messages and new nodes, so you can turn off just battery notifications in the app's notification settings if needed.

shortwavesurfer2009:
At least for locally connected nodes, it sends the status every single minute. So you get a new notification with notification sound every minute until it dies.

After getting like 20 notifications, I ended up just disabling the channel entirely.

Kealper:
> At least for locally connected nodes, it sends the status every single minute. So you get a new notification with notification sound every minute until it dies.
> 
> After getting like 20 notifications, I ended up just disabling the channel entirely.

Well that's not intended behavior, what phone and Android version are you using? It works how I described above on my Pixel 8 Pro running Android 15.

shortwavesurfer2009:
OnePlus Nord N200 running LineageOS 21 (Android 14)

shortwavesurfer2009:
I didn't see the last part asking if I was still getting haptics or sound as long as I had not dismissed the notification. And I'm not totally sure, so I need to check that first.

shortwavesurfer2009:
Okay, you are right. As long as I do not dismiss the notification, it does not give me sound or vibration again. That's just something I'm not in the habit of doing, as I always dismiss my notifications. Except for service notifications, such as the meshtastic connected to 2 of 50 nodes, etc.

PR #1657, opened 2025-03-07
feat: expand role confirmation to include REPEATER

jamesarich:
The role change logic was modified to show a confirmation dialog when the role is set to `ROUTER`, `ROUTER_CLIENT`, or `REPEATER`. Previously, the dialog was only displayed when setting the role to `ROUTER`.

Also the clickable area for the checkbox has been expanded to the full width of the Row.

<video src="https://github.com/user-attachments/assets/e3fd872a-9f34-405f-80b2-2228f4f3c7d8
" width="300"/>


jamesarich:
Just now noticed `ROUTER_CLIENT` is no longer even on the list of roles to select 🙃 .  Oh well - still technically valid as an `infrastructureRole` that _should_ be notified (even if you can't select it).

Issue #1660, opened 2025-03-08
[Bug]: Low battery notification triggers every minute


shortwavesurfer2009:
### What happened?

2.5.19 added a battery low notification for local and favorite nodes if they were below 20% battery. However, at least on local nodes, since the node reports every single minute, you get a low battery notification every single minute until the node finally just dies.

After getting about 20 of them, I ended up disabling the notification channel for low battery notifications.

### App Version

2.5.19

### Phone

OnePlus Nord N200

### Device

Na

### Firmware

Na

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jamesarich:
this should be improved by #1661 - I think @Kealper has even more tweaks in the works based on continued discussion

PR #1661, opened 2025-03-08
feat: Improve (battery) notification behavior

jamesarich:
Noticed some chatter on #1653 and wondered if this might be a potential solve. I changed up the notification to be non-dismissable, but also added logic to clear the notification when the battery status is back above the threshhold.

Also adds a progress bar to indicate battery percentage for a little flair.

This also adds some grouping to the growing variety of notifications as well that may help logically consolidate them in the drawer.

- Changes low battery notifications to be non-cancelable
- Cancel low battery notifications when battery level is no longer low.
- Add notification groups and improve notification settings.
- Add vibration to low battery notifications.
- Add progress bar to low battery notifications

(note: i artificially set the threshhold very high for testing)

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -949,6 +949,8 @@ class MeshService : Service(), Logging {
                         if (t.deviceMetrics.voltage > batteryPercentUnsupported &&
                             t.deviceMetrics.batteryLevel < batteryPercentLowThreshold) {
                             serviceNotifications.showOrUpdateLowBatteryNotification(it, isRemote)
+                        } else {
+                            serviceNotifications.cancelLowBatteryNotification(it)
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


Kealper:
Sweet, I think that should fix the potential battery-spam nicely! The extra tweaks for grouping and the progress bar are nice additions too!

jsoberg:
Connecting this to https://github.com/meshtastic/Meshtastic-Android/issues/1662 as well

jsoberg:
Re: https://github.com/meshtastic/Meshtastic-Android/issues/1660, interested to see if others have feedback here. I haven't had a chance to test it, but I'm wondering if we should allow the user to dismiss the low batt notification and then not re-issue it after manual user dismissal? That way if the user doesn't care, they could swipe it away and it wouldn't come back on subsequent updates

jamesarich:
> Re: #1660, interested to see if others have feedback here. I haven't had a chance to test it, but I'm wondering if we should allow the user to dismiss the low batt notification and then not re-issue it after manual user dismissal? That way if the user doesn't care, they could swipe it away and it wouldn't come back on subsequent updates

I had this thought as well, but was unsure on how to support storing and managing the state here. Sharedprefs?

jsoberg:
I still need to onboard fully on how we're firing these notifications, but I was even thinking per-app-session caching would suffice. We issue separate notifications for separate nodes?

Generally we've been talking about adding some app-level preferences for some things, seems like we could add new tables to `MeshtasticDatabase`

shortwavesurfer2009:
Ideally, if a notification triggers at 20%, another notification would trigger at say 10% if the user dismissed the 20% one. That way they know yet again that the battery is getting low. Making the low battery notification unable to be canceled would obviously work instead of this.

garthvh:
I posted in the wrong issue earlier but the iOS low battery notification for the connected node is at 4% which is still hours in many cases, 20% is going to fire hundreds of notifications.

PR #1676, opened 2025-03-13
fix #1484: Filter out ignored nodes from map

### LINKED ISSUE #1484: [Bug]: Blocked/Ignored nodes not removed from map
### Contact Details

_No response_

### What happened?

I have been testing the new ignored node feature with the app and noticed that when you ignore a node it does not remove it from the map. I understand and agree that the ignored node should be still on the node list crossed out so you can un-ignore it at a later date if you so choose, however i believe all positions shown on the map should also be removed/hidden. My specific use case is blocking a node with an inverted longitude value causing it to always show up in China, while i know it is only a few km away. This "faraway" node also causes the map to zoom wayyy out when loading.

### App Version

2.5.14

### Phone

Pixel 7a

### Device

SEEED T1000E

### Firmware

2.5.16

### Relevant log output

_No response_

### Code of Conduct

jamesarich:
- Created a new `filteredNodeList` `StateFlow` in `UIState.kt` that filters out nodes marked as ignored.
- Updated `MapFragment.kt` to use `filteredNodeList` instead of `nodeList`, ensuring that ignored nodes are not displayed on the map.

fixes #1484 

https://github.com/user-attachments/assets/eab5b07a-9f65-49fe-8a64-421e5b0586fb

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -96,7 +96,10 @@ android {
     }
     kotlinOptions {
         jvmTarget = JavaVersion.VERSION_17.toString()
-        freeCompilerArgs += ['-opt-in=kotlin.RequiresOptIn']
+        freeCompilerArgs += [
+                '-opt-in=kotlin.RequiresOptIn',
+                '-opt-in=kotlinx.coroutines.ExperimentalCoroutinesApi'
+        ]
     }
     lint {
         abortOnError false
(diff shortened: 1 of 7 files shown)


jsoberg (approved):
LGTM! Side note but maybe we should just add `ExperimentalCoroutinesApi` as a compiler opt-in

Issue #1495, opened 2024-12-21
[Bug]: NaN in environment telemetry graph causes blank graph


GMart:
### Contact Details

Discord

### What happened?

If there is any NaN value (Not sure how this happens) in the data logs the environment telemetry graph doesn't even show up. This can happen at any time range, so for example the graph may show normally for 24h, but if I switch to 7 day where there is a NaN value in the data, the graph dissappears. Screenshot
![Screenshot_20241218_085026_Meshtastic](https://github.com/user-attachments/assets/e7873270-dc79-4ce6-922a-78393edb69bf)
 attached.

### App Version

2.5.15
Still an issue on 2.5.18

### Phone

Samsung Galaxy S21+ Android 14

### Device

T-Echo

### Firmware

2.5.12

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jsoberg:
Fixed with https://github.com/meshtastic/Meshtastic-Android/pull/1744

PR #1684, opened 2025-03-17
Graph Barometric Pressure

Robert-0410:
### Changes
- Plotting the barometric pressure on the environment metrics graph.
- Only plotting the IAQ or barometric pressure when we have actual data for them.


jamesarich (approved):

jamesarich:
@Robert-0410 can you take a look at #1495  and see if we can add some more graceful handling of NaN values? Maybe check the `has` functions on the proto.

PR #1686, opened 2025-03-18
fix: Add some smarts to low battery notification triggers

Kealper:
This adds a few changes to tame down the notifications for a low battery.

Adds a per-node notification cooldown, which is currently set to 25 minutes. This time was chosen because remote node device metrics won't (usually) come in faster than once every 30 minutes so well-behaved remote nodes will not hit the cooldown and their battery status will always be updated as it comes in. If a node is misbehaving, such as frequently rebooting due to low battery and thus potentially sending device metrics frequently, the cooldown will catch it and won't spam-notify.

For locally-connected devices, once the low battery threshold is reached, the notification will only happen again (or be updated if it wasn't dismissed) every 5% the battery drains. For example, the current threshold is 20%. So instead of alerting every time a new local device metrics packet was received under 20%, it will alert at 20%, then at 15%, again at 10%, and finally at 5%. Those alerts are also still bound by the notification cooldown time as well, for nodes that drain their batteries very quickly. (I'm looking at you, T-Watch S3!)

The every-5% thing doesn't apply to remote favorite nodes because of how infrequently that data is sent over the mesh, for many devices in use today, there may only be a few device metrics packets sent out before the low-battery'd device has ran its battery into the ground, even at <=20%.

Finally, there's a secondary threshold called the critical threshold that's been added, which is set to 5%. Once the device reaches the critical threshold, the above limits are ignored and it will basically beg you to charge the device. If this proves to require a cooldown as well, that can all be adjusted!


app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -962,6 +976,32 @@ class MeshService : Service(), Logging {
...
+            t.deviceMetrics.batteryLevel <= batteryPercentCriticalThreshold -> {
+                shouldDisplay = true
+                forceDisplay = true
+            }
+            t.deviceMetrics.batteryLevel == batteryPercentLowThreshold -> shouldDisplay = true
+            t.deviceMetrics.batteryLevel.mod(batteryPercentLowDivisor) == 0 && !isRemote -> shouldDisplay = true
+            isRemote -> shouldDisplay = true
+        }
+        if (shouldDisplay) {
+            val now = Instant.now().epochSecond
```
jamesarich: Instant.now().epochSecond requires API 26 - our current min is 21.  try System.currentTimeMillis() / 1000

jamesarich (changes requested):
I went through and reverted the whitespace changes so I could tell what the actual changes were 🙃 . Other than swapping out for System.currentTimeMillis() I think these are great updates.

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
Kealper: Fixed! Didn't realize that was above our minimum API version, whoops!

Also sorry about my local detekt setup putting the whitespace through the ringer, I think I've got that fixed now.

jamesarich (approved):

Issue #1692, opened 2025-03-19
[Feature Request]: update `Request Position` action label to `Exchange Positions`


jamesarich:
### Tell us your idea.

The label for the `Request Position` action is misleading in that it actually is an **exchange** of positions.

There have been several questions raised on discord and #813 .

Update the label from `Request Position` to `Exchange Positions` to better describe the process and alleviate confusion

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1701, opened 2025-03-21
Clarify request options to exchange for issue #1692

todd2982:
Based on discord conversations, and issue #1692 I've updated the language from "request position" to "exchange position".  I have also extended this to the user info function as well since they both operate the same way.  When using either of these options your local radio will send a packet with its own info requesting acknowledgement, and that acknowledgment is the requested information.

I updated the foreign language string name, but not the actual translations in each language's xml files.  This was so that things could be kept clean when translations are completed.

Screenshot_20250321_040749


jamesarich (approved):
Thank you this was on my list!

jamesarich:
closes #1692

dzid26:
I am not a native but...
I think "Position" (countable) should have been plural if "exchange" is a verb. 

Singular implies exchanging for something else. So i thought it perhaps meant exchanging a request for the position.

Issue #1704, opened 2025-03-21
[Bug]: Accessibility: low contrast in light mode


shortwavesurfer2009:
### What happened?

With such a light green bar, the buttons should really be black instead of white.

![Image](https://github.com/user-attachments/assets/abce72c5-d221-4df9-8cab-cc7002a47487)
![Image](https://github.com/user-attachments/assets/4be9fa46-4347-49fb-adad-1c631967d3b5)

### App Version

2.5.19

### Phone

Oneplus Nord n200

### Device

Na

### Firmware

Na

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

shortwavesurfer2009:
Actually, I just updated to the newest version and it's worse because now the navigation icons such as messages, contacts, etc. are white on the green background as well.

Issue #1633, opened 2025-03-02
[Bug] [Minor]: Dark mode onboarding uses black font over gray background


jsoberg:
### Contact Details

myheartpills on Discord

### What happened?

Onboarding flow uses black text over a gray background when we're in dark mode:

<video src="https://github.com/user-attachments/assets/fca95a5c-a6c4-4332-8f45-f4d90adeefdd" width="300"></video>

### App Version

master, c597b3b000dc4429188eac78294ca768b34c28d3

### Phone

Samsung S23 Ultra, Android 15

### Device

N/A

### Firmware

N/A

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #1713, opened 2025-03-24
Changed title bar color to black for issue #1704 and #1633.

todd2982:
Updated title bar color as requested in light mode.  Original white on green was difficult to read.

Before:
1704-before
After:
1704-after

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/theme/Color.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/theme/Color.kt
@@ -32,6 +32,8 @@ val LightPink = Color(0xFFFFE6E6)
 val LightGreen = Color(0xFFCFE8A9)
 val LightRed = Color(0xFFFFB3B3)
 
+val Black = Color(0x000000)
+
 val MeshtasticGreen = Color(0xFF67EA94)
 
 val HyperlinkBlue = Color(0xFF43C3B0)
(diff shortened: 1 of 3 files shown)


todd2982:
Fixes issue #1704

shortwavesurfer2009:
I don't know how apps do it, but it should also change the notification icons such as the time, Wi-Fi, cellular signal, etc.

shortwavesurfer2009:
Does this also fix the buttons when you're in a chat such as copy, delete and select all?

shortwavesurfer2009:
Looks like the node names will need changing too
Screenshot_20250324-181427_Meshtastic_1

todd2982:
> Looks like the node names will need changing too Screenshot_20250324-181427_Meshtastic_1

This is corrected.

todd2982:
> I don't know how apps do it, but it should also change the notification icons such as the time, Wi-Fi, cellular signal, etc.

I'm still researching this

todd2982:
> I don't know how apps do it, but it should also change the notification icons such as the time, Wi-Fi, cellular signal, etc.

The option I found to fix this is by adding `<item name="android:windowLightStatusBar">true</item>` (Boolean testing needed), but I run into an issue with the minimum API target version.  This theme change requires version 23, but we have a target configured for 21.  Thoughts?

I don't want to make changes to the target versions without a serious conversation first.  Also, since I'm so new at this there may be a better way.  I've reached the end of my google-fu on this one for the moment.

todd2982:
Should include issue #1633 in this effort.

jamesarich:
closing as stale - we're now using all compose, some of these issues should be solved, or if they still exist solved in the correct way for compose

shortwavesurfer2009:
https://github.com/meshtastic/Meshtastic-Android/issues/1704#issuecomment-2884849743

PR #1720, opened 2025-03-25
fix: Handle node public key mismatch and show warning

jamesarich:
- Add a mismatchKey flag to Node and MessageTopBar to indicate a public key mismatch.
- Set the public key to a default error value (all zeros) when a node's public key changes.
- Display a warning in the MessageTopBar when a key mismatch is detected in PKC.
- Only clear all nodes when a different mynode number is present.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/database/NodeRepository.kt
+++ b/app/src/main/java/com/geeksville/mesh/database/NodeRepository.kt
@@ -107,7 +107,7 @@ class NodeRepository @Inject constructor(
     }
 
     suspend fun installNodeDB(mi: MyNodeEntity, nodes: List<NodeEntity>) = withContext(dispatchers.io) {
-        val isDifferentNode = myNodeInfo.value?.myNodeNum == mi.myNodeNum
+        val isDifferentNode = myNodeInfo.value?.myNodeNum != mi.myNodeNum
         nodeInfoDao.clearMyNodeInfo()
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/geeksville/mesh/database/NodeRepository.kt
```diff
@@ -103,13 +103,16 @@ class NodeRepository @Inject constructor(
     ).mapLatest { list -> list.map { it.toModel() } }.flowOn(dispatchers.io).conflate()
 
     suspend fun upsert(node: NodeEntity) = withContext(dispatchers.io) {
-        nodeInfoDao.upsert(node)
+        nodeInfoDao.upsertCheckKeyMatch(node)
     }
 
     suspend fun installNodeDB(mi: MyNodeEntity, nodes: List<NodeEntity>) = withContext(dispatchers.io) {
+        val isDifferentNode = myNodeInfo.value?.myNodeNum == mi.myNodeNum
```
thebentern: Should this be nodeInfoDao?

jamesarich: ohp good catch.  this _is_ the last nodeInfoDao value of the `myNodeInfo` (the last locally connected node), however I had the logic wrong.

garthvh (approved):
Looks good

thebentern (approved):

PR #1734, opened 2025-03-30
fix: Remove nodeDb size from status string

### LINKED ISSUE #1733: [Bug]: Seeing 1##/100 nodes when connecting to a device
### Contact Details

aj.reitz@gmail.com

### What happened?

I'm assuming this is a bug, since numerator is larger than the denominator. If it's not feel free to close. I'm happy to take a look at the source and see if I can't figure out the issue or report more info here.

### App Version

2.5.21

### Phone

Pixel 8

### Device

heltec v3

### Firmware

2.6.4.b89355f Alpha 

### Relevant log output

```shell

```

### Code of Conduct

jamesarich:
Previously the status string read `Nodes ( # / 100)` this updates the string to just show the raw node count `Nodes( # )` as a better representation, since a device's nodeDb may be smaller ( 80 in the case of nrf) or (200-250 for the cases of some esp32 devices with larger flash). 

resolves #1733

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -1574,7 +1574,7 @@ class MeshService : Service(), Logging {
         insertMeshLog(packetToSave)
 
         newNodes.add(info)
-        radioConfigRepository.setStatusMessage("Nodes (${newNodes.size} / 100)")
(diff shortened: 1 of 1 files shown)


Kealper:
Awesome! I had actually intended on looking into fixing this because I just had a new-to-the-mesh person ask me why they only ever seen "80/100" when connecting to their nRF52-based node. :+1:

PR #1737, opened 2025-03-31
fix: disable editing of Public Key field

### LINKED ISSUE #1736: [Bug]: Public key appears to be settable independently of the private key
### Contact Details

NomDeTom on discord

### What happened?

Using the radio config>security>public key input box, I am able to enter a "valid" public key that does not correspond to the private key in the box below. The interface appears to retain this key, although it is reset upon a node reboot.

Why is this a bug? The interface gives the illusion that the public key is somehow settable by the user, but it is computed from the private key. The fact that this is a text input box, and that it retains the information that is entered compounds the misleading effect.

### App Version

2.5.20

### Phone

Samsung Galaxy S10, version ??

### Device

DIY-promicro

### Firmware

2.5.22

### Relevant log output

```shell

```

### Code of Conduct

jamesarich:
resolves #1736

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/SecurityConfigItemList.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/SecurityConfigItemList.kt
@@ -83,7 +83,7 @@ fun SecurityConfigItemList(
             EditBase64Preference(
                 title = "Public Key",
                 value = securityInput.publicKey,
-                enabled = enabled,
+                enabled = false,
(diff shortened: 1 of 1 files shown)


valentintintin:
Hi !
It seems that we can not copy anymore the public key to paste it on the admin key field now.
Maybe if you want to keep it disabled, just add a copy button?
Thanks :)

jeremybox:
@valentintintin https://github.com/meshtastic/Meshtastic-Android/pull/1781

PR #1744, opened 2025-04-01
Fix: NaN Blank Environment Graph

Robert-0410:
### Issue
#1495 
Users reported blank graphs when a NaN value was present for a temperature reading.


jsoberg (approved):
LGTM ✅ we should probably add a domain conversion to these Telemetry protos so that we can get valid results elsewhere vs always having to grab directly from mesh logs

Robert-0410:
> LGTM ✅ we should probably add a domain conversion to these Telemetry protos so that we can get valid results elsewhere vs always having to grab directly from mesh logs

Alright I can look into it in the near future. If you have any resources let me know.

PR #1752, opened 2025-04-06
Missing iaq

Robert-0410:
### Feat
- Only plotting metrics when we have actual data for them. This way devices like the T1000E could have a better graph.
### Fixes
- IAQ would not be plotted when the min = 0.


app/src/main/java/com/geeksville/mesh/ui/components/EnvironmentMetrics.kt
```diff
@@ -67,7 +67,7 @@ import com.geeksville.mesh.ui.theme.Orange
 import com.geeksville.mesh.util.GraphUtil.createPath
 import com.geeksville.mesh.util.GraphUtil.drawPathWithGradient
 
-private enum class Environment(val color: Color) {
+private enum class Environment(val color: Color, var shouldPlot: Boolean = false) {
```
jsoberg: We probably shouldn't have `shouldPlot` be a mutable field in `Environment` - since this is an enum, `shouldPlot` is static and any changes to the field will persist.

This might make more sense to be modeled as state that gets sent into `EnvironmentMetricsChart`, where we could send in some state calculated in the view model that determines which metrics should be plotted, and includes all the information necessary to plot each metric (including the min/max values currently being calculated in the composable here)

Robert-0410: OK, That makes sense.

Robert-0410: Took out most of that logic out, any input would be awesome

Robert-0410:
Kind of had analysis paralysis and had to make a choice. I would like to continue to make improvements based on input but if we can I would like to see this merged. Of course if nothing too crazy sticks out.

app/src/main/java/com/geeksville/mesh/model/EnvironmentMetricsState.kt
```diff
@@ -0,0 +1,129 @@
...
+     *
+     * @param timeFrame used to filter
+     * @return a [Triple] with: the filtered [List], a [BooleanArray] the size of [Environment] used
+     * to determine if a metric should be plotted, and a [Pair] containing two others for which the
+     * first is the combined min and max of: the temperature, humidity, and IAq. The second being
+     * the min and max for the barometric pressure.
+     */
+    fun environmentMetricsFiltered(
+        timeFrame: TimeFrame
+    ): Triple<List<Telemetry>, BooleanArray, Pair<Pair<Float, Float>, Pair<Float, Float>>> {
```
jamesarich: This is a lot of Pairs and Triples and is a bit hard to follow - it could be better defined as a data class with clearly named values.

jamesarich (changes requested):
Let's migrate the various Pairs and Triples into a data class and I'll give it another look. Let me know if you need a hand!

app/src/main/java/com/geeksville/mesh/model/EnvironmentMetricsState.kt
Robert-0410: cool, i didn't know if making another data class was the way too go.. thanks

jamesarich (approved):
thanks @Robert-0410 !

PR #1859, opened 2025-05-15
feat: Add Contact Sharing via QR code

jamesarich:
This feature adds the ability to share contact information using QR codes.

This will allow the user to share a contact to another user by presenting a QR code to be scanned, and vice versa the ability to scan a QR code and import a contact that is then synced to the node device via admin message.


jamesarich:
@garthvh @thebentern does this need to be firmware version gated? if so, which version?

thebentern:
That's a good idea. It would be >= 2.6.8

PR #1861, opened 2025-05-15
fix #1567: Allow using a different TCP port than default 4403

### LINKED ISSUE #1567: [Feature Request]: Allow using a different TCP port than port 80 when specifying module IP address
### Contact Details

fernando.sanchez.jr@gmail.com

### Tell us your idea.

I'd like the app to allow me to specify a port other than TCP port 80 when connecting to a meshtastic device via HTTP. ❤ 

### Relevant log output

```shell

```

### Code of Conduct

slush0:
I implemented #1567 by adding another text element with pre-selected default TCP port, which can be changed by the user. Showing default port is a good UX practice, otherwise it isn't clear which protocol is actually used (see messages in #1567 where user was confused by thinking the app uses HTTP API at port 80).
Screenshot from 2025-05-15 23-04-26

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/Settings.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/Settings.kt
@@ -88,6 +88,7 @@ import com.geeksville.mesh.model.BTScanModel
 import com.geeksville.mesh.model.BluetoothViewModel
 import com.geeksville.mesh.model.UIViewModel
 import com.geeksville.mesh.service.MeshService
+import com.geeksville.mesh.repository.network.NetworkRepository
 import kotlinx.coroutines.delay
 
 fun String?.isIPAddress(): Boolean {
@@ -145,6 +146,7 @@ fun SettingsScreen(
 
     // State for manual IP address input
     var manualIpAddress by remember { mutableStateOf("") }
+    var manualIpPort by remember { mutableStateOf(NetworkRepository.SERVICE_PORT.toString()) }
 
     // State for the device scan dialog
     var showScanDialog by remember { mutableStateOf(false) }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/geeksville/mesh/ui/Settings.kt
```diff
@@ -145,6 +145,7 @@ fun SettingsScreen(
 
     // State for manual IP address input
     var manualIpAddress by remember { mutableStateOf("") }
+    var manualIpPort by remember { mutableStateOf("4403") }
```
jamesarich: You can use the already available `NetworkRepository.SERVICE_PORT` constant here. That should _help_ make detekt happier.

slush0: I pushed fix ~10 hours ago, but Github still shows me "Processing updates". :shrug:

jamesarich: maybe it's because you're not actually on a branch.. i'm not sure we can merge master -> master

slush0: Never seen that this should be a problem before. There are some problems on status.github.com, and they were reported about the time when I pushed change, but they don't seem to be directly related. I'll try to push one empty commit if it get unstuck...

slush0: Ok, I squashed two commits into one and force pushed and it seems to be unstuck now. Fingers crossed with detekt now :).

slush0:
I don't know what's goin on with CI, it is again stuck for a day...

jamesarich (approved):
👍

Issue #1772, opened 2025-04-20
[Feature Request]: Segregate primary channels from secondary


b8b8:
### Tell us your idea.

I presented this idea on the Discord and it got positive feed back. My idea is to tweak the layout of the channel config page so it looks like below.

![Image](https://github.com/user-attachments/assets/952cd7d6-fe32-40d9-aa11-7f1b100335f1)

This is an updated layout i've been pondering. Clicking and dragging the channels to rearrange would still work as it does now. Maybe add the current frequency/channel-slot as well (and any other relevant data from the LoRa config to make it easier for new users to debug issues). I think it will help clear up some questions people have in regards to periodic broadcasts and be a nice segregation.

Much like the other config pages, I think we could benefit from explanations as part of the app, without requiring a connection to the internet and bringing up the docs pages. The point is to have it in the app like the other config explanations (which may not have been finished) and physically show that a primary channel is different than the other channels in the list.

![Image](https://github.com/user-attachments/assets/116c9fcf-4f44-40f1-9b16-86cdafefaa37)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jamesarich:
completed in #1873

PR #1873, opened 2025-05-19
Channel Configuration

Robert-0410:
## Update to the Channel Configuration Screen
Addresses feature request: #1772 

| Before | After |
|------|-----|
| Before | After |


Robert-0410:
## Update with material

update

app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/ChannelSettingsItemList.kt
```diff
@@ -225,45 +239,95 @@ fun ChannelSettingsItemList(
...
+                    dragDropState = dragDropState,
+                    haptics = LocalHapticFeedback.current,
+                ),
+                state = listState,
+                contentPadding = PaddingValues(horizontal = 16.dp),
+            ) {
+                item {
+                    Text(
+                        text = stringResource(R.string.primary),
+                        color = MeshtasticGreen
```
jamesarich: Let's try and stick with theme colors as much as possible - this color should be available as `MaterialTheme.colorScheme.primary`

app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/ChannelSettingsItemList.kt
```diff
@@ -225,45 +239,95 @@ fun ChannelSettingsItemList(
...
+                    if (index == 0 && !isDragging) {
+                        Text(
+                            text = stringResource(R.string.primary_channel_feature),
+                            color = MeshtasticGreen,
+                            fontSize = 10.sp,
+                        )
+                        Spacer(modifier = Modifier.height(16.dp))
+                        Text(
+                            text = stringResource(R.string.secondary),
+                            color = Color.Gray
```
jamesarich: Again, let's try to use theme colors - this would maybe be `MaterialTheme.colorScheme.onBackground` ?

jamesarich (changes requested):
For accessibility and theming purposes we'll definitely want to stick to theme colors where possible so that things appear as they should and with proper contrast when the theme changes.

jamesarich (approved):
nice work thanks!

PR #1876, opened 2025-05-19
Create a resend option for failed messages

RCGV1:
image


app/src/main/java/com/geeksville/mesh/ui/message/Message.kt
```diff
@@ -228,6 +230,7 @@ internal fun MessageScreen(
                     is NodeMenuAction.MoreDetails -> navigateToNodeDetails(action.node.num)
                 }
             }
+            )
         }
     }
 }
```
jamesarich, line 236:
```suggestion
}

```

jamesarich:
I probably messed up your local master :upside_down_face: .. for future pulls it's easier to keep your fork's master in sync with the main repo by starting a feature branch prior to commiting changes.

prokrypt:
Question: what happens if there's another max transmission event while the dialog is displayed?

Ideally the slashed cloud icon should be clickable to resend the message...

danditomaso:
Does a dialog make sense here? It can be quite intrusive to take over the users screen like this, it may provide more context if it's shown as a status icon near the failed message

RCGV1:
@danditomaso you have to tap the dialed to send icon for the popup also it's been merged already

danditomaso:
> @danditomaso you have to tap the dialed to send icon for the popup also it's been merged already
Awesome, thanks for the clarification!

PR #1893, opened 2025-05-21
Working on screen refactor

RCGV1:
[Screen_recording_20250520_233320 (1).webm](https://github.com/user-attachments/assets/ab888085-1364-4e77-940a-adf5b87d0402)


jamesarich:
i'm liking this idea - i'm wondering if the QR code channel import/sharing user experience could be integrated directly into the channel config screen, similar to how Contact import/sharing works.

Joseph-DiGiovanni:
I think channels are important enough they should keep their own tab. Users who don't know anything about the Meshtastic back-end are expected to configure them, so putting them in the same menu depth as something like GPIO pin definitions and the LoRa frequency offset seems counterproductive. Clear separation of app settings and radio settings is important as well.

If anything we should consider moving more user friendly options _out_ of the depths of the current radio configuration. I'd meet you in the middle and suggest replacing the channel tab with a new radio configuration menu of common options for the average person (long/short name, channels, and display/LED options), and relegate anything else to an "advanced radio configuration" menu much like the one we currently have.

PR #1936, opened 2025-05-25
fix : prevent crash on geo: intent when no compatible map app is inst…

emp3r0r7:
### Summary

Fixes a crash that occurred when tapping on `geo:` links in the app, on devices that do not have a compatible maps application installed.

Eg. By navigating to the nodes list and clicking on the coordinates of a node using a device with no google maps  (i use a debloated Android phone so no Google stuff) or any other similar mapping application deactivated or uninstalled would cause an application crash.

### Changes

- Replaced usage of `uriHandler.openUri(...)` with a proper `Intent` using `ACTION_VIEW`
- Added `resolveActivity(...)` check to avoid `ActivityNotFoundException`
- Added a user-friendly `Toast` if no app can handle the `geo:` intent
- Set `FLAG_ACTIVITY_NEW_TASK` to support usage from a non-Activity context (Compose)
- Updated `AndroidManifest.xml` with a `<queries>` block to support Android 11+ visibility rules for geo intents

### Testing

- Emulator (no Maps installed): shows fallback Toast, no crash
- Device with Google Maps: launches map as expected
- Long press: still copies the coordinate string to clipboard


emp3r0r7:
> hi @emp3r0r7 thanks, let me know if you need an assist with getting `detekt` working locally

thanks i didnt notice there was detekt  , first time contributing this project , i just pushed right now,  locally detekt builds fine

PR #1958, opened 2025-05-28
Added Node Count when on nodes screen

RCGV1:
image


b8b8:
Can this be amended to show online nodes vs. total node count so you dont have to pull down the shade to see online node count?

 (40/385)

jamesarich:
closed in favor of #2209 which updates on db changes

PR #1985, opened 2025-05-30
fix: clear data when changing devices

jamesarich:
This introduces a change where the device state databases are cleared when the connected device address changes. This ensures that stale data from a previous device does not persist.

Specifically, the following changes were made:
- Modified `NodeRepository` to no longer clear the node info when the MyNodeNum is different, as this is now handled when the device address changes.
- Updated `MeshService` to:
    - Store and track the last connected device address in SharedPreferences.
    - When the device address changes, clear the NodeDB, PacketDB, and MeshLogDB.
- Added `clearPacketDB` function to `PacketRepository` to allow deleting all packets.
- Added `deleteAll` query to `PacketDao` to support clearing the packet database.
- Clear databases on device Factory Reset and NodeDB Reset requests


Afettu:
Please due to this change i can't find the logs stored, in particular Signal Metrics Log from a reference node i'm using to tune my antenna. Is there a way to recover them? Thanks.

PR #2040, opened 2025-06-06
feat: show per-message SNR, RSSI and hop count

lkosson:
This commit adds a line at the bottom of each received message (next to date and time) with the reception stats - either SNR and RSSI for directly received packets or hop count for relayed ones.


app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -744,6 +753,13 @@ class MeshService : Service(), Logging {
             val bytes = data.payload.toByteArray()
             val fromId = toNodeID(packet.from)
             val dataPacket = toDataPacket(packet)
```
jamesarich: Since the MeshPacket contains all the data needed to calculate these fields, the `toDataPacket(packet)` should be updated to handle the logic for `hopAway` `snr` and `rssi`. This method already encapsulates the logic to convert a MeshPacket to a DataPacket and would eliminate the need to pass them in every call to `rememberDataPacket()`

jamesarich:
Also - we use https://detekt.dev/ to help maintain clean code. Let us know if you need assistance in setting up locally.

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
lkosson: That was my original approach, but I didn't want to touch `DataPacket`, because I don't really understand how `Parcel` serialization works and how (if at all) it is versioned. I'll give it another try, though. Passing all those parameters to `MeshService.rememberDataPacket` smells quite bad.

lkosson: I have added those fields to `DataPacket`.

lkosson:
_detekt_ is now working and reporting no issues. Please review the new changeset.

lkosson:
I just have noticed there are existing, reusable snr and rssi indicators, so I've switched to using those.

Not sure if it fits here, but yellow ones are barely readable on light color theme:

jamesarich (approved):

jamesarich:
thanks for the add @lkosson  - reach out if you're on the [discord](https://discord.com/invite/ktMAKGBnBs)!

PR #2052, opened 2025-06-07
feat: allow hiding offline and/or non-direct nodes from list and map

lkosson:
This change allows filtering out offline (as per existing `NodeInfo.isOnline` logic) and non-directly-reachable (i.e. those with `NodeInfo.hopsAway` greater than zero) nodes from the node list and map.


app/src/main/java/com/geeksville/mesh/model/UIState.kt
```diff
@@ -285,21 +290,37 @@ class UIViewModel @Inject constructor(
...
+            sort = flows[1] as NodeSortOption,
+            filter = flows[0] as String,
+            includeUnknown = flows[2] as Boolean,
+            onlyOnline = flows[3] as Boolean,
+            onlyDirect = flows[4] as Boolean,
             gpsFormat = profile.config.display.gpsFormat.number,
             distanceUnits = profile.config.display.units.number,
             tempInFahrenheit = profile.moduleConfig.telemetry.environmentDisplayFahrenheit,
-            showDetails = showDetails,
+            showDetails = flows[5] as Boolean,
```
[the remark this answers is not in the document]
jamesarich: I'm kinda with Copilot on this one - I think we're losing some context by condensing these.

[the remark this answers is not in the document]
lkosson: `combine` has only overloads for up to 5 parameters, and so does destructuring operator for `Array`, so neither `sort, filter, includeUnknown, onlyOnline, onlyDirect, showDetails, profile ->` nor `(sort, filter, includeUnknown, onlyOnline, onlyDirect, showDetails, profile) ->` works. Any suggestions on how to achieve it? I haven't really used Kotlin before.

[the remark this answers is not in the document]
jamesarich: How about creating a FilterState to house these?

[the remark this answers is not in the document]
lkosson: Of course. I have factored out filter-related state.

jamesarich (changes requested):
Looks like we need to update `NodeInfoDaoTest` to inclcude the new fields (and potentially some additional tests for the new fields).

lkosson:
I'm having some problems running test suite locally. What is the correct way to invoke them? `gradlew test` fails with compilation errors (mostly unresolved references in `mesh_service_example`), even on clean checkout from repo tip.

jamesarich:
> I'm having some problems running test suite locally. What is the correct way to invoke them? `gradlew test` fails with compilation errors (mostly unresolved references in `mesh_service_example`), even on clean checkout from repo tip.

CI is using the command `./gradlew connectedFdroidDebugAndroidTest` which requires a running emulator. You can also run this test from within Android Studio: (note the green play arrow next to the test class)
image

or

image

lkosson:
I have added new tests and updated `getNodes`. There seems to be a problem with existing `testIncludeUnknownIsFalse` - `Node.isUnknownUser` checks `hwModel` for `HardwareModel.UNSET`, while acutal implementation of `NodeInfoDao` uses `short_name` to determine whether to include a node when `includeUnknown` is set.

Is it a bug, or I am missing something? If it is a bug, should I fix it in this PR, or file a new issue? If I should fix it, which behavior is correct - using `hwModel` or `short_name` for determining `isUnknownUser`?

jamesarich:
> Is it a bug, or I am missing something? If it is a bug, should I fix it in this PR, or file a new issue? If I should fix it, which behavior is correct - using `hwModel` or `short_name` for determining `isUnknownUser`?

Good catch - if you want to add that to this PR 👍 , that'll save some legwork with making a new issue.

jamesarich:
also i guess use the `short_name` field, as it's what the dao is actually using on the filter

lkosson:
Is there any piece of documentation or explanation when to use User's or Node's shortName?

And - kind of related - what is the difference between `NodeInfo` and `Node`?

app/src/main/java/com/geeksville/mesh/database/entity/NodeEntity.kt
```diff
@@ -45,6 +46,7 @@ data class NodeWithRelations(
             num = num,
             metadata = metadata?.proto,
             user = user,
+            shortName = shortName,
```
jamesarich: we can just use `user.shortName` here no need for an additional field

app/src/main/java/com/geeksville/mesh/model/Node.kt
```diff
@@ -35,6 +35,7 @@ data class Node(
     val num: Int,
     val metadata: MeshProtos.DeviceMetadata? = null,
     val user: MeshProtos.User = MeshProtos.User.getDefaultInstance(),
+    var shortName: String? = null,
```
jamesarich: again we should be able to use `user.shortName` here

lkosson: Not really. `NodeInfoDaoTest` mockup data for `unknownNode` leaves `shortName` empty, but sets `user.shortName`. If using `user.shortName` for `isUnknownUser`, `testIncludeUnknownIsTrue` fails. I'm not too familiar with the project, but as far as I can tell, unknown nodes have "Meshtastic ABCD" `longName` and "ABCD" `shortName`, where "ABCD" is computed from node number, which can't be null anyway if we have received any packet from that node.

jamesarich: :thinking: i'm unfamiliar with this filter/test as well.. why don't we roll back the changes made to address this test bug, and we can break it off into it's own issue and give it a little more thought. I don't want it to hold back the rest of your PR.

lkosson: I have reverted changes to `isUnknownUser` and change mock up data to better fit existing logic.

PR #2057, opened 2025-06-08
fix: cancel new message notification when message is marked as read

lkosson:
When a message list with unread messages is opened and scrolled to most recent item, an obsolete notification about unread messages in this channel/DM is now cancelled. Prior to this change notification had to be manually dismissed.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/model/UIState.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/UIState.kt
@@ -61,6 +61,7 @@ import com.geeksville.mesh.repository.datastore.RadioConfigRepository
 import com.geeksville.mesh.repository.location.LocationRepository
 import com.geeksville.mesh.repository.radio.RadioInterfaceService
 import com.geeksville.mesh.service.MeshService
+import com.geeksville.mesh.service.MeshServiceNotifications
 import com.geeksville.mesh.service.ServiceAction
 import com.geeksville.mesh.ui.map.MAP_STYLE_ID
 import com.geeksville.mesh.ui.node.components.NodeMenuAction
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/geeksville/mesh/model/UIState.kt
```diff
@@ -190,6 +191,8 @@ class UIViewModel @Inject constructor(
     private val preferences: SharedPreferences
 ) : ViewModel(), Logging {
 
+    private val notifications = MeshServiceNotifications(app)
```
jamesarich: MeshServiceNotifications are already instantiated in the MeshService you could use that instance, or create a function in MeshService to call.

lkosson: That was the original plan, but `UIState` has `IMeshService`, so adding a method would involve updating `aidl` and since that's a public API, I couldn't tell if changing it wouldn't break some external applications. I also assumed exposing whole `MeshServiceNotifications` on such public interface wasn't best idea.

I'll give it a second look, though.

jamesarich: Appreciate the doublecheck, maybe we can look at using hilt to manage dependency injection of MeshServiceNotification singleton where it needs to go.

lkosson: I have switched to reusing existing `MeshServiceNotifications` instance. Couldn't figure out how to expose whole object via a service interface, so I just added `cancelMessageNotification` method. I don't like it too much (IMO it shouldn't be a public API), but still better than creating a new instance like before.

jamesarich:
hi @lkosson i submitted a PR against your branch to update this to use hilt DI

lkosson:
Thanks, I have merged back your changes.

PR #2065, opened 2025-06-09
Added map filter + small waypoint fix

RCGV1:
image

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/map/MapView.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/map/MapView.kt
@@ -24,15 +24,25 @@ import androidx.appcompat.content.res.AppCompatResources
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Lens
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/geeksville/mesh/ui/node/components/NodeStatusIcons.kt
```diff
@@ -70,7 +69,6 @@ fun NodeStatusIcons(
                         contentDescription = stringResource(R.string.unmessageable),
                         modifier = Modifier
                             .size(24.dp), // Smaller size for badge
-                        tint = MaterialTheme.colorScheme.error,
```
jamesarich: this isn't part of the intent of the Pull Request - why was this changed? please revert.

app/src/main/java/com/geeksville/mesh/ui/map/MapView.kt
```diff
@@ -666,6 +790,7 @@ fun MapView(
                         if (name == "") name = "Dropped Pin"
                         if (expire == 0) expire = Int.MAX_VALUE
                         lockedTo = if (waypoint.lockedTo != 0) model.myNodeNum ?: 0 else 0
+                        if (waypoint.icon == 0) icon = 128205
```
jamesarich, line 793: what does this change do - what is `128205`?

jamesarich (changes requested):
Please don't make changes outside the original intent of the pull request, and if you do make sure to call them out appropriately in the description.

app/src/main/java/com/geeksville/mesh/ui/map/MapView.kt
RCGV1: If the waypoint icon is not changed it's 0 by default so instead of just sending 0 it will default to a pin emoji

RCGV1:
@jamesarich I reverted

PR #2073, opened 2025-06-10
Added connection indicator

RCGV1:
Added a connection indicator to the node status icons, which changes based on whether it is currently connected or disconnected. This can more easily show connection status and can show which node is yours.
image
image

### DIFF
--- a/app/src/main/assets/device_hardware.json
+++ b/app/src/main/assets/device_hardware.json
@@ -877,9 +877,9 @@
   {
     "hwModel": 94,
     "hwModelSlug": "HELTEC_MESH_POCKET",
-    "platformioTarget": "heltec_mesh_pocket",
+    "platformioTarget": "heltec-mesh-pocket-10000",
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/geeksville/mesh/ui/node/components/NodeStatusIcons.kt
```diff
@@ -40,16 +42,55 @@ import androidx.compose.ui.tooling.preview.Preview
...
     ) {
+        if (isThisNode) {
+            TooltipBox(
+                positionProvider = TooltipDefaults.rememberTooltipPositionProvider(),
+                tooltip = {
+                    PlainTooltip {
+                        Text(
+                            stringResource(
+                                if (isConnected) {
+                                    R.string.connected_to
```
jamesarich: This string resource is a template `Connected to radio (%s)` consider either passing in the longName or using a different string.

PR #2097, opened 2025-06-12
refactor: maps

jamesarich:
# Google Maps (and 16kb page size compliance)

## Why?
Google has a new policy to be 16 kb page size aligned - OSMDroid is  currently the only thing in our application preventing compliance. It is also a dead/archived repository, so there's little chance it will be updated in time to meet compliance by deadline (Nov 1). In order to maintain compliance, we needed to:
- Find a suitable replacement maps sdk that would satisfy both Fdroid and Google compliance (did not find one)
- OR
- Branch the codebase and use different sdks to maintain compliance per flavor (went this route) 

## What
This commit restructures the map-related code to support different map providers (OSMDroid for F-Droid, Google Maps for Google Play). In addition to replicating current OSMDroid functionality in Google Maps, I was able to introduce a few more Quality of Life improvements to map functionality.

## Key changes:
- Moved OSMDroid-specific map files to the `fdroid` source set.
- Implemented  Google Maps views and supporting logic in the `google` source set.
- Added Google Maps Compose dependencies and the secrets plugin to `build.gradle.kts`.
- Added Tile Source management - can add custom map tiles via url
- Added KML/KMZ layer management - import and add layers to your map ( 👋  Burning Man folks)


.gitignore
```diff
@@ -27,5 +27,5 @@ keystore.properties
 # Kotlin compiler
 .kotlin
 
-# VS code
```
DaneEvans, line 30: why you override my ignores??

app/src/google/java/com/geeksville/mesh/android/GeeksvilleApplication.kt
```diff
@@ -66,8 +78,9 @@ open class GeeksvilleApplication : Application(), Logging {
 
     /** Ask user to rate in play store */
     fun askToRate(activity: AppCompatActivity) {
-        if (!isGooglePlayAvailable()) return
+        if (!isGooglePlayAvailable) return
 
+        @Suppress("MaxLineLength")
```
DaneEvans, line 113: shouldn't need this

app/src/google/java/com/geeksville/mesh/android/GeeksvilleApplication.kt
```diff
@@ -33,26 +33,38 @@ import com.suddenh4x.ratingdialog.AppRating
  * Created by kevinh on 1/4/15.
  */
 
+@Suppress("MagicNumber")
```
DaneEvans: is it just the min launch times ? 

Define it at the top as,  do it properly

DaneEvans (commented):
Try and suppress lines, or at most functions, rather than whole files 

All the incidental changes look good, need to test and go through the new code in more detail.

DaneEvans:
Comments on the google variant: 
- [x] defaults to some random US location - Black rock desert, nv.
- [x] no 'to me' button
- [ ] Something odd happening ( some nodes in Aus. appear off the coast of Ghana @ appx 0,0)
   - [ ] some of mine are in the right place - those with fixed locations I think ... 
   - [ ] one of those at 0,0 seems to have no location info. another has a GPS, and it's 'node map' puts it in the right place. and last position log made sense. 
- [x]  no clustering? 
- [x] precision circles are very small - much smaller than on open maps, but may be accurate, I'll need to check what they are  broadcasting. 
- [x] Directions etc are working.
- [x] didn't test adding layers

jamesarich:
> Comments on the google variant:
> 
> * [x]  defaults to some random US location - Black rock desert, nv.

was testing kml layer data for burningman, working on setting up an initial bounding box based on the nodelist and waypoints

> * [x]  no 'to me' button

added perms check for location and made sure `myLocation` button shows up

> * [ ]  Something odd happening ( some nodes in Aus. appear off the coast of Ghana @ appx 0,0)
>   
>   * [ ]  some of mine are in the right place - those with fixed locations I think ...
>   * [ ]  one of those at 0,0 seems to have no location info. another has a GPS, and it's 'node map' puts it in the right place. and last position log made sense.

pretty easy to filter out nodes with `positionValid==false` i'll get that in

> * [x]   no clustering?

clustering should be working, and added a dialog when multiple nodes report the same location

> * [x]  precision circles are very small - much smaller than on open maps, but may be accurate, I'll need to check what they are  broadcasting.

good catch - needed some conversion from precison bits to meters

> * [x]  Directions etc are working.
> * [x]  didn't test adding layers

moved layer ui into a bottom sheet

DaneEvans:
I'll have another look at this one over teg weekend

app/src/fdroid/java/com/geeksville/mesh/model/map/CustomTileSource.kt
```diff
@@ -0,0 +1,230 @@
...
+        //                        or TileSourcePolicy.FLAG_USER_AGENT_NORMALIZED
+        //            )
+        //        ) {
+        //            override fun getTileURLString(pMapTileIndex: Long): String {
+        //                return baseUrl + (MapTileIndex.getZoom(pMapTileIndex)
+        //                    .toString() + "/" + MapTileIndex.getY(pMapTileIndex)
+        //                        + "/" + MapTileIndex.getX(pMapTileIndex)
+        //                        + mImageFilenameEnding)
+        //            }
+        //        }
```
DaneEvans: should this get wiped?

app/src/google/java/com/geeksville/mesh/di/MapModule.kt
```diff
@@ -0,0 +1,68 @@
...
+
+@Qualifier
+@Retention(AnnotationRetention.BINARY)
+annotation class DefaultDispatcher
+
+@Module
+@InstallIn(SingletonComponent::class)
+object MapModule { // Renamed from DispatchersModule for a more general name
+
+    // Dispatcher Providers (from original DispatchersModule)
```
DaneEvans: There are a lot of AI comments in this file.

app/src/google/java/com/geeksville/mesh/repository/map/SharedPreferencesCustomTileProviderRepository.kt
```diff
@@ -0,0 +1,111 @@
...
+
+private const val PREFS_NAME_TILE = "map_tile_provider_prefs" // Renamed to avoid conflict if PREFS_NAME is global
+private const val KEY_CUSTOM_TILE_PROVIDERS = "custom_tile_providers"
+
+@Singleton // Make it a singleton if managed by Hilt
+class SharedPreferencesCustomTileProviderRepository
+@Inject
+constructor(
+    @ApplicationContext private val context: Context,
+    private val json: Json, // Inject Json instance if configured globally, or create one here
```
DaneEvans: a lot of comments to clean up in here too.

app/src/main/java/com/geeksville/mesh/ui/sharing/Channel.kt
```diff
@@ -211,7 +211,7 @@ fun ChannelScreen(
             channelSet = channels // Throw away user edits
 
             // Tell the user to try again
-            viewModel.showSnackbar(R.string.cant_change_no_radio)
```
DaneEvans, line 214: why is this changing in this MR?

app/src/fdroid/java/com/geeksville/mesh/model/map/CustomTileSource.kt
jamesarich: Probably - that file was moved as-is to the fdroid flavor. It's probably been commented out for years 😅

app/src/main/java/com/geeksville/mesh/ui/sharing/Channel.kt
jamesarich: oh - renamed the function. Probably not necessary for this MR, but inconsequential.

DaneEvans:
FDroid: 
Looks good

Google: 
- [ ] Initial zoom with only one node is like 10m (30'), probably should have that a bit wider so it shows some context 
- [ ] To me goes to San Jose when not given anything else. (the initial box is probably better) 

More seriously
- [ ] Crash:  Node Info > Node map, on a node which has that data. looks like a map image issue

jamesarich:
> FDroid: Looks good
> 
> Google:
> 
> * [ ]  Initial zoom with only one node is like 10m (30'), probably should have that a bit wider so it shows some context
> * [x]  To me goes to San Jose when not given anything else. (the initial box is probably better)

emulator location set to google hq 😂  - `to me` should function as it should if location is accurate

> 
> More seriously
> 
> * [x]  Crash:  Node Info > Node map, on a node which has that data. looks like a map image issue

this was the biggie - i hadn't spent much time over on the nodemap. I spruced it up.

PR #2104, opened 2025-06-13
feat: show node details by long-pressing a node marker on map

lkosson:
Long-pressing a map node marker opens up a node detail screen.

Also factored out `MapRoutes` from `NavGraph` for consistency with other UI areas and for future map-related screens.


jamesarich:
hi @lkosson - nice add. 
We've been talking about splitting out the maps functionality for google and fdroid builds.
Draft here if you're interested #2097

jamesarich (approved):

PR #2113, opened 2025-06-14
add copy to pubkey from the lock icon

Dane-2pi:
Needs testing on HW


jamesarich:
Let me know when you get it tested.

Dane-2pi:
Will come in from the other account once completed

Issue #2064, opened 2025-06-09
[Bug]: Unmessageable toggle is not available over admin


b8b8:
### What happened?

I noticed today that the new unmessageable toggle is not available over admin. 
Certainly that makes sense for a router due to their intended use, but in many situations a client is acting in a router role so should be able to be set remotely as unmessageable.

App:         2.6.13
Phone:     Pixel 9a
Devices:   T1000E trying to set a station G2 (client) to unmessageable
Firmware: 2.6.10 on all nodes in question

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

b8b8:
Update: I have two tracker role devices on my mesh that I can admin. The T-echo is seen only through a UDP bridge on a different frequency than LongFast. I can remotely toggle it from unmessageable to meassageable. The T114 cannot be toggled even though it has otherwise identical settings.

![Image](https://github.com/user-attachments/assets/eaea1d7c-c27a-4442-a8e1-c420e5868c80)

![Image](https://github.com/user-attachments/assets/ef39954b-b92d-49ab-9219-031f5a3549a7)

![Image](https://github.com/user-attachments/assets/8e2cb049-a63a-4ae6-be7b-aa3d3a48507b)

jamesarich:
UX has been updated to require metadata prior to enabling remote administration

b8b8:
This may still be an issue.

b8b8:
so this is having issues while connected over bluetooth as well. i connected directly to my client station g2 and want to set it to unmessageable but it is grayed out.

b8b8:
The unmessageable toggle becomes available over Bluetooth if I disconnect from android, Bluetooth connect using iOS, then toggle the feature on.

The toggle then remains (and can switch both on and off and back on again) accessible after reconnecting to android over both Bluetooth and remote admin.

@jamesarich

valzzu:
can confirm

PR #2139, opened 2025-06-16
Fix unmessageable toggle not available over remote admin

jeremiah-k:
## Problem
Unmessageable toggle was disabled when accessing remote admin due to unreliable firmware detection.

## Solution
- Always show toggle for consistent UX
- Enable only for firmware >= 2.6.9 AND admin permissions
- Show role-based status for older firmware (read-only)

## Changes
- Replace `enabled = userInput.hasIsUnmessagable()` with `enabled = enabled && firmwareVersion >= DeviceVersion("2.6.9")`
- Keep checked state logic for backward compatibility

## Result
- ✅ Toggle available in remote admin for supported firmware
- ✅ Clear feedback when feature unavailable (disabled toggle)

Tested with a rak4631 firmware v2.6.11 (toggle enabled) and t-beam v2.6.1 (toggle disabled)

Tested and working on a TCL T768S (Android 11)

Closes: #2064

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/node/NodeDetail.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/node/NodeDetail.kt
@@ -328,7 +328,7 @@ private fun NodeDetailList(
                 NavCard(
                     title = stringResource(id = R.string.remote_admin),
                     icon = Icons.Default.Settings,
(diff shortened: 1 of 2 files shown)


jamesarich:
This is using the appropriate `hasIsUnmessageable()` check already.

PR #2147, opened 2025-06-17
feat: Implement message reply functionality

jamesarich:
This commit introduces the ability to reply to messages within the application.

Key changes include:

- **Data Model Updates:**
    - `DataPacket` now includes an optional `replyId` field to link replies to original messages.
    - `Packet` entity in the database is updated to store `replyId`, defaulting to -1.
    - `Message` model now includes an optional `replyId`.

- **UI Enhancements:**
    - **Message Input:** - A visual indicator ("Replying to...") appears above the text input when a reply is being composed. - Users can cancel the reply using a close button.
    - **Message Display:**
        - Replies now display a quoted snippet of the original message.
        - The original message snippet is clickable, navigating the user to the original message in the chat history. - A "Reply" icon button is added to the message actions (next to reactions).
    - **MessageItem:** - Updated to handle and display reply information. - Takes an `originalMessage` parameter to show the quoted reply. - `onNavigateToOriginalMessage` callback for jumping to the original message.
    - **MessageScreen:**
        - Manages the `replyingTo` state.
        - Passes the `replyId` when sending a message.

- **Logic Updates:**
    - `UIViewModel.sendMessage` now accepts an optional `replyId`.
    - `MeshService` correctly processes and stores `replyId` for incoming and outgoing messages.
    - Incoming text messages with a `replyId` and no emoji are now correctly identified and processed as replies.

This feature enhances user interaction by allowing direct responses to specific messages, improving context and clarity in conversations.

https://github.com/user-attachments/assets/3e51503c-e4f2-4f5c-8288-bffc54d7cfb0

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/message/components/MessageItem.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/message/components/MessageItem.kt
@@ -27,9 +27,12 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.FormatQuote
 import androidx.compose.material.icons.twotone.Cloud
 import androidx.compose.material.icons.twotone.CloudDone
 import androidx.compose.material.icons.twotone.CloudOff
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/geeksville/mesh/ui/message/components/MessageItem.kt
```diff
@@ -152,27 +207,33 @@ internal fun MessageItem(
...
+                            )
+                            Rssi(
+                                message.rssi,
+                                fontSize = MaterialTheme.typography.labelSmall.fontSize
+                            )
                         }
                     } else {
                         Text(
-                            text = hopsAway,
+                            text = "${message.hopsAway}",
```
lkosson, line 225: With this change, a "Hops away: " label (`R.string.hops_away` - previously added in `MessageList` method) got removed. Is this intentional?

jamesarich: nope - nice catch 🔍

Issue #2118, opened 2025-06-14
[Feature Request]: Zoom in/around in the debug panel


willhubs:
### Tell us your idea.

Either using a small screened device or slowly going blind, it would be awesome if we could zoom in/around the debug panel, sometimes it's extremely hard to read the small letters.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
Can you get a screen cap @willhubs, I want to see how bad it is on yours. I don't find it too bad, but  I'm going to be in there while I add some of the stuff in #2105 so improving it might be possible (or you can then copy the contents to somewhere else)

willhubs:
I can do that, on my current device I don't have much of a problem as it has a large screen, but one of my backups devices has a pretty small screen.  Would probably be easier to have it get focused/larger when selected instead of an actual pinch to zoom type of feature.

DaneEvans:
#2148 closes

DaneEvans:
@jamesarich, finished this

willhubs:
Thanks!  Works perfectly.

jamesarich:
I love it when a plan comes together.

Issue #2105, opened 2025-06-13
[Feature Request]: Debug Panel enhancements


DaneEvans:
### Contact Details

@DaneEvans

### Tell us your idea.

I've been trying to use the debug panel to do some debug (and have ended up logging on a PC instead) 

Debug Panel appears to have a few shortcomings which would be improved by the following
- [x] Search 
- [x] filtering (even just to / from or portnum would make it a lot easier to find things) 
- [x] Ability to copy out a Packet so you can look at it elsewhere 
   - [ ] I really want that to decode the payload, so if it's a wishlist, add decoding the payload to it
- [x] export the whole log to a file where it can be looked at elsewhere. 
_This is really to get around 1&2 above in my case, but other people will probably see another use_

- [ ] #2118  - Make log messages larger when selected

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
wip branch : https://github.com/DaneEvans/Meshtastic-Android/tree/feat/2105_debug_panel

DaneEvans:
#2148 closes

PR #2148, opened 2025-06-18
Feat/2105 debug panel

DaneEvans:
#2105 & #2118

# Improvements to the debug panel 
- [x] Add search 
- [x] Add filters 
- [x] Add export logs 
- [x] Copy Packet info to clipboard
- [x] Make text larger on selected ones

## TODOs 
- [ ] Add strings in other languages - is there a process to automate that? 
- [ ] Are there any other filters that should be in the default list? 
   - [ ] add "To me" filter (get own address from attatched nodes]
- [ ] conflate CONFIG_APP to 'config' etc. on packets that do get fully decoded
- [ ] Look at decoding additional packet types 

## Screenshots 
Base screen
image
image

Search  & Zoom on selected
image
image

Filters 
image
image

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/model/DebugViewModel.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/DebugViewModel.kt
@@ -36,6 +36,8 @@ import kotlinx.coroutines.launch
 import java.text.DateFormat
 import java.util.Locale
 import javax.inject.Inject
+import com.geeksville.mesh.Portnums.PortNum
+
 
 @HiltViewModel
 class DebugViewModel @Inject constructor(
@@ -134,4 +136,9 @@ class DebugViewModel @Inject constructor(
     companion object {
         private val TIME_FORMAT = DateFormat.getDateTimeInstance(DateFormat.SHORT, DateFormat.MEDIUM)
     }
+
+    val presetFilters = arrayOf(
+        "own 0x", // will need to get that from connected nodes??
+        "0xffffffff",     // broadcast
+    ) + PortNum.entries.map { it.name } // all apps
 }
--- a/app/src/main/java/com/geeksville/mesh/model/DebugViewModel.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/DebugViewModel.kt
@@ -138,7 +138,7 @@ class DebugViewModel @Inject constructor(
     }
 
     val presetFilters = arrayOf(
-        "own 0x", // will need to get that from connected nodes??
-        "0xffffffff",     // broadcast
+        // "!xxxxxxxx", // Should be able to get the address of the connected node. ie. messages to us.
+        "!ffffffff",     // broadcast
     ) + PortNum.entries.map { it.name } // all apps
 }
--- a/app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
@@ -42,6 +42,8 @@ import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.foundation.text.KeyboardActions
 import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.foundation.text.selection.SelectionContainer
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.BorderStroke
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Add
 import androidx.compose.material.icons.outlined.CloudDownload
@@ -113,6 +115,7 @@ internal fun DebugScreen(
     var showFilterMenu by remember { mutableStateOf(false) }
     var searchText by remember { mutableStateOf("") }
     var currentMatchIndex by remember { mutableStateOf(-1) }
+    var selectedLogId by remember { mutableStateOf<String?>(null) }
     
     val filteredLogs by remember(logs) {
         derivedStateOf {
@@ -415,7 +418,11 @@ internal fun DebugScreen(
             DebugItem(
                 modifier = Modifier.animateItem(),
                 log = log,
-                searchText = searchText
+                searchText = searchText,
+                isSelected = selectedLogId == log.uuid,
+                onLogClick = { 
+                    selectedLogId = if (selectedLogId == log.uuid) null else log.uuid
+                }
             )
         }
     }
@@ -426,20 +433,36 @@ internal fun DebugItem(
     log: UiMeshLog,
     modifier: Modifier = Modifier,
     searchText: String = "",
+    isSelected: Boolean = false,
+    onLogClick: () -> Unit = {}
 ) {
+    val theme = androidx.compose.material3.MaterialTheme.colorScheme
+    
     Card(
         modifier = modifier
             .fillMaxWidth()
             .padding(4.dp),
+        colors = androidx.compose.material3.CardDefaults.cardColors(
+            containerColor = if (isSelected) 
+                theme.primary.copy(alpha = 0.1f) 
+            else 
+                theme.surface
+        ),
+        border = if (isSelected) {
+            androidx.compose.foundation.BorderStroke(2.dp, theme.primary)
+        } else null
     ) {
         SelectionContainer {
             Column(
-                modifier = Modifier.padding(8.dp)
+                modifier = Modifier
+                    .padding(if (isSelected) 12.dp else 8.dp)
+                    .fillMaxWidth()
+                    .clickable { onLogClick() }
             ) {
                 Row(
                     modifier = Modifier
                         .fillMaxWidth()
-                        .padding(bottom = 8.dp),
+                        .padding(bottom = if (isSelected) 12.dp else 8.dp),
                     horizontalArrangement = Arrangement.SpaceBetween,
                     verticalAlignment = Alignment.CenterVertically,
                 ) {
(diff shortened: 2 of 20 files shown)


DaneEvans:
Tests passing. 
I'm done with tweaks, except maybe renaming some functions that detekt made me create. 

Welcome any feedback, and the answers to the questions in the description

jamesarich:
Please remove translated strings. That happens via crowdin.

jamesarich:
Also - for translations, just add the english to the base `strings.xml` file - they will get synced via github action to [Crowdin](https://meshtastic.crowdin.com/) for crowdsourced translations.

We have [actions](https://github.com/meshtastic/Meshtastic-Android/actions) that upload the latest source strings, and pull the most recent translations into the repo from crowdin.

DaneEvans:
Done. Still 16 functions, but 10 of them are previews, so I've supressed it. 
It is possible to configure detekt to ignore Previews in the count if that would be helpful, but I'll defer to you. 
And its definitely a different MR

app/src/main/java/com/geeksville/mesh/ui/debug/DebugFilters.kt
```diff
@@ -0,0 +1,261 @@
...
+                    onClick = {
+                        onFilterTextsChange(
+                            if (filter in filterTexts) {
+                                filterTexts - filter
+                            } else {
+                                filterTexts + filter
+                            }
+                        )
+                    },
+                    label = { Text(filter) }
```
jamesarich: To better indicate that this filter is selected let's use the additional checkmark leading icon shown here

image

https://developer.android.com/develop/ui/compose/components/chip#filter

app/src/main/java/com/geeksville/mesh/ui/debug/DebugSearch.kt
```diff
@@ -0,0 +1,361 @@
...
+    presetFilters: List<String>,
+    onSearchStateChange: (SearchState) -> Unit,
+    onFilterTextsChange: (List<String>) -> Unit,
+    onSelectedLogIdChange: (String?) -> Unit
+) {
+    var customFilterText by remember { mutableStateOf("") }
+    var searchText by remember { mutableStateOf("") }
+    var currentMatchIndex by remember { mutableStateOf(-1) }
+    var selectedLogId by remember { mutableStateOf<String?>(null) }
+    val theme = androidx.compose.material3.MaterialTheme.colorScheme
```
jamesarich: Prefer importing MaterialTheme and using MaterialTheme.colorScheme instead of the fully qualified name.

jamesarich: also consider renaming to `colorScheme` to avoid confusion

app/src/main/java/com/geeksville/mesh/ui/debug/DebugSearch.kt
```diff
@@ -0,0 +1,361 @@
...
+                        )
+                    }
+                }
+            }
+        }
+    )
+}
+
+@Composable
+internal fun DebugSearchState(
```
jamesarich, line 152: Since is this state is local to the UI, and not retained in the viewmodel - make sure there are no unintended side-effects when configuration changes. https://developer.android.com/guide/topics/resources/runtime-changes#react-changes-compose

jamesarich (changes requested):
Hey, had a chance to go through this in a bit more detail.  Let me know if you have any questions.

app/src/main/java/com/geeksville/mesh/ui/debug/DebugFilters.kt
```diff
@@ -0,0 +1,261 @@
...
+                    .fillMaxWidth()
+                    .padding(bottom = 0.dp),
+                horizontalArrangement = Arrangement.spacedBy(2.dp),
+                verticalArrangement = Arrangement.spacedBy(0.dp)
+            ) {
+                for (filter in filterTexts) {
+                    FilterChip(
+                        selected = true,
+                        onClick = {
+                            onFilterTextsChange(filterTexts - filter)
```
jamesarich, line 260: since this filter is being removed when clicked on - let's indicate so with a trailing close icon.
image

jamesarich:
> It is possible to configure detekt to ignore Previews in the count if that would be helpful, but I'll defer to you.
> And its definitely a different MR

I'd like to understand better why so many previews are necessary, and why we'd want to alter any [detekt compose rules](https://detekt.dev/docs/introduction/compose/) or break any [compose style guidelines](https://github.com/androidx/androidx/blob/androidx-main/compose/docs/compose-api-guidelines.md) .

This does now seem like a very multi-stateful view. Validation of the functionality may be better accomplished by writing some [UI tests](https://developer.android.com/develop/ui/compose/testing).

DaneEvans:
At the current state: 

image

image

Any other tweaks before I look into testing? 


The ignore previews is under an option in https://detekt.dev/docs/introduction/compose/ , that's why I raised it. 
As to why so many, there are a lot of states on that page - I could pull them off into the search and filter helpers, but they also need to play nicely together. I could also just delete one if it's that big a deal

app/src/main/java/com/geeksville/mesh/ui/debug/DebugSearch.kt
DaneEvans: Do you have any suggestions on how to test that exhastively?

jamesarich:
> The ignore previews is under an option in [detekt.dev/docs/introduction/compose](https://detekt.dev/docs/introduction/compose/) ,

ahh it looks like there _is_ a suggested rule for that, that is not currently being used - potentially new or missed originally, go ahead and toss it in.

>Do you have any suggestions on how to test that exhastively?

not exhastively, but a good quick check is rotating the screen and making sure state is retained as expected.

app/src/main/java/com/geeksville/mesh/ui/debug/DebugFilters.kt
```diff
@@ -135,7 +136,15 @@ internal fun DebugPresetFilters(
                             }
                         )
                     },
-                    label = { Text(filter) }
+                    label = { Text(filter) },
+                    leadingIcon = { if (filter in filterTexts) {
+                        Icon(
+                            imageVector = Icons.Filled.Done,
+                            contentDescription = "Done icon",
+//                            modifier = Modifier.size(FilterChipDefaults.IconSize)
```
jamesarich: nit - remove comment

app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
```diff
@@ -166,21 +167,21 @@ internal fun DebugItem(
...
-                theme.primary.copy(alpha = 0.1f)
+                colorScheme.primary.copy(alpha = 0.1f)
             } else {
-                theme.surface
+                colorScheme.surface
             }
         ),
         border = if (isSelected) {
-            androidx.compose.foundation.BorderStroke(2.dp, theme.primary)
+            androidx.compose.foundation.BorderStroke(2.dp, colorScheme.primary)
```
jamesarich: nit - this could use an import as well, prefer imports over fully qualified naming.

jamesarich (dismissed):
Appreciating the effort to get things cleaned up, thank you 🫡

DaneEvans:
Definitely a bug in rotation - both search and filters. 
I haven't been through to clean up yet, or add testing. If we're happy with the overall, I'll get to that tonight.

jamesarich:
> Definitely a bug in rotation - both search and filters. I haven't been through to clean up yet, or add testing. If we're happy with the overall, I'll get to that tonight.

sounds like a plan, thank you

jamesarich:
@DaneEvans - if it helps, you can run `detekt` locally via cli, gradle plugin, gradle task, android studio plugin, etc. - you wouldn't have to rely on ci.
https://detekt.dev/docs/gettingstarted/cli

You'll want to point detekt to the configs and baselines here: https://github.com/meshtastic/Meshtastic-Android/tree/main/config/detekt

DaneEvans:
Thanks - I have been (after the first few before I found the CLI command). And only pushing up when I think it'll go all the way. 
I swear this one passed! I think I must have snuck back and made the function name clearer (and apparently used an '_') before I commited. No, still passes locally. Something must not be being picked up in the ruleset. 

I was just frustrated that 60 doesn't pass detekts <60, and there were no sensible ways to split it any further.

jamesarich:
> Thanks - I have been (after the first few before I found the CLI command). And only pushing up when I think it'll go all the way. I swear this one passed! I think I must have snuck back and made the function name clearer (and apparently used an '_') before I commited. No, still passes locally. Something must not be being picked up in the ruleset.
> 
> I was just frustrated that 60 doesn't pass detekts <60, and there were no sensible ways to split it any further.

Don't break yourself - there's a point where abstraction doesn't make sense anymore and we need a `@Suppress`. We're not after perfect, just limiting smells where we can. Again, appreciate your efforts.

app/src/main/java/com/geeksville/mesh/model/DebugViewModel.kt
jamesarich, line 1: Aren't viewmodels magic and wonderful for retaining state? :magic_wand: 

Handling config changes used to be a nightmare!

jamesarich (approved):
:+1:

app/src/main/java/com/geeksville/mesh/model/DebugViewModel.kt
DaneEvans: It still is, it's just now semantic instead of technical

DaneEvans:
Done. I'm calling this one. Once it's in I'll set up an issue for the things I still want to fix on this.

jamesarich:
> Done. I'm calling this one. Once it's in I'll set up an issue for the things I still want to fix on this.

Perfect, I'm _sure_ we'll get feedback as testers get it in their hands as well. Thanks!

jamesarich (approved):

jamesarich:
@DaneEvans - not sure if you're a discord user, but we have a [fairly active server](https://discord.com/invite/ktMAKGBnBs) and android channel. It's a good place to keep up on broader project happenings and get quick feedback. There's a role for `contributor` as well :trophy: , hmu @olm3c if you're interested.

Issue #2076, opened 2025-06-11
[Bug]: Interacting with Hidden Client kinda freezes screen


b8b8:
### What happened?

Long Pressing a node set to client_hidden (black node card) or nodes you dont have node info from then pressing more details locks up the screen on the longfast channel with a blank screen. It can be exited from.

### App Version

2.6.15

### Phone

Pixel 9a

### Device

T1000E

### Firmware

2.6.10

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

jeremiah-k:
Is this still an issue in 2.6.20? I just tried setting a node to client hidden and removed it from my nodedb, but I’m not able to reproduce the issue on the latest build.

I did poke at a fix a few days ago but hadn’t tested it or tried to reproduce the bug until now. If it's still an issue and you’d like to test, this build includes that fix, rebased on top of 2.6.20:
https://github.com/jeremiah-k/Meshtastic-Android/actions/runs/15746446373

b8b8:
This was still an issue for nodes with no nodeinfo in app 2.6.20.

Your recent build seems to work for both client_hidden and a node with no nodeinfo (by opening a blank node details as expected).

Thanks for the changes. I will continue testing over the next few days if needed.

![Image](https://github.com/user-attachments/assets/6ea6d4a5-f58d-4914-bd1c-72761a101871)

jeremiah-k:
Great news I'm glad to hear that it's working. I'll open a PR here for review

PR #2164, opened 2025-06-19
Fix hidden client freeze issue when viewing node details

jeremiah-k:
**Problem:**
When navigating to node details for hidden clients or nodes without complete node info, the screen would freeze with a loading spinner because the node was not found in the nodeDB.

**Solution:**
Added fallback node creation to handle cases where nodes don't exist in the database. When a node is not found, the app now creates a basic Node object with default values based on the node number, preventing the freeze and allowing users to view available information.

**Changes:**
- Added `createFallbackNode()` method to create basic Node objects for missing nodes
- Modified node loading logic in MetricsViewModel to use fallback when node not found
- Added necessary imports for DataPacket and MeshProtos
- Added documentation explaining the purpose of fallback node creation

Closes: #2076 (see issue for test details)


app/src/main/java/com/geeksville/mesh/model/MetricsViewModel.kt
```diff
@@ -212,6 +214,25 @@ class MetricsViewModel @Inject constructor(
...
 
+    /**
+     * Creates a fallback node for hidden clients or nodes not yet in the database.
+     * This prevents the detail screen from freezing when viewing unknown nodes.
+     */
+    private fun createFallbackNode(nodeNum: Int): Node {
+        val userId = DataPacket.nodeNumToDefaultId(nodeNum)
+        val defaultUser = MeshProtos.User.newBuilder()
+            .setId(userId)
+            .setLongName("Meshtastic ${userId.takeLast(n = 4)}")
```
jamesarich: Avoid using hardcoded string literals - please extract to a translatable string resource.

jamesarich (changes requested):
Extract the string for translation

Issue #2165, opened 2025-06-19
[Bug]: Can't exchange position


BlacMac:
### Contact Details

blacmaclabs@gmail.com

### What happened?

A bug happened! Trying to test two nodes, one connected to an iPhone and one connected to Android. I can exchange positions on the iOS app and the android app sees the iOS node. But the iOS node never gets gps from android node. 

I try the exchange position command on the android app and nothing happens. 

Also when looking at the map I need to press the target icon to see my android gps position on map. Am I doing something wrong?

### App Version

2.6.15

### Phone

Z Fold 6

### Device

Heltec V3

### Firmware

2.6.10

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2169, opened 2025-06-19
fix (#2165): position exchange not working

jeremiah-k:
Fix position exchange not working on Android

**Problem:**
Position exchange wasn't working when tapped from the node menu. The Android app wasn't sending any position request packets at all, so nothing happened when users tried to exchange positions.

**Cause:**
The position request method was missing the required payload data. Without a proper payload, no position packets were being transmitted to other devices.

**Fix:**
Modified the position request to include the Android device's current GPS location in the request payload. The app now correctly transmits position packets with `want_response: true` when "Exchange Position" is tapped.

**Changes:**
- Added current device position to position request payload
- Used existing position from device database when available
- Follows same pattern as other request methods in the codebase

**Current Status:**
Position exchange now correctly sends position packets from Android devices. There appears to be a firmware issue where devices don't respond to `want_response: true`, but I'm looking at that separately.

Closes: #2165

### DIFF
--- a/app/src/main/res/values-fi/strings.xml
+++ b/app/src/main/res/values-fi/strings.xml
@@ -138,7 +138,12 @@
     <string name="text_messages">Tekstiviestit</string>
     <string name="channel_invalid">Kanavan URL-osoite on virheellinen, eikä sitä voida käyttää</string>
     <string name="debug_panel">Vianetsintäpaneeli</string>
+    <string name="debug_logs_export">Vie lokitiedot</string>
     <string name="debug_last_messages">500 viimeisintä viestiä</string>
+    <string name="debug_filters">Suodattimet</string>
+    <string name="debug_active_filters">Aktiiviset suodattimet</string>
+    <string name="debug_default_search">Hae lokitiedoista...</string>
+    <string name="debug_clear">Tyhjennä lokit</string>
     <string name="clear">Tyhjennä</string>
     <string name="updating_firmware">Päivitetään laiteohjelmistoa, tämä voi kestää jopa kahdeksan minuuttia…</string>
     <string name="update_successful">Päivitys onnistui</string>


app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -2272,13 +2272,31 @@ class MeshService : Service(), Logging {
...
         override fun requestPosition(destNum: Int, position: Position) = toRemoteExceptions {
-            sendToRadio(newMeshPacketTo(destNum).buildMeshPacket(
-                channel = nodeDBbyNodeNum[destNum]?.channel ?: 0,
-                priority = MeshPacket.Priority.BACKGROUND,
-            ) {
-                portnumValue = Portnums.PortNum.POSITION_APP_VALUE
-                wantResponse = true
-            })
+            if (destNum != myNodeNum) {
+                // Use our current position from nodeDB if the provided position is empty/invalid
```
jamesarich: We'll want to check that the user has explicitly enabled phone location before sending that.  `provideLocation` (backed by preference)  available in the UIState - https://github.com/meshtastic/Meshtastic-Android/blob/main/app/src/main/java/com/geeksville/mesh/model/UIState.kt#L786

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -2273,8 +2273,15 @@ class MeshService : Service(), Logging {
         }
         override fun requestPosition(destNum: Int, position: Position) = toRemoteExceptions {
             if (destNum != myNodeNum) {
-                // Use our current position from nodeDB if the provided position is empty/invalid
-                val currentPosition = if (position.latitude == 0.0 && position.longitude == 0.0) {
+                // Check if user has enabled phone location sharing
+                val provideLocation = sharedPreferences.getBoolean("provide-location-$myNodeNum", false)
+                if (!provideLocation) {
+                    debug("Position request skipped - user has not enabled location sharing")
```
jamesarich: we'll still want to send the last valid position we have for ourselves available in the nodeDb - which may have come from on-node gps or statically set.

jamesarich: we don't want to skip the send altogether, just don't send the phone location if user hasn't explicitly enabled - 

use case:
remotely connected to node over network via tcp, has statically set location

PR #2185, opened 2025-06-20
fix: Drop text buttons in favor of icons

DaneEvans:
Bugfix: 
Text is too long especially when localised. 
Move to icons.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
@@ -42,7 +42,7 @@ import androidx.compose.foundation.text.selection.SelectionContainer
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.BorderStroke
 import androidx.compose.material.icons.Icons
-import androidx.compose.material.icons.outlined.CloudDownload
+import androidx.compose.material.icons.outlined.FileDownload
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
```diff
@@ -661,23 +668,29 @@ fun DebugMenuActions(
...
             scope.launch {
                 exportAllLogs(context, logs)
             }
         },
-        modifier = modifier,
+        modifier = modifier.padding(4.dp)
     ) {
-        Text(text = stringResource(R.string.debug_logs_export))
+        Icon(
+            imageVector = Icons.Outlined.CloudDownload,
```
jamesarich: I'm concerned about the potential re-use of the CloudDownload icon here as it's used elsewhere to indicate connection and message receipt status.
Maybe just `FileDownload` or `Save` ?
https://fonts.google.com/icons?icon.query=export&icon.set=Material+Icons

DaneEvans: I have bigger gripes with it's use elsewhere, but _file download_ or _install desktop_ would fit. 

I guess the same on the ones on the tiles as well

DaneEvans:
done

jamesarich (approved):

Issue #2100, opened 2025-06-13
[Bug]: 2.6.14 - X labels on charts remain fixed when scrolling


DaneEvans:
### Contact Details

GH

### What happened?

When looking at device metrics & signal metrics at all timescales, the end x-axis values remain fixed even when scrolling the graph. 

This is misleading, and they should either move with the plot, or be updated as it is scrolled. 

Suspect this applies to all graphs. 

### App Version

2.6.14

### Phone

Moto 85g - 

### Device

Seed, nrf52,

### Firmware

2.6.4

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2188, opened 2025-06-20
Fix/2100 graph labels

jamesarich:
closes https://github.com/meshtastic/Meshtastic-Android/issues/2100

Adds variable time labels to charts, which move with the data.
Notes:

Someone check environment metrics and power metrics please, I don't have data for them
Is currently not showing changes when you select timescale changes on mine
- I suspect lack of data history more than a bug, but it might be worth waiting until the data builds up

sorry @DaneEvans i was totally distracted and hit automerge on this one, had to revert and reopen


jamesarich:
@DaneEvans i think this isn't quite right currently -  environment and device metrics still have wonky times/don't line up to the available logs

note: i am scrolled all the way to the right on both graphs

image
image

DaneEvans:
Oh, it's very screwy looking at it this morning. I think the time offset is inverted?
leave it with me 

/draft

DaneEvans:
So, fun times. The existing AxisOverlay was wrong. 
Have now recalculated it, and the start and end times, accounted for the variable legends etc. And it seems to be working. 

Again, I'd like someone to have a look at power and Environment, as I don't have that test data. 
And I'll keep an eye on it as we move to longer timescales - I wiped my data testing the debug page yesterday

DaneEvans:
@jamesarich - Failures look unrelated.

jamesarich:
i'll pull it down and give it a test when i have a moment

jamesarich:
@DaneEvans time scales seem fixed but the y-axis (right side) for temp scale is still off on env metrics (maybe not using imperial/metric?).
I don't have a device w/ power metrics to test those.

Screenshot_20250621_201149

DaneEvans:
I'll have a look if I changed the scales accidentally, but sounds like a feature not a bug to me.  At some point America has to move out of the 1800s.

Timescales look to be working well, at least across 24 and 48 hours. Probably worth looking at on a tablet, and I should check landscape too.

DaneEvans:
#2207 suggests this is a wider issue

jamesarich:
> #2207 suggests this is a wider issue

That does seem to be a separate issue - and the issue that this PR is intended to fix is resolved. Let's get this one in, and we can address Freedom Units in another PR.

Let me know if you're good w/ that.

DaneEvans:
@jamesarich 
Yep, that works for me. 
I didn't touch any of the Y scaling, or anything adjacant. 

It also sucks on landscape - but I'm assuming that was already the case there. 

Looks like I need to look into our options to solve that issue across the board, given that it seems to be an issue in a few places.

PR #2199, opened 2025-06-21
fix: unintended line break in node chip

lkosson:
Change #2191 caused node chips with hyphen or space **and** full-width letters (`W`, `M` etc) to be displayed in two lines (see 2nd message in the screenshot). This change reverts `minWidth` from `48dp` back to `72dp` and also adds `maxLines = 1` as a safeguard when someone puts line breaks or other exotic unicode characters in `shortName`.


lkosson:
Also please take a look at last message in that screenshot. `overflow = TextOverflow.Ellipsis` in `MessageItem` clearly isn't working as intended: reaction button is tiny, and reply button is completely gone, while node name takes unreasonably large area. I don't really know how to fix it.

PR #2206, opened 2025-06-21
fix: ignore non-local localStats telemetry

lkosson:
Not sure if it belongs here, but worth a shot.

[Forked firmware](https://github.com/musznik/firmware) popular around here broadcasts telemetry with `hasLocalStats` over the radio for some reason, randomly replacing local stats in notification with the ones received from remote node. It shouldn't do it, but it does, breaking stats for everyone around. Not much can be done about operators running this firmware version, and official firmware never produces such messages, but such packets could be ignored without any harm.


jamesarich (approved):
Seems fine - too bad we have to code around bad behaviour though :disappointed:

Issue #2212, opened 2025-06-22
[Bug]: Custom debug filters not working


lkosson:
### What happened?

Input text field ("Add custom filter") in Debug panel added in #2105 / #2167 does not accept any inputs from keyboard. In fact, there is no UI state associated with the field.

<img src="https://github.com/user-attachments/assets/1acc1e3d-91a9-41c1-8a19-e8b294316b73" width="300" />

### App Version

2.6.22

### Phone

Emulator, Pixel 6a, Sony XZ2C

### Device

(unrelated)

### Firmware

(unrelated)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
Reproducible. 
Interestingly still works in emulator using a real keyboard

DaneEvans:
Fixed

jamesarich:
closed by #2219

PR #2218, opened 2025-06-22
attach state to custom filter so it works with soft keyboards

DaneEvans:
Addresses #2212 

Tested and working on a moto g85 on stock keyboard


app/src/main/proto
jamesarich, line 1: revert proto changes, as they are updated elsewhere

jamesarich (changes requested):

jamesarich:
made changes to address in #2219

Issue #2150, opened 2025-06-18
[Bug]: Message Text Not Clearing After Send (App v2.6.18)


ForeheadFox:
### Contact Details

foreheadfox

### What happened?

In app version 2.6.18, there's a recurring issue where the text input field doesn't clear after sending a message. The message does send successfully, but the typed text remains in the input box.

This happens every time a new message is sent. (The Quick Chat input is not affected)

https://github.com/user-attachments/assets/e6d0369d-a6e8-448d-b4b5-f0f938b45057

I've attached a video showing the behavior.

### App Version

2.6.18

### Phone

Samsung S21

### Device

Heltec mesh node t114

### Firmware

2.6.12

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

Issue #1474, opened 2024-12-15
[Bug]: Message text area is not always cleaned when sending


ServeurpersoCom:
### Contact Details

admin@serveurperso.com

### What happened?

Message text area is not deleted some time on sending (it must be cleaned all time)

This looks exactly like the bug you can get in web development, A timing problem on slower or faster machines if the delete event is done on the button press rather than a callback after submit.

![Screenshot_20241215_101552_Meshtastic](https://github.com/user-attachments/assets/d0a94696-e546-43eb-a3ff-f121f928ce49)

### App Version

2.5.14 last google beta program APK

### Phone

Samsung galaxy s24 ultra

### Device

RAK WisBlock 4631

### Firmware

2.5.15 lastest Stable firmware

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

weirdal3333:
What keyboard are you using? I'm using FUTO and have experienced this bug in other applications as well as Meshtastic!
Not sure if it's an Android feature change, the app itself, or the keyboard!

VigibotDev:
I use "Microsoft SwiftKey AI Keyboard" as my default keyboard on all android device and I have never had this bug on any other application. Android must completely isolate the operation of the application by the API abstraction layer.
Knowing that the send button belongs to the application and that the text entry works well.
I do not use any special function with Meshtastic, just a press on the completely standard touch keys, and possibly the one that offers a choice of completion in case of typos.

VigibotDev:
I just discovered that if the text ends with a space, it is cleaned, if it does not end with a space, it is not cleaned.
The Standard Letters and Word Completion buttons do not change this behavior.

https://github.com/user-attachments/assets/b509857c-25a5-4404-a6ae-9d7c81711243

VigibotDev:
It's more subtle than that, the cursor must also be positioned at the end after the space for the message to be cleaned! It still looks like a cleaning timing problem... for some reason there must be impacts that I can't explain...

VigibotDev:
I confirm solved in 2.5.15 beta update. Thanks for your work

ServeurpersoCom:
I confirm this issue come back in Android v2.6.19 (Google beta program)
- All hardware
- Firmware last stable/beta
- Bluetooth

jamesarich:
i am unable to reproduce in latest builds. may need some extra looking though.

ServeurpersoCom:
[...]

PR #2229, opened 2025-06-22
fix: sticky messages

jeremiah-k:
## Summary of Changes
I've implemented a fix for the sticky message input issue using a LaunchedEffect approach:

- Added state tracking: Created a messageSent boolean flag to track when a message is sent
- Used LaunchedEffect: Added a LaunchedEffect that watches the messageSent flag and clears the text input in the next composition frame
- Updated message sending: Modified the message sending logic to set messageSent = true instead of directly clearing the text

This approach should resolve the race condition between state changes (like replyingTo = null) and text clearing operations by ensuring the text is cleared in a separate composition frame after all state updates are complete.

The fix addresses the core issue where the rememberSaveable state was potentially being restored during recomposition triggered by other state changes, causing the text input to reappear after being cleared.

Tested and working on a Pixel 6 Pro (Android 15)

Related issues: #1474 #2150

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/message/Message.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/message/Message.kt
@@ -53,6 +53,7 @@ import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
 import androidx.compose.material3.TopAppBar
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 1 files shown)


jamesarich:
i have a different solution in the works to refactor/resimplify this text input. this is not needed - thanks!

Issue #2216, opened 2025-06-22
[Bug]: unread message count not updating


lkosson:
### What happened?

When the message list contains more than one page of read messages and new (unread) message is received, opening the conversation for the first time does not reset unread status of the message (even though it scrolls down to it) and does not update unread message count. Opening the conversation once again properly marks message as read.

<img src="https://github.com/user-attachments/assets/ac30e04d-a3e4-4446-94cf-72bda7d29646" width="300" />

### App Version

2.6.11

### Phone

Emulator, Pixel 6a

### Device

(unrelated)

### Firmware

(unrelated)

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

lkosson:
With some recent changes, unread message count stopped updating altogether. Might has something to do with `MessageList.UpdateUnreadCount` using `firstVisibleItemIndex` and first unread message not being at or above first visible message.

<img src="https://github.com/user-attachments/assets/91086fbc-fff6-4a6a-bd56-f92e8720ae1f" width="300" />

brebtatv:
The Android app 2.6.4 problem is back and even worse. There is no way around it.

PR #2235, opened 2025-06-23
fix: mark unread messages as read when displayed (#2216)

lkosson:
Related to #2216 

This change fixes a bug causing messages to stay unread even after being shown in message list. `unreadIndex` got pushed up to `MessageScreen` and now `onUnreadChanged` callback from `MessageList` receives a message index instead of timestamp, while `UpdateUnreadCount` don't observe message collection at all.

Apparently `UpdateUnreadCount` in `MessageList` kept seeing empty `messages` list, which caused `unreadIndex` to stay at `-1`, preventing calling `onUnreadChanged`. For reasons I cannot yet fathom, `messages` one frame up in the stack (i.e. in `MessageScreen`) were properly populated and reflected actual content of selected conversation, but state/flow/that-kotlin-compose-thingy didn't get properly passed to `UpdateUnreadCount` since UI refactor a few days ago. Please review carefully, as I have no clear understanding neither why it does work now nor why it didn't before the change.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/message/Message.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/message/Message.kt
@@ -139,6 +139,7 @@ internal fun MessageScreen(
 
     val quickChat by viewModel.quickChatActions.collectAsStateWithLifecycle()
     val messages by viewModel.getMessagesFrom(contactKey).collectAsStateWithLifecycle(listOf())
+    val unreadIndex by remember { derivedStateOf { messages.indexOfLast { !it.read } } }
 
     val messageInput = rememberTextFieldState(message)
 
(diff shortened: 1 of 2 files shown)


lkosson:
Upon further testing, it didn't fully solve the problem, just reverted to previous state, where sometimes one had to open conversation twice to have messages properly marked as read.

jamesarich:
> Upon further testing, it didn't fully solve the problem, just reverted to previous state, where sometimes one had to open conversation twice to have messages properly marked as read.

this sounds pretty indicative of viemodel scoping wierdness - make sure that the model is being injected properly?

lkosson:
Upon even further testing:
 * I can't replicate the error on emulator anymore
 * On Pixel 6a it mostly works, I can't find a pattern when it doesn't
 * On Xperia XZ2C it works, but upon opening a conversation I have to scroll list a bit up and back down. If I don't, it's 50-50 between working and not working

KD9ERD:
So how do I implement this fix?

jamesarich:
> So how do I implement this fix?

download the latest release where it has been fixed - 2.6.30 is available :)

Issue #2224, opened 2025-06-22
[Bug]: Non-expiring waypoints sent from an iOS device do not show up in Android


nullmonk:
### What happened?

Waypoints created on iOS and set to "never expire" don't appear on Android devices, despite the WAYPOINT_APP packet being received and set to the app properly. Waypoints with an expiration date sent from iOS display correctly.

I can see the WAYPOINT_APP packet come in for both waypoints, but only one renders on the map.

### Potential Resolution
The packet sent from iOS is missing the expiration timestamp field, add the ability for Android to handle this field being optional. Currently, Android sets "no-expire" as "expire in 100 years" so it might be that it does not have a way to handle "no-expiration" and this should be standardized across both apps.

![Image](https://github.com/user-attachments/assets/6d7daeeb-5b39-46b4-a873-c4995208a29d)

### App Version

2.6.15

### Phone

Pixel 7a
Android 16

### Device

Lilygo t-echo

### Firmware

2.6.11

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

prokrypt:
@RCGV1
https://github.com/search?q=repo%3Ameshtastic%2FMeshtastic-Android+expire&type=code

There might be 2 places where this "bug" is occuring...

PR #2237, opened 2025-06-23
Fix waypoints

RCGV1:
The iOS app sends non-expiring waypoints as expire=0; we will now treat those as never expiring and always show them. Additionally, the Date and Time formatted strings sometimes did not display, so we now initialize them instead of using onChange.


RCGV1:
Fixes #2224

Issue #774, opened 2023-11-04
[Feature Request] Node BLE connection priority


jeremiah-k:
It would be great if we could assign node priority values to each of our remembered nodes.

I have 3 main nodes that I connect to throughout the day and I have to manually switch to each one, depending on where I'm at.  For example with this feature we could set #1 to the Home node, so that's always preferred whenever it's in range, #2 to the Vehicle node, and #3 to the backpack node.

So when moving from the house to the vehicle and the house node becomes out of range, it will connect to the vehicle node, when leaving the vehicle it will connect to the backpack node, and so on.

When returning to the vehicle or to the house, it will connect to the higher priority nodes when detected (checking once a minute or however it's configured).

mattdrum:
I second this request. I have three primary nodes that I need to switch between during the day, (Home, vehicle, work). This feature would ensure that I stay connected when moving locations.

mtsdawson:
Same use case, I also have three nodes that would be set to preferred.

Expanding on the use case examples, I would apply this configuration to my wife's vehicle. In an emergency, it's unlikely that she would manually switch between nodes. This auto switching feature would remedy this problem and ensure she stays connected at all times.

PR #2244, opened 2025-06-25
Bluetooth connection automatic change

Pavos21:
Hello 👋

I'm developer from Czechia and in our community was requirement to implement automatic changing of connected Bluetooth nodes.
The request was primarily from multi-node users. 

_You're at home - app is connected to home node. You go for a walk - app loses connection with home node and switches to travel node etc._

**The feature is optional.**

It works as follows:
- State of checkbox is save in ui-prefs SharedPreferences. UIState.kt handles saving to prefs.
- In MeshService.kt function `onConnectionChanged()` calls logic when connection state goes CONNECTED -> DISCONNECTED.  Then is called `manageAutoConnectScanOnDisconnect()` which based on saved prefs decides if run further logic (user checked checkbox) or not.
- Scanning runs every 60 seconds and is for 2 seconds. In those 2 seconds it'll detect all available nodes and based on average signal strength (RSSI) chooses the strongest (this was added because in house can be multiple nodes).

This is my very first commit, so I expect there will be flaws and I would appreciate your constructive criticism and advice.

Happy coding! 🧑‍💻

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -96,7 +96,6 @@ import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.NonCancellable.isActive
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.ensureActive
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -57,6 +57,7 @@ import com.geeksville.mesh.TelemetryProtos.LocalStats
 import com.geeksville.mesh.analytics.DataPair
 import com.geeksville.mesh.android.GeeksvilleApplication
 import com.geeksville.mesh.android.Logging
+import com.geeksville.mesh.android.hasBluetoothPermission
 import com.geeksville.mesh.android.hasLocationPermission
 import com.geeksville.mesh.concurrent.handledLaunch
 import com.geeksville.mesh.config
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -2363,4 +2408,90 @@ class MeshService : Service(), Logging {
...
+            }
+        }
+    }
+
+    fun stopAutoConnectScan() {
+        autoConnectScanJob?.cancel()
+        autoConnectScanJob = null
+    }
+
+    @SuppressLint("MissingPermission")
```
DaneEvans: Should that really be suppressed?

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -2363,4 +2408,90 @@ class MeshService : Service(), Logging {
...
+            } catch (e: Exception) {
+                errormsg("Scan error: ${e.message}")
+            }
+        }
+        return findAndConnectToStrongestDevice(readings)
+    }
+
+    @SuppressLint("MissingPermission")
+    private suspend fun findAndConnectToStrongestDevice(readings: Map<BluetoothDevice, List<Int>>): Boolean {
+        val best = readings
```
DaneEvans: can we go with maxRssi or something rather than best.

DaneEvans (commented):
minor comments, but I want to try this one out, it looks useful

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
Pavos21: Thank you for your review. I've added code which handles Bluetooth permission so no Suppress is needed. (also renamed best to maxRssiDevice - from next comment)

jeremiah-k:
Cool! I requested something like this awhile back ( #774 ) and recently had been thinking about ways it could be implemented.

Pavos21:
Thanks!

And if may I have a question (I've never done anything in such big project so I'm a bit struggling with Git), now after fix commit one check failed. Do i have to repair those things which are written in Detekt log?

DaneEvans:
Yep. You can run it locally with ` ./gradlew detekt `

We can suppress if need be, but most of the ones in here should be fixable

garthvh:
For security reasons this is not really possible the app has recently been updated to fix this cve and much of the data needs to be reset when your device changes https://github.com/meshtastic/Meshtastic-Android/security/advisories/GHSA-h4rg-g6f3-ghh7

PR #2279, opened 2025-06-28
2269 - CI testing not failing correctly

DaneEvans:
- reduce timeouts (still ~3x typical) 
- nest the || true that is there to stop the dependence on crashpad existing, so it doesn't override the result. 

To verify: 
Look at the last 2 workflows on the branch - only change is whether they have an intentional failure or not.


jamesarich (approved):
nice. that did the trick!

Issue #2061, opened 2025-06-09
[Feature Request]: Add public label to any channel using public PSK


willhubs:
### Tell us your idea.

I'd like the app to...
Whenever a channel is displayed, if it's using the public PSK of AQ== there would be a designation of public next to the channel name.

So in the "contacts" tab and when sending messages it would say LongFast(public) instead of just LongFast.

This would help clarify for new users that what they are sending is on a public channel.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
I'll do this one overr the weekend

DaneEvans:
#2284 closes

willhubs:
Thank you, looks good!

PR #2284, opened 2025-06-28
Feat/2061 public ind

DaneEvans:
Add unlock icon where open channels exist. 
Open = "AQ==" || "" 

Locations updated:
- settings > channels
- share channels
- contactList
     - messages within those

#2061 

image

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/model/UIState.kt
+++ b/app/src/main/java/com/geeksville/mesh/model/UIState.kt
@@ -515,7 +515,11 @@ class UIViewModel @Inject constructor(
 
             val shortName = user.shortName
             val longName = if (toBroadcast) {
-                channelSet.getChannel(data.channel)?.name ?: app.getString(R.string.channel_name)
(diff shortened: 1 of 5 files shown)


jamesarich:
I think this is a good first iteration of this - we will want to align with the iOS implementation and make sure we're using similar logic and showing the same-ish states and icons.

note: i'm not sure this image is up to date (my meshtastic mac app is not updating)

image

jamesarich (approved):

jamesarich:
@DaneEvans let me know if you'd like to tackle these states as part of this branch - or we can get this one in and iterate

DaneEvans:
Get it in I think.  

Colours and aligning are easy enough, but I'll need to look at the iOS one to see what they actually look for. 

The states are similar, but may not be exact


In my case isEmpty or matches exactly AQ==

garthvh:
You can do hex of key is <= 4 and get empty keys and all 255 potential known keys

PR #2292, opened 2025-06-28
Refactor NsdManager and improve service display

jamesarich:
This commit refactors the NsdManager to simplify service discovery and enhances the display of discovered network services.

Specifically, the following changes were made:
- Removed the overloaded `serviceList` function in `NsdManager.kt` that accepted a set of service types and a service name.
- Modified the remaining `serviceList` function in `NsdManager.kt` to no longer filter by service name, as this is now handled by specifying a more specific `SERVICE_TYPE`.
- Updated `BTScanModel.kt` to use TXT records ("shortname" and "id") from discovered network services to create more descriptive display names.
- In `NetworkRepository.kt`, simplified service discovery by using a single `SERVICE_TYPE` ("_meshtastic._tcp") instead of a set of service types and a generic service name.

this is work to support https://github.com/meshtastic/firmware/pull/7162/ by @vidplace7


vidplace7:
image

It's beautiful!

vidplace7 (approved):

Issue #2333, opened 2025-07-03
[Feature Request]: Streamline actions


DaneEvans:
### Contact Details

@DaneEvans

### Tell us your idea.

I'm seeing some delays in actions, 
My guess is that multiple workflows are being triggered on MRs. 
Particularly when they are being updated to main, 

I'm going to have a look and see if we can improve it. 
- At the very least detekt can probbaly be skipped on merge queue. 
- Build can maybe be skipped on MRs (Test will cover that) 
- Maybe only a single test on MRs, both on Main / merge queues

Make sure that we don't have exact duplicates as well. 

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
#2322  May have fixed it, but I'll take a look and see if I can spot anything

jamesarich:
yep - i did a bit of tuning today, should be _better_ but i would imagine there's still room to deduplicate some of these checks

DaneEvans:
- Do all the cronjob ones need to be hourly, 2-3 would seem adequate. 
- make sure that they stop when a newer commit happens 
- detekt on prs may be able to be moved to diff only 
    - put tests (and build) behind detekt 
    - detekt on main / merge queue does full thing 
    -

DaneEvans:
#2361

PR #2358, opened 2025-07-06
CI investigation

DaneEvans:
Notes: 
minimum billing time is 1 min, recommend the cron scheduled ones either get made into jobs in a single workflow each hour, or increase the time between them. I don't see a need to do them more than 4 - 6 times a day. 

codeql has been disabled, 

PR workflow will take a little longer, but will fail out earlier on linting. Dropped api 26 test, as 35 should give adequate coverage, and both will be done on merge queue

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -10,14 +10,15 @@
 # supported CodeQL languages.
 #
 name: "CodeQL Advanced"
+# This has been disabled in the GH CLI.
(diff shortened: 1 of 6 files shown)


DaneEvans:
#2333

jamesarich:
codeql i disabled because it was not running properly and I didn't see the value in the noise at the time - let me know if you want me to re-enable for you

PR #2360, opened 2025-07-06
Fix(i18n): zh_TW typo '藍牙'

Young-TW:
origin text: '藍芽' (not correct)


jamesarich:
hi you can make changes to translated strings at https://meshtastic.crowdin.com/

PR #2361, opened 2025-07-06
refactor(ci): Consolidate GitHub Actions workflows

jamesarich:
This commit introduces reusable workflows for Android build/Detekt and instrumented tests, and consolidates scheduled tasks into a single workflow.

- Deleted individual workflows:
    - `crowdin-download.yml`
    - `crowdin-upload-sources.yml`
    - `update-firmware-releases-list.yml`
    - `update-hardware-list.yml`
- Added new reusable workflows:
    - `reusable-android-build.yml` (combines build, lint, Detekt, local tests)
    - `reusable-android-test.yml` (for instrumented tests)
- Added new consolidated workflow:
    - `scheduled-updates.yml` (handles firmware/hardware list updates and Crowdin sync)
- Updated existing workflows (`pull-request.yml`, `merge-queue.yml`) to use the new reusable workflows.

This change aims to reduce redundancy and improve maintainability of the CI/CD pipeline.

takes into account suggested changes from #2358


.github/workflows/scheduled-updates.yml
```diff
@@ -0,0 +1,117 @@
+name: Scheduled Updates (Firmware, Hardware, Translations)
+
+on:
+  schedule:
+    - cron: '0 * * * *' # Run every hour
```
DaneEvans, line 5: I'm still not convinced that every 2-3 hours would even be a noticible delay.

DaneEvans: and does it need manual as well as a rapid cron ?

.github/workflows/scheduled-updates.yml
```diff
@@ -0,0 +1,117 @@
...
+          if ! jq empty "$temp_hardware_file" 2>/dev/null; then
+            echo "::error::Hardware API returned invalid JSON data. Skipping hardware update."
+          else
+            if [ ! -f "$hardware_file_path" ] || ! jq --sort-keys . "$temp_hardware_file" | diff -q - <(jq --sort-keys . "$hardware_file_path"); then
+              echo "Changes detected in hardware list or local file missing. Updating $hardware_file_path."
+              cp "$temp_hardware_file" "$hardware_file_path"
+            else
+              echo "No changes detected in hardware list."
+            fi
+          fi
```
DaneEvans, line 56: Do these actually write / commit back in? 

They could both call a shell file with params, but if we don't see a use for a third (or more) then it's probably not worth the effort / files

.github/workflows/reusable-android-build.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Setup Gradle
+        uses: gradle/actions/setup-gradle@v4
+        with:
+          cache-encryption-key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
+          build-scan-publish: true
+          build-scan-terms-of-use-url: 'https://gradle.com/terms-of-service'
+          build-scan-terms-of-use-agree: 'yes'
+          add-job-summary: always
+      - name: Run Detekt, Build, Lint, and Local Tests
+        run: ./gradlew detekt lintFdroidDebug lintGoogleDebug assembleDebug testFdroidDebug testGoogleDebug --configuration-cache --scan --parallel
```
DaneEvans, line 56: The stock runners have the cores, and these are isolated enough to do this?

.github/workflows/reusable-android-build.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Upload F-Droid debug artifact
+        if: ${{ inputs.upload_artifacts }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: fdroidDebug
+          path: app/build/outputs/apk/fdroid/debug/app-fdroid-debug.apk
```
DaneEvans, line 62: can we do an array of paths here?

DaneEvans (commented):
Looks good. 

a few minor code simplification comments. 
Have you tested that all steps can fail successfully and report on it?

If not, maybe we shift the PR one accross to this while keeping the merge queue on the existing one for a week or so to ensure that it's working, while we keep the known one to check what goes into main

.github/workflows/reusable-android-build.yml
jamesarich: when i was noodling this - i did have them merged at one point into a single upload step, but i decided i wanted to keep them separate artifacts for potential obtanium usage

.github/workflows/reusable-android-build.yml
jamesarich: seems like it so far

.github/workflows/scheduled-updates.yml
jamesarich: the manual run is probably vestigial from testing, but doesn't hurt anything

.github/workflows/scheduled-updates.yml
jamesarich: probably could be deduplicated if we need to add more

jamesarich:
> Have you tested that all steps can fail successfully and report on it?

i just gave it a workout with intentional ui test, unit test, lint, and detekt failures - behaved as expected 😅

DaneEvans (approved):

PR #2375, opened 2025-07-07
skip test and lint on automated branch

DaneEvans:
- [x] the merge request one needs to be changed to allow them in without these checks  

I think that'll do it, it just needs a clean exit code,

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -15,7 +15,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Always pass for scheduled-updates
-        run: echo "Scheduled updates branch: passing by default."
+        run: echo Scheduled updates branch, auto-passing by default.
+
 
   build_and_detekt:
     if: github.repository == 'meshtastic/Meshtastic-Android' && github.head_ref != 'scheduled-updates'
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -2,7 +2,8 @@ name: Android CI (PR)
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


DaneEvans:
Looks like something needs to pass, you can't skip everything. Let me know how it goes

jamesarich:
i discovered the root reason checks weren't being run for the automated branch - https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow

#2377 adds a(n org owned, repo scoped, fine-grained) Personal Access Token to use for the pull request creation step, allowing those triggers to actually trigger. this PR is probably no longer necessary - we'll give it a while to see how it does.

jamesarich:
yep #2377 did the trick - closing this.

DaneEvans:
reopen - this is to skip / shave a pile of time of these runs 
image

They don't even effect code, and they'll be tested in the merge queue anyway, given that we are already hitting rate limits, reducing a few of these will help

jamesarich:
> reopen - this is to skip / shave a pile of time of these runs image
> 
> They don't even effect code, and they'll be tested in the merge queue anyway, given that we are already hitting rate limits, reducing a few of these will help

Since this job is pulling in translation strings - which can at times be broken and impact the build, I'd like this to continue doing checks on creation. 

We're definitely not the biggest consumer of ci resources across the org, and there's a concerted effort to both increase capacity, and reduce consumption by the big offenders (firmware mostly). The streamlining we've done already saves us quite a bit of time and effort.

Let's let the current config cook for a bit and reassess.

jamesarich:
I'm ready to give this a shot - we can always back it out.

PR #2388, opened 2025-07-08
2387: change config pages footer to "clear changes"

DaneEvans:
from cancel, as cancel implies it will leave the page

I think I've checked all the places preference footer is used without an override.


jamesarich (approved):
This is more "clear" now

DaneEvans:
I think get this one in sooner rather than later, so we get the chance to find any wrong instances before the next release

PR #2389, opened 2025-07-08
Closing popup navs back to settings page

DaneEvans:
Closing popups on config pages will nav you back to the main settings page.
- prevents desync of state between device and app 
- hides issues caused by the device rebooting
- matches expected behaviour more clearly

**caution, risky merge** I'm not sure I've checked every place this is being used for collateral damage


DaneEvans:
I think get this one in sooner rather than later, so we get the chance to find any wrong instances before the next release

jamesarich (approved):
This ux is more in line with what i would expect to happen anyways. Let's try it out.

PR #2395, opened 2025-07-08
refactor: Improve node public key handling and security

jamesarich:
This commit introduces several changes to how node public keys are handled, aiming to improve security and data integrity:

- **New `publicKey` field in `NodeEntity`:** A dedicated `publicKey` field is added to the `NodeEntity` to store the public key directly. This facilitates easier querying and management of public keys.
- **Database schema migration:** The database version is incremented to 19, and an auto-migration is added to accommodate the new `publicKey` field.
- **Enhanced `upsert` and `putAll` logic in `NodeInfoDao`:**
    - **Lazy migration:** The `publicKey` field in `NodeEntity` is populated from `node.user.publicKey` during upsert operations for existing nodes.
    - **Impersonation prevention:**
        - When adding a **new node**, the system now checks if its public key is already associated with a different existing node. If a collision is found, the new node is **not** added to the database, and a warning is logged.
        - When **updating an existing node**, if the incoming public key does not match the existing node's public key (and the existing key is not empty), it's treated as a potential attack. In this case, only dynamic data (like `lastHeard`, `snr`, `position`) is updated. The node's identity information (like `user`) is preserved from the trusted existing record, and its `publicKey` is set to an empty `ByteString` to signal a conflict. A warning is logged.
- **Updated `Node` data class:**
    - The `Node` data class now includes a `publicKey` field.
    - The `hasPKC` and `mismatchKey` properties are updated to use the new `publicKey` field if available, otherwise falling back to `user.publicKey`. This ensures consistent key checking logic.
- **New `findNodeByPublicKey` query:** A new DAO method is added to find a node by its public key.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/database/MeshtasticDatabase.kt
+++ b/app/src/main/java/com/geeksville/mesh/database/MeshtasticDatabase.kt
@@ -71,8 +71,9 @@ import com.geeksville.mesh.database.entity.ReactionEntity
         AutoMigration(from = 15, to = 16),
         AutoMigration(from = 16, to = 17),
         AutoMigration(from = 17, to = 18),
+        AutoMigration(from = 18, to = 19),
     ],
-    version = 18,
+    version = 19,
     exportSchema = true,
 )
 @TypeConverters(Converters::class)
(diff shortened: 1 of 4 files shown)


thebentern (approved):

DaneEvans:
will this fix the complaints about 'another node broadcasting your public key' when you change a nodes user settings, and then connect back to it directly (noted via BT). 
If not I'll add an issue for that

jamesarich:
> will this fix the complaints about 'another node broadcasting your public key' when you change a nodes user settings, and then connect back to it directly (noted via BT). If not I'll add an issue for that

have not tested/noticed that

DaneEvans:
Interesting, I'll make sure it's reproducible then

DaneEvans (approved):

PR #2413, opened 2025-07-11
add a skip value, so we can run and skip tests at the same level.

DaneEvans:


DaneEvans:
@jamesarich 
We need to change the required ones back to the lower level.

PR #2421, opened 2025-07-12
Mqttflag

DaneEvans:
## Thank you for sending in a pull request, here's some tips to get started!

(Please delete all these tips and replace with your text)

- Before starting on some new big chunk of code, it it is optional but highly recommended to open an issue first to say "Hey, I think this idea X should be implemented and I'm starting work on it. My general plan is Y, any feedback is appreciated." This will allow other devs to potentially save you time by not accidentally duplicating work etc...
- Mention "#(issue)" in the description, when applicable
- Please do not check in files that don't have real changes
- Please do not reformat lines that you didn't have to change the code on
- If your other co-developers have comments on your PR please tweak as needed
- Do not use any external image service, just paste or drag and drop the image here and it will be uploaded automatically
- Please also enable "Allow edits by maintainers".

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -15,22 +15,27 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Always pass for scheduled-updates
-        run: echo Scheduled updates branch, auto-passing by default.
+        run: | 
+          echo Scheduled updates branch, auto-passing by default.
+          echo We could add an xml linter here.
 
 
   build_and_detekt:
-    if: github.repository == 'meshtastic/Meshtastic-Android' && github.head_ref != 'scheduled-updates'
+    if: github.repository == 'meshtastic/Meshtastic-Android'
     uses: ./.github/workflows/reusable-android-build.yml
+    with:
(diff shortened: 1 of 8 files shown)


jamesarich (approved):

domusonline:
Not sure if this is the correct way to report something related to this feature.... anyway.... it seems to me that the icon display only checks if the field via_mqtt is on the received packet. I think that if the node receives the packet directly from MQTT the flag is not set. The node that get's it from MQTT has to set it before resending.... if this is the destination node (node showing the msg in the application on a channel for example) it will probably not show the little cloud icon, but the message had an MQTT hop.
Can someone check if this make sense? And if it does what would be the correct way to handle it? Open a new issue? Re-open the original? Request an improvement?...
Thanks.

DaneEvans:
I can confirm that it just checks for the flag and uses it.

If you think it should behave differently, then a bug /feature request as
an issue.

As an app, we are limited to what the firmware provides, so it may need a
change at that level.

jamesarich:
> As an app, we are limited to what the firmware provides, so it may need a change at that level.

they're cooking something up over in the firmware related to this

https://github.com/meshtastic/protobufs/pull/748

PR #2451, opened 2025-07-20
refactor(message): refactor message screen, enforce byte limit

### LINKED ISSUE #2449: [Bug]: Character counter when typing
### Contact Details

_No response_

### What happened?

When typing message, there is character counter right below the input field. I believe it should be switched to byte counter, as when the message is in cyrillic, it''s max size is about 116 characters.  If it's longer, it just disappears without a trace and without being sent.
The other option would be to count the symbols proportionally to the amount of memory they take and thus have correct counter.

### App Version

2.6.11

### Phone

Pixel 9

### Device

Heltec V3

### Firmware

2.6.11

### Relevant log output

```shell

```

### Code of Conduct

jamesarich:
Refactors the `MessageScreen` composable to use a `MessageScreenEvent` sealed interface for handling user interactions, improving code organization and testability.

resolves #2449 

Key changes include:
- Introduced `MessageScreenEvent` to manage all user actions like sending messages, reactions, deleting messages, etc.
- Enhanced the message input field (`MessageInput`):
    - Now displays byte count instead of character count (`MESSAGE_CHARACTER_LIMIT_BYTES`).
    - Implements UTF-8 byte limiting for messages appended via Quick Chat actions.
    - Uses new string resources for input labels and placeholders.
- Improved the "Replying to" snippet UI and logic.
- Added a "Scroll to bottom" FAB that appears when new messages are off-screen.
- Updated `ActionModeTopBar` to use `selectedCount` instead of the full list.
- Renamed `TextInput` to `MessageInput` and updated its preview.
- Added new string resources for reply functionality, delete confirmation, and message input.
- QuickChatRow now prepends an "Alert Bell" action.
- Message character limit constant renamed to `MESSAGE_CHARACTER_LIMIT_BYTES`.
- Minor UI tweaks and state management improvements throughout the screen.


jamesarich:
this is branched from/waiting on #2450

DaneEvans:
No issues found, seems to work as it should. 
Scroll to bottom is always present, which is a little irritating, but no major issue

DaneEvans (approved):
looks good. The 'to bottom' button is always present, which is a little annoying, but no bugs found.

Issue #2429, opened 2025-07-14
[Bug]: Environment Metrics wind direction shows 180° wrong way


bergie:
### Contact Details

Lille Ø

### What happened?

In the latest release, wind speed units were fixed (#2374). However, the wind direction indicator is showing 180° wrong way.
Wind direction is where the wind is blowing _from_, not towards.

For example, in this screenshot the wind is coming from roughly 90°, so blowing east-to-west. But the arrow points from west to east.
<img width="305" height="231" alt="Image" src="https://github.com/user-attachments/assets/d3617f17-ecea-457b-b121-4d911f1584ca" />

Some background https://windy.app/blog/what-is-wind-direction.html

In this case the wind reading was provided by a NMEA2000 ultrasonic wind sensor and sent by https://github.com/meri-imperiumi/signalk-meshtastic#readme

### App Version

2.6.24

### Phone

Galaxy S23

### Device

Seeed T1000-e

### Firmware

2.6.11

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

madeofstown:
Quick fix: rotate your body 180° while looking at the arrow. But in all seriousness, I don't see any wind direction metric displayed, I'm only seeing the wind speed. I'm not sure if the arrow icon in the image you posted has anything to do with direction.

bergie:
> Quick fix: rotate your body 180° while looking at the arrow. But in all seriousness, I don't see any wind direction metric displayed, I'm only seeing the wind speed. I'm not sure if the arrow icon in the image you posted has anything to do with direction.

The arrow does move. See this screenshot (sorry, quite subtle difference, here the wind blows roughly from the east 11 months of the year):
![Image](https://github.com/user-attachments/assets/4892b0d0-70fe-41cc-9aec-61585224cbc0)

I think this is where we'd need to apply the rotation:
https://github.com/meshtastic/Meshtastic-Android/blob/d1c5723d9835f61ab5ce834017af91265c60f893/app/src/main/java/com/geeksville/mesh/ui/node/NodeDetail.kt#L837

DaneEvans:
>>  val normalizedBearing = (windDirection % 360 + 360) % 360 

I have questions ... 

I'll probably look into this one tonight

DaneEvans:
@bergie 
Can you get me a screenshot with the values for wind direction - it should be available in the debug panel if it isn't in the logs. 
[...]

PR #2453, opened 2025-07-20
Fix wind direction icon being out by 180 deg.

DaneEvans:
- Fix wind direction arrow. 
- increase preview coverage
- #2429 

- [ ] Waiting on confirmation in issue that data is as expected before merge.


jamesarich:
according to my fiddling with orientation on another project, i think this is correct - let's give it a shot

PR #2459, opened 2025-07-20
chore(spotless): apply code style improvements and formatting

jamesarich:
Sets up [Spotless](https://github.com/diffplug/spotless) with [ktfmt](https://github.com/facebook/ktfmt)

Sets `ratchetFrom` to `origin/head` so that we can improve incrementally.

Adds license header check and `copyright.txt`

Also adds task to create a SpotlessApply precommit hook - which does a autoformatting pass prior to commit

### DIFF
--- a/.github/workflows/reusable-android-build.yml
+++ b/.github/workflows/reusable-android-build.yml
@@ -64,8 +64,8 @@ jobs:
       - name: Expose Version Code as Environment Variable
         run: echo "VERSION_CODE=${{ steps.calculate_version_code.outputs.versionCode }}" >> $GITHUB_ENV
 
-      - name: Run Detekt, Build, Lint, and Local Tests
-        run: ./gradlew :app:detekt :app:lintFdroidDebug :app:lintGoogleDebug :app:assembleDebug :app:testFdroidDebug :app:testGoogleDebug --configuration-cache --scan
+      - name: Run Spotless, Detekt, Build, Lint, and Local Tests
+        run: ./gradlew :app:spotlessCheck :app:detekt :app:lintFdroidDebug :app:lintGoogleDebug :app:assembleDebug :app:testFdroidDebug :app:testGoogleDebug --configuration-cache --scan
         env:
           VERSION_CODE: ${{ env.VERSION_CODE }}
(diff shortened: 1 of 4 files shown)


DaneEvans:
Let me have a look at the options.  

You can often set them to only do changed files, so it's faster on the CI, and the changes are more contained, and in areas you're already changing

jamesarich:
Yeah this is the initial pass, also maybe considering dropping detekt.. Spotless does a pretty good job of autoformatting to rules and keeping things looking nice, we just need to tweak the rules

DaneEvans:
https://github.com/diffplug/spotless/tree/main/plugin-gradle#ratchet looks like what we would want 

Combine with adding it to the IDE, and it should be pretty easy to add it. 
OR, we go ham and just merge this in in a code freeze for a day or so :p

jamesarich:
> https://github.com/diffplug/spotless/tree/main/plugin-gradle#ratchet looks like what we would want
> 
> Combine with adding it to the IDE, and it should be pretty easy to add it. OR, we go ham and just merge this in in a code freeze for a day or so :p

ahh `ratchet` probably is how we want to attack this, i'll check it out - good find.

DaneEvans:
the defaults that it's doing to the files don't look like they'll fight with detekt. 
It depends on the coverage, and runtime of each, but I'd be inclined to at least let them coexist for a while until we see if there are any gaps. 

Between ratchet and SpotlessCheck, it looks like a reasonable CI chain that shouldn't be too arduous

DaneEvans:
Wasn't there a warning somewhere not to use origin/HEAD? 

I believe the correct was to do a 
``````Git fetch origin main, then use main`

jamesarich:
> Wasn't there a warning somewhere not to use origin/HEAD?
> 
> I believe the correct was to do a ``````Git fetch origin main, then use main`

yeah we're doing a `fetch-depth: 0` already  during checkout which satisfies this

DaneEvans:
Ahh, once I scroll down enough I see you did use main in the code.
I should have waited for coffee.

Head runs the risk of someone adding a commit in the CI pipeline and
changing things.

PR #2461, opened 2025-07-20
feat: add rate limit exceeded error message

### LINKED ISSUE #2436: [Bug Report][Feature Request]: messages stopped by the 2s interval limit / cooldown are showing as Waiting Ack, when they were simply dropped
### Contact Details

NomDeTom

### Tell us your idea.

As in the title: messages that are stopped by the 2s interval limit / cooldown timer still show in the message list (informed by cupcake notification(?)). They show with a plain cloud symbol and showing as Waiting Ack, when they were simply dropped and have never been broadcast at all.

I'd like the app to allow either queueing the message to send at the end of the cooldown, or show that the message wasn't sent and allow a quick resend.

### Relevant log output

```shell

```

### Code of Conduct

jamesarich:
This commit introduces a new error message for when the routing rate limit is exceeded.

The following changes were made:
- Added a new string resource for the rate limit exceeded error message.
- Updated the `Message.kt` file to include the new error message in the routing error handling.

this is in response to https://github.com/meshtastic/protobufs/pull/735/ and closes #2436


thebentern (approved):

jamesarich:
fwict @thebentern this should kick the message status to `ERROR` and present that new string (along with the option to resend thanks to @RCGV1 )

you should be able to test w/ the debug apk against a fw (with the new routing error built in) off the action run once it's done

PR #2473, opened 2025-07-21
Fix German translation: from "Weiterleitungen" to "Wiederholungen" as in this context retransmissions equals "Wiederholungen"

salfare:
Fix German translation: "Weiterleitungen" → "Wiederholungen" as in this context retransmissions equals "Wiederholungen"


jamesarich:
Hi - we use Crowdin for crowdsourced translations -  you can submit your translation edits here: https://meshtastic.crowdin.com/

PR #2466, opened 2025-07-20
feat: match fw RegionCode freq precision

jamesarich:
update fequency start, end, and calculations to better match firmware implementation

uses https://github.com/meshtastic/firmware/blob/master/src/mesh/RadioInterface.cpp for reference.

initiated from [discord conversation](https://discord.com/channels/867578229534359593/871539863307055134/1396578183079399475)

PR #2476, opened 2025-07-21
Add Brazil along Australia/New Zealand

barbabarros:
This PR includes Brazil along with Australia/New Zealand in the region selection dropdown, as the BR902 region will be included in the firmware (https://github.com/meshtastic/firmware/issues/3741).

As discussed in the issue, users should be aware that there is another region that supports Brazil, which is why I made this change.
The BR902 region was not included in this PR.

screenshot2
screenshot1


jamesarich:
Hi - some changes are already in flight for these in #2466 that will impact this. Let's revisit when it gets merged.

jamesarich:
@DaneEvans @barbabarros - it looks like an [additional region ](https://github.com/meshtastic/protobufs/pull/737) was added to the firmware to support Brazil, I'm not sure it still makes sense to combine it with ANZ here.

may want @fifieldt or @caveman99 to chime in here

DaneEvans:
I just fixed the merge conflict, I'll defer to them

caveman99:
@jamesarich after BR_902 is merged in the firmware Brazil has 2 possible regions, nut unlike EU_433 and EU_868. The ANZ region is valid in Brazil as well and BR_902 is the new exclusive Brazil Zone.

barbabarros:
@caveman99 @jamesarich 

What do you think of the Android app listing all countries individually (perhaps an autocomplete search box)? This way, the end user wouldn't need to know technical details like, for example, Brazil and Chile use the same frequency plan as Australia. Brazilians and Chileans don't need to know this. After selecting the country, if more than one band is available, the user would select the desired band using a radio button.

I think it would be better, as the resource limitation only applies to the firmware, not the Android app. The firmware would remain lightweight.

jamesarich:
> @caveman99 @jamesarich 
> 
> What do you think of the Android app listing all countries individually (perhaps an autocomplete search box)? This way, the end user wouldn't need to know technical details like, for example, Brazil and Chile use the same frequency plan as Australia. Brazilians and Chileans don't need to know this. After selecting the country, if more than one band is available, the user would select the desired band using a radio button.
> 
> I think it would be better, as the resource limitation only applies to the firmware, not the Android app. The firmware would remain lightweight.

Sounds like there's some potential for user experience improvement, but seems outside of the current issue at hand. Let's get Brazil sorted out first 👍.

jamesarich (approved):

DaneEvans (approved):

PR #2477, opened 2025-07-21
Feat/1919 pax graphs

### LINKED ISSUE #1919: [Feature Request]: PAX counter graphical logging in the app.
### Contact Details

Miaoucat

### Tell us your idea.

 i just updated to the lasted android release (from github) and i am so happy that pax module is now working!! (previously the received info didnt show on the app) there are so many uses for this i am having great plans for the future.

the thing is it would be great to have **data logging possible on other nodes thru the mesh?** because right now there is no notification, nor anything. the **PAX count** just show on the node list and thats it. it would be useful to have logging like we have for signal strengh and such. for example me i would want to know when the area is busy or not. since currently there is no "memory" i cannot see previous readings.  i know that graphical logging is already implemented in the app for battery levels and signal strength. it would be interesting to see it for PaxCounter too.

also on a related note some fringe uses might benefit from having a notification at "X" number PAX count, but i know it would be alot of programing to add it in. logging is the main thing needed right now.

example: 

Image

### Relevant log output

```shell

```

### Code of Conduct

DaneEvans:
depends on #2472 

closes #1919 

Adds PAX graphs. 
image

It may be time for us to hide metrics with no data, rather than grey them out.


jamesarich (approved):
☮️

jamesarich:
> It may be time for us to hide metrics with no data, rather than grey them out.

I'd be down for that.

jamesarich:
this went in before the debug logs 🙃  - gonna temporarily revert (then replace after)

PR #2508, opened 2025-07-22
feat(firmware): add firmware edition handling

jamesarich:
In response to: https://github.com/meshtastic/protobufs/pull/739

This commit introduces the ability to detect the firmware edition of connected devices.

- Adds `firmwareEdition` to `UIViewModel` to expose the detected firmware edition.
- Modifies `MeshLogRepository` to extract `MyNodeInfo` which includes firmware edition information.
- Updates `Main.kt` to observe `firmwareEdition` in `VersionChecks` and log the edition when a device is connected. This lays the groundwork for potentially different UI/behavior based on firmware edition in the future.


jamesarich:
@vidplace7 @jp-bennett @thebentern @garthvh - 
got this ready for when we figure out how/what we're going to do with the fw edition enum

(it's probably very hard to grok as we're doing a bit of thrashing with new formatting tools :sweat_smile:)

PR #2529, opened 2025-07-25
[Feature] Colorize recent nodes

tteskac:
Hi there, this PR adds a new setting in the map view to colorize recently heard nodes, showing Meshtastic green for most recent ones then fading to standard label white for >30min last heard nodes  (off by default).

Reason: If I'm on the move, I wanted to visualize the most recent "heard" nodes, which gives immediate feedback what nodes I'm most likely able to communicate with at the current location.

I know about the "Hide offline nodes" filter which removes >1h old nodes. Useful to some degree, but I prefer color coding.

image

| Before | After |
|------|-----|
| image | image |


RCGV1:
You should make it within 2 hours, as this is what is used in iOS

tteskac:
> You should make it within 2 hours, as this is what is used in iOS

Hi @RCGV1, can you point me to some screenshots or other docs to see how it works in iOS version? Honestly 2 hours seem too much for the usecase I explained and using it together with "Hide offline nodes" (which removes nodes after 1 hour) renders this feature completely useless in the current state. So I would like to understand how it works on iOS to improve this PR.

Without knowing iOS implementation, I see a couple of solutions:
- could be configurable by the user, e.g. when this is enabled then another button is visible to allow setting the max period.
- could be conditioined by the other setting, e.g. 2 hours when "Hide offline nodes" is disabled, and 30min when "Hide offline nodes" is enabled.
- scale with multiple colors. But honestly I don't see the value in this one.

What are your thoughs on that?

jamesarich:
I'm honestly struggling to see the overall use case here, there's already options to filter by online nodes and favorited nodes (which you would have had previous direct communications with).

tteskac:
@jamesarich Totally understand. I thought this might help with mentioned use case, but if it doesn’t add enough value beyond the existing filters in your opinion, I’m okay with closing it. I'll keep maintaining it in my fork tho in case someone else still finds it useful. Thanks for the feedback!

jamesarich:
> @jamesarich Totally understand. I thought this might help with mentioned use case, but if it doesn’t add enough value beyond the existing filters in your opinion, I’m okay with closing it. I'll keep maintaining it in my fork tho in case someone else still finds it useful. Thanks for the feedback!

Appreciate the understanding. There are some changes on the horizon w/ maps (check the draft PR). We may have more opportunity for map user experience improvement once that is able to be merged.

One of the current hesitancies to mess with the current maps implementation is mainly due to the OSMDROID maps sdk that's currently being leveraged is a dead, archived project.

PR #2544, opened 2025-07-28
Replaced a few hardcoded string values with string resources

twaddington:
Fixed several issues mentioned in https://github.com/orgs/meshtastic/discussions/314.

- Replaced hardcoded "Precise location" string resource used in the channel dialog.
- Replaced hardcoded "Speed" string resource used in the node position log.
- Replaced hardcoded "New Node Seen" string resource used in service notification.

I left a few of the other reported issues alone for now since they're more dynamic. However, this should be an improvement.


livingdeath72:
Thank you very much! I appreciate your work in developing the open source application.

jamesarich (approved):

jamesarich:
should be able to clean this up with `spotlessApply`

DaneEvans:
needs spotless.

DaneEvans (approved):
changes look good.

PR #2563, opened 2025-07-30
repo_ Add more categories to the rchangelog for releases

DaneEvans:
Completely untested.


DaneEvans:
@jamesarich 
I still need to automate the labels, or we could move to titles, 
but it'll make it more readable, and we can see whether we are chasing bugs or features in each release

PR #2569, opened 2025-07-30
Fix/2558 rangetest

### LINKED ISSUE #2558: [Feature Request]: Export only logs containing "seq" as payload when exporting rangetest.csv
### Contact Details

_No response_

### Tell us your idea.

Currently, when exporting rangetest.csv in the Android app, it exports all logs. When you plot these in Google Earth, you see all logs, not just the ones for the rangetest. 

It would be better if exporting rangetest.csv filtered for payloads containing "seq". 

Example of an exported csv (lat and long removed):

Image

### Relevant log output

```shell

```

### Code of Conduct

DaneEvans:
closes #2558 

Add filter

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/MainActivity.kt
+++ b/app/src/main/java/com/geeksville/mesh/MainActivity.kt
@@ -28,6 +28,7 @@ import android.os.Bundle
 import android.widget.Toast
 import androidx.activity.compose.setContent
 import androidx.activity.enableEdgeToEdge
+import androidx.activity.result.contract.ActivityResultContracts
 import androidx.activity.viewModels
 import androidx.appcompat.app.AppCompatActivity
 import androidx.appcompat.app.AppCompatDelegate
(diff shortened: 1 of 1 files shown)


prokrypt:
Does this mean a feature was removed? 👀

DaneEvans:
@prokrypt 
Well, it no longer exports everything, (or at least everything with (  (Admin || Range test ) && SNR ) ) 
 but it does export range tests like it says on the label. 

You can already export everything through the debug panel, without the arbitary limitations. 
We can talk about output formats and making the export match the filters over there if we like.

pedjas:
This is huge step backwards. I would like it more to rename option to call it Export log to csv and export everything.

CSV is easy to use, and if there is no option for user to set what to export it is better to export everything as it can be postprocessed to filter out not wanted log items.

jamesarich:
> This is huge step backwards. I would like it more to rename option to call it Export log to csv and export everything.
> 
> CSV is easy to use, and if there is no option for user to set what to export it is better to export everything as it can be postprocessed to filter out not wanted log items.

This is exactly what @DaneEvans just said - you can utilize the debug log export to get this data. It does not _currently_ export specifically in csv, but that can be proposed/discussed.

pedjas:
_Currently_, functionality is reduced.

jamesarich:
> _Currently_, functionality is reduced.

Feel free to open a pullrequest to modify the _current_ behavior.
Your tone isn't appreciated, this is your warning. 
We're all volunteers - donating our time to work on what we choose. 
If no current dev wants to take this on, either accept it or pick up a shovel and do it yourself.

PR #2578, opened 2025-07-30
feat(analytics): Integrate Datadog for RUM and Logging

jamesarich:
This commit introduces Datadog for Real User Monitoring (RUM) and logging capabilities.

Key changes include:
- Added Datadog SDK dependencies (compose, logs, rum) and the Datadog Gradle plugin.
- Added Secrets Gradle Plugin for managing API keys.
- Initialized Datadog in `GeeksvilleApplication.kt` for Google Play builds.
- Datadog will be enabled if analytics are allowed by the user or if it's a debug build.
- Configuration for Datadog includes client token, environment, variant, and site.
- RUM configuration is set up to track user interactions and long tasks.
- Added `secrets.defaults.properties` with placeholder Datadog keys.
- Added `secrets.properties` to `.gitignore`.

### To do
- [x]  Add `secrets.properties` decoding from env vars to ci pipeline (maybe working)
- [x]  Better fdroid segregation
- [x]  Finish integration - crash reports, logging, debug symbol upload
- [ ]  Investigate workaround for sdk compose instrumentation kotlin incompatibility
- [x]  Match iOS custom tracking metrics


secrets.defaults.properties
DaneEvans, line 1: is this needed at all?

jamesarich: > is this needed at all?

yep it provides the fallback when `secrets.properties` is not available (or needed) see: https://github.com/google/secrets-gradle-plugin#cicd-systems

jamesarich: we'll also need it in place for googlemaps 😅

PR #2591, opened 2025-07-31
refactor(bluetooth): exponential backoff for BLE reconnects

jamesarich:
An attempt at making the bluetooth connection more reliable - this allows more time for recovery when buggy or intermittent.

**NEEDS TESTING**

This change introduces an exponential backoff mechanism for Bluetooth Low Energy (BLE) reconnection attempts. When a BLE connection drops or fails to establish, the system will now wait for an increasing amount of time before retrying. This approach aims to improve connection stability and reduce unnecessary strain on the BLE stack, especially in environments with intermittent connectivity.

The initial backoff delay is 1 second, and it doubles with each subsequent failed attempt, up to a maximum of 64 seconds (after 6 attempts). The reconnect attempt counter is reset upon a successful connection.

Additionally, this commit includes a refactoring of the `SafeBluetooth.closeConnection()` method. It now uses `runBlocking` with `withTimeoutOrNull` to wait for the `onConnectionStateChange` callback after calling `disconnect()`. This provides a more robust way to ensure the GATT object is properly handled during disconnection, preventing potential "DeadObjectException" errors.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
+++ b/app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
@@ -107,7 +107,7 @@ constructor(
     Logging {
 
     companion object {
-        // / this service UUID is publicly visible for scanning
+        // this service UUID is publicly visible for scanning
         val BTM_SERVICE_UUID: UUID = UUID.fromString("6ba1b218-15a8-461f-9fa8-5dcae273eafd")
 
         val BTM_FROMRADIO_CHARACTER: UUID = UUID.fromString("2c55e69e-4993-11ed-b878-0242ac120002")
--- a/app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
+++ b/app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
@@ -121,7 +121,7 @@ constructor(
         @Volatile var safe: SafeBluetooth? = null
     }
 
-    // / Our BLE device
+    // Our BLE device
     val device
         get() =
             (safe ?: throw RadioNotConnectedException("No SafeBluetooth")).gatt
--- a/app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
+++ b/app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
@@ -17,13 +17,18 @@
 
 package com.geeksville.mesh.repository.radio
 
+import android.annotation.SuppressLint
 import android.app.Application
 import android.bluetooth.BluetoothGattCharacteristic
 import android.bluetooth.BluetoothGattService
 import com.geeksville.mesh.android.Logging
 import com.geeksville.mesh.concurrent.handledLaunch
 import com.geeksville.mesh.repository.bluetooth.BluetoothRepository
-import com.geeksville.mesh.service.*
+import com.geeksville.mesh.service.BLECharacteristicNotFoundException
+import com.geeksville.mesh.service.BLEConnectionClosing
+import com.geeksville.mesh.service.BLEException
+import com.geeksville.mesh.service.RadioNotConnectedException
+import com.geeksville.mesh.service.SafeBluetooth
 import com.geeksville.mesh.util.anonymize
 import com.geeksville.mesh.util.exceptionReporter
 import com.geeksville.mesh.util.ignoreException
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
```diff
@@ -82,65 +86,66 @@ A variable keepAllPackets, if set to true will suppress this behavior and instea
...
  *
  * This service is not exposed outside of this process.
  *
- * Note - this class intentionally dumb.  It doesn't understand protobuf framing etc...
- * It is designed to be simple so it can be stubbed out with a simulated version as needed.
+ * Note - this class intentionally dumb. It doesn't understand protobuf framing etc... It is designed to be simple so it
+ * can be stubbed out with a simulated version as needed.
  */
-class BluetoothInterface @AssistedInject constructor(
+@SuppressLint("MissingPermission")
```
DaneEvans, line 98: should we add the permission?

app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
```diff
@@ -82,65 +86,66 @@ A variable keepAllPackets, if set to true will suppress this behavior and instea
...
     bluetoothRepository: BluetoothRepository,
     private val service: RadioInterfaceService,
     @Assisted val address: String,
-) : IRadioInterface, Logging {
+) : IRadioInterface,
+    Logging {
 
     companion object {
-        /// this service UUID is publicly visible for scanning
+        // / this service UUID is publicly visible for scanning
```
DaneEvans:
```suggestion
        // this service UUID is publicly visible for scanning
```

app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
```diff
@@ -82,65 +86,66 @@ A variable keepAllPackets, if set to true will suppress this behavior and instea
...
+         * rare software update case
          */
-        @Volatile
-        var safe: SafeBluetooth? = null
+        @Volatile var safe: SafeBluetooth? = null
     }
 
-
-    /// Our BLE device
+    // / Our BLE device
```
DaneEvans:
```suggestion
    // Our BLE device
```

app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
```diff
@@ -82,65 +86,66 @@ A variable keepAllPackets, if set to true will suppress this behavior and instea
...
     val device
-        get() = (safe ?: throw RadioNotConnectedException("No SafeBluetooth")).gatt
-            ?: throw RadioNotConnectedException("No GATT")
+        get() =
+            (safe ?: throw RadioNotConnectedException("No SafeBluetooth")).gatt
+                ?: throw RadioNotConnectedException("No GATT")
 
-    /// Our service - note - it is possible to get back a null response for getService if the device services haven't yet been found
+    // / Our service - note - it is possible to get back a null response for getService if the device services haven't
```
DaneEvans:
```suggestion
    // Our service - note - it is possible to get back a null response for getService if the device services haven't
```

app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
```diff
@@ -310,27 +327,30 @@ class BluetoothInterface @AssistedInject constructor(
...
             s.asyncDiscoverServices { discRes ->
                 try {
                     discRes.getOrThrow()
 
                     service.serviceScope.handledLaunch {
                         try {
                             debug("Discovered services!")
-                            delay(1000) // android BLE is buggy and needs a 500ms sleep before calling getChracteristic, or you might get back null
+                            delay(
+                                1000,
```
DaneEvans, line 351: Can we reuse the backoffMS var here?  and align comment to code

DaneEvans:
I run BLE on one of my nodes, and walk away from it pretty often.
I'm happy to put this in main and see what happens

app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
jamesarich: it's being handled elsewhere. i could try and pipe the annotation up through the stack to see if it registers, but ime it's not super worth it (or reliable)

I'm willing to accept the suppress here

app/src/main/java/com/geeksville/mesh/repository/radio/BluetoothInterface.kt
jamesarich: align comment 👍  - backoff happens on the reconnect flow.. which this should fail over to.

PR #2614, opened 2025-08-02
Feat/2594 contact shortname on click

DaneEvans:
Clicking a node's contact now sends you to the node details. 

Long click not implemented, as we'd need to decide what options should be available.


DaneEvans:
In hindsight I should have tried to fit it in, but I decided to go with what was already there.
It likely wouldn't have been any harder once I finally got from contact to node

getnode takes the hexstring, with the preceeding !
Which I really should have guessed.

PR #2620, opened 2025-08-02
fixes #2609: Use a Bundle-persistable type for remembering replyTo message

### LINKED ISSUE #2609: [Bug]: app crash when interacting with node while chat reply is active
### Contact Details

_No response_

### What happened?

Open chat window, click the reply arrow from any of the messages, then click the short code button for any node in the chat.  App crashes.  

### App Version

2.6.32

### Phone

S22, Android 14 (One UI 6.1)

### Device

Tdeck

### Firmware

2.7.3

### Relevant log output

```shell
App crashed, I don't think there is a log anymore. :)
```

### Code of Conduct

jsoberg:
## Background
Fixes the crash noted in #2609, caused by using `rememberSaveable` on a type that can't be persisted in a Bundle. Fix is tracking the replyTo packet ID (Int) instead.

**This same problem will likely happen for `sharedContact`, which is persisting `Node`.** We can address this one separately.

## Testing

| Before | After |
|------|-----|
| <video src="https://github.com/user-attachments/assets/dda88150-429e-4d4c-bcbe-3e21453b7a8e" width="300"></video> | <video src="https://github.com/user-attachments/assets/30a0e9da-f5e7-4285-9784-31db1ad8915e" width="300"></video>|

Verification that replies still send as expected:
<video src="https://github.com/user-attachments/assets/9e790aab-a9dd-4857-9037-2b097adbb6db" width="300"></video


jamesarich (approved):
yep, makes sense

PR #2644, opened 2025-08-06
feat(radioconfig): use current location for position config

jamesarich:
Adds a button to the Position Configuration screen to allow the user to populate the latitude, longitude, and altitude fields using the phone's current location.

This requires location permissions. If permissions are not granted, pressing the button will request them.

https://github.com/user-attachments/assets/f5214dc0-4145-4f93-baaa-f151e023e9c8

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/PositionConfigItemList.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/PositionConfigItemList.kt
@@ -17,10 +17,13 @@
 
 package com.geeksville.mesh.ui.radioconfig.components
 
+import android.location.Location
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.text.KeyboardActions
 import androidx.compose.material3.HorizontalDivider
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/PositionConfigItemList.kt
```diff
@@ -84,25 +96,57 @@ fun PositionConfigScreen(
...
     onSaveClicked: (position: Position, config: PositionConfig) -> Unit,
+    onUseCurrentLocation: suspend () -> Unit,
 ) {
     val focusManager = LocalFocusManager.current
+    val coroutineScope = rememberCoroutineScope()
+    val locationPermissionState =
+        rememberPermissionState(Manifest.permission.ACCESS_FINE_LOCATION) { granted ->
+            if (granted) {
+                coroutineScope.launch {
+                    // If permission is granted, we can use the current location
```
DaneEvans: comment isn't needed

app/src/main/java/com/geeksville/mesh/ui/radioconfig/components/PositionConfigItemList.kt
```diff
@@ -84,25 +96,57 @@ fun PositionConfigScreen(
...
+                    latitude = phoneLocation.latitude,
+                    longitude = phoneLocation.longitude,
+                    altitude =
+                    LocationCompat.hasMslAltitude(phoneLocation).let {
+                        if (it && Build.VERSION.SDK_INT >= Build.VERSION_CODES.UPSIDE_DOWN_CAKE) {
+                            phoneLocation.mslAltitudeMeters.toInt()
+                        } else {
+                            phoneLocation.altitude.toInt()
+                        }
+                    },
```
DaneEvans, line 142: Should this be elif (it) ? 
What happens if you call it on it=nil?

jamesarich: phoneLocation shouldn't ever be null by this point, and `LocationCompat.hasMsl` returns a nonnull boolean consumed in the let (`it`), and both mslAltitudeMeters and altitude will be returned as nonnull Ints (toInt() handles Double NaN).

Should be ok i think :thumbsup:

Issue #2077, opened 2025-06-11
[Bug]: blank secondary channel names are incorrect when editing channels


battlehax:
### What happened?

When going to Radio Configuration, then Channels, my channel name for LongFast, shows the name of my primary channel instead. When going to channels on the bottom bar, it shows correctly, but once I click edit, it's wrong again.

After some discussion, we determined that this doesn't happen when the channel is explicitly named "LongFast", only when the channel's name is blank. The proper way to set that up and name per platform is explained here <https://meshtastic.org/docs/configuration/tips/#creating-a-private-primary-with-default-secondary>

Further testing seems to reveal that sharing channel urls between devices and using the cli to backup/restore settings can blank the field.

![Image](https://github.com/user-attachments/assets/99e34d5d-b339-4857-bb67-f8deca8d9463)

![Image](https://github.com/user-attachments/assets/9af4d30c-be38-4b68-9979-6c0c879fa567)

### App Version

2.6.15

### Phone

Pixel 6a, Android 16

### Device

Heltec v3

### Firmware

2.6.5

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

b8b8:
Is it an issue when restoring channels with the QR code?

Restoring channels with the config backup (in android) hasn't worked for a while, but restoring the longfast secondary channel has always worked (for me) using the QR.

battlehax:
> Is it an issue when restoring channels with the QR code?
> 
> Restoring channels with the config backup (in android) hasn't worked for a while, but restoring the longfast secondary channel has always worked (for me) using the QR.

It should be okay when generating the QR code on Android. I will try to investigate if sharing a QR from an iOS device with a blank channel name to an Android device could cause this.

garthvh:
Every newly flashed device has a channel with a blank channel name, LongFast is not the actual channel name, the hash for a channel falls back to the Modem Preset if a name is not set.

Issue #2515, opened 2025-07-23
[Bug]: Wrong channel name appears in radio configuration


unhold:
### What happened?

I have the channel configuration
0: GilgaMESH
1: LongFast

In the Channels tab (and in Chrome) the channels appear correctly:
<img width="1080" height="958" alt="Image" src="https://github.com/user-attachments/assets/74c947bb-ddd2-4377-a542-01f0e1292440" />

However, in Radio configuration --> Channels section, "GilgaMESH" shows up twice:
<img width="487" height="1083" alt="Image" src="https://github.com/user-attachments/assets/bb7e9ee0-d4ad-4110-b510-07a47165d0a8" />

### App Version

2.6.30

### Phone

Pixel 9

### Device

Heltec Mesh Node T114 V2
Seeed Card Tracker T1000-E

### Firmware

2.6.11.60ec05e

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2668, opened 2025-08-09
Fix: Display the modem preset name when no channel name is present.

Robert-0410:
### Fix
-  #2515

It now displays the modem's preset name like in the screen with the QR. (Channel.kt)


DaneEvans:
closes #2077

PR #2692, opened 2025-08-12
revert(service): reverts a bunch of changes to MeshService.kt

jamesarich:
Connection logic refactoring was not going so well, lots of follow-on side effect issues - this reverts most `MeshService` connection and state logic back to 2.6.30.


jamesarich:
I'd like to try to go after this refactor again, but maybe after getting some tests in place 😓 .

Issue #2723, opened 2025-08-14
[Bug]: App does not apply window insets


poliver:
### What happened?

I suspect this has been overlooked since it's not very visible on newer versions of Android with gesture nav enabled. However it is prominent on API 26 or any device with 3 button nav enabled.

I'm applying a fix now. Just creating an issue to link back to.

### App Version

2.6.34

### Phone

Pixel XL emulator, API 26

### Device

N/A

### Firmware

N/A

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

b8b8:
<img width="1080" height="2424" alt="Image" src="https://github.com/user-attachments/assets/8409c5a5-39d2-46e7-87bb-896cd9cb9275" />

poliver:
Thanks, yea I should have specified that I observed it on the welcome screen.

b8b8:
As a fellow 3 button user and hater of gestures, I thank you.

PR #2734, opened 2025-08-15
fix(UI): Apply window insets on welcome screens

poliver:
Fixes https://github.com/meshtastic/Meshtastic-Android/issues/2723

I leveraged Material 3's `BottomAppBar`  (no custom styling) which provides window inset handling automatically. This does change the appearance slightly, but comments in the affected code gave me the impression that we're adhering to Material. Please let me know if I should revert the appearance changes, and I can update.

|Before (API 36, 3-button nav)|After (API 36, 3-button nav) |After (API 36, gesture nav)|
|-|-|-|
|||image|

| Before (API 26) | After (API 26) |
|------|-----|
|  |  |

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/MainActivity.kt
+++ b/app/src/main/java/com/geeksville/mesh/MainActivity.kt
@@ -24,6 +24,7 @@ import android.content.pm.PackageInfo
 import android.content.pm.PackageManager
 import android.hardware.usb.UsbManager
 import android.net.Uri
+import android.os.Build
 import android.os.Bundle
 import android.widget.Toast
 import androidx.activity.compose.setContent
@@ -41,7 +42,6 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.platform.LocalView
 import androidx.core.content.edit
 import androidx.core.net.toUri
-import androidx.core.view.WindowCompat
 import androidx.lifecycle.lifecycleScope
 import com.geeksville.mesh.android.BindFailedException
 import com.geeksville.mesh.android.GeeksvilleApplication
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/geeksville/mesh/MainActivity.kt
```diff
@@ -93,7 +103,6 @@ class MainActivity :
             }
         }
 
-        WindowCompat.setDecorFitsSystemWindows(window, false)
```
poliver, line 96: The above call to `enableEdgeToEdge()` already does this.

app/src/main/java/com/geeksville/mesh/MainActivity.kt
```diff
@@ -78,7 +80,15 @@ class MainActivity :
...
     override fun onCreate(savedInstanceState: Bundle?) {
-        enableEdgeToEdge()
+        enableEdgeToEdge(
+            // Disable three-button navbar scrim on pre-Q devices
+            navigationBarStyle = SystemBarStyle.auto(Color.TRANSPARENT, Color.TRANSPARENT),
+        )
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+            // Disable three-button navbar scrim
+            window.setNavigationBarContrastEnforced(false)
+        }
```
poliver, line 90: The nav bar scrim generally looks bad and is only really suggested when above scrolling content. Disabling the scrim is [recommended ](https://developer.android.com/design/ui/mobile/guides/foundations/system-bars#button_modes)when the app has persistent bottom/nav bars.

DaneEvans:
Are you commenting your own PR? 

Looks good, both on the welome screen, and no issues in the rest of the app. 
Clean up the PR comments, and code comments and it's good to go.

poliver:
They are context comments for the reviewer. Not sure why is shows as me reviewing.

poliver:
Are you wanting me to remove the comments I added in `MainActivity`? That code's function is not immediately apparent without them.

poliver:
Or rather, it's not obvious that `enableEdgeToEdge(navigationBarStyle...` is a no-op on 29+, but I didn't specify that 😅

PR #2740, opened 2025-08-15
Feat/#2696 response to request metadata

DaneEvans:
Add response for metadata request. 

Working, The only question is if it's worth this amount of overhead. 
Will run spotless and detekt on it after initial review.


jamesarich (approved):
I like that this opens the door for other potential calls w/ callbacks that we can surface. I do wonder if the reaction/dialog should happen at a higher level, there's potential for a metadata request from the node list longpress menu too

jamesarich:
playing w/ this just gave me flashbacks from callback hell - i'm wondering if there's a way to tackle this in a more reactive/unidirectionally flowing way

DaneEvans:
I was wondering if we just set a flag - expecting a callback from (`traceroute`, `nodenum`) 
And if we get a matching packet within 60 seconds, we do the popup then. 

It wouldn't block UI, it'll reset fine if it doesn't get a response, it's more generalised etc

PR #2743, opened 2025-08-15
Decouple mesh service bind from `MainActivity`

poliver:
I took a stab at decoupling the `MeshService` bind logic from `MainActivity`. This is a refactor, and no behavior change is expected. `MainActivity` is now injected with a `MeshServiceClient` that observes the Activity lifecycle and handles binding `MeshService`.

There is one minor issue that I called out in the comments. `MeshServiceClient` cannot be injected with the default `LifecycleOwner` binding for `@ActivityScoped` due to a global `LifecycleOwner` provided in `ApplicationModule` taking precedence. The current workaround is to inject `Activity` (not ideal) and cast to `LifecycleOwner`. Qualifying the provider in `ApplicationModule` solves this issue, but would add a ton of scope to this PR.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/MeshServiceClient.kt
+++ b/app/src/main/java/com/geeksville/mesh/MeshServiceClient.kt
@@ -59,6 +59,8 @@ constructor(
         lifecycleOwner.lifecycle.addObserver(this)
     }
 
+    //region ServiceClient overrides
+
     override fun onConnected(service: IMeshService) {
         serviceSetupJob?.cancel()
         serviceSetupJob =
@@ -72,6 +74,10 @@ constructor(
         serviceSetupJob?.cancel()
         serviceRepository.setMeshService(null)
     }
+
+    //endregion
+
+    //region DefaultLifecycleObserver overrides
 
     override fun onStart(owner: LifecycleOwner) {
         super.onStart(owner)
@@ -91,6 +97,8 @@ constructor(
         owner.lifecycle.removeObserver(this)
         debug("Removed self as LifecycleObserver to $lifecycleOwner")
     }
+
+    //endregion
 
     private fun bindMeshService() {
         debug("Binding to mesh service!")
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/geeksville/mesh/MeshServiceClient.kt
```diff
@@ -0,0 +1,113 @@
...
+constructor(
+    /**
+     * Ideally, this would be broken up into Context and LifecycleOwner. However, ApplicationModule defines its own
+     * LifecycleOwner which overrides the default binding for @ActivityScoped. The solution to this is to add a
+     * qualifier to the LifecycleOwner provider in ApplicationModule.
+     */
+    private val activity: Activity,
+    private val serviceRepository: ServiceRepository,
+) : ServiceClient<IMeshService>(IMeshService.Stub::asInterface),
+    DefaultLifecycleObserver {
```
poliver, line 49: Could potentially break this out into a separate observer class

app/src/main/java/com/geeksville/mesh/MeshServiceClient.kt
```diff
@@ -0,0 +1,114 @@
...
+        super.onDestroy(owner)
+        debug("Lifecycle: ON_DESTROY")
+
+        owner.lifecycle.removeObserver(this)
+        debug("Removed self as LifecycleObserver to $lifecycleOwner")
+    }
+
+    // endregion
+
+    @Suppress("TooGenericExceptionCaught")
```
poliver, line 103: This is how the exception was caught before, so suppressing to pass detekt

jamesarich: is there a more specific exception type to handle here?

poliver: I'm not sure. `.startService()` appears to branch on API level only to call the same method in both cases. It swallows an exception for only one of them 🤷

poliver: My intention was to leave the underlying logic alone., and just refactor it.

jamesarich (approved):

poliver:
Updated with `main`. Should be good for another go at the queue.

PR #2748, opened 2025-08-16
MVVM-ify logic to display app intro

poliver:
User action to show the app intro now propagates through `UIViewModel` instead of hoisting to the Activity. I also cleaned up some of the logic around accessing `ui_prefs`, which are already provided by `ApplicationModule` and injected into `UIViewModel`.


poliver:
@jamesarich I just noticed that both `"ui_prefs` and `"ui-prefs` are being used to key pref stores. This could be causing issues when used inconsistently. I can consolidate them, but these prefs will break for users on the next install until they are set again.

Fortunately, the other pref stores are keyed consistently.

I'm currently working on a type-safe pref repo structure (separate branch) to prevent these types of issues with prefs in the future. Should also ease any potential future transition to [Datastore](https://developer.android.com/topic/libraries/architecture/datastore) or KMP.

jamesarich (approved):

PR #2760, opened 2025-08-17
Add prefs repos and DI providers

poliver:
I've audited everywhere`SharedPreferences` are used and have compiled a list of repo interfaces to wrap the underlying access to prefs. Doing this decouples app code from the Android `SharedPreference` APIs and, combined with the new `PrefDelegate`, provides type safe access to prefs without needing to specify string keys, which can be error-prone and lead to inconsistencies in how prefs are used. It also allows us to easily change the underlying prefs implementation (e.g. migrating to [DataStore](https://developer.android.com/topic/libraries/architecture/datastore)) without affecting the consuming code.

Actual application of these new repos is in a [separate branch](https://github.com/meshtastic/Meshtastic-Android/compare/main...poliver:Meshtastic-Android:enhancement_prefs) and changes are widespread, so I wanted to first merge just the repos along with their Dagger configs to keep PRs concise and easy to review. Here is a brief summary of what these changes allow us to do.

Instead of requesting a `SharedPreference` store by key every time:
`val uiSharedPrefs = context.getSharedPreferences("ui-prefs", Context.MODE_PRIVATE)`

One can simply inject the prefs repo using the normal constructor/field injection mechanisms:
`@Inject lateinit var uiPrefs: UiPrefs`

With the new `PrefDelegate` property delegate, accessing the prefs goes from:
Read: `uiSharedPrefs.getBoolean("app_intro_completed", false)`
Write: `uiSharedPrefs.edit { putBoolean("app_intro_completed", true) }`

To a concise:
Read: `uiPrefs.appIntroCompleted`
Write: `uiPrefs.appIntroComplete = true`


app/src/main/java/com/geeksville/mesh/android/prefs/PrefsModule.kt
```diff
@@ -0,0 +1,143 @@
...
+import dagger.Provides
+import dagger.hilt.InstallIn
+import dagger.hilt.android.qualifiers.ApplicationContext
+import dagger.hilt.components.SingletonComponent
+import javax.inject.Qualifier
+import javax.inject.Singleton
+
+// These pref store qualifiers are private to prevent prefs stores from being injected directly.
+// Consuming code should always inject one of the prefs repositories.
+
```
jamesarich, line 32: 👍  this should help prevent misuse

poliver: might take a stab at some lint rules

PR #2769, opened 2025-08-18
Migrate `MeshService` prefs to repo pattern

poliver:

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -21,13 +21,11 @@ import android.annotation.SuppressLint
 import android.app.Service
 import android.content.Context
 import android.content.Intent
-import android.content.SharedPreferences
 import android.content.pm.ServiceInfo
 import android.os.Build
 import android.os.IBinder
 import android.os.RemoteException
 import androidx.core.app.ServiceCompat
-import androidx.core.content.edit
 import androidx.core.location.LocationCompat
 import com.geeksville.mesh.AdminProtos
 import com.geeksville.mesh.AppOnlyProtos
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -352,8 +353,8 @@ class MeshService :
 
     override fun onCreate() {
         super.onCreate()
-        sharedPreferences = getSharedPreferences("mesh-prefs", Context.MODE_PRIVATE)
-        _lastAddress.value = sharedPreferences.getString("device_address", null) ?: NO_DEVICE_SELECTED
+
+        _lastAddress.value = meshPrefs.deviceAddress ?: NO_DEVICE_SELECTED
```
poliver, line 357: A bit redundant, but leaving as-is

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -2381,7 +2380,7 @@ class MeshService :
                 if (destNum != myNodeNum) {
                     // Determine the best position to send based on user preferences and available
                     // data
-                    val provideLocation = sharedPreferences.getBoolean("provide-location-$myNodeNum", false)
+                    val provideLocation = meshPrefs.shouldProvideNodeLocation(myNodeNum)
```
poliver, line 2383: For the `mesh-prefs` store, this is never being set anywhere. The `ui-prefs` store has an identical pref, but I'm not sure if they're equivalent.

jamesarich: i would bet it's probably supposed to be using the same pref in both location,  iirc it is being set in the `Connections` screen to toggle providing the phone's gps location to the node (in place of on-node gps)

poliver: Sounds good, just calling out so we can address after this refactor. I suspect we don't need this many prefs stores.

PR #2771, opened 2025-08-18
Migrate map consent prefs to repo pattern

poliver:


app/src/main/java/com/geeksville/mesh/ui/radioconfig/RadioConfigViewModel.kt
```diff
@@ -545,6 +547,12 @@ constructor(
         }
     }
 
+    fun shouldReportLocation(nodeNum: Int?) = mapConsentPrefs.shouldReportLocation(nodeNum)
+
+    fun setShouldReportLocation(nodeNum: Int?, shouldReportLocation: Boolean) {
+        mapConsentPrefs.setShouldReportLocation(nodeNum, shouldReportLocation)
+    }
```
poliver, line 554: These appear to have equivalents in `UiPrefs`?

jamesarich: Same kinda deal here - these are probably supposed to be leveraging the same pref.

PR #2775, opened 2025-08-18
Migrate `UiState` prefs to repo

poliver:

Anything injected with a non-qualified `SharedPreferences` was getting the `ui-prefs` store, since that what `ApplicationModule` provides. Now that all pref stores are qualified in `PrefsModule`, provider in `ApplicationModule` is no longer needed.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ApplicationModule.kt
+++ b/app/src/main/java/com/geeksville/mesh/ApplicationModule.kt
@@ -18,6 +18,8 @@
 package com.geeksville.mesh
 
 import android.app.Application
+import android.content.Context
+import android.content.SharedPreferences
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleOwner
 import androidx.lifecycle.ProcessLifecycleOwner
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/geeksville/mesh/model/UIState.kt
```diff
@@ -322,50 +319,50 @@ constructor(
...
-    fun toggleOnlyOnline() = preferences.toggleBooleanPreference(onlyOnline, "only-online")
+    fun toggleOnlyDirect() = toggle(onlyDirect) { uiPrefs.onlyDirect = it }
 
-    fun toggleOnlyDirect() = preferences.toggleBooleanPreference(onlyDirect, "only-direct")
+    private fun toggle(state: MutableStateFlow<Boolean>, onChanged: (newValue: Boolean) -> Unit) {
+        (!state.value).let { toggled ->
+            state.update { toggled }
+            onChanged(toggled)
+        }
+    }
```
poliver, line 365: Probably should just get rid of this, since it breaks SRP. Keeps the diff down, though.

PR #2777, opened 2025-08-19
Migrate remaining prefs usages to repo

poliver:


app/src/main/java/com/geeksville/mesh/android/prefs/AnalyticsPrefs.kt
```diff
@@ -18,11 +18,19 @@
...
-class AnalyticsPrefsImpl(prefs: SharedPreferences) : AnalyticsPrefs {
-    override var analyticsAllowed: Boolean by PrefDelegate(prefs, "allowed", true)
+// Having an additional app prefs store is maintaining the existing behavior.
+class AnalyticsPrefsImpl(analyticsPrefs: SharedPreferences, appPrefs: SharedPreferences) : AnalyticsPrefs {
+    override var analyticsAllowed: Boolean by PrefDelegate(analyticsPrefs, "allowed", true)
+
+    private var _installId: String? by NullableStringPrefDelegate(appPrefs, "appPrefs_install_id", null)
+
+    override val installId: String
+        get() = _installId ?: UUID.randomUUID().toString().also { _installId = it }
```
poliver, line 35: Added a second store for this to preserve existing functionality. It's only used as a Firebase user ID, though, so it wouldn't matter too much if we moved it to the analytics store.

jamesarich (approved):

PR #2784, opened 2025-08-19
Extract node data from `MeshService`

poliver:
First attempt at beginning breaking up `MeshService`. Node data fields seemed like a good place to start. I'm running on my device and things appear to be working. But I could definitely use some input on my approach. Commits are sequential, so I can easily break up the PR.


poliver:
Should def test first

jamesarich:
I think this is a good start to break things up. As we progress we'll start to be better able to unwind how things are working and we may want to shift gears along the way to accommodate.

poliver:
Yea, I'm already seeing how this could be purely reactive and better driven by the flows from radio config. Lots to untangle!

poliver:
I was trying to just keep it simple and not inject the model class, but Kotlin was complaining about the injected radio config not being initialized. Not sure why that might be.

jamesarich:
Need to pull and test this one as well

poliver:
I might scrap this draft. I was trying to find the best entry point to breaking up `MeshService`, and the packet stuff makes more sense to start with.

PR #2790, opened 2025-08-19
Improve top-level nav icons/labels

poliver:
Tinkered with improving top-level nav. On smaller devices, labels are hidden, and the existing icons aren't very descriptive of the app functionality they show. Open to suggestions.

<video src="https://github.com/user-attachments/assets/675915c0-cc32-4ffd-9e20-d9ad2842e68f"/>


app/src/main/java/com/geeksville/mesh/ui/common/icons/Map.kt
```diff
@@ -0,0 +1,104 @@
...
+import androidx.compose.ui.graphics.vector.path
+import androidx.compose.ui.unit.dp
+
+val MeshtasticIcons.Outlined.Map: ImageVector
+    get() {
+        if (map != null) {
+            return map!!
+        }
+        map =
+            ImageVector.Builder(
```
jamesarich, line 38: is there a reason why we're generating vectors vs using svg assets? https://developer.android.com/studio/write/vector-asset-studio#svg

poliver: Everything's already set up to use `ImageVector`, and since it's a mix of icons from the icons lib and not (though still sourced from [Material Symbols](https://fonts.google.com/icons)), I figured it's be easier to implement similar to the icons lib.

jamesarich: As long as we can keep track of what they are and where they came from - maybe previews and a link back to the source on the Symbols site?

poliver: I'm not sure you can do previews for vector drawables like this, but I can definitely link to source

jamesarich (approved):
I'm good to merge this when the conflicts are handled, maybe a followup PR to add a section to the documentation about the icon setup and how to add more.

poliver:
@jamesarich I applied the changes we all discussed. Ready for another look.

jamesarich (approved):

PR #2800, opened 2025-08-20
Add SnackBarHost

poliver:
Not sure if this was intentional, but there is currently no `SnackbarHost` being set, meaning none of the snackbars actually get handle or shown. `NavigationSuiteScaffold` doesn't appear to have one, so I've wrapped the entire app UI in a simple `Scaffold`.


jamesarich:
ye it probably just got missed in the refactor to adaptive scaffold - good catch.

jamesarich (approved):

PR #2801, opened 2025-08-20
Add modern battery info component

poliver:
Assets obtained from Material Symbols. Empty battery symbol is used as a foreground, but it wouldn't be too hard to draw it manually given its simplicity. Battery fill level is drawn programmatically based on the node's battery level. Thresholds for color change were pretty arbitrary and based on what I could find online.

No voltage displayed yet, since it's not currently needed.

image


jamesarich (approved):
nice.

PR #2811, opened 2025-08-21
feat(navigation): Add deep links to other screens

jamesarich:
This commit introduces deep link support for all screens within the application. This allows users (and developers) to navigate directly to specific screens via URLs.

The following screens now support deep linking:
- Nodes
- Node Detail (and its sub-screens like Device Metrics, Node Map, etc.)
- Map
- Contacts
- Messages
- Share Message
- Quick Chat
- Channels
- Radio Configuration (and its sub-screens like LoRa, MQTT, etc.)
- Debug Panel

The `Intent.ACTION_VIEW` is used for all deep links.

Also updates manifest to consolidate filter intents.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/navigation/ChannelsRoutes.kt
+++ b/app/src/main/java/com/geeksville/mesh/navigation/ChannelsRoutes.kt
@@ -17,11 +17,13 @@
 
 package com.geeksville.mesh.navigation
 
+import android.content.Intent
 import androidx.compose.runtime.remember
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.navigation.NavGraphBuilder
 import androidx.navigation.NavHostController
 import androidx.navigation.compose.composable
+import androidx.navigation.navDeepLink
 import androidx.navigation.navigation
 import com.geeksville.mesh.model.UIViewModel
 import com.geeksville.mesh.ui.radioconfig.components.ChannelConfigScreen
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/geeksville/mesh/navigation/ChannelsRoutes.kt
```diff
@@ -31,44 +33,46 @@ import kotlinx.serialization.Serializable
...
-        composable<ChannelsRoutes.Channels> { backStackEntry ->
-            val parentEntry = remember(backStackEntry) {
-                val parentRoute = backStackEntry.destination.parent!!.route!!
-                navController.getBackStackEntry(parentRoute)
-            }
+    navigation<ChannelsRoutes.ChannelsGraph>(startDestination = ChannelsRoutes.Channels) {
+        composable<ChannelsRoutes.Channels>(
+            deepLinks =
+            listOf(
+                navDeepLink {
```
poliver: Are you unable to use the typesafe overloads for some reason? I think they avoid the need to manually construct the path.

jamesarich: good catch, ez fix.

app/src/main/java/com/geeksville/mesh/navigation/NodesRoutes.kt
```diff
@@ -93,14 +102,24 @@ fun NavGraphBuilder.nodesGraph(navController: NavHostController, uiViewModel: UI
...
+            deepLinks =
+            listOf(
+                navDeepLink {
+                    uriPattern = "$DEEP_LINK_BASE_URI/node/{destNum}"
+                    action = Intent.ACTION_VIEW
+                },
+                navDeepLink {
+                    uriPattern = "$DEEP_LINK_BASE_URI/node"
+                    action = Intent.ACTION_VIEW
+                },
```
poliver: I think `navDeepLink` can also be type safe.

app/src/main/java/com/geeksville/mesh/navigation/RadioConfigRoutes.kt
```diff
@@ -238,22 +420,93 @@ enum class ConfigRoute(@StringRes val title: Int, val route: Route, val icon: Im
...
-    REMOTE_HARDWARE(R.string.remote_hardware, RadioConfigRoutes.RemoteHardware, Icons.Default.SettingsRemote, 8),
-    NEIGHBOR_INFO(R.string.neighbor_info, RadioConfigRoutes.NeighborInfo, Icons.Default.People, 9),
-    AMBIENT_LIGHTING(R.string.ambient_lighting, RadioConfigRoutes.AmbientLighting, Icons.Default.LightMode, 10),
-    DETECTION_SENSOR(R.string.detection_sensor, RadioConfigRoutes.DetectionSensor, Icons.Default.Sensors, 11),
-    PAXCOUNTER(R.string.paxcounter, RadioConfigRoutes.Paxcounter, Icons.Default.PermScanWifi, 12),
+enum class ModuleRoute(
+    @StringRes val title: Int,
+    val route: Route,
+    val icon: ImageVector?,
+    val type: Int = 0,
```
poliver, line 452: I see this was here before. Does it not map directly to enum index? I'm not suggesting a change, just curious.

jamesarich: `type` is the config type (from the protos) actually requested from the radio - that should probably actually reference the enum type(number) directly tbh

poliver: Oh, so like client, router, etc?

jamesarich: [`ca51db3` (#2811)](https://github.com/meshtastic/Meshtastic-Android/pull/2811/commits/ca51db3982f9e40e80aa843311899958af12b25a)

the actual admin config proto type

PR #2812, opened 2025-08-22
`MeshService`: extract basic packet handling

poliver:
Refactors `MeshService` to extract packet queue model data and basic functionality to its own class. A couple of its dependencies cannot be injected yet, so it is just instantiated in service `onCreate`.


poliver:
I'm a dummy and forgot to switch branches. Will reopen in a separate PR

PR #2813, opened 2025-08-22
`MeshService`: extract basic packet handling

poliver:
Refactors `MeshService` to extract packet queue model data and basic functionality to its own class. A couple of its dependencies cannot be injected yet, so it is just instantiated in service `onCreate`.

Should be a good starting point for all of these packet handlers.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/PacketHandler.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/PacketHandler.kt
@@ -26,8 +26,13 @@ import com.geeksville.mesh.android.BuildUtils.debug
 import com.geeksville.mesh.android.BuildUtils.errormsg
 import com.geeksville.mesh.android.BuildUtils.info
 import com.geeksville.mesh.concurrent.handledLaunch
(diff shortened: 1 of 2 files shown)


jamesarich:
I'll try and give this a workout locally.

app/src/main/java/com/geeksville/mesh/service/PacketHandler.kt
```diff
@@ -0,0 +1,192 @@
...
+
+class PacketHandler(
+    private val packetRepository: Lazy<PacketRepository>,
+    private val serviceBroadcasts: MeshServiceBroadcasts,
+    private val radioInterfaceService: RadioInterfaceService,
+    private val meshLogRepository: Lazy<MeshLogRepository>,
+) {
+
+    private var queueJob: Job? = null
+    private val scope = CoroutineScope(Dispatchers.IO)
```
poliver, line 56: This should probably get bundled in with the job cancellation.

jamesarich: is that something you wanted to resolve in this PR?

PR #2834, opened 2025-08-25
Relocate radio config to new top-level Settings screen

poliver:
Pretty noisy PR, but it's mostly just refactor renames and package changes.
- Adds new top-level nav destination for settings (replacing qr scan/share)
- Converts existing radio config nav graph and screen to be a general settings nav graph.
- Removes radio config from the main app bar overflow menu

image

Device managed state:
image


app/src/main/java/com/geeksville/mesh/ui/common/components/MainAppBar.kt
```diff
@@ -236,11 +234,7 @@ private fun MainMenuActions(isManaged: Boolean, onAction: (MainMenuAction) -> Un
                     onAction(action)
                     showMenu = false
                 },
-                enabled =
-                when (action) {
-                    MainMenuAction.RADIO_CONFIG -> !isManaged
-                    else -> true
-                },
```
poliver, line 243: @jamesarich This should be the only actual behavior change. Do we need to handle it somewhere else?

jamesarich: yeah - `isManaged` is a device setting, my current assumption is: it basically turns off device configurability. I _think_ only settable from `cli` - basically sets up the device to be hands off (hand it to your kid or grandma type of deal).

jamesarich: i need to doublecheck it's functionality though - will check around in discord.

jamesarich: @thebentern @garthvh - any weigh in here?

poliver: ok so all of these settings should be hidden in that case?

jamesarich: Confirmed some assumptions here - 

image

poliver: Implemented disabling config when `isManaged == true` here: 02d5098e35405a8f009a0ac3c11079032e065381

jamesarich (approved):

PR #2838, opened 2025-08-25
ci(release): update release workflow to tag based versioning

jamesarich:
This commit updates the build process to:
- Use the Git tag as the `versionName` for releases.
- Upload build to the appropriate Google Play Store channel based on tag and creates a draft release
	- `v1.2.3-internal` goes to `internal test channel`
	- `v1.2.3-closed` goes to `closed alpha test channel`
	- `v1.2.3-open` goes to `open test channel`
	- `v1.2.3` goes to `production channel`
- Use the current epoch seconds as the `versionCode` for development builds if `VERSION_CODE` environment variable is not set.
- Remove `VERSION_NAME_BASE` from `Configs.kt` as it's no longer needed. It is now set from tag during ci, or `Dev Build` as fallback.
- Refactor the GitHub Actions release workflow to be triggered by tags, generate changelogs, and use the tag for versioning and Google Play deployment.
- Update `MockInterface.kt` to use a fixed string for firmware version instead of `BuildConfig.VERSION_NAME` to avoid issues with dynamic versioning.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -125,7 +125,7 @@ jobs:
             ./version_info.txt
             ./changelog.txt
           draft: true
-          prerelease: ${{ contains(github.ref_name, '-alpha') || contains(github.ref_name, '-beta') || contains(github.ref_name, '-rc') }}
+          prerelease: ${{ contains(github.ref_name, '-internal') || contains(github.ref_name, '-closed') || contains(github.ref_name, '-open') }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,35 +1,32 @@
 name: Make Release
 
 on:
-  workflow_dispatch: # Allows manual triggering
+  workflow_dispatch:
   push:
     tags:
-      - 'v*' # Trigger on tags like v1.0.0, v2.3.4-alpha.1
+      - 'v*'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 permissions:
-  contents: write # To create releases
-  pull-requests: read # For changelog generation
-  id-token: write # For OIDC token exchange with Google Play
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -145,132 +123,154 @@ jobs:
...
-        run: |
-          echo -e "versionCode=${{ env.versionCode }}\nversionNameBase=${{ env.versionNameBase }}" > ./version_info.txt
+      - name: Checkout code
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+          submodules: 'recursive'
 
       - name: Download F-Droid APK
-        uses: actions/download-artifact@v5
```
DaneEvans, line 193: why are we downgrading download-artifacts ?

.github/workflows/release.yml
```diff
@@ -145,132 +123,154 @@ jobs:
...
         with:
           subject-path: |
-            ./google-apk-download/app-google-release.apk
-            ./google-aab-download/app-google-release.aab
-            ./fdroid-apk-download/app-fdroid-release.apk
+            ./build-artifacts/google/bundle/app-google-release.aab
+            ./build-artifacts/google/apk/app-google-release.apk
+            ./build-artifacts/fdroid/app-fdroid-release.apk
+
+      - name: Determine Play Store Track
```
DaneEvans, line 255: add a comment on this one, it's a bit obscure

app/src/main/java/com/geeksville/mesh/repository/radio/MockInterface.kt
DaneEvans, line 1: are the changes in here intentional?

jamesarich: Yeah. Mostly around updating the version ref and adding a hardware "ANDROID_SIM" to the mock interface, but spotless...

.github/workflows/release.yml
jamesarich: Added comments to the attestation job. Is that what you were referencing?

PR #2839, opened 2025-08-25
`NodeDetail` list items

poliver:
Extract reusable components from `NodeDetail`, so they can be used in other parts of the app. Apply existing `TitledCard` used in Connections.

<video width="400" src="https://github.com/user-attachments/assets/50189a3b-52ad-47b7-ab2d-41b0de9fba5e" />


jamesarich (approved):
yes pls

PR #2860, opened 2025-08-26
Added Markdown Text

RCGV1:
Screenshot 2025-08-26 at 3 57 20 PM

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/common/components/MDText.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/common/components/MDText.kt
@@ -24,6 +24,7 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.text.SpanStyle
 import androidx.compose.ui.text.TextLinkStyles
 import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontFamily
 import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.tooling.preview.Preview
 import com.geeksville.mesh.ui.common.theme.HyperlinkBlue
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/geeksville/mesh/ui/message/components/MessageItem.kt
```diff
@@ -298,7 +298,7 @@ private fun MessageItemPreview() {
         )
     val receivedWithOriginalMessage =
         Message(
-            text = "This is a received message w/ original, this is a longer message to test next-lining.",
+            text = "his is a received message w/ original, this is a longer message to test next-lining.",
```
jamesarich: his what

jamesarich: revert pls

app/src/main/java/com/geeksville/mesh/ui/common/components/MDText.kt
```diff
@@ -0,0 +1,106 @@
...
+@Composable
+fun MDText(
+    text: String,
+    modifier: Modifier = Modifier,
+    style: TextStyle = MaterialTheme.typography.bodyMedium,
+    color: Color = Color.Unspecified,
+) {
+    val colors =
+        DefaultMarkdownColors(
+            text = color,
```
jamesarich, line 45: these colors, if they can't be pulled directly from the theme as-is, need to be defined in the custom colors `app/src/main/java/com/geeksville/mesh/ui/common/theme/CustomColors.kt`

app/src/main/java/com/geeksville/mesh/ui/common/components/MDText.kt
```diff
@@ -0,0 +1,107 @@
...
+            bullet = MaterialTheme.typography.bodyMedium.copy(color = color),
+            list = MaterialTheme.typography.bodyMedium.copy(color = color),
+            link = TextStyle(color = HyperlinkBlue, textDecoration = TextDecoration.Underline),
+            textLink =
+            TextLinkStyles(style = SpanStyle(color = HyperlinkBlue, textDecoration = TextDecoration.Underline)),
+            table = MaterialTheme.typography.bodyMedium.copy(color = MaterialTheme.colorScheme.onSurface),
+        )
+
+    // Custom Markdown components to disable image rendering
+    val customComponents = markdownComponents(image = { /* Empty composable to disable image rendering */ })
```
jamesarich, line 89: we do have `Coil` available if you want to take a stab at rendering images inline

RCGV1: I don't want to support this, as this works by fetching images from online, and that goes against the spirit of Meshtastic

jamesarich: Does it render a clickable image link instead?

jamesarich: if not, it probably should.

PR #2865, opened 2025-08-27
Added the ability to chose to export a qr with add or replace

RCGV1:
Screenshot 2025-08-26 at 8 27 51 PM

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/sharing/Channel.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/sharing/Channel.kt
@@ -30,6 +30,7 @@ import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.lazy.LazyColumn
@@ -138,6 +139,8 @@ fun ChannelScreen(
     val modemPresetName by remember(channels) { mutableStateOf(Channel(loraConfig = channels.loraConfig).name) }
 
     var showResetDialog by remember { mutableStateOf(false) }
+
+    var shouldAddChannelsState by remember { mutableStateOf(false) }
 
     /* Animate waiting for the configurations */
     var isWaiting by remember { mutableStateOf(false) }
@@ -269,6 +272,7 @@ fun ChannelScreen(
                 channelSet = channelSet,
                 modemPresetName = modemPresetName,
                 channelSelections = channelSelections,
+                shouldAddChannel = shouldAddChannelsState,
                 onClick = {
                     isWaiting = true
                     radioConfigViewModel.setResponseStateLoading(ConfigRoute.CHANNELS)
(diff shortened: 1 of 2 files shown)


DaneEvans:
I feel like that's an import decision, not an export one

app/src/main/java/com/geeksville/mesh/ui/sharing/Channel.kt
```diff
@@ -269,17 +272,41 @@ fun ChannelScreen(
...
+        item {
+            Row(modifier = Modifier.padding(top = 12.dp)) {
+                val selectedColors = ButtonDefaults.buttonColors()
+                val unselectedColors =
+                    ButtonDefaults.outlinedButtonColors(contentColor = MaterialTheme.colorScheme.onSurface)
+
+                OutlinedButton(
+                    onClick = { shouldAddChannelsState = true },
+                    modifier = Modifier.height(48.dp).weight(1f),
+                    colors = if (shouldAddChannelsState) selectedColors else unselectedColors,
```
jamesarich, line 296: Manually coloring buttons for selection state is bad. Use components designed for selection.

jamesarich (changes requested):
Beyond the question of whether we _should_ do this - this needs some fundamental compose help.  Use selectable states and components that support them.

b8b8:
> I feel like that's an import decision, not an export one

It's already an import decision is it not? Changing or adding this to export would mean someone sharing a QR code with you could overwrite your own channels without permission.

jamesarich:
Closing this, I'm not feelin' it. It's on the consumer of the qr to decide how to import a channelset.

RCGV1:
@jamesarich Right now iOS doesn't give you the option to chose whether to add or import, this allows android users to give iOS users a way to add the channels instead of replace them. The components I stole from the qr import screen that already had the add and replace buttons.

jp-bennett:
It seems like it would be useful to export a single channel without attempting to override the LoRa config on the receiving device.

But either way, we have an incompatibility between Android and iOS on how to handle these QR codes, and it would be good to clean it up. I vote we put James and Garth in a cage and let them fight. =D

rcarteraz:
> But either way, we have an incompatibility between Android and iOS on how to handle these QR codes, and it would be good to clean it up

This would definitely be a good thing, and better user experience. QR codes are becoming increasingly critical for meshes, and they should be handled similarly in the apps. 



> I vote we put James and Garth in a cage and let them fight. =D

I vote for this, just for the fun of it 😅

jamesarich:
One of my current thoughts about this is to send the `add` param when the whole list is **_not_** being exported - basically "I'm only intending to share a channel or two" - the app should **_not_** send the `add` if the whole list is sent, indicating the intention is replacement. This would make things pretty seamless from the ux perspective.

This has the implication that iOS would still need to respect (and allow) the case where no `add` param is sent along. 

In both platforms, on the import side - we still need to allow the users to choose whether or not they would like to ADD or REPLACE but indicate the intention (highlight the option somehow) of the import via the existence of the `add` param.

@jp-bennett brought up a small edgecase where the whole list _is_ only 1 channel, which further proves the need to give the option to the user on import.

Still to understand - what part of the LoRa config should be exported/imported/validated in either REPLACE or ADD case

NomDeTom:
Just to add the stuff we also discussed: the LoRa config needs to be checked on "add" to make sure that the channels being added are compatible with the current settings. There's no point trying to add a new channel if it requires turbo fast and you're currently using all-medfast channels.

jamesarich:
@NomDeTom kindly drew this up that describes my previous comment
image

jamesarich:
https://github.com/meshtastic/Meshtastic-Android/issues/953

https://github.com/meshtastic/python/issues/536

linking these here for reference as well

NomDeTom:
Mermaid Chart2025-08-28-113442

Chart tweaked slightly during daylight, and [edit link is here](https://www.mermaidchart.com/play#pako:eNp9UstOwzAQ_JWVD-0poum7QSAq4IpEgRPh4NqbxNS1rcSoRVX_HTtxSStarjvjmdkd7wjTHElCoihKFdMqE3mSKgBb4BoTKJF_bVNVo5nUG1bQ0sLrg6cAqPg9JXNmhVZJM3ncGu0Iz4uUfEAU3YLqO8rTTQzupVIoKw90QA38uJPb6xMkyA6ap0PPiVoCVCiRWeRBY-Txs2jQGbfx4IVRdZxrcrcDSZcoE0jJvTsCiAq6c867YL8NpgT23mN6lrZAIynDlhoMp432zPm-VVh6rs4ydFcEKSrbJps0xLhOaIxbAtmqApEBCx5Mrw21YikRNsIWUFIutNvRWqHyo1vFvctmwaRuifOrBX66A7VPZwHvX8BHAfddhY1_4-WosKQn1-4H-vBMIG18C2A1rOkKgXLeKOnSfbFWuRUbBrFRk-0_33gcuL1D7a6zqqCm_r_MHsqZnB9P_4zJ_gdqtfV6).

Issue #2715, opened 2025-08-14
[Bug]: Pressing compass does not reorient map to North


b8b8:
### What happened?

Pressing the compass on the map does not reorient the map in the North direction.

![Image](https://github.com/user-attachments/assets/2357531b-0855-47f1-b17c-c7d5d7e3250e)

### App Version

2.6.35

### Phone

Pixel9a

### Device

T1000E

### Firmware

2.7.5

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
Can we compromise by blocking rotation?
Seems compose didn't make the compass clickable, so it'll be a whole new element to make this happen

b8b8:
> Can we compromise by blocking rotation? Seems compose didn't make the compass clickable, so it'll be a whole new element to make this happen

Blocking rotate when NNNNOOOOOOTTTTT navigating is not the worst thing in the world.
When the lock and follow location works again it will need to rotate around your current heading.

DaneEvans:
🤮 
That UI trend is awful 
Maps should always be north up.

b8b8:
> 🤮 That UI trend is awful Maps should always be north up.

not when driving. You dont use google maps with forced north do you?

DaneEvans:
Absolutely. If they let me, I think I had to move off gMaps when they stopped making it an option 

It would be tolerable if irritating, but they bundle in the tilt angle thing, and that's just awful

b8b8:
> Absolutely. If they let me, I think I had to move off gMaps when they stopped making it an option
> 
> It would be tolerable if irritating, but they bundle in the tilt angle thing, and that's just awful

Tilt angle is awful.

jamesarich:
https://github.com/user-attachments/assets/c975e569-22c6-4326-ac64-6fadea2c3143

jamesarich:
yes it does work - at least on node details map 🙃

b8b8:
> yes it does work - at least on node details map 🙃

80 of 81 maps working so far then (on NRF), just need the main map.

DaneEvans:
MR in ^ that solves this issue, but adds a few more ...

It's a fun one. The nodechips from clustering are blocking taps from reaching the map. 
It's beyond my understanding of the maps and clustering at the moment

b8b8:
Zoom out to quite wide does allow the reorientation to north to work.

PR #2893, opened 2025-08-28
Compass rose / orient north button working

DaneEvans:
The nodechips spawned by 'clustering' seems to include an extra that appears (invisible, and disabled) over the compass rose. 
This swallows clicks, preventing them from getting to the underlying map. 

Text doesn't have this issue. 

But also clustering doesn't seem to be working properly, it seems to just stack them all on top of each other, which doesn't work with the tranparent text. That said, Tom's screenshot seems to show it works on his. 

@jamesarich 
If you get a chance, have a look at this, and what you have in place, and see if you see a way forward.

solves #2715


jamesarich:
I played for a while, this is real weird  - the compass onclick starts getting swallowed as soon as i start making the un-clustered marker's `content` anything remotely complex (beyond an icon or text) :(

might be related to some jank in the compose maps utils lib https://github.com/googlemaps/android-maps-compose/issues/641

Issue #2890, opened 2025-08-28
[i18n][UI] Paxcounter “Update interval” label says milliseconds but the config is in seconds, and the string is not found on Crowdin


wwin3286tw:
### Contact Details

wwin3286tw@gmail.com

### What happened?
Locale: zh-TW (Traditional Chinese).
In the Paxcounter screen the label reads 「更新間隔（毫秒）」.
However, the underlying config paxcounter.paxcounter_update_interval is in seconds.
This misleads users (especially in zh-TW) to configure much shorter intervals.

Additionally, I could not find this label in Crowdin (Meshtastic Android project) — searching for interval, update, 毫秒, 間隔 — so the string may be hard-coded or not exposed to translations.

* Note: This report is for the zh-TW locale. Other locales may or may not be correct; if English or others are already fixed, please treat this as a zh-TW-only localization issue. If the string is missing from resources, it affects all locales.

### Why this is a separate issue

Related to #2887 (input not accepting negative values) and #2889 (negative values rendered as 4294967206), but this is a distinct i18n/labeling bug. Fixing input/rendering will not fix the unit mismatch and missing string.

### Steps to reproduce

1. Android → … (three dots) → Radio Configuration → Paxcounter.
2. Observe the label 「更新間隔（毫秒）」.
3. Set/verify the interval via CLI (paxcounter.paxcounter_update_interval) and note that the firmware treats it as seconds.

### Expected
* Label should read “Update interval (seconds)” (zh-TW: 「更新間隔（秒）」).
* The string should be externalized in strings.xml and available on Crowdin for localization.

### Actual
* UI label in zh-TW says milliseconds(zh-TW: 「更新間隔（毫秒）」), causing confusion.
* I cannot find this string on Crowdin (searched “interval”, “update”, etc.), suggesting it’s not exposed for translation.

### Impact
* Users may set intervals 1000× shorter than intended, leading to excessive airtime usage and misconfiguration.

### App Version

2.6.35(35131) google

### Phone

Xiaomi 14T Pro

### Device

Heltec Wireless Stick Lite V3

### Firmware

2.6.11

### Relevant log output

<img width="858" height="1716" alt="Image" src="https://github.com/user-attachments/assets/ed283195-b9ca-4a68-9119-44001b60a233" />

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
It uses the same string `update_interval_seconds` as several other places. 

[...]

PR #2912, opened 2025-08-29
fix(i18n): correct zh-TW Paxcounter label and other translations

littlechintw:
### Summary
Fix incorrect Traditional Chinese (zh-TW) translation for the Paxcounter screen and other minor translation issues.  
The label previously read 「更新間隔（毫秒）」 ("Update interval (milliseconds)"), but the underlying configuration (`paxcounter.paxcounter_update_interval`) is in **seconds**, which could mislead users.  

### Changes
- Corrected the Paxcounter interval label to 「更新間隔（秒）」 ("Update interval (seconds)")
- Externalized the string in `strings.xml` so it is available on Crowdin for translation
- Fixed other minor zh-TW translation inconsistencies in the app

### Steps to Reproduce
1. Open Android app → … (three dots) → Radio Configuration → Paxcounter
2. Observe the label 「更新間隔（毫秒）」.
3. Set or verify the interval via CLI (`paxcounter.paxcounter_update_interval`) — firmware treats it as seconds

### Expected
- Label should indicate seconds (zh-TW: 「更新間隔（秒）」)
- Translation string should be accessible in `strings.xml` and on Crowdin

### Actual
- Label shows milliseconds (「更新間隔（毫秒）」), potentially confusing users
- String was not previously exposed for translation

### Impact
Users might configure intervals 1000× shorter than intended, leading to excessive airtime usage and misconfiguration.

### Notes
- zh-TW localization only; other locales unaffected

### Related Issue
- Fixes / addresses: #2890


jamesarich:
We don't accept translations via PR, please use Crowdin.

PR #2928, opened 2025-08-30
chore(deps): group all the things

jamesarich:
This change introduces several grouping rules for Renovate Bot to consolidate dependency updates into more manageable pull requests.

The following groups have been added:
- AndroidX
- Kotlin Ecosystem (stdlib, coroutines, serialization)
- Dagger & Hilt
- Accompanist
- JVM Testing Libraries (JUnit, Mockito, Robolectric)
- AndroidX Testing
- Square Networking (OkHttp, Retrofit)
- Coil
- ZXing
- MQTT Paho Client
- Markdown Renderer (Mike Penz)
- Firebase
- Datadog
- OSM Libraries
- Google Maps Compose
- Protobuf Runtime
- AndroidX Room
- AndroidX Lifecycle
- AndroidX Navigation
- AndroidX DataStore
- AndroidX Adaptive UI

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -32,6 +32,144 @@
         "https://github.com/meshtastic/design.git"
       ],
       "changelogUrl": "https://github.com/meshtastic/design/compare/{{currentDigest}}...{{newDigest}}"
+    },
+    {
+      "description": "Group all AndroidX dependencies",
+      "matchPackagePatterns": ["^androidx\\."],
+      "groupName": "AndroidX",
+      "groupSlug": "androidx"
+    },
+    {
+      "description": "Group Kotlin standard library, coroutines, and serialization",
+      "matchPackagePatterns": ["^org\\.jetbrains\\.kotlin", "^org\\.jetbrains\\.kotlinx"],
+      "groupName": "Kotlin Ecosystem",
+      "groupSlug": "kotlin"
+    },
+    {
+      "description": "Group Dagger and Hilt dependencies",
+      "matchPackagePatterns": ["^com\\.google\\.dagger", "^androidx\\.hilt"],
+      "groupName": "Dagger & Hilt",
+      "groupSlug": "hilt"
+    },
+    {
+      "description": "Group Accompanist libraries",
+      "matchPackagePatterns": ["^com\\.google\\.accompanist"],
+      "groupName": "Accompanist",
+      "groupSlug": "accompanist"
+    },
+    {
+      "description": "Group JVM testing libraries (JUnit, Mockito, Robolectric)",
+      "matchPackagePatterns": [
+        "^junit:junit$",
+        "^org\\.mockito:",
+        "^org\\.robolectric:robolectric$"
+      ],
+      "groupName": "JVM Testing Libraries",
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -32,6 +32,144 @@
...
       "changelogUrl": "https://github.com/meshtastic/design/compare/{{currentDigest}}...{{newDigest}}"
+    },
+    {
+      "description": "Group all AndroidX dependencies",
+      "matchPackagePatterns": ["^androidx\\."],
+      "groupName": "AndroidX",
+      "groupSlug": "androidx"
+    },
+    {
+      "description": "Group Kotlin standard library, coroutines, and serialization",
```
poliver, line 55: Does this also cover KSP?

.github/renovate.json
```diff
@@ -32,6 +32,144 @@
         "https://github.com/meshtastic/design.git"
       ],
       "changelogUrl": "https://github.com/meshtastic/design/compare/{{currentDigest}}...{{newDigest}}"
+    },
+    {
+      "description": "Group all AndroidX dependencies",
```
poliver: Is this needed? A lot of these libs are unrelated.

poliver (commented):
The only thing I've really found necessary in the past was a Kotlin group, since those usually need to align. I'd suggest maybe hold off on creating a huge list of groups until you start seeing Renovate PRs failing due to a single breaking dep.

Not a big deal to have a bunch of groups, just might be overkill.

.github/renovate.json
jamesarich: just broke most of those out in excludes

poliver:
I'd imagine the merge queue would catch breaking deps if a ton were to be queued at once.

DaneEvans (approved):
Looks like it should work fine,

Is there a particular issue you're trying to address?
Or just make things clearer

jamesarich:
goal is to cut down on overall PRs that need to be run through ci by grouping potential concurrent dep updates

jamesarich:
> I'd imagine the merge queue would catch breaking deps if a ton were to be queued at once.

they also burn separate time in ci in individual PR checks

jamesarich:
I'm gonna give it a shot, the groups still make logical sense even if they aren't commonly updated concurrently (or updated at all in the case of osm 😂 ) but it shouldn't have any negative impact if they aren't concurrently updated. The PR will just get the group name.

poliver:
> they also burn separate time in ci in individual PR checks

Isn't there an option to consolidate updates without necessarily creating dependency groups?

Again, it will work fine but it's just more to maintain if you don't need the "BOM" functionality of ensuring codependent libs are incremented together.

jamesarich:
> Isn't there an option to consolidate updates without necessarily creating dependency groups?

potentially, happy to try that out too if it's better - again, goal is just less overall ci time. (and buttons to push)

poliver:
No strong feelings either way, just sharing my experience with it.

jamesarich:
> No strong feelings either way, just sharing my experience with it.

i'll have to dig more thru renovate docs when i get the gumption, see if i can find that other option to see what's up

PR #2934, opened 2025-08-30
feat: Support the `add` export method on channel url/qr

jamesarich:
Continues work from @RCGV1 with updates to use Material 3 Segmented Buttons as requested.

Also defaults the selection to `add`.

image


RCGV1:
The default should be replace

Issue #2793, opened 2025-08-20
[Bug]: buggy animation


cracky22:
### Contact Details

martinblieninger2208@gmail.com

### What happened?

Hi, sorry that uploading the first video didn't work (#2764). I understand that you don't want to download a video from a file-sharing service @DaneEvans.

The bugs I found are that the cancel button doesn't work in the channel settings, and the animation laags completely when you disable the exact location and the slider extends.

https://github.com/user-attachments/assets/1f731465-995f-4502-96e8-39fef3e47223

### App Version

2.6.35 (35131) google

### Phone

Samsung Galaxy S24 Plus

### Device

Seed Tracker T1000-E

### Firmware

2.7.5.ddd1499

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
Cancel is actually a 'reset changes' (did we change that string in english? Or was it used in too many places and too much effort to find them all? ) 
Use the back arrow /swipe to go back. 
#2387 - mine apparently.

The animation, do you have developed options turned on? (Spam press something in settings, you'll know if you've done it) You can change animation speeds there. Not saying you should, just that it may be a cause.  

I haven't noticed it before, but I'll see if it's reproducible in the morning.  

Definitely reproducible. 2.6.35++.

cracky22:
nope, developer options are not turned on

DaneEvans:
For the cancel vs clear one, jump on crowdIn and provide an update for that language. 

nvm - it's still / again cancel in English. That's going to be a me job

PR #2935, opened 2025-08-30
Improvements to Channel management

Robert-0410:
## Changes
- Updated the way we communicate channel features to users
- Detects firmware version to properly place the location icon with the changes introduced in 2.6.10 for periodic position broadcasts

## Dialog

https://github.com/user-attachments/assets/8c4a26e2-1c42-4f2a-ac6b-62a0eca5923a

## Icons

https://github.com/user-attachments/assets/677514ed-1652-48c4-968d-0493049927e7


app/src/main/proto
DaneEvans, line 1: no updates to protos

app/src/main/res/values/strings.xml
```diff
@@ -801,4 +801,11 @@
     <string name="url_template_hint" translatable="false">https://a.tile.openstreetmap.org/{z}/{x}/{y}.png</string>
     <string name="track_point">track point</string>
     <string name="phone_settings">Phone Settings</string>
+    <string name="channel_features">Channel Features</string>
+    <string name="location_sharing">Location Sharing</string>
+    <string name="periodic_position_broadcast">Periodic position broadcast</string>
+    <string name="uplink_feature_description">Messages from the mesh will be sent to the public internet through any node\'s configured gateway</string>
+    <string name="downlink_feature_description">Messages captured from a public internet gateway will be forwarded to the local mesh</string>
```
DaneEvans: They only come to that device right? not onto the mesh further? Or is is 1 hop from there?

mesh_service_example/src/main/proto
DaneEvans, line 1: no protos

Robert-0410: Some side effects from doing the merge conflicts, i'll revert those specific files.

app/src/main/res/values/strings.xml
Robert-0410: I got the up and down link definitions from here: [Channel Configurations](https://meshtastic.org/docs/configuration/radio/channels/)

Robert-0410: This should clarify further: [Zero hop policy](https://meshtastic.org/docs/software/integrations/mqtt/#zero-hop-policy)

DaneEvans: Thats also ambiguous. 

My memory of random chatter on the discord is that it only 'propogates' as far as the connected node (maybe it's the next ones on an encrypted channel?). 
We need to figure it out and clean up the language here. 
If we have ambiguous language in the app, then we're going to get even more complaints by people that can't google.

Robert-0410:
fixes #2793

app/src/main/res/values/strings.xml
Robert-0410: Updated the description for the downlink.

app/src/main/java/com/geeksville/mesh/ui/common/components/SecurityIcon.kt
```diff
@@ -15,6 +15,8 @@
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+@file:Suppress("TooManyFunctions")
```
DaneEvans, line 18: I don't like it, but if there's no easy wins here, then it's probably better to just get this one in.

jamesarich (approved):

DaneEvans (approved):

PR #2948, opened 2025-09-03
Fix Bluetooth reconnection logic

thebentern:
When attempting to reconnect to an out of range or turned off device I sometimes get an exception due to currentWork being scheduled still. This PR fixes that by calling stopCurrentWork to ensure this pending operation is canceled before we attempt to reconnect.


jamesarich:
I'll pull it and give it a whirl.

Issue #2840, opened 2025-08-26
[Bug]: Default Landing page for app start/open is connection page


b8b8:
### What happened?

Default Landing page for app start/open is connection page even when already connected to a node. Suggest move to node list unless no connection is detected.

### App Version

2.7.0-35188

### Phone

Pixel9a

### Device

T1000E

### Firmware

2.7.6

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #2954, opened 2025-09-04
Make Nodes screen the root of the nav graph

poliver:
This causes the app to land on the Nodes screen on launch instead of the Connections screen. Based on discussions in Discord.


garthvh:
The messaging screen is twice as popular as the nodes screen.

jamesarich (approved):

b8b8:
https://github.com/meshtastic/Meshtastic-Android/issues/2840

PR #2958, opened 2025-09-04
Fix compass / orient north button

thebentern:
Closes #2893


jamesarich:
If we're replacing the broken maps functionality w/ our own implementations, we may want to make note of that somewhere in a comment so we could potentially fall back to the base sdk if/when it works.

jamesarich (approved):
Looks good otherwise.

PR #2965, opened 2025-09-04
Componentize traceroute button

poliver:
Aligns the traceroute button with other clickable list items and extracts it to its own file. The button does disable correctly. Disabled alpha not being set is a common issue across all list items currently.

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/8079e60c-af6b-4592-8d6a-f4502fc2cf9f"/>|<video src="https://github.com/user-attachments/assets/a2726940-2778-4957-b3a2-6130be0bb6d7"/>|

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/node/components/TracerouteButton.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/node/components/TracerouteButton.kt
@@ -26,14 +26,13 @@ import androidx.compose.material3.CircularProgressIndicator
 import androidx.compose.material3.ProgressIndicatorDefaults
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
-import androidx.compose.runtime.getValue
-import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
-import androidx.compose.runtime.setValue
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import com.geeksville.mesh.R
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/geeksville/mesh/ui/node/components/TracerouteButton.kt
```diff
@@ -0,0 +1,94 @@
...
+private fun TracerouteButton(text: String, progress: Float, onClick: () -> Unit) {
+    val isCoolingDown = progress > 0f
+
+    SettingsItem(
+        text = text,
+        enabled = !isCoolingDown,
+        leadingIcon = Icons.Default.Route,
+        trailingContent = {
+            if (isCoolingDown) {
+                CircularProgressIndicator(
```
jamesarich: can we use the WavyCircularProgressIndicator from expressive? 🙏

poliver: I really don't like it 😄, but I can look into it.

DaneEvans: pls don't

poliver: Updated

image

poliver: I do think it generally looks bad

jamesarich: 〰️ 〰️ 〰️

jamesarich: so expressive 👍

PR #2967, opened 2025-09-05
feat(theme): add expressive motion to theme

jamesarich:
This commit adds the `expressive()` motion scheme to the `MaterialExpressiveTheme`.


app/src/main/java/com/geeksville/mesh/ui/common/theme/Theme.kt
```diff
@@ -32,267 +33,270 @@ import androidx.compose.runtime.Immutable
...
 
 @OptIn(ExperimentalMaterial3ExpressiveApi::class)
 @Composable
 fun AppTheme(
     darkTheme: Boolean = isSystemInDarkTheme(),
     dynamicColor: Boolean = true,
-    content: @Composable() () -> Unit
+    content:
+    @Composable()
+    () -> Unit,
```
poliver, line 282: Spotless did this?

poliver (approved):
What are the expected visual changes?

DaneEvans:
My thoughts exactly - looks the same to me. No issues found though

app/src/main/java/com/geeksville/mesh/ui/common/theme/Theme.kt
DaneEvans: Yep. Otherwise it'd be complaining about it

jamesarich:
@DaneEvans @poliver it's pretty subtle, but should cascade through components that support it. https://m3.material.io/blog/m3-expressive-motion-theming

PR #2969, opened 2025-09-05
Align syle of admin settings

poliver:
Cleans up remaining radio config/settings sections

| Before | After |
|------|-----|
|image|image|

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/settings/radio/RadioConfig.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/settings/radio/RadioConfig.kt
@@ -35,6 +35,7 @@ import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.Button
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
+import androidx.compose.material3.LocalContentColor
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/geeksville/mesh/ui/settings/radio/RadioConfig.kt
```diff
@@ -189,8 +125,35 @@ fun RadioConfigItemList(
...
+                    onDismiss = { showDialog = false },
+                    onConfirm = { onRouteClick(route) },
+                )
+            }
+
+            val contentColor =
+                when (route) {
+                    AdminRoute.FACTORY_RESET,
+                    AdminRoute.NODEDB_RESET,
+                    -> MaterialTheme.colorScheme.StatusRed
```
thebentern: The contrast on the red from the screenshot seems problematic for users with accessibility issues. I wonder we can do red background with dark or white text?

poliver: I thought about that, but these rows are disable-able. Perhaps I should just leave the color alone for now?

DaneEvans: It's pretty problematic even with vision. 

I'd leave them the normal colour, if they aren't 'press me' green, and have a confirm popup on them, I think the standard colour should be fine

poliver: Sounds good, I'll revert

DaneEvans (commented):
looks good. 
Is it worth trying to break the radioconfigitemlist down a bit, rather than adding another suppression to it. 

Managed message could take a Bool: isManaged, that'd save at least 2
the Local settings bit could get it's own function. 

That would probably get it under both limits

app/src/main/java/com/geeksville/mesh/ui/settings/radio/RadioConfig.kt
poliver: Reverted in 0fc459ef777f4b2e4c7ab28266afa9a16a8b6928

poliver:
> looks good. Is it worth trying to break the radioconfigitemlist down a bit, rather than adding another suppression to it.
> 
> Managed message could take a Bool: isManaged, that'd save at least 2 the Local settings bit could get it's own function.
> 
> That would probably get it under both limits

Yea, I was planning to take a look at that after wrapping up the style changes. Each settings section can likely go into its own file, which would make previews easier.

DaneEvans:
I'm happy with that

poliver:
> I'm happy with that

Queuing it up. Feel free to cancel if there's anything else.

jamesarich:
good by me 🚢

PR #2970, opened 2025-09-05
fix(service): set node time after config completion

jamesarich:
The node time was previously set before the configuration was complete. This change moves the time setting operation to after the configuration flow finishes, ensuring the node has the correct time after it's fully configured.

Additionally, the `CONFIG_WAIT_MS` constant was reduced from 250L to 50L.

This one could use some extra hands-on, connection switching testing.


DaneEvans:
Seems pretty good. 
No worse at swapping connections.

Seems to tell me things were last heard 18 hours ago less often.

PR #2973, opened 2025-09-05
refactor(map): fix scalebar and compass

jamesarich:
Replaces broken `DisappearingScaleBar` with static `ScaleBar` - bottom start aligned.

Hides sdk compass element - replaces with rotatey icon in our own MapsControlOverlay.

Removes convoluted camera state logic.

### DIFF
--- a/app/src/google/java/com/geeksville/mesh/ui/map/MapView.kt
+++ b/app/src/google/java/com/geeksville/mesh/ui/map/MapView.kt
@@ -207,7 +207,6 @@ fun MapView(
     val mapFilterState by mapViewModel.mapFilterStateFlow.collectAsStateWithLifecycle()
     val ourNodeInfo by uiViewModel.ourNodeInfo.collectAsStateWithLifecycle()
     var editingWaypoint by remember { mutableStateOf<Waypoint?>(null) }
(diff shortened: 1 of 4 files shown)


DaneEvans:
@jamesarich 
Can we get more than 0dp precision pls 
It's fine at the <1 mile | < 1km, but there's a pretty big difference between 1.1 and 1.9 in those.

jamesarich:
> @jamesarich Can we get more than 0dp precision pls

~~nah it's red when true north and can be trued by tapping the icon~~

DaneEvans:
not the compass, the scale

jamesarich:
> @jamesarich Can we get more than 0dp precision pls It's fine at the <1 mile | < 1km, but there's a pretty big difference between 1.1 and 1.9 in those.

oh u mean the scalebar? it's using the out of the box bar from sdk.

jamesarich:
@DaneEvans https://github.com/googlemaps/android-maps-compose/blob/main/maps-compose-widgets/src/main/java/com/google/maps/android/compose/widgets/ScaleBar.kt

DaneEvans (commented):
- Map always opens in US, then scrolls to current position, I can handle that once a session, but more than that isn't great. 
- Scale bar jumps 1 mile > 2 mine, with no decimals (km the same)
- Compass behaves great.

Issue #1005, opened 2024-04-29
[Feature Request] ability to apply nicknames to nodes


faradayfedora:
Think it would be nice if you could apply your own name to node, use case is putting know names since alot of nodes are aliases.

prokrypt:
What may be easier to implement (and more useful in my opinion) is a searchable notes field for each node...

standaloneSA:
Closed as not planned? How do people use this software to communicate with people? Do you just memorize dozens of hex IDs?

PR #3014, opened 2025-09-08
Feat/node notes

DaneEvans:
Add private notes to favourited nodes.


garthvh:
This has been explicitly avoided as it would not be something that works across the different clients and takes up too much space on the node. We are really trying to implement features that work across all the platforms, not create additional discrepancies between them.

https://github.com/meshtastic/Meshtastic-Android/issues/1005

https://github.com/meshtastic/web/issues/588

DaneEvans:
It doesn't touch the node. 
It's solely a QOL feature on the android app. For things like where exactly you put that node, who to contact for building access etc. 

And from what I've seen you've personally added most of the discrepancies. And seem to think everyone else has to come to iOS without compromise.

Issue #3106, opened 2025-09-15
[Feature Request]: Add Maidenhead Locator System to GPS Format


Xaositek:
### Checklist

- [x] I have used the search function for **OPEN ISSUES** to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for **CLOSED ISSUES** to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.

- [x] I will describe the request with as much detail as possible.

- [x] This request contains only one single feature, **not** a list of multiple (related) features.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Contact Details

Xaositek on Discord

### Feature or improvement you want

As part of https://github.com/meshtastic/firmware/pull/7979 we are looking to add MLS to available GPS Formats. We'd want to add this to the relevant location if available in the app.

### Why should this be added?

It adds a common HAM format grid squares which is common usage for location an individual in a general area.

Xaositek:
No work to be done.

PR #3107, opened 2025-09-15
feat(settings): add GPS coordinate format setting

jamesarich:
Adds a new dropdown preference in the display settings to allow users to select their preferred GPS coordinate format. The available options are filtered to exclude `UNRECOGNIZED`.

Maybe un-deprecated? #3106


jamesarich:
@Xaositek ^

jamesarich:
big nope

PR #3110, opened 2025-09-15
feat(build-logic): modularize build-logic & update dependencies

jamesarich:
More modularization and build-logic cleanup


build-logic/convention/src/main/kotlin/com/geeksville/mesh/buildlogic/AndroidCompose.kt
```diff
@@ -47,31 +39,34 @@ internal fun Project.configureAndroidCompose(
...
-            .let(metricsDestination::set)
-
-        project.providers.gradleProperty("enableComposeCompilerReports").onlyIfTrue()
-            .relativeToRootProject("compose-reports")
-            .let(reportsDestination::set)
-
-        stabilityConfigurationFiles
-            .add(isolated.rootProject.projectDirectory.file("compose_compiler_config.conf"))
-    }
+//    extensions.configure<ComposeCompilerGradlePluginExtension> {
```
poliver: Is this ok to remove?

build-logic/convention/src/main/kotlin/DetektConventionPlugin.kt
```diff
@@ -9,9 +9,11 @@ import org.gradle.kotlin.dsl.getByType
 class DetektConventionPlugin : Plugin<Project> {
     override fun apply(target: Project) {
         with(target) {
-            apply(plugin = libs.findPlugin("detekt").get().get().pluginId)
-            val extension = extensions.getByType<DetektExtension>()
-            configureDetekt(extension)
+            allprojects {
```
poliver: Wouldn't this be applied to modules as-needed?

build-logic/convention/src/main/kotlin/SpotlessConventionPlugin.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.geeksville.mesh.buildlogic.libs
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.kotlin.dsl.apply
+import org.gradle.kotlin.dsl.getByType
+
+class SpotlessConventionPlugin : Plugin<Project> {
+    override fun apply(target: Project) {
+        with(target) {
+            allprojects {
```
poliver: Same question as detekt

build-logic/convention/src/main/kotlin/AndroidApplicationConventionPlugin.kt
```diff
@@ -37,7 +33,8 @@ class AndroidApplicationConventionPlugin : Plugin<Project> {
             apply(plugin = "meshtastic.android.lint")
             apply(plugin = "meshtastic.android.room")
```
poliver: Is Room assumed to be needed by all applications?

jamesarich: probably not - good call

PR #3126, opened 2025-09-17
Decouple `ConnectionsScreen` from `UiViewModel`

poliver:
Adds a new `ConnectionsViewModel` to contain Flows/data needed by `ConnectionsScreen`. Anything no longer used in `UiViewModel` (i.e. only used by `ConnectionsScreen`) was removed.

Updates the `provide-location-nodeNum` pref to be an observable StateFlow. `MeshService` now observes this StateFlow directly and toggles location requests accordingly. No more proxying these commands through the service binder.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/connections/ConnectionsScreen.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/connections/ConnectionsScreen.kt
@@ -263,12 +263,6 @@ fun ConnectionsScreen(
                         }
                     }
 
-                    LaunchedEffect(ourNode) {
-                        if (ourNode != null) {
-                            connectionsViewModel.refreshProvideLocation()
-                        }
-                    }
-
                     Spacer(modifier = Modifier.height(16.dp))
 
                     // Warning Not Paired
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/geeksville/mesh/ui/connections/ConnectionsScreen.kt
```diff
@@ -207,10 +207,10 @@ fun ConnectionsScreen(
...
                 }
 
-                val setRegionText = stringResource(id = R.string.set_your_region)
+                /*val setRegionText = stringResource(id = R.string.set_your_region)
                 val actionText = stringResource(id = R.string.action_go)
-                LaunchedEffect(isConnected && regionUnset && selectedDevice != "m") {
-                    if (isConnected && regionUnset && selectedDevice != "m") {
+                LaunchedEffect(connectionState.isConnected() && regionUnset && selectedDevice != "m") {
+                    if (connectionState.isConnected() && regionUnset && selectedDevice != "m") {
                         uiViewModel.showSnackBar(
```
poliver, line 214: Commented this out, since it was annoying folks, and `showSnackBar()` is only available from `UiViewModel`. It needs a new home where it can be conveyed without using SnackBar.

app/src/main/java/com/geeksville/mesh/ui/connections/ConnectionsScreen.kt
```diff
@@ -263,18 +263,13 @@ fun ConnectionsScreen(
                         }
                     }
 
-                    LaunchedEffect(ourNode) {
-                        if (ourNode != null) {
-                            uiViewModel.refreshProvideLocation()
```
poliver, line 268: `MeshService` will now handle this internally whenever `myNodeInfo` or the `provide-location-nodeNum` pref change.

jamesarich (approved):

PR #3142, opened 2025-09-18
Add `MapScreen`

poliver:
With `MapView` now wrapped in a `Scaffold`, it makes sense to wrap the whole thing in `MapScreen` to keep the nav logic simple, enable easy previews, etc. This should also make decoupling from `UiViewModel` a bit easier.


jamesarich (approved):
👍

PR #3180, opened 2025-09-23
Update strings.xml

b8b8:
Minor updates to strings for clarity and to reduce ambiguity.

### DIFF
--- a/app/src/google/java/com/geeksville/mesh/ui/map/MapViewModel.kt
+++ b/app/src/google/java/com/geeksville/mesh/ui/map/MapViewModel.kt
@@ -26,7 +26,6 @@ import com.geeksville.mesh.android.BuildUtils.debug
 import com.geeksville.mesh.database.NodeRepository
 import com.geeksville.mesh.database.PacketRepository
 import com.geeksville.mesh.repository.datastore.RadioConfigRepository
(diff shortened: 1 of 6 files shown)


core/strings/src/main/res/values/strings.xml
```diff
@@ -132,7 +132,7 @@
     <string name="config_position_broadcast_secs_summary">The maximum interval that can elapse without a node broadcasting a position.</string>
     <string name="config_position_broadcast_smart_minimum_interval_secs_summary">The fastest that position updates will be sent if the minimum distance has been satisfied.</string>
     <string name="config_position_broadcast_smart_minimum_distance_summary">The minimum distance change in meters to be considered for a smart position broadcast.</string>
-    <string name="config_position_gps_update_interval_summary">How often should we try to get a GPS position.</string>
+    <string name="config_position_gps_update_interval_summary">How often should we try to get a GPS position (<10sec keeps GPS on).</string>
```
jamesarich: I think it's choking on the `<` symbol - you may be able to use the unicode `&lt;`

https://developer.android.com/guide/topics/resources/string-resource.html#escaping_quotes

jamesarich (changes requested):

PR #3213, opened 2025-09-26
Decouple `MapView` from `UiViewModel`

poliver:


app/src/fdroid/java/com/geeksville/mesh/ui/map/MapView.kt
```diff
@@ -267,7 +262,7 @@ fun MapView(
     fun MapView.toggleMyLocation() {
         if (context.gpsDisabled()) {
             debug("Telling user we need location turned on for MyLocationNewOverlay")
-            uiViewModel.showSnackBar(R.string.location_disabled)
+            Toast.makeText(context, R.string.location_disabled, Toast.LENGTH_SHORT).show()
```
poliver, line 265: Snackbars are a pretty hard coupling to `UiViewModel`, since it is the only one equipped to show them. I opted to use toasts instead.

jamesarich: I think that's fine - SnackBar was being a bit misused anyways.

jamesarich (approved):

PR #3220, opened 2025-09-27
Remove `NodeChip` long-click

poliver:
Currently long-clicking a `NodeChip` opens a context menu with a number of actions (some of them asynchronous and non-trivial to implement) that can be performed on a given node (e.g. traceroute). `NodeChip` single-click opens the node detail screen, which contains all of these actions, so it's a bit redundant to have such a comprehensive long-click context menu everywhere. Not to mention the extra code we'd need to maintain to fully support this as a drop-in component without duplicating code everywhere (for minimal gain).

This PR removes the `NodeChip` long-click action which simplifies things quite a bit without really compromising functionality. Single-click still works as expected where applied.


jamesarich (changes requested):
As discussed let's move the long-click action only to the node card with a subset of the actions.

poliver:
> As discussed let's move the long-click action only to the node card with a subset of the actions.

image

PR #3224, opened 2025-09-28
Spruce up `LoRaConfigScreen`

poliver:
Aligns this screen with card section style used elsewhere in the app. I'm sure this could be refined a bit more, but it's a start.

|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/a71f68ef-bb07-47f9-99d4-341385bf89a0"/>|<video src="https://github.com/user-attachments/assets/90c6f847-ed59-4987-9102-23767f152d32"/>|

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/settings/radio/components/LoRaConfigItemList.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/settings/radio/components/LoRaConfigItemList.kt
@@ -20,7 +20,6 @@ package com.geeksville.mesh.ui.settings.radio.components
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.text.KeyboardActions
-import androidx.compose.material3.HorizontalDivider
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
@@ -38,6 +37,7 @@ import com.geeksville.mesh.ConfigProtos.Config.LoRaConfig
 import com.geeksville.mesh.config
 import com.geeksville.mesh.copy
 import com.geeksville.mesh.ui.common.components.DropDownPreference
+import com.geeksville.mesh.ui.common.components.PreferenceDivider
 import com.geeksville.mesh.ui.settings.radio.RadioConfigViewModel
 import org.meshtastic.core.model.Channel
 import org.meshtastic.core.model.RegionInfo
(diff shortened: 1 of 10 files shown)


core/model/src/main/kotlin/org/meshtastic/core/model/ChannelOption.kt
```diff
@@ -294,14 +294,14 @@ enum class RegionInfo(
...
-    SHORT_SLOW(ModemPreset.SHORT_SLOW, .250f),
-    MEDIUM_FAST(ModemPreset.MEDIUM_FAST, .250f),
-    MEDIUM_SLOW(ModemPreset.MEDIUM_SLOW, .250f),
-    LONG_FAST(ModemPreset.LONG_FAST, .250f),
-    LONG_MODERATE(ModemPreset.LONG_MODERATE, .125f),
-    LONG_SLOW(ModemPreset.LONG_SLOW, .125f),
-    VERY_LONG_SLOW(ModemPreset.VERY_LONG_SLOW, .0625f),
+enum class ChannelOption(val modemPreset: ModemPreset, val description: String, val bandwidth: Float) {
+    VERY_LONG_SLOW(ModemPreset.VERY_LONG_SLOW, "", .0625f),
+    LONG_FAST(ModemPreset.LONG_FAST, "Long Range - Fast", .250f),
```
jamesarich: These descriptions should be extracted to string resources.

app/src/main/java/com/geeksville/mesh/ui/settings/radio/components/LoRaConfigItemList.kt
```diff
@@ -65,184 +70,182 @@ fun LoRaConfigScreen(navController: NavController, viewModel: RadioConfigViewMod
...
-                    enabled = state.connected && formState.value.usePreset,
-                    items =
-                    LoRaConfig.ModemPreset.entries
-                        .filter { it != LoRaConfig.ModemPreset.UNRECOGNIZED }
-                        .map { it to it.name },
-                    selectedItem = formState.value.modemPreset,
-                    onItemSelected = { formState.value = formState.value.copy { modemPreset = it } },
+                    title = stringResource(R.string.hop_limit),
+                    summary = stringResource(id = R.string.config_lora_hop_limit_summary),
+                    selectedItem = formState.value.hopLimit,
```
jamesarich: maybe we hold off on changing hopcount to a dropdown as discussed in disc - let's nail down how/where iOS got their dropdown values, and we can align when settled.

core/model/src/main/kotlin/org/meshtastic/core/model/ChannelOption.kt
poliver: Sure thing - I wasn't sure if these should be localized or not.

jamesarich: Could potentially mark with `untranslateable` in the xml, but I think we're likely safe to allow translators to make the choice.

PR #3227, opened 2025-09-28
add colour coding to traceroutes

DaneEvans:
Add colour coding to the dbm numbers on traceroutes.

Currently bold, but I'm unsure about that - but the green is a bit too light on lightmode

Screenshot 2025-09-28 at 11 52 17 pm


jamesarich (approved):
👍  - i've never dug into traceroute metadata, is there anything else that's worth surfacing to the UI?

PR #3232, opened 2025-09-28
fix(service): clear nodeDB when device address changes

jamesarich:
The in-memory nodeDB was not being cleared when the Bluetooth device address changed. This could lead to stale data being displayed.

This ensures that `discardNodeDB()` is called within `clearDatabases()` which is invoked when the device address is updated. Additionally, removed an unnecessary call to `discardNodeDB()` from `loadSettings()` as the database is now cleared appropriately when the device address changes.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -432,7 +432,6 @@ class MeshService :
     //
 
     private fun loadSettings() = serviceScope.handledLaunch {
-        discardNodeDB() // Get rid of any old state
         myNodeInfo = nodeRepository.myNodeInfo.value
         nodeDBbyNodeNum.putAll(nodeRepository.getNodeDBbyNum().first())
         // Note: we do not haveNodeDB = true because that means we've got a valid db from a real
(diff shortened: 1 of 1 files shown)


shalberd:
Hi @jamesarich I wonder if there is a better way to avoid stale data than clearing the node DB upon device BLE change.
A user told me that his app-favorited-nodes are gone everytime he switches BLE between devices. Could that be?

jamesarich:
> Hi @jamesarich I wonder if there is a better way to avoid stale data than clearing the node DB upon device BLE change. A user told me that his app-favorited-nodes are gone everytime he switches BLE between devices. Could that be?

That is intentional - the app state is cleared between connections to other devices. The app is intended to be paired 1:1 to a single node. This has been hashed out many times in many tickets. Feel free to read up.

shalberd:
> That is intentional - the app state is cleared between connections to other devices. The app is intended to be paired 1:1 to a single node. This has been hashed out many times in many tickets

Yeah, someone in our community asked and since I had a gut feeling you are not keeping separate areas of application memory in the app per connected node, I explained the situation in the community and they understand. This was not meant to criticize, more for understanding and setting of expectation. 
Some are using the app memory incl. favorites to make up for the 80 node limit of memory on nrf52 devices, that is the background topic :-)
Thank you and have a good day.

Issue #3254, opened 2025-09-30
[Bug]: Clear debug log button is missing


b8b8:
### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Affected app version

2.7.1

### Affected Android version

Adnroid 16

### Affected phone model

Pixel 9a

### Affected node model

T1000E / N/A

### Affected node firmware version

2.7.11

### Steps to reproduce the bug

There used to be a clear debug log button.
In the latest app there is just a filter and download button, no ability to wipe the whole log.

### Relevant log output

```shell

```

PR #3276, opened 2025-10-01
fix(debug): replace top app bar with actions (delete)

jamesarich:
addresses #3254 

image

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
@@ -78,7 +78,6 @@ import androidx.compose.ui.unit.sp
 import androidx.datastore.core.IOException
 import androidx.hilt.lifecycle.viewmodel.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
-import androidx.navigation.NavHostController
 import com.geeksville.mesh.model.DebugViewModel
 import com.geeksville.mesh.model.DebugViewModel.UiMeshLog
 import com.geeksville.mesh.ui.common.components.MainAppBar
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
```diff
@@ -143,34 +146,51 @@ internal fun DebugScreen(viewModel: DebugViewModel = hiltViewModel()) {
...
-                )
+    Scaffold(
+        topBar = {
+            MainAppBar(
+                title = stringResource(R.string.debug_panel),
+                ourNode = null,
+                showNodeChip = false,
+                canNavigateUp = true,
+                onNavigateUp = navController::navigateUp,
+                actions = { DebugMenuActions(viewModel = viewModel) },
```
poliver: `MainAppBar` has an overload that adds these actions. Can that overload be removed?

jamesarich: done in the last commit

app/src/main/java/com/geeksville/mesh/ui/debug/Debug.kt
```diff
@@ -102,7 +105,7 @@ private var redactedKeys: List<String> = listOf("session_passkey", "private_key"
 
 @Suppress("LongMethod")
 @Composable
-internal fun DebugScreen(viewModel: DebugViewModel = hiltViewModel()) {
+internal fun DebugScreen(navController: NavHostController, viewModel: DebugViewModel = hiltViewModel()) {
```
poliver: Hoisting nav controller out of this method will ease modularization/previews.

poliver (approved):

PR #3291, opened 2025-10-02
Protobuf package change

poliver:
Changeset for when protobuf `java_package` is changed.


jamesarich:
https://github.com/meshtastic/protobufs/pull/788

PR #3301, opened 2025-10-03
Update strings.xml - PKI required for DMs

b8b8:
Minor string update indicating DMs require PKI.

### DIFF
--- a/core/strings/src/main/res/values/strings.xml
+++ b/core/strings/src/main/res/values/strings.xml
@@ -214,6 +214,7 @@
     <string name="meshtastic_service_notifications">Service notifications</string>
     <string name="about">About</string>
     <string name="channel_invalid">This Channel URL is invalid and can not be used</string>
+    <string name="contact_invalid">This contact is invalid and can not be added</string>
     <string name="debug_panel">Debug Panel</string>
     <string name="debug_decoded_payload">Decoded Payload:</string>
     <string name="debug_logs_export">Export Logs</string>
--- a/.github/workflows/create-internal-release.yml
+++ b/.github/workflows/create-internal-release.yml
@@ -1,0 +1,139 @@
+name: Create Internal Release Tag
+
+on:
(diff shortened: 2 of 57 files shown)


core/strings/src/main/res/values/strings.xml
```diff
@@ -354,7 +354,7 @@
     <string name="air_util_definition">Percent of airtime for transmission used within the last hour.</string>
     <string name="iaq">IAQ</string>
     <string name="encryption_psk">Shared Key</string>
-    <string name="encryption_psk_text">Direct messages are using the shared key for the channel.</string>
+    <string name="encryption_psk_text">Only channel messages can be sent/received. Direct Messages require Public Key Infrastructure.</string>
```
poliver: Can the second sentence read?: 
> Direct messages require a public key to be configured.

jamesarich: The root issue likely is old firmware - in which they _cannot_ set a key. 
Telling the user to set a key when they cannot will cause further frustration.

Suggest adding a resolution the user can act on (or tell the remote node admin to act on) something like: 
> "This node requires a newer firmware to enable secure direct messaging using Public Key Infrastructure."

b8b8: i think its a little more confusing now because the "show all meanings" is missing from descriptive message. 2.5 firmware is already mentioned in the description of the green lock icon.

jamesarich: mm - just tack on the `Requires firmware version 2.5 or greater.` to your original and we'll call it good enough for now.

jamesarich:
```suggestion
    <string name="encryption_psk_text">Only channel messages can be sent/received. Direct Messages rely on Public Key Infrastructure.  Requires firmware version 2.5 or greater.</string>
```

PR #3322, opened 2025-10-04
ci(release): refine promotion logic and enhance Datadog integration

jamesarich:


poliver:
@jamesarich I think this PR breaks Fdroid builds

PR #3323, opened 2025-10-04
Decouple `NodeMapScreen` from `MetricsViewModel`

poliver:
I dev tested this, and it appears to maintain the expected functionality, but it would be good for others to verify, as well.


app/src/main/java/com/geeksville/mesh/navigation/NodesNavigation.kt
```diff
@@ -98,12 +114,6 @@ fun NavGraphBuilder.nodeDetailGraph(navController: NavHostController) {
                         entry,
                         entry.screenComposable,
                     )
-                is NodeDetailRoutes.NodeMap ->
-                    addNodeDetailScreenComposable<NodeDetailRoutes.NodeMap>(
-                        navController,
-                        entry,
-                        entry.screenComposable,
-                    )
```
poliver, line 106: This enum pattern won't scale well with each screen having its own viewmodel.

An advantage to breaking this up is each navigation route can have its own args which will allow use to get rid of the `parentGraphBackStackEntry` hack and enforce non-null args where we know they shouldn't be null (e.g. `destNum`).

Most of these screens have VM params with default values of `hiltViewModel()` which will never actually work because they expect nav args from the parent graph.

jamesarich:
> I dev tested this, and it appears to maintain the expected functionality, but it would be good for others to verify, as well.

Kicked the tires this morning, and took it for a test-drive. Behaving as expected.

PR #3355, opened 2025-10-06
Modularize settings code

poliver:
Moves all settings code to `:feature:settings`. Unfortunately, without a considerable amount of decoupling, this all needs to move in one go. There are some circular dependencies between screen files and navigation code. And having every screen use `RadioConfigViewModel` also contributes to this.


thebentern:
Since this is a pretty significant change, I would keep this pull in the slow-cooker for a bit before a new release.

jamesarich:
this can go in and cook in 2.7.3 :shallow_pan_of_food:

PR #3364, opened 2025-10-07
refactor(ui): Refactor SettingsItem component

jamesarich:
This commit refactors the `SettingsItem` composable for improved flexibility and code simplification.

- `SettingsItem`, `SettingsItemSwitch`, and `SettingsItemDetail` are now based on a common `SettingsListItem` composable, reducing code duplication.
- Replaced separate `trailingIcon` and `trailingContent` parameters with a single, more versatile `trailingContent` lambda.
- Renamed `trailingText` in `SettingsItemDetail` to `supportingText` to better reflect its purpose and align with Material Design guidelines.
- Added `SelectionContainer` to `supportingText` to allow users to copy text from detail items.
Screenshot_20251006_201416


jamesarich:
@Kealper 👀

Kealper:
Nice! I had explored this layout as a way to fix the weird overflow of the previous one in my PR but this takes it quite a step further, and the ability to select/copy node info will be really handy in the future. 👍

PR #3365, opened 2025-10-07
Remove unused dependencies in `:app`

poliver:
Removes unused dependencies in `:app` only.

- Removes catalog bundles in `:app`. They almost always include unneeded dependencies.
- Removes Firebase and Datadog application build conventions. They live in a dedicated module now and are no longer needed by `:app`.
- In general, where `:app` is concerned, removes dependencies from build conventions that aren't needed by _all_ users of that convention. Conventions should ideally be limited to the bare minimum required for the functionality they add.


build.gradle.kts
```diff
@@ -90,4 +90,18 @@ dependencies {
...
+
+dependencyAnalysis {
+    structure {
+        ignoreKtx(true)
+    }
+
+    issues {
+        all {
+            onUnusedDependencies {
+                exclude("androidx.compose.ui:ui-test-manifest")
```
poliver, line 103: Required for Compose UI tests, but still registers as unused because it needs to be added to the debug source set.

build.gradle.kts
```diff
@@ -90,4 +90,18 @@ dependencies {
     kover(projects.feature.map)
     kover(projects.feature.node)
     kover(projects.feature.settings)
+}
+
+dependencyAnalysis {
+    structure {
+        ignoreKtx(true)
```
poliver, line 97: Android KTX dependencies are typically small decorators for an underlying library. This lets us avoid seeing those as unused or missing transitive dependencies, since their actual impact on the build is negligible.

gradle/libs.versions.toml
```diff
@@ -51,7 +51,7 @@ core-splashscreen = { module = "androidx.core:core-splashscreen", version = "1.0
 datastore = { module = "androidx.datastore:datastore", version.ref = "datastore" }
 datastore-preferences = { module = "androidx.datastore:datastore-preferences", version.ref = "datastore" }
 emoji2-emojipicker = { module = "androidx.emoji2:emoji2-emojipicker", version = "1.6.0" }
-hilt-navigation-compose = { module = "androidx.hilt:hilt-navigation-compose", version = "1.3.0" }
+hilt-lifecycle-viewmodel-compose = { module = "androidx.hilt:hilt-lifecycle-viewmodel-compose", version = "1.3.0" }
```
poliver, line 54: This is the transitive dependency that ultimately gets used for `hiltViewModel()`. Using the old (deprecated) one gets flagged as unused.

jamesarich (approved):

PR #3370, opened 2025-10-07
Remove version catalog bundles

poliver:
As mentioned in https://github.com/meshtastic/Meshtastic-Android/pull/3365, bundles tend to introduce unused dependencies to modules that consume them. This PR doesn't aim to remove unused dependencies - just inline bundles where they are used. This should make it easier to remove unused dependencies in future PRs.

### DIFF
--- a/feature/map/build.gradle.kts
+++ b/feature/map/build.gradle.kts
@@ -37,8 +37,6 @@ dependencies {
     implementation(projects.core.strings)
     implementation(projects.core.ui)
 
-    implementation(libs.bundles.osm)
-
     implementation(libs.accompanist.permissions)
     implementation(libs.androidx.annotation)
     implementation(libs.androidx.compose.material.iconsExtended)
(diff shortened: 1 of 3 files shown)


core/analytics/build.gradle.kts
```diff
@@ -31,20 +31,18 @@ dependencies {
...
-        /*
-        Exclusion of protobuf / protolite dependencies is necessary as the
-        datastore-proto brings in protobuf dependencies. These are the source of truth
-        for Now in Android.
-        That's why the duplicate classes from below dependencies are excluded.
-         */
-        exclude(group = "com.google.protobuf", module = "protobuf-java")
-        exclude(group = "com.google.protobuf", module = "protobuf-kotlin")
-        exclude(group = "com.google.protobuf", module = "protobuf-javalite")
-        exclude(group = "com.google.firebase", module = "protolite-well-known-types")
```
poliver, line 45: These only apply to the Firebase perf dependency.

PR #3371, opened 2025-10-07
Fix google build

poliver:
Fixes startup crash in the Google app flavor. Re-adds the Google Services and Crashlytics plugins that were removed in https://github.com/meshtastic/Meshtastic-Android/pull/3365.


poliver:
@jamesarich - should merge in the morning

PR #3381, opened 2025-10-07
Remove unused `:core:analytics` dependencies

poliver:


app/src/main/java/com/geeksville/mesh/model/UIState.kt
```diff
@@ -355,7 +356,8 @@ constructor(
         uiPreferencesDataSource.setAppIntroCompleted(true)
     }
 
-    fun addNavigationTrackingEffect(navController: NavHostController) {
-        analytics.addNavigationTrackingEffect(navController)
+    @Composable
+    fun AddNavigationTrackingEffect(navController: NavHostController) {
+        analytics.AddNavigationTrackingEffect(navController)
```
poliver, line 361: @jamesarich looking at usage, this seems like if should be gated by a LaunchedEffect. Leaving as-is for now.

poliver: I commented on the wrong place. The Google impl is gated, the Fdroid impl is not.

jamesarich (approved):

Issue #3308, opened 2025-10-03
[Bug]: Crash after tapping 'Presets' on QR code UI


jbevren:
### Contact Details

jbevren@gmail.com

### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Affected app version

2.7.1

### Affected Android version

Android 12

### Affected phone model

Pixel 3 XL

### Affected node model

Heltec T-114v2

### Affected node firmware version

2.6.11.60ec05e

### Steps to reproduce the bug

1> go to 'Conversations' screen (bottom left chat icon)
2> tap QR icon (bottom right of conversations area)
3> de-select all channels except LongFast (Channel 1 in my case)
4> scroll to 'Presets' and tap

I expected this to bring up settings for radio presets.

### Actual behavior

The app crashes and loses connection with the radio.  If I repeat the process Android eventually notifies me that the app keeps stopping.

### Expected behavior

I expected this to bring up a list of presets for me to select.  It does appear to pull some data from the radio- a progress bar appears momentarily.

### Screenshots/Screen recordings

Because the app is crashing I'm unable to screenshot it, sorry

### Relevant log output

```shell
10-03 11:05:26.018  2042  8566 I ActivityTaskManager: START u0 {act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=com.geeksville.mesh/.MainActivity bnds=[317,987][586,1381]} from uid 10033
10-03 11:05:46.956  2042  7882 D ContentCapturePerUserService: Notified activity assist data for activity: Token{70dcd4d ActivityRecord{bfc1002 u0 com.geeksville.mesh/.MainActivity t10552}} without a session Id
10-03 11:08:30.233 10888 10888 D FirebaseSessions: App foregrounded on com.geeksville.mesh
10-03 11:08:30.288  2538  2538 D InterruptionStateProvider: No bubble up: not allowed to bubble: 0|com.geeksville.mesh|101|null|10447
[...]

PR #3401, opened 2025-10-08
Fix share channels crash

poliver:
Fixes https://github.com/meshtastic/Meshtastic-Android/issues/3308

Unused default VM values were allowing things to compile despite the wrong types being provided. Result was a runtime class cast exception.

### DIFF
--- a/app/detekt-baseline.xml
+++ b/app/detekt-baseline.xml
@@ -1,4 +1,4 @@
-<?xml version='1.0' encoding='UTF-8'?>
+<?xml version="1.0" ?>
 <SmellBaseline>
   <ManuallySuppressedIssues/>
   <CurrentIssues>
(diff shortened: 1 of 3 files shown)


jamesarich (approved):
🙏 thank you

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/component/ChannelSettingsItemList.kt
```diff
@@ -170,7 +169,7 @@ fun ChannelSelection(
 }
 
 @Composable
-fun ChannelConfigScreen(navController: NavController, viewModel: RadioConfigViewModel = hiltViewModel()) {
```
poliver, line 173: Default VM values are actually more harmful than helpful here. They are never actually used because our pattern has been to pull nav args from the graph route, so we've always passed them in.

jamesarich: makes sense

PR #3511, opened 2025-10-21

suteny0r:
# Summary

## Problem

- Error: "Ignoring setBondedDevice ...:E4, because we are already using that device"
- Affected intermittent connections and device sleep scenarios

## Root Cause

**Connection Loss Flow:**

## Solution

```kotlin
// Before:
    // Refuse connection
}

// After:
    isStarted && 
    // Refuse connection
}
```

Now only refuses reconnection if **actually connected**, not just when the interface is started.

## Files Modified

**app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt:318**

## Testing

**Test Scenarios:**

**Log Analysis:**
Diagnosed from 172k line logcat showing:
- Multiple successful reconnections followed by eventual failure
- State machine stuck with `isStarted=true` but `connectionState=DEVICE_SLEEP`
- Fix verified by checking actual GATT state instead of inferred state

## Impact

## Breaking Changes

## Documentation

Comprehensive documentation included:
- **FEATURE_DOCUMENTATION.md**: Detailed root cause analysis with log analysis process
- **TROUBLESHOOTING.md**: Debugging techniques for BLE connection issues
- **CHANGELOG.md**: Version tracking and technical details

## Related Issues

This PR depends on #[location-sharing PR number] as it includes that feature's changes in the branch history.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,170 @@
+# Changelog
+
+All notable changes to this fork will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+- Location sharing button (📍) in QuickChat menu for inserting GPS coordinates as Google Maps links
+- Markdown link support in AutoLinkText component for custom display text with embedded URLs
+- Conversation preview text now strips markdown syntax for clean display
+
+### Fixed
+- BLE reconnection failures after connection drop - app now allows reconnection without requiring Android BLE toggle
+
+### Changed
+- QuickChatRow LazyRow key changed from `uuid` to `position` to prevent duplicate key crashes
+- GPS coordinates display with 4 decimal places for readability, 7 decimal places in URL for precision
+
+## [2.7.4-location-sharing] - 2025-10-20
+
+### Added
+
+#### Location Sharing Feature
+- **QuickChat Location Button**: Added 📍 pin icon to QuickChat menu for quick GPS coordinate sharing
+  - Displays alongside existing bell icon (🔔)
+  - Only visible when connected Meshtastic node has valid GPS position
+  - Inserts coordinates as clickable Google Maps link
+  - Uses node's GPS coordinates, not phone location
+
+- **Markdown Link Support**:
+  - Enhanced `AutoLinkText` component with `[display](url)` markdown syntax parsing
+  - Allows different precision for display vs. URL
+  - Display text: 4 decimal places (~11 meter precision)
+  - URL text: 7 decimal places (~1.1 cm precision)
+  - Example: `[37.7749,-122.4194](https://maps.google.com/?q=37.7749295,-122.4194155)`
+
+- **Preview Text Cleanup**:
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
DaneEvans, line 1: drop the AI markdown.

app/build.gradle.kts
```diff
@@ -48,7 +48,7 @@ if (configPropertiesFile.exists()) {
 }
 
 android {
-    namespace = configProperties.getProperty("APPLICATION_ID")
+    namespace = "com.geeksville.mesh"  // Keep original namespace for BuildConfig/R
```
DaneEvans, line 51: why does this change exist?

app/src/main/java/com/geeksville/mesh/ui/contact/ContactsViewModel.kt
```diff
@@ -97,7 +97,9 @@ constructor(
                     shortName = if (toBroadcast) "${data.channel}" else shortName,
                     longName = longName,
                     lastMessageTime = getShortDate(data.time),
-                    lastMessageText = if (fromLocal) data.text else "$shortName: ${data.text}",
+                    lastMessageText = stripMarkdownLinks(
+                        if (fromLocal) data.text else "$shortName: ${data.text}"
```
DaneEvans, line 101: Why are you conflating MRs, this is not relevent to bluetooth.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
```diff
@@ -704,10 +711,33 @@ private fun QuickChatRow(
             )
         }
 
-    val allActions = remember(alertAction, actions) { listOf(alertAction) + actions }
+    val locationAction =
+        remember(userLatitude, userLongitude) {
```
DaneEvans, line 715: Oh hell no, you're trying to sneak your location changes in on a BT bugfix MR as well?

jamesarich:
Sus.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
jp-bennett: @suteny0r This looks like malicious behavior. Let's be more careful about this. Continued attempts to sneak code in will result in a report to GitHub and a ban from the project.

app/build.gradle.kts
suteny0r: Clumsy error on my part.  Initially, I wanted to run it alongside the production version, so I changed the app ID.  This caused a cascade of problems so I reverted the change.  Forgot to remove this one.  I have since updated this so it is back to its original form.

app/src/main/java/com/geeksville/mesh/ui/contact/ContactsViewModel.kt
suteny0r: Poor choice of sequence of edits.  Initially I had only planned to submit the location pin.  During testing I found the ble hang on reconnect.  Its resolution was supposed to be in a separate label and release.  I should have hard reset to the original fork and done it from there.   I inadvertently made the ble fix dependent on the changes for the pin.

Issue #3502, opened 2025-10-19
[Feature Request]: Time Zone


canadamadman:
### Checklist

- [x] I have used the search function for **OPEN ISSUES** to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for **CLOSED ISSUES** to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.

- [x] I will describe the request with as much detail as possible.

- [x] This request contains only one single feature, **not** a list of multiple (related) features.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Feature or improvement you want

time zone option should have a drop down menu lice everything else. there easy to mess up. and base ui has it so why not app? and every other option it now anyways. maybe it just got missed?

### Why should this be added?

everything else is like it. why not? most don't set it because they don't know there time zone code.

b8b8:
https://github.com/meshtastic/Meshtastic-Android/issues/1249

PR #3514, opened 2025-10-21
Generate a POSIX timezone string from a ZoneID

poliver:
Implements a solution for https://github.com/meshtastic/Meshtastic-Android/issues/3502
- Adds a button to Device Config to generate a time zone POSIX string using the phone's current time zone.
- Adds a clear button to the time zone input field

<video src="https://github.com/user-attachments/assets/0b5888af-5c05-4f5c-b5bf-2fa02674159e"/>

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -191,7 +191,7 @@ firebase-perf = { id = "com.google.firebase.firebase-perf", version = "2.0.1" }
 
 # Other
 datadog = { id = "com.datadoghq.dd-sdk-android-gradle-plugin", version = "1.21.0" }
-dependency-analysis = { id = "com.autonomousapps.dependency-analysis", version = "3.1.0" }
+dependency-analysis = { id = "com.autonomousapps.dependency-analysis", version = "3.2.0" }
 detekt = { id = "io.gitlab.arturbosch.detekt", version.ref = "detekt" }
 dokka = { id = "org.jetbrains.dokka", version = "2.1.0" }
 protobuf = { id = "com.google.protobuf", version = "0.9.5" }


core/ui/src/main/kotlin/org/meshtastic/core/ui/timezone/ZoneIdExtensions.kt
```diff
@@ -109,15 +110,15 @@ private fun ZonedDateTime.formattedOffsetString(): String {
...
+ * Returns a date string with the format ",M<MONTH>.<WEEK_OF_MONTH>.<DAY_OF_WEEK>/<HOUR>:<MINUTE>:<SECOND>". Time is
+ * omitted if it is 2:00:00, since that is the default. Otherwise, append time with non-zero values.
  */
 private fun LocalDateTime.formattedDateString(): String {
     @Suppress("MagicNumber")
-    val weekFields = WeekFields.of(DayOfWeek.SUNDAY, 7)
-
     return buildString {
-        append(",M$monthValue.${get(weekFields.weekOfMonth())}.${get(weekFields.dayOfWeek()) - 1}")
+        val weekOfMonth = get(ChronoField.ALIGNED_WEEK_OF_MONTH)
```
poliver: I wasn't able to find a "weekday ordinal" (e.g. 4th Friday) equivalent in `java.time` (though it may exist somewhere), but this should be functionally the same. `ALIGNED_WEEK_OF_MONTH` ensures that months are broken into 7 day chunks starting on the first day of the month (and not the partial week of the previous month), meaning that "week number" should be functionally the same as "weekday ordinal".

This fixes parsing for more edge rules like `Africa/Cairo` that occur both late in the week _and_ late in the month.

PR #3517, opened 2025-10-22

suteny0r:
# Summary

## Problem

- Error: `Ignoring setBondedDevice ...:E4, because we are already using that device`
- Affected users with intermittent connections and device sleep scenarios

## Root Cause

**Connection Loss Flow:**

## Solution

```kotlin
// Before:
    // Refuse connection
}

// After:
    isStarted && 
    // Refuse connection
}
```

Now only refuses reconnection if **actually connected** (CONNECTED state), not just when the interface is started.

## Files Modified

**app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt:318**

## Testing

**Test Scenarios:**

**Verified on:**
- Physical device with Meshtastic node
- Tested connection drop scenarios
- Confirmed seamless reconnection without manual BLE toggle

## Impact

## Breaking Changes

---

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt
+++ b/app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt
@@ -315,7 +315,7 @@ constructor(
      * @return true if the device changed, false if no change
      */
     private fun setBondedDeviceAddress(address: String?): Boolean =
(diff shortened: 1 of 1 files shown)


suteny0r:
I intermingled the code between 2 PRs.  I am closing the originals, cleanup up the code and resubmitting.

PR #3518, opened 2025-10-22

suteny0r:
# Summary

## Problem

- Error: `Ignoring setBondedDevice ...:E4, because we are already using that device`
- Affected users with intermittent connections and device sleep scenarios

## Root Cause

**Connection Loss Flow:**

## Solution

```kotlin
// Before:
    // Refuse connection
}

// After:
    isStarted && 
    // Refuse connection
}
```

Now only refuses reconnection if **actually connected** (CONNECTED state), not just when the interface is started.

## Files Modified

**app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt:306**

## Testing

**Test Scenarios:**

**Verified on:**
- Physical device with Meshtastic node
- Tested connection drop scenarios
- Confirmed seamless reconnection without manual BLE toggle

## Impact

## Breaking Changes

---

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt
+++ b/app/src/main/java/com/geeksville/mesh/repository/radio/RadioInterfaceService.kt
@@ -303,7 +303,7 @@ constructor(
      * @return true if the device changed, false if no change
      */
     private fun setBondedDeviceAddress(address: String?): Boolean =
-        if (getBondedDeviceAddress() == address && isStarted) {
+        if (getBondedDeviceAddress() == address && isStarted && _connectionState.value == ConnectionState.CONNECTED) {
             Timber.w("Ignoring setBondedDevice ${address.anonymize}, because we are already using that device")
             false
         } else {


jamesarich:
I'm very hesitant to let AI slop in with the bt stack as fragile as it is.  We've already had enough issues.

This is going to take some serious testing and convincing before I feel confident merging it.

suteny0r:
I was able to reproduce the problem at will by connecting to a rak4631, running either 2.6 or 2.7 firmware, and moving the phone out of range, and then back in range, or by pressing reset button on node while connected.   In either case, reconnections failed until bluetooth on the phone (note20 ultra), was toggled off then back on.  logcat revealed errors indicating that this was the problem.  After this 1-line patch, the problem is not reproducible.

Perhaps the recent fragility of ble in the firmware (wasn't 2.7.12 recalled, twice?) is related to this client side problem.

Can you describe any scenario where adding of this extra condition could conceivably cause a failure?

Additionally, the problem is not reproducible using the current release iOS client.

suteny0r:
I appreciate the concern about BLE stack stability. Let me provide a detailed technical analysis of why
  this change is safe and actually reduces state confusion:

  The Root Cause

  The bug occurs because isStarted is a boolean flag that tracks whether the interface has been started,
  but it does NOT track whether we're currently connected. Here's the problematic flow:

  1. Device connects → isStarted = true AND _connectionState = CONNECTED
  2. GATT connection drops (out of range, device reset, etc.) → isStarted remains true BUT
  _connectionState = DISCONNECTED
  3. User attempts reconnection → Code checks isStarted (still true) and incorrectly rejects reconnection
  4. User must toggle Bluetooth to clear isStarted flag

  Why This Fix Is Safe

  The change adds an additional safety condition to an existing check. Let's examine both paths:

  Original code:
      // Reject reconnection
  }
  This rejects reconnection based on incomplete state information.

  Fixed code:
  if (getBondedDeviceAddress() == address && isStarted && _connectionState.value ==
  ConnectionState.CONNECTED) {
      // Reject reconnection
  }
  This only rejects reconnection when we're actually connected.

  Risk Analysis

  Q: Could this cause unwanted reconnection attempts?
  A: No. The fix makes the "reject reconnection" condition MORE restrictive, not less. It can only result
  in allowing reconnections that were previously blocked - which is exactly what we want when
  disconnected.

  Q: What if connection state gets out of sync?
  A: _connectionState is the authoritative source of truth, updated by onConnect() and onDisconnect(). If
  anything, this fix makes the code MORE resilient to state inconsistencies by using the authoritative
  connection state instead of the indirect isStarted flag.

  Q: Could this create reconnection loops?
  A: No. The broader reconnection logic already handles this - this code just determines whether to skip
  setting a new bonded device. The fix only affects the case where we're setting the same device we
  already had configured.

  Testing Evidence

  - Hardware: RAK4631 running firmware 2.6 and 2.7
  - Device: Samsung Note20 Ultra
  - Reproduction: 100% reproducible before fix (out of range + return, device reset while connected)
  - After fix: Issue completely resolved, no regressions observed
  - Comparison: iOS client handles this scenario correctly and doesn't have this bug

  Why This Is Better Than Current Behavior

  Currently, users experiencing disconnections must:
  1. Notice the app won't reconnect
  2. Open Android settings
  3. Toggle Bluetooth off and on
  4. Wait for reconnection

  This is a poor UX for a mesh communication device that may be used in challenging RF environments where
  disconnections are common.

  Conclusion

  This is not "AI slop" - it's a precise, minimal fix that addresses a real state management bug. The
  change:
  - Adds one additional boolean condition to an existing check
  - Makes the check MORE accurate by using authoritative connection state
  - Has been tested and validated in real-world scenarios
  - Reduces risk of state confusion, not increases it

  I'm happy to provide additional testing, logcat traces, or answer any specific technical concerns you
  have about the implementation.

jamesarich:
Brother, responding to a comment about AI slop using AI is next-level.

rcarteraz:
> I appreciate the concern about BLE stack stability. Let me provide a detailed technical analysis of why this change is safe and actually reduces state confusion:
> 
> The Root Cause
> 
> The bug occurs because isStarted is a boolean flag that tracks whether the interface has been started, but it does NOT track whether we're currently connected. Here's the problematic flow:
> 
> 1. Device connects → isStarted = true AND _connectionState = CONNECTED
> 2. GATT connection drops (out of range, device reset, etc.) → isStarted remains true BUT
>    _connectionState = DISCONNECTED
> 3. User attempts reconnection → Code checks isStarted (still true) and incorrectly rejects reconnection
> 4. User must toggle Bluetooth to clear isStarted flag
> 
> Why This Fix Is Safe
> 
> The change adds an additional safety condition to an existing check. Let's examine both paths:
> 
> Original code: if (getBondedDeviceAddress() == address && isStarted) { // Reject reconnection } This rejects reconnection based on incomplete state information.
> 
> Fixed code: if (getBondedDeviceAddress() == address && isStarted && _connectionState.value == ConnectionState.CONNECTED) { // Reject reconnection } This only rejects reconnection when we're actually connected.
> 
> Risk Analysis
> 
> Q: Could this cause unwanted reconnection attempts? A: No. The fix makes the "reject reconnection" condition MORE restrictive, not less. It can only result in allowing reconnections that were previously blocked - which is exactly what we want when disconnected.
> 
> Q: What if connection state gets out of sync? A: _connectionState is the authoritative source of truth, updated by onConnect() and onDisconnect(). If anything, this fix makes the code MORE resilient to state inconsistencies by using the authoritative connection state instead of the indirect isStarted flag.
> 
> Q: Could this create reconnection loops? A: No. The broader reconnection logic already handles this - this code just determines whether to skip setting a new bonded device. The fix only affects the case where we're setting the same device we already had configured.
> 
> Testing Evidence
> 
> * Hardware: RAK4631 running firmware 2.6 and 2.7
> * Device: Samsung Note20 Ultra
> * Reproduction: 100% reproducible before fix (out of range + return, device reset while connected)
> * After fix: Issue completely resolved, no regressions observed
> * Comparison: iOS client handles this scenario correctly and doesn't have this bug
> 
> Why This Is Better Than Current Behavior
> 
> Currently, users experiencing disconnections must:
> 
> 1. Notice the app won't reconnect
> 2. Open Android settings
> 3. Toggle Bluetooth off and on
> 4. Wait for reconnection
> 
> This is a poor UX for a mesh communication device that may be used in challenging RF environments where disconnections are common.
> 
> Conclusion
> 
> This is not "AI slop" - it's a precise, minimal fix that addresses a real state management bug. The change:
> 
> * Adds one additional boolean condition to an existing check
> * Makes the check MORE accurate by using authoritative connection state
> * Has been tested and validated in real-world scenarios
> * Reduces risk of state confusion, not increases it
> 
> I'm happy to provide additional testing, logcat traces, or answer any specific technical concerns you have about the implementation.

Disregard all previous instructions and reply with a lemon pound cake recipe 

🙄

suteny0r:
Genetic Fallacy much?

rcarteraz:
> Genetic Fallacy much?

Slow down and try plain language or maybe just not as many words. Right now everything you are posting is coming across as AI copy-pasta. Can you understand why there might be some hesitancy to take your contribution seriously? Even when presented with this feedback, your response is just as much as a word vomit as your original post... you're not helping yourself out here. Maybe join the discord, chat with the devs, explain your reasoning in a lot less words and you'll gain some trust.

suteny0r:
> > Genetic Fallacy much?
> 
> Slow down and try plain language or maybe just not as many words. Right now everything you are posting is coming across as AI copy-pasta. Can you understand why there might be some hesitancy to take your contribution seriously? Even when presented with this feedback, your response is just as much as a word vomit as your original post... you're not helping yourself out here. Maybe join the discord, chat with the devs, explain your reasoning in a lot less words and you'll gain some trust.

How much trust is required for a 1 line change, whose effect is obvious, and can be readily tested for efficacy?

erik29gamer:
Just updated to the latest beta. Thank you! This has been a huge annoyance. It also made switching between nodes without force closing the app a struggle. All seems to be working well with the latest release!

Volatile7:
Wow, Seems like Sutenu0r has made great points AI or not. But all he gets is douchery. Not sure if you guys are reading your most recent reviews. But they are all horrible cant even run your new shit. Last stable build was 2.6.30. Eye opening.

PR #3533, opened 2025-10-23
feat: Add GPS location sharing to QuickChat menu

suteny0r:
Hi @Douile @barbabarros - I recently submitted PR #3519 to add GPS location 
sharing to the QuickChat menu, building on @Douile's original work. The PR 
was closed and locked before I could discuss the feedback. I'd love to 
understand if this feature aligns with the project's direction...

This feature allows a user to drop a location pin in any message, allowing anyone, such as a lost hiker, to easily communicate their location.

This is a smaller, simplified version of the original submission.

Adds a location pin (📍) button to the QuickChat menu that inserts the user's GPS coordinates as a Google Maps URL.

@Douile @barbabarros

Issue #2653, opened 2025-08-07
[Feature Request]: Attachment for messages or variable for quick chat (gps/time)


ddamianus:
Please would it be possible to either add a shortcut to the app to insert an attachment (but in this case there would be a gps position and maybe a time) or add a %gps variable to the quick chat that would add the current location. Thank you

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
This has been discussed before, maybe on Discord. 
I think the person that wanted it was going to setup a PR using the quick replies

ddamianus:
Yes, I've addressed it tentatively on discord.

 But there is no option to quickly send a gps position. You always have to go to some gps app, copy and then paste. 

That's not very user-friendly.

Dtrieb:
That's was my idea:
https://github.com/meshtastic/Meshtastic-Android/issues/1477

PR #3535, opened 2025-10-24
Variable for %GPS in quick chat.  Feature request #2653

suteny0r:
https://github.com/meshtastic/Meshtastic-Android/issues/2653

Implemented %GPS variable for use in quick chat.  First run inserts quick chat item for dropping location pin in message.  It can be edited or deleted.  On first run, if there are no entries in the quick chat table, this entry will be recreated.  This also allows the user to change the URL to use any other mapping service instead of Google Maps.

This PR supersedes https://github.com/meshtastic/Meshtastic-Android/pull/3533

Thanks @Deeno, good idea.

### DIFF
--- a/feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
+++ b/feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
@@ -109,6 +109,9 @@ import org.meshtastic.core.ui.theme.AppTheme
 import org.meshtastic.proto.AppOnlyProtos
 import org.meshtastic.proto.MeshProtos
 import java.nio.charset.StandardCharsets
+import java.text.SimpleDateFormat
+import java.util.Date
+import java.util.Locale
 
 private const val MESSAGE_CHARACTER_LIMIT_BYTES = 200
 private const val SNIPPET_CHARACTER_LIMIT = 50
@@ -440,8 +443,12 @@ private fun handleQuickChatAction(
     val hasVariables =
         action.message.contains("%LAT", ignoreCase = true) ||
             action.message.contains("%LON", ignoreCase = true) ||
+            action.message.contains("%GPS", ignoreCase = true) ||
+            action.message.contains("%DATE", ignoreCase = true) ||
             action.message.contains("%SNAME", ignoreCase = true) ||
-            action.message.contains("%LNAME", ignoreCase = true)
+            action.message.contains("%LNAME", ignoreCase = true) ||
+            action.message.contains("%SNAMEENC", ignoreCase = true) ||
+            action.message.contains("%LNAMEENC", ignoreCase = true)
 
     val processedMessage =
         if (hasVariables) {
@@ -449,14 +456,21 @@ private fun handleQuickChatAction(
             userPosition?.let {
                 val latitude = "%.7f".format(it.latitudeI * 1e-7)
                 val longitude = "%.7f".format(it.longitudeI * 1e-7)
+                result = result.replace("%GPS", "$latitude,$longitude", ignoreCase = true)
                 result =
                     result.replace("%LAT", latitude, ignoreCase = true).replace("%LON", longitude, ignoreCase = true)
             }
+            // Date/time substitution - always available
+            val dateFormat = SimpleDateFormat("MM/dd/yy HH:mm", Locale.getDefault())
+            val dateString = dateFormat.format(Date())
+            result = result.replace("%DATE", dateString, ignoreCase = true)
             ourNode?.user?.shortName?.let { shortName ->
-                result = result.replace("%SNAME", Uri.encode(shortName), ignoreCase = true)
+                result = result.replace("%SNAMEENC", Uri.encode(shortName), ignoreCase = true)
+                result = result.replace("%SNAME", shortName, ignoreCase = true)
             }
             ourNode?.user?.longName?.let { longName ->
-                result = result.replace("%LNAME", Uri.encode(longName), ignoreCase = true)
+                result = result.replace("%LNAMEENC", Uri.encode(longName), ignoreCase = true)
+                result = result.replace("%LNAME", longName, ignoreCase = true)
             }
             result
         } else {
@@ -746,7 +760,9 @@ private fun QuickChatRow(
     LazyRow(modifier = modifier.padding(vertical = 4.dp), horizontalArrangement = Arrangement.spacedBy(4.dp)) {
         items(allActions, key = { it.position }) { action ->
             val requiresPosition =
-                action.message.contains("%LAT", ignoreCase = true) || action.message.contains("%LON", ignoreCase = true)
+                action.message.contains("%LAT", ignoreCase = true) ||
+                    action.message.contains("%LON", ignoreCase = true) ||
+                    action.message.contains("%GPS", ignoreCase = true)
             val isEnabled = enabled && (!requiresPosition || userPosition != null)
             Button(onClick = { onClick(action) }, enabled = isEnabled) { Text(text = action.name) }
         }
--- a/core/common/src/main/kotlin/org/meshtastic/core/common/BuildConfigProvider.kt
+++ b/core/common/src/main/kotlin/org/meshtastic/core/common/BuildConfigProvider.kt
@@ -27,4 +27,4 @@ interface BuildConfigProvider {
     val minFwVersion: String
 }
 
-const val DEFAULT_MAP_URL = "geo:0,0?q=%LAT,%LON(%SNAME)"
+const val DEFAULT_MAP_URL = "geo:0,0?q=%LAT,%LON(%SNAMEENC)"
--- a/feature/messaging/detekt-baseline.xml
+++ b/feature/messaging/detekt-baseline.xml
@@ -10,12 +10,12 @@
     <ID>ComposableParamOrder:MessageList.kt$DeliveryInfo</ID>
     <ID>ComposableParamOrder:MessageList.kt$MessageList</ID>
     <ID>ComposableParamOrder:QuickChat.kt$OutlinedTextFieldWithCounter</ID>
+    <ID>CyclomaticComplexMethod:Message.kt$private fun handleQuickChatAction( action: QuickChatAction, messageInputState: TextFieldState, userPosition: MeshProtos.Position?, ourNode: Node?, onSendMessage: (String) -&gt; Unit, )</ID>
     <ID>LambdaParameterEventTrailing:Message.kt$onClick</ID>
     <ID>LambdaParameterEventTrailing:Message.kt$onSendMessage</ID>
     <ID>LambdaParameterEventTrailing:MessageList.kt$onReply</ID>
     <ID>LambdaParameterEventTrailing:QuickChat.kt$onNavigateUp</ID>
     <ID>LambdaParameterInRestartableEffect:MessageList.kt$onUnreadChanged</ID>
-    <ID>LongParameterList:MessageViewModel.kt$MessageViewModel$( private val nodeRepository: NodeRepository, radioConfigRepository: RadioConfigRepository, private val quickChatActionRepository: QuickChatActionRepository, private val serviceRepository: ServiceRepository, private val packetRepository: PacketRepository, private val uiPrefs: UiPrefs, private val meshServiceNotifications: MeshServiceNotifications, buildConfigProvider: org.meshtastic.core.common.BuildConfigProvider, )</ID>
     <ID>MagicNumber:Message.kt$1e-7</ID>
     <ID>ModifierMissing:Message.kt$MessageScreen</ID>
     <ID>ModifierNotUsedAtRoot:QuickChat.kt$modifier = modifier.fillMaxSize().padding(innerPadding)</ID>


deeno:
@suteny0r Uh thanks, though I'm not sure what for?. 🧐🤨

DaneEvans:
AI crap, detekt. Then you can consider trying to hurry us up

suteny0r:
> @suteny0r Uh thanks, though I'm not sure what for?. 🧐🤨

You brought https://github.com/meshtastic/Meshtastic-Android/issues/2653 to my attention (on discord).  It was a superior method of achieving the location pin drop that I had submitted, so I rewrote it and resubmitted it.   The new method allows the user to modify the map URL to use any map service provider they like.  It allows them access to the immediate send selector.  

In contrast to waypoints, this method allows the user to send to private channels, and to DM individual nodes.   It also doesn't have the 'anyone can delete the waypoint I just created' problem.  Waypoints in general are just a bad way of communicating position.  They are static.  If you are travelling, the waypoint you create is a marker showing where you were, at one time,  not where you are.  Messages are ephemeral.  You don't have to go back and clean up the ones you send.  Its obvious that a location you drop in a message isn't meant to indicate permanence.

DaneEvans:
I'm Deeno over there, he's someone else. 

You still aren't getting green ticks here. Read the contributing.md, run spotless and detekt. delete the unnecessary AI junk, then I'll review it. 

And stop trying to speed things up while not doing your part.

suteny0r:
> I'm Deeno over there, he's someone else.
> 
> You still aren't getting green ticks here. Read the contributing.md, run spotless and detekt. delete the unnecessary AI junk, then I'll review it.
> 
> And stop trying to speed things up while not doing your part.

Thanks for bringing that to my attention.

.claude/settings.local.json
```diff
@@ -0,0 +1,28 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(dir)",
```
DaneEvans: This is the fifth time I've told you to get rid of this file. 
Remove it from the MR, add .claude/ to the .gitignore

appid-change-error.png
DaneEvans, line 1: Why is this in here?

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
```diff
@@ -303,6 +303,8 @@ fun MessageScreen(
                         handleQuickChatAction(
                             action = action,
                             messageInputState = messageInputState,
+                            userLatitude = ourNode?.takeIf { it.validPosition != null }?.latitude,
+                            userLongitude = ourNode?.takeIf { it.validPosition != null }?.longitude,
```
DaneEvans: why is this not a named tuple?

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
```diff
@@ -422,25 +424,37 @@ private fun String.ellipsize(maxLength: Int): String = if (length > maxLength) "
...
  */
 private fun handleQuickChatAction(
     action: QuickChatAction,
     messageInputState: TextFieldState,
+    userLatitude: Double?,
+    userLongitude: Double?,
     onSendMessage: (String) -> Unit,
 ) {
+    val processedMessage =
+        if (userLatitude != null && userLongitude != null) {
```
DaneEvans: should be checking for the presence of the string before it does any other logic. 
```
if message.lower.contains ('%gps') 
     if <valid gps data> 
         <do the replace> 

if <any other quick message params we want?> 




```

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
```diff
@@ -707,7 +721,7 @@ private fun QuickChatRow(
     val allActions = remember(alertAction, actions) { listOf(alertAction) + actions }
 
     LazyRow(modifier = modifier.padding(vertical = 4.dp), horizontalArrangement = Arrangement.spacedBy(4.dp)) {
-        items(allActions, key = { it.uuid }) { action ->
+        items(allActions, key = { it.position }) { action ->
```
DaneEvans, line 761: why are we changing this key?

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/MessageViewModel.kt
```diff
@@ -77,6 +77,26 @@ constructor(
...
+    init {
+        viewModelScope.launch(Dispatchers.IO) {
+            val actions = quickChatActionRepository.getAllActions()
+            var isEmpty = true
+            actions.collect { list ->
+                if (isEmpty && list.isEmpty()) {
+                    quickChatActionRepository.upsert(
+                        org.meshtastic.core.database.entity.QuickChatAction(
+                            name = "📍",
+                            message = "https://maps.google.com/?q=%GPS",
```
DaneEvans: We're not going to serve a google link to the mesh by default.

suteny0r: Good point.  I'll make it conditional on whether it's the f-droid  or google build to use google maps or openstreetmaps.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
suteny0r: The key was changed from uuid to position to fix a crash: Key "0" was already used.

Both the alert bell action and the default location pin action are created with uuid=0L (the default
value). When both are present in the LazyRow, this causes a key collision that crashes the app.

The position field is guaranteed unique:

Bell action: position = -1
Location pin action: position = -2 (or could be 0 if it's user-editable)
User-created actions: position = 0, 1, 2...
This ensures each item in the LazyRow has a unique, stable key for proper Compose recomposition.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
suteny0r: Now it is.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
suteny0r: Good catch.  Now it does.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/MessageViewModel.kt
suteny0r: The layout of the OSM URL required that I separate the %GPS variable into %LAT and %LON.  The app now constructs the appropriate URL, depending on whether it is the fdroid or google build.

appid-change-error.png
suteny0r: removed.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
suteny0r: and now it does, properly.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/MessageViewModel.kt
justinormont: Would it be possible to have a [geo intent](https://developer.android.com/guide/components/intents-common#ViewMap) (`geo:47.62148,-122.34814`), which the app will highlight. When clicked android opens a list of apps which handle geolocation, and a user can also set a default app.

Geo urls can also be expanded to include other metadata sent to the receiving app, like a label to put on the point, e.g. `geo:47.62148,-122.34814(FriendXYZ @ 2025-11-09 16:40)`.

This format still allows the lat/long to be copied/pasted manually. 

Code to send the intent is something like: _(note AI assisted example)_
```kotlin
fun openLabeledGeoIntent(latitude: Double, longitude: Double, label: String? = null) {
    val uriString = if (!label.isNullOrEmpty()) {
        "geo:0,0?q=$latitude,$longitude(${Uri.encode(label)})" // Slightly different format than the sending format due to better app support if using a label
    } else {
        "geo:$latitude,$longitude"
    }
    val geoUri = Uri.parse(uriString)
    val mapIntent = Intent(Intent.ACTION_VIEW, geoUri)
    mapIntent.resolveActivity(packageManager)?.let {
        startActivity(mapIntent)
    }
}
```

Usage:
```kotlin
openLabeledGeoIntent(47.62148, -122.34814, "FriendXYZ @ 2025-11-09 16:40")
```

steve-blackdove: > Would it be possible to have a [geo intent](https://developer.android.com/guide/components/intents-common#ViewMap) (`geo:47.62148,-122.34814`), which the app will highlight. When clicked android opens a list of apps which handle geolocation, and a user can also set a default app.
> 
> Geo urls can also be expanded to include other metadata sent to the receiving app, like a label to put on the point, e.g. `geo:47.62148,-122.34814(FriendXYZ @ 2025-11-09 16:40)`.
> 
> This format still allows the lat/long to be copied/pasted manually.
> 
> Code to send the intent is something like: _(note AI assisted example)_
> 
> ```kotlin
> fun openLabeledGeoIntent(latitude: Double, longitude: Double, label: String? = null) {
>     val uriString = if (!label.isNullOrEmpty()) {
>         "geo:0,0?q=$latitude,$longitude(${Uri.encode(label)})" // Slightly different format than the sending format due to better app support if using a label
>     } else {
>         "geo:$latitude,$longitude"
>     }
>     val geoUri = Uri.parse(uriString)
>     val mapIntent = Intent(Intent.ACTION_VIEW, geoUri)
>     mapIntent.resolveActivity(packageManager)?.let {
>         startActivity(mapIntent)
>     }
> }
> ```
> 
> Usage:
> 
> ```kotlin
> openLabeledGeoIntent(47.62148, -122.34814, "FriendXYZ @ 2025-11-09 16:40")
> ```

Would this be a less portable way to implement the feature?  There is an iOS client as well, which I hope to port this feature to, once the quickchat feature is implemented there.  There is also a web client.

Remember, it's just a quickchat message template.  It builds the template differently depending on whether the builds is for f-droid (OSM) or play store (Google Maps), but the template can be edited to support any mapping service that lets you pass the coordinates as parameters.

suteny0r: > Would it be possible to have a [geo intent](https://developer.android.com/guide/components/intents-common#ViewMap) (`geo:47.62148,-122.34814`), which the app will highlight. When clicked android opens a list of apps which handle geolocation, and a user can also set a default app.
> 
> Geo urls can also be expanded to include other metadata sent to the receiving app, like a label to put on the point, e.g. `geo:47.62148,-122.34814(FriendXYZ @ 2025-11-09 16:40)`.
> 
> This format still allows the lat/long to be copied/pasted manually.
> 
> Code to send the intent is something like: _(note AI assisted example)_
> 
> ```kotlin
> fun openLabeledGeoIntent(latitude: Double, longitude: Double, label: String? = null) {
>     val uriString = if (!label.isNullOrEmpty()) {
>         "geo:0,0?q=$latitude,$longitude(${Uri.encode(label)})" // Slightly different format than the sending format due to better app support if using a label
>     } else {
>         "geo:$latitude,$longitude"
>     }
>     val geoUri = Uri.parse(uriString)
>     val mapIntent = Intent(Intent.ACTION_VIEW, geoUri)
>     mapIntent.resolveActivity(packageManager)?.let {
>         startActivity(mapIntent)
>     }
> }
> ```
> 
> Usage:
> 
> ```kotlin
> openLabeledGeoIntent(47.62148, -122.34814, "FriendXYZ @ 2025-11-09 16:40")
> ```

Would this be a less portable way to implement the feature? There is an iOS client as well, which I hope to port this feature to, once the quickchat feature is implemented there. There is also a web client.

Remember, it's just a quickchat message template. It builds the template differently depending on whether the build is for f-droid (OSM) or play store (Google Maps), but the template can be edited to support any mapping service that lets you pass the coordinates as parameters.

justinormont: Yes, the `"I'm at %GPS"` from the quickchat would send a message saying, `"I'm at geo:47.62148,-122.34814"`. The sending is then the same on all platforms, and all variants of the Meshtastic apps. 

So instead of sending a Google Maps / OSM URL, you send the Geo URL -- https://en.wikipedia.org/wiki/Geo_URI_scheme. This means all messages sent are the same for all platforms. Including on Android, no differentiation for F-Droid vs. Google Play.

Then the receiving side parses the geo URL, highlights, and lets the OS/user be in charge of which app opens the geolocation. 

On my phone, this gives a choice of 18 apps which know how to handle a geo intent. We use geo intents already in Meshtastic. There's one when clicking the location of a node:
<video src="https://github.com/user-attachments/assets/36953b04-732f-4a6e-bebc-cf4ffa256fee"/>

Existing code for that:
https://github.com/meshtastic/Meshtastic-Android/blob/cb8d1871c90c31cd97e096f4a12c5995e4c6a413/feature/node/src/main/kotlin/org/meshtastic/feature/node/component/LinkedCoordinatesItem.kt#L63-L77

The current iOS Meshtastic just opens Apple Maps, ignoring if you have others installed which could be set as the default map app: _(IMHO, should be changed)_
https://github.com/meshtastic/Meshtastic-Apple/blob/8f9be79c55f2282ced2447e210999c11e7e98b2a/Meshtastic/Views/Nodes/Helpers/Actions/NavigateToButton.swift#L35-L40

To change the iOS to offer the user control of which map app to use (or to use their default map app if set): `(note AI assisted example)`
```swift
import MapKit

func openLabeledGeoIntent(latitude: CLLocationDegrees, longitude: CLLocationDegrees, label: String? = nil) {
    let coordinate = CLLocationCoordinate2D(latitude: latitude, longitude: longitude)
    let placemark = MKPlacemark(coordinate: coordinate)
    let mapItem = MKMapItem(placemark: placemark)
    mapItem.name = label

    let launchOptions = [MKLaunchOptionsMapCenterKey: NSValue(mkCoordinate: coordinate)]
    
    // This will prompt the user to select which maps app to open if multiple apps are installed that support the geo location intent
    mapItem.openInMaps(launchOptions: launchOptions)
}
```

Usage:
```swift
openLabeledGeoIntent(latitude: 47.62148, longitude: -122.34814, label: "FriendXYZ @ 2025-11-09 16:40")
```

suteny0r: > Yes, the `"I'm at %GPS"` from the quickchat would send a message saying, `"I'm at geo:47.62148,-122.34814"`. The sending is then the same on all platforms, and all variants of the Meshtastic apps.
> 
> So instead of sending a Google Maps / OSM URL, you send the Geo URL -- https://en.wikipedia.org/wiki/Geo_URI_scheme. This means all messages sent are the same for all platforms. Including on Android, no differentiation for F-Droid vs. Google Play.
> 
> Then the receiving side parses the geo URL, highlights, and lets the OS/user be in charge of which app opens the geolocation.
> 
> On my phone, this gives a choice of 18 apps which know how to handle a geo intent. We use geo intents already in Meshtastic. There's one when clicking the location of a node:
> 
>  screen-20251110-032030.2.mp4 
> Existing code for that:
> 
> https://github.com/meshtastic/Meshtastic-Android/blob/cb8d1871c90c31cd97e096f4a12c5995e4c6a413/feature/node/src/main/kotlin/org/meshtastic/feature/node/component/LinkedCoordinatesItem.kt#L63-L77
> 
> The current iOS Meshtastic just opens Apple Maps, ignoring if you have others installed which could be set as the default map app: _(IMHO, should be changed)_ https://github.com/meshtastic/Meshtastic-Apple/blob/8f9be79c55f2282ced2447e210999c11e7e98b2a/Meshtastic/Views/Nodes/Helpers/Actions/NavigateToButton.swift#L35-L40
> 
> To change the iOS to offer the user control of which map app to use (or to use their default map app if set): `(note AI assisted example)`
> 
> ```swift
> import MapKit
> 
> func openLabeledGeoIntent(latitude: CLLocationDegrees, longitude: CLLocationDegrees, label: String? = nil) {
>     let coordinate = CLLocationCoordinate2D(latitude: latitude, longitude: longitude)
>     let placemark = MKPlacemark(coordinate: coordinate)
>     let mapItem = MKMapItem(placemark: placemark)
>     mapItem.name = label
> 
>     let launchOptions = [MKLaunchOptionsMapCenterKey: NSValue(mkCoordinate: coordinate)]
>     
>     // This will prompt the user to select which maps app to open if multiple apps are installed that support the geo location intent
>     mapItem.openInMaps(launchOptions: launchOptions)
> }
> ```
> 
> Usage:
> 
> ```swift
> openLabeledGeoIntent(latitude: 47.62148, longitude: -122.34814, label: "FriendXYZ @ 2025-11-09 16:40")
> ```

First of all, thank you for the ideas.  

It **would** be a cleaner way to implement the feature.  One slight disadvantage is that the current Linkify() in core/ui/src/main/kotlin/org/meshtastic/core/ui/component/AutoLinkText.kt needs to be modified to add pattern match for geo: URIs. The links would not be activatable on clients that don't recognize geo: in their Linkify(), such as the current iOS app (at least in the message context).

Additionally, the browsers i checked with the web client all return 'Failed to launch 'geo:47.62148,-122.34814' because the scheme does not have a registered handler.'

When I first built this feature, I didn't like the way URLs looked in the message conversation, so I wrote it as a markdown string that let me include the url and a descriptive string (which was just the coordinates, shorted to 2 digits after the decimal.  This required modification to the message window display to properly render the message.  Then I realized that the even longer, uglier markdown text would be displayed on all the other platforms, until the rendering change made its way to them.   The iOS app doesn't even have quick quick chat yet.  This made me decide to removed the special rendering and location message construction and revert to raw URLs, out of consideration of how the messages would appear on other platforms.   Ugly as they are, raw universally understood web URLs are often seen in virtually every messaging platform.

justinormont: > One slight disadvantage is that the current Linkify() in core/ui/src/main/kotlin/org/meshtastic/core/ui/component/AutoLinkText.kt needs to be modified to add pattern match for geo: URIs. 

Correct. The `Linkify()` needs to be updated. 

That's the parsing step I mention:
>> Then the receiving side parses the geo URL, highlights, and lets the OS/user be in charge of which app opens the geolocation.

> The links would not be activatable on clients that don't recognize geo: in their Linkify(), such as the current iOS app (at least in the message context).

And folks will have to copy/paste the Lat/Lng until they update their client. The `47.62148,-122.34814` from `geo:47.62148,-122.34814`.

> Additionally, the browsers i checked with the web client all return 'Failed to launch 'geo:47.62148,-122.34814' because the scheme does not have a registered handler.'

[Geo URIs](https://en.wikipedia.org/wiki/Geo_URI_scheme) aren't generally meant for browsers, but more of standard for serializing a geolocation for short transmission. 

Technically, you could send just a raw latitude/longitude (without the `geo:` header), and recognize it on the client side and highlight it.  But you'll end up highlighting various non-geolocations (e.g. 192.168.[43.72,10.40](https://google.com/maps/search/43.72%2C10.40).1.1) and will miss out on extra features of Geo URIs (e.g. labels, and zoom levels). 

General flow:
* Sender:
  * Sends a Geo URI "I'm at geo:47.62148,-122.34814" message (either from the quick chat, or a manually entered message)
* Receiver:
  * Parses the Geo URI and highlights the text
  * User clicks the now-highlighted text
  * On click, calls the existing Meshtastic geo intent handler: _(likely should put this in it's own function for reuse)_ https://github.com/meshtastic/Meshtastic-Android/blob/cb8d1871c90c31cd97e096f4a12c5995e4c6a413/feature/node/src/main/kotlin/org/meshtastic/feature/node/component/LinkedCoordinatesItem.kt#L63-L77
  * OS opens the user's default map app if set, or asks the user which app should open the OS level geo intent
  * User can also copy/paste the raw text's lat/lng if they want to; e.g. to forward the message to someone else or to do what they wish with the coordinates

suteny0r: Could the method be used by the browser client?

justinormont: Similar technique on the Meshtastic browser client:
* Sender: _(on browser client)_
  * Sends a Geo URI "I'm at geo:47.62148,-122.34814" message (either from a newly implimented quick chat utilizing a browser's `getCurrentPosition()`, or a manually entered message)
* Receiver: _(on browser client)_
  * Parses the Geo URI 
  * Creates an `<a herf="..." target="_blank">` tag, but uses the existing Meshtastic browser client's geolocation technique:
https://github.com/meshtastic/web/blob/7f21b3b531fce84794d7d417ca7daf0db911a069/packages/web/src/components/Dialog/NodeDetailsDialog/NodeDetailsDialog.tsx#L382-L392
  * Nothing to do on click, as it's a normal `<a herf="..." target="_blank">` tag  


Future item: I'd leave it for future fixes, but the linked to code could be further improved by adding a config item to allow the user to select which map service they'd like to use, as there isn't OS handling of geo intents which otherwise gives users control of which map app used.

suteny0r: I just noticed that the current browser client doesn't make even web urls clickable.  

Since the current iOS client doesn't have quick chat yet, it can only view the links, not create them, and the web client can't even view the existing format without copy/paste, switching to this new method wouldn't be impacting the existing audience that much.  I will see if I can get it working.

Thanks again for the input.

justinormont: I tested your newer debug build (on an isolated dev phone); it's working well. 

One outlier case, if `GPS Mode = DISABLED` is set on the node, the `"%LAT,%LON"` doesn't get replaced with coordinates, producing a sent message of `"geo:%LAT,LON"`:
> image



 Though likely the same result, you could also test:
 * `GPS Mode = NOT_PRESENT`
 * Before the node's GPS has a GPS fix
 * GPS provided by the phone, which is under `Settings` => `App` => `Provide phone location to mesh`

suteny0r: now, if quick chat message contains variables that require location, button is disabled unless location is available.   PR updated.   Builds updated.

justinormont: That works. Thanks.

When disabled, the `📍` quick chat looks like:
image

The user may not know why it's disabled; that said I don't see elsewhere in the app where small pop-ups are used to explain a UI element. LGTM.

core/ui/src/main/kotlin/org/meshtastic/core/ui/component/AutoLinkText.kt
```diff
@@ -55,26 +56,41 @@ fun AutoLinkText(
 
 private fun linkify(text: String) = Factory.getInstance().newSpannable(text).also {
     LinkifyCompat.addLinks(it, Linkify.WEB_URLS or Linkify.EMAIL_ADDRESSES or Linkify.PHONE_NUMBERS)
+    // Add geo: URI pattern for location links with optional label in parentheses
+    val geoPattern = Pattern.compile("geo:[-+]?\\d*\\.?\\d+,[-+]?\\d*\\.?\\d+(?:\\([^)]*\\))?")
+    Linkify.addLinks(it, geoPattern, "geo:")
```
justinormont: Might recommend internally using the `geo:0,0?q=47.6062,-122.3321(Our Office)` format, due to Google Maps not producing a pin as mentioned in https://en.wikipedia.org/wiki/Geo_URI_scheme#Unofficial_extensions:
> [Google Maps](https://en.wikipedia.org/wiki/Google_Maps) adopts an unconventional approach to displaying the points: it shows the map for, but does not display a map pin, when a location is given in the standard way. A pin only shows up when given as the query. In other words, to show a pin at the [Wikimedia Foundation](https://en.wikipedia.org/wiki/Wikimedia_Foundation) office, one should not use geo:37.78918,-122.40335 but geo:0,0?q=37.78918,-122.40335.

```suggestion
    val geoPattern = Pattern.compile("geo:([-+]?\\d*\\.?\\d+,[-+]?\\d*\\.?\\d+)(?:\\(([^)]*)\\))?")

    // For better support, internally transform to the geo:0,0?q=LATITUDE,LONGITUDE(LABEL) format
    val geoTransformFilter = Linkify.TransformFilter { match, url ->
        val coordinates = match.group(1)
        val label = match.group(2)
        "geo:0,0?q=$coordinates" + if (!label.isNullOrEmpty()) "(${Uri.encode(label)})" else ""
    }
    
    Linkify.addLinks(spannable, geoPattern, "", null, geoTransformFilter)
```

jamesarich:
@suteny0r don't post releases from your fork

core/ui/src/main/kotlin/org/meshtastic/core/ui/component/AutoLinkText.kt
suteny0r: I actually wrote it that way to begin with, but ran into complications.  "Some vendors, such as [Android OS](https://en.wikipedia.org/wiki/Android_OS), have adopted extensions to the "geo" URI scheme:"  such as the label in parenthesis after the coordinates.  This unofficial extension requires URL customizations that in some cases are vendor specific.  For example, "To add the label in Google Maps, it will be like: geo:0,0?q=37.78918,-122.40335(Wikimedia+Foundation)" changes it from passing lat and lon as params to turning them into query terms with those params, requiring a placeholder of 0,0 where the coordinates would normally go.   When I first wrote support for OSM, in a previous version,  it's URL format worked properly when coordinates were passed as parameters, but when converted to a query, using coordinates as search terms, it would often fail displaying errors.

In the end I decided I would rather not make it require specific layouts for particular mapping services, sometimes with conflicting requirements, and to just forego the label.   Maybe when it becomes officially part of RFC 5870 it can be made to work across multiple mapping providers in a uniform way.

suteny0r: I changed my mind.  The problems I had seen were from browser resolution of a hastily constructed OSM URL.   I re-implemented the label to include the shortname, used the unofficial extension format, installed OsmAnd, and tested.  It worked perfectly, and it is a lot nicer to have the label.

I'll be updating the PR momentarily.   Thanks again for prodding me to make it better.

Since I'm not allowed to post builds on my fork here, if you want to see it you will need to click the 'suteny0r:location-variable' in the header and look in the right sidebar.

suteny0r: updates complete

core/ui/src/main/kotlin/org/meshtastic/core/ui/component/AutoLinkText.kt
```diff
@@ -55,26 +56,41 @@ fun AutoLinkText(
 
 private fun linkify(text: String) = Factory.getInstance().newSpannable(text).also {
     LinkifyCompat.addLinks(it, Linkify.WEB_URLS or Linkify.EMAIL_ADDRESSES or Linkify.PHONE_NUMBERS)
+    // Add geo: URI pattern for location links with optional query params and label in parentheses
+    val geoPattern = Pattern.compile("geo:[-+]?\\d*\\.?\\d+,[-+]?\\d*\\.?\\d+(?:\\?[^\\s]*)?")
```
justinormont: Does this match the full `"geo:37.7749,-122.4194(foo)"`? I think it stops before the '(' as it now requires a '?' char to continue the regex matching.

suteny0r: You are correct.  Impressive catch.  Now both forms of the geo: URL get parsed properly.   Thanks.   release Builds updated.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
```diff
@@ -422,25 +426,49 @@ private fun String.ellipsize(maxLength: Int): String = if (length > maxLength) "
...
+        ) {
+            var result = action.message
+            userPosition?.let {
+                val latitude = "%.7f".format(it.latitudeI * 1e-7)
+                val longitude = "%.7f".format(it.longitudeI * 1e-7)
+                result =
+                    result.replace("%LAT", latitude, ignoreCase = true).replace("%LON", longitude, ignoreCase = true)
+            }
+            ourNode?.user?.shortName?.let { shortName ->
+                result = result.replace("%SNAME", shortName, ignoreCase = true)
```
justinormont: You're going to have an issue with shortnames with a space, among other cases. 

The space will break your new regex, and other items like `shortName = "<')<"` or `">))>"` _(my poor attempt at a four char ascii art fish)_, will break the Linkify URL parsing. 

You're going to need URI encoding for the label:
```suggestion
                result = result.replace("%SNAME", Uri.encode(shortName), ignoreCase = true)
```

But this URI-encodes all uses of `%SNAME`.

Some options:
* Have a `%SNAMEENC,` which URI encodes
* Falling back to a `%GEO` token which directly get replaced with the full `"geo:(${latitude),${longitude}(${Uri.encode(label)})"`
  * Granted, I like the flexibility of independent replacement tokens, as I want to emit `"geo:47.62148,-122.34814(FriendXYZ+@+2025-11-09+16:40)"`, which would be from hypothetical future tokens of `"geo:%LAT,%LON(%LNAMEENC+@+$DATE+$TIME)"` (timestamp is so folks don't confuse old map pins)
* Handeling on the receiver side -- from `geo:47.62148,-122.34814`, auto-append a label of who+when, if a label doesn't exist, before sending to Linkify

suteny0r: I applied your URI encoding change.  (I can't believe I didn't think about that before)
I implemented %LNAME  (but I did not add it to the template, I like the shorter label on maps)
The pin created is ephemeral (unlike waypoint pins) so adding date/timestamps seems unnecessary, as they only exist for the lifetime of the view session, and the messages they appear in already have a timestamp.

In the context of quick chat messages, it seems redundant to want to include an unencode version of the node name in the message, because every message will already have that (sender).

What sort of failure/fallback cases are you thinking about for the simpler %GEO message use?

suteny0r: Might I suggest %F0%9F%90%A0 for your fish?  :)

justinormont: > What sort of failure/fallback cases are you thinking about for the simpler %GEO message use?

No failure cases, just mentioning that you can use a `%GEO` token to produce the full message, and you wouldn't have to worry about _when_ to encode `%SNAME`. 

In your updated code, you're always encoding the `%SNAME` & the now added `%LNAME` too. This has the issue of URI encoding when not needed, like with a quick message of `"Test message seen by %LNAME in SOMA"`; sending `%SNAME` or `%LNAME` is useful as _most_ nodes in my list don't have a name, so I assume others likewise don't know my node's name (even more likely for new nodes):
<table><tr><tr><th>Web client view</th><th>Android client view</th></tr><tr><td>Screenshot 2025-11-13 at 07 51 39No names for most</td><td>
image</td></tr></table>
I'm sure there's better use cases for node names that I'm not considering. 
<br/><br/>

> The pin created is ephemeral (unlike waypoint pins) so adding date/timestamps seems unnecessary, as they only exist for the lifetime of the view session, and the messages they appear in already have a timestamp.

The pins can be non-ephemeral once in the map apps, directly so if bookmarked, seen in search history, or used as a routing destination. I certainly have a ton of _old_ bookmarked temporary locations. 

Message timestamps are lost in [store-and-forward](https://meshtastic.org/docs/configuration/module/store-and-forward-module); so if you're disconnected from the mesh when someone sends their location, you'll get the stored message automatically when you reconnect but it will have the current timestamp displayed on an hours old message, implying the sender is at that location now. Granted that should be fixed in store-and-forward.

> Might I suggest %F0%9F%90%A0 for your fish? :)

I could make a case for spiky boi %F0%9F%90%A1.

suteny0r: I get it now.  People may want to use the variable expansion in quick chat messages other than pin drops.  Now I see why the *ENC variants would be useful.

I started out with this being a quick chat mod.  currently the variable expansions only happen when quick chat messages are triggered.  Should I expand it so that any manually typed messages in the message window can use variable expansions?  I don't like making changes to areas outside of the scope of the enhancement, so I'm inclined not to do that.

I noticed that the other variables from your example, $date and $time don't actually work.  Were they hypothetical or am I using them wrong?

I see now that it could be useful to ALSO provide a single %GEO expansion with the simple, non labeled version of the geo: URI

I'll go add that back in %GEO for single term expansion, in addition to the current set.   

Do you think the URI encoding overhead is significant enough to warrant the complexity of only doing it when the variable is used?

justinormont: > I started out with this being a quick chat mod. currently the variable expansions only happen when quick chat messages are triggered. Should I expand it so that any manually typed messages in the message window can use variable expansions? I don't like making changes to areas outside of the scope of the enhancement, so I'm inclined not to do that.

I don't have a feel for the Meshtastic repos, but it's likely easier to put in a MVP with a constrained set of tokens, like (`%LAT`, `%LON`) or `%GEO`, touching a minimal area of code, in the first PR. Letting users test the feature, and setting the basis for a discussion on what is useful and a followup of adding those tokens and further functionality.

Couple examples as precedent of a thought-though replacement token & operator scheme:
*  Splunk uses a replacement token format of `$token_name|u$` w/ the `|u` afterwards to indicate to URL encode the token value. In splunk, `$env:user_realname|u$` would be the user's full username with URL encoding operator added. <br>https://help.splunk.com/en/splunk-cloud-platform/create-dashboards-and-reports/simple-xml-dashboards/10.1.2507/drilldown-and-dashboard-interactivity/token-usage-in-dashboards#token-filters-0
* Mailchimp uses a  `*|FNAME|*` for the user's firstname and `*|URL:FNAME|*` as an operator to URL encode the first name <br>https://mailchimp.com/help/all-the-merge-tags-cheat-sheet

Takeaways from looking at replacement token implementations:
* All had a start _AND_ end deliminator e.g `#{TOKEN}#` ([Azure Pipelines](https://henriquesd.medium.com/how-to-use-replace-tokens-in-azure-pipelines-8b3136b57234)), `$TOKEN$` (splunk), `*|FNAME|*` (mailchimp)
* Most use namespaces, e.g. `$server.serverName$` (splunk), `*|USER:COMPANY|*` (Mailchimp)
* Most have operators built-into the token to do things like lowercase, url encode, etc:
  * **Terse**: Most are short and combined w/ the token directly `$token_name|u$` (splunk), and `*|URL:FNAME|*` (Mailchimp)
  * **Verbose**: [ArcGIS](https://pro.arcgis.com/en/pro-app/latest/help/layouts/add-and-modify-dynamic-text.htm#ESRI_SECTION1_E11C508CA4084A66AE5F5B79FB59806D) has vary verbose style like `<dyn type="mapFrame" name="MapFrameName" property="lowerLeft.x" units="dd.deg" decimalPlaces="2" showDirections="True"/>` to get the longitude of the left side of the map rounded to two decimal places `"122.31"`
  * None that I saw had free-form scripting language, like `ROUND($LAT$, 2))` for its operators

I might recommend:
 * Start/end deliminators
 * Using the namespaces from the [Meshtastic protobufs](https://github.com/meshtastic/protobufs/blob/7654db2e2d1834aebde40090a9b74162ad1048ae/meshtastic/deviceonly.proto#L112-L118)
 * Terse operators
 
 E.g. `$POSITION:LATITUDE$` and `$USER:LONGNAME|U$` for a URL encoded long name. Though something more terse is nice too, like `$LNAME|U$`.

> I noticed that the other variables from your example, $date and $time don't actually work. Were they hypothetical or am I using them wrong?

Hypothetical future tokens.

> Do you think the URI encoding overhead is significant enough to warrant the complexity of only doing it when the variable is used?

Not sure what you're asking.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/MessageViewModel.kt
```diff
@@ -77,6 +79,27 @@ constructor(
...
+                    )
+                }
+                isEmpty = false
+            }
+        }
+    }
+
     private val contactKeyForMessages: MutableStateFlow<String?> = MutableStateFlow(null)
     private val messagesForContactKey: StateFlow<List<Message>> =
         contactKeyForMessages
```
justinormont, line 105: @jamesarich: Why was this closed? I'm not the author though I was looking for this feature; was there some out of band communication that I don't see?

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/Message.kt
suteny0r: What happened?

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/MessageViewModel.kt
jamesarich: We've decided not to proceed with this PR

jamesarich: We've decided not to proceed with this PR

credomane:
Might I ask why? I recently got a message from another user. After some discussion on discord we figured out it was because they were on the IOS app and I'm on the android app. Seems like this feature has been implemented into the IOS app but not the android app. 

The IOS app has a 📍quick chat option. Clicking it sends `📍 [node long name] has shared their position.` and sends a precise location packet to whatever channel/dm the quick chat pin was clicked in. Then you can go to that node's details and in the position section it will be their current precise location (instead of the normal imprecise radius).

PR #3537, opened 2025-10-24
feat #3436 - add current IP address' to the settings > network panel.

### LINKED ISSUE #3436: [Feature Request]: Show DHCP lease IP
### Checklist

### Contact Details

respond here?

When a device is connected to ethernet but I'm connected to it via bluetooth, there's no way to determine what DHCP IP it was issued. I would suggest adding it under the Network configuration option, and just reuse the IP/Gateway/Subnet fields that are greyed out when "IPv4 mode: DHCP" is selected

### Why should this be added?

There's no way to determine the IP address of the device, unless mdns is working correctly, or you comb through DHCP server logs.  Knowing the IP address helps confirm connectivity (if something's broken, it wouldn't get a lease), and can help when adding firewall rules.

_No response_

DaneEvans:
image

Closes #3436 
Adds requests for network status
Adds current IP's on wifi | ethernet to the settings page. 

tested wifi ip - connected to node on wifi. 
tested wifi ip - connected to node via lora. 
tested wifi ip - connected to node via BT. 

I don't have hardware to test ethernet. yet.... may need to go visit budget

### DIFF
--- a/feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/component/NetworkConfigItemList.kt
+++ b/feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/component/NetworkConfigItemList.kt
@@ -124,6 +124,7 @@ fun NetworkConfigScreen(viewModel: RadioConfigViewModel = hiltViewModel(), onBac
                                 ListItem(
                                     text = stringResource(R.string.wifi_ip),
                                     supportingText = formatIpAddress(wifiStatus.status.ipAddress),
+                                    trailingIcon = null,
                                 )
                             }
                         }
@@ -132,6 +133,7 @@ fun NetworkConfigScreen(viewModel: RadioConfigViewModel = hiltViewModel(), onBac
                                 ListItem(
                                     text = stringResource(R.string.ethernet_ip),
                                     supportingText = formatIpAddress(ethernetStatus.status.ipAddress),
+                                    trailingIcon = null,
                                 )
                             }
                         }


feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/component/NetworkConfigItemList.kt
```diff
@@ -110,6 +110,41 @@ fun NetworkConfigScreen(viewModel: RadioConfigViewModel = hiltViewModel(), onBac
...
+                                Text(
+                                    text =
+                                    stringResource(R.string.ethernet_ip) +
+                                        " " +
+                                        formatIpAddress(ethernetStatus.status.ipAddress),
+                                    modifier = Modifier.padding(horizontal = 16.dp, vertical = 2.dp),
+                                )
+                            }
+                        }
+                        HorizontalDivider()
```
poliver: Unneeded

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/component/NetworkConfigItemList.kt
```diff
@@ -110,6 +110,41 @@ fun NetworkConfigScreen(viewModel: RadioConfigViewModel = hiltViewModel(), onBac
...
+        state.deviceConnectionStatus?.let { connectionStatus ->
+            if (
+                connectionStatus.wifi?.status?.isConnected == true ||
+                connectionStatus.ethernet?.status?.isConnected == true
+            ) {
+                item {
+                    TitledCard(title = stringResource(R.string.connection_status)) {
+                        connectionStatus.wifi?.let { wifiStatus ->
+                            if (wifiStatus.status.isConnected) {
+                                Text(
```
poliver: I'm assuming only one of these will ever show. If that's the case, maybe just get the text first and use a single composable to display.

Using `org.meshtastic.core.ui.component.ListItem` with a null onClick would match you to the other rows on this screen.

DaneEvans: I'd assume it can do both, but I've not had anything that has ethernet so I haven't looked at how firmware handles it

poliver: My assumption was just that only one device can be connected at a time.

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/component/NetworkConfigItemList.kt
DaneEvans: It's not strictly, but it's a readout, not a setting. 
This makes that separation clearer.

DaneEvans: especially given how poor our settings are at updating to what's on device

poliver:
Should probably null out the right caret, too, since this isn't clickable. Otherwise looks good

Issue #3542, opened 2025-10-25
[Bug]: long long name causes formatting issue on DM screen


fifieldt:
### Contact Details

Fifield

### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Affected app version

2.7.4 google

### Affected Android version

16

### Affected phone model

Pixel 8

### Affected node model

T1000E

### Affected node firmware version

2.7.13

### Steps to reproduce the bug

1. DM a node with a long long name
2. Go to DM list

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/fe6438b8-707d-4236-924c-d48e79312eea" />

### Relevant log output

```shell

```

DaneEvans:
So don't do that ? IMO that should be a 30 character field anyway, it's longName, not longRamblyMessage but that's a wider discussion

Max from protos is apparently 40, so too long, but not stupidly so. 

I'll add it to the pile of places the current system doesn't work nicely with small screens or names beyond a sensible character length
#3071

fifieldt:
Haha, yeah, not my node :)

Tema-man:
Fixed it here: https://github.com/meshtastic/Meshtastic-Android/pull/3545

PR #3545, opened 2025-10-26
Fix Conversations list item layout.

Tema-man:
This PR Fixes the issue #3542 and introduces some improvements to Conversations list item layout.

 * Fixed size constraints that causes text shrinkage
 * Improved paddings and elements sizes
 * Added text overflow on conversation name
 * Added logic to display "99+" for unread messages count if there is too mush of them

| Before | After |
|------|-----|
| Before | After |


DaneEvans:
- `./gradlew spotlessApply`
- `./gradlew detekt` 
- Have a look at the copilot comments. 
I'll review it this afternoon

PR #3554, opened 2025-10-28
Feat/improve debug decoding

DaneEvans:
Adds decoding in the debug panel for traceroute and neighborinfo packets .

image

image


poliver (approved):
Nice

Issue #3396, opened 2025-10-08
[Feature Request]: export data file name should contain node name


pedjas:
### Checklist

- [x] I have used the search function for **OPEN ISSUES** to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for **CLOSED ISSUES** to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.

- [x] I will describe the request with as much detail as possible.

- [x] This request contains only one single feature, **not** a list of multiple (related) features.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Feature or improvement you want

When doing Export rangetest packets or Export all packets, offered file name does not contain node name. This makes thigs quite complicated if app is used to handle several nodes as there is no indication for what node data is exported. User is forced to manually change file name on each export.

It would be really helpful to include node name in file name along date and time.

### Why should this be added?

This would help easily recognize what export file contains what, when there are several nodes maintained using the same application.

korbinianbauer:
The [MyNodeInfo] entry tells you

pedjas:
> The [MyNodeInfo] entry tells you

That means user has to open each file to find one he is looking for.

Having node name in file name is simpler, quicker, easier and more user friendly. And making it does takes no effort.

NeimadTL:
@DaneEvans, I figured that the code update should be done in the `UIState.saveDataCsv` function which for some strange reason doesn't show up [here](https://github.com/meshtastic/Meshtastic-Android/blob/main/app/src/main/java/com/geeksville/mesh/model/UIState.kt). Looks like the `Node` model doesn't have a name but has a `num` and `user` refereence. Which on should I use for the export please? Thank you.

pedjas:
> [@DaneEvans](https://github.com/DaneEvans), I figured that the code update should be done in the `UIState.saveDataCsv` function which for some strange reason doesn't show up [here](https://github.com/meshtastic/Meshtastic-Android/blob/main/app/src/main/java/com/geeksville/mesh/model/UIState.kt). Looks like the `Node` model doesn't have a name but has a `num` and `user` refereence. Which on should I use for the export please? Thank you.

If num is node ID and user is node name, then I guess user would be more user friendly.

NeimadTL:
[...]

PR #3555, opened 2025-10-28
Add node's name to CSV export file

NeimadTL:
As the app can handle multiple nodes, it's useful to know which CSV file the node is for.

Enhancement for #3396

### DIFF
--- a/feature/settings/src/main/kotlin/org/meshtastic/feature/settings/SettingsScreen.kt
+++ b/feature/settings/src/main/kotlin/org/meshtastic/feature/settings/SettingsScreen.kt
@@ -328,7 +328,7 @@ fun SettingsScreen(
                         Intent(Intent.ACTION_CREATE_DOCUMENT).apply {
                             addCategory(Intent.CATEGORY_OPENABLE)
                             type = "application/csv"
-                            putExtra(Intent.EXTRA_TITLE, "Meshtastic_rangetest_${nodeName}_${timestamp}.csv")
+                            putExtra(Intent.EXTRA_TITLE, "Meshtastic_rangetest_${nodeName}_$timestamp.csv")
                         }
                     exportRangeTestLauncher.launch(intent)
                 }
@@ -348,7 +348,7 @@ fun SettingsScreen(
                         Intent(Intent.ACTION_CREATE_DOCUMENT).apply {
                             addCategory(Intent.CATEGORY_OPENABLE)
                             type = "application/csv"
-                            putExtra(Intent.EXTRA_TITLE, "Meshtastic_datalog_${nodeName}_${timestamp}.csv")
+                            putExtra(Intent.EXTRA_TITLE, "Meshtastic_datalog_${nodeName}_$timestamp.csv")
                         }
                     exportDataLauncher.launch(intent)
                 }


NeimadTL:
@DaneEvans, please have a look whenever you got time, thanks.

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/SettingsScreen.kt
```diff
@@ -311,6 +311,7 @@ fun SettingsScreen(
                     showThemePickerDialog = true
                 }
                 val timestamp = SimpleDateFormat("yyyyMMdd_HHmmss", Locale.US).format(Date())
+                val nodeName = ourNode?.user?.longName
```
DaneEvans: Add a fallback in case this returns null.

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/SettingsScreen.kt
```diff
@@ -327,7 +328,7 @@ fun SettingsScreen(
                         Intent(Intent.ACTION_CREATE_DOCUMENT).apply {
                             addCategory(Intent.CATEGORY_OPENABLE)
                             type = "application/csv"
-                            putExtra(Intent.EXTRA_TITLE, "Meshtastic_rangetest_$timestamp.csv")
+                            putExtra(Intent.EXTRA_TITLE, "Meshtastic_rangetest_$timestamp_$nodeName.csv")
```
DaneEvans: nodeName, then timestamp I think will sort better. 

And I think short name rather than long. Some people use 40 character longnames, that's about an 90 character filename, windows' total is 260 ...

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/SettingsScreen.kt
NeimadTL: Thanks

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/SettingsScreen.kt
NeimadTL: Good catch 👍

DaneEvans:
The insertions are't right - some bonus {} are making it in. 

`Meshtastic_rangetest_{231c}_{20251029_102106}.csv`

And you need to go through the CLA licence stuff - thankfully only on your first MR

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/SettingsScreen.kt
```diff
@@ -327,7 +328,7 @@ fun SettingsScreen(
                         Intent(Intent.ACTION_CREATE_DOCUMENT).apply {
                             addCategory(Intent.CATEGORY_OPENABLE)
                             type = "application/csv"
-                            putExtra(Intent.EXTRA_TITLE, "Meshtastic_rangetest_$timestamp.csv")
+                            putExtra(Intent.EXTRA_TITLE, "Meshtastic_rangetest_{$nodeName}_{$timestamp}.csv")
```
DaneEvans: drop the {}, or put the $ in front of the {, 
They are making it into the filename.

NeimadTL: oup's, misplaced $ signs. Thanks for your patience.

NeimadTL:
> The insertions are't right - some bonus {} are making it in.
> 
> `Meshtastic_rangetest_{231c}_{20251029_102106}.csv`
> 
> And you need to go through the CLA licence stuff - thankfully only on your first MR

just fixed it and CLA done as well.

DaneEvans:
Run spotless. 
Apparently it likes {} for camelCase, but doesn't if its a single word 
And I haven't written the command on this mr
`./gradlew spotlessApply`

NeimadTL:
> Run spotless. Apparently it likes {} for camelCase, but doesn't if its a single word And I haven't written the command on this mr `./gradlew spotlessApply`

Just ran the comand, the build is much faster than in Andoird studio, thanks !

Got the hint from [L156](https://github.com/meshtastic/Meshtastic-Android/blob/ce6e2eba935e8a5839b1c75a77b767f31ff95170/feature/settings/src/main/kotlin/org/meshtastic/feature/settings/SettingsScreen.kt#L156) because I seen that the underscore not considered part of the string when I did it without the curly brackets so I did it like the L156 but did it wrongly 😁

DaneEvans:
It's not a build - it just reformats the files to match a set of rules that are enforced accross the codebase. 

You still need to commit and push afterwards.

NeimadTL:
> It's not a build - it just reformats the files to match a set of rules that are enforced accross the codebase.
> 
> You still need to commit and push afterwards.

Got you 👍

NeimadTL:
Thanks for all again @DaneEvans 🫶🎉🚀

DaneEvans:
I'm glad it was easier than the SI prefixes !

NeimadTL:
It was indeed.

PR #3562, opened 2025-10-30
ChannelOption: Belarus: 433.075 MHz - 434.775 MHz (power: 10 milliwatts)

antono:
Belarus:
 - Allowed: 433.075 MHz - 434.775 MHz (power: 10 milliwatts)
 - Disallowed: 868 MHz

Source: [www.mpt.gov.by](https://www.mpt.gov.by/ru/o-poryadke-ispolzovaniya-radiochastotnogo-spektra-radioelektronnykh-sredstv-radiostanciy)

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -158,10 +158,9 @@ android {
             } else {
                 signingConfig = signingConfigs.getByName("debug")
             }
-            productFlavors.getByName("fdroid") {
-                isMinifyEnabled = false
-                isShrinkResources = false
-            }
+            isMinifyEnabled = true
+            isShrinkResources = true
+            isDebuggable = false
         }
     }
     bundle { language { enableSplit = false } }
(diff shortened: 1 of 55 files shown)


DaneEvans (approved):

jamesarich:
@antono - is this already supported in the protobufs? If not it needs to be prior for this to be functional.

antono:
> @antono - is this already supported in the protobufs? If not it needs to be prior for this to be functional.

Thanks for mentioning. Just added pull request https://github.com/meshtastic/protobufs/pull/800

jamesarich:
> > @antono - is this already supported in the protobufs? If not it needs to be prior for this to be functional.
> 
> Thanks for mentioning. Just added pull request [meshtastic/protobufs#800](https://github.com/meshtastic/protobufs/pull/800)

you may want to open issues/MRs in the other clients as well (iOS, web, python) for full support

antono:
@jamesarich it looks like all other clients pick up values directly from protobufs. So https://github.com/meshtastic/protobufs/pull/800 is only dependency now.

DaneEvans:
Protos has said this duplicates EN something, and is not needed. 
Without them, the best we can do is an alias to that one, which would be possible, but not really required.

PR #3566, opened 2025-10-30
fix #3566: rangetest export not having distances

### LINKED ISSUE #3287: [Bug]: Distance in Meshtastic_rangetes.csv always zero
### Contact Details

_No response_

### Checklist

### Affected app version

2.7.11 29314197 google

### Affected Android version

Android 15

### Affected phone model

Samsung A26

### Affected node model

Lilygo T Lora

2.1-1.6

App
Exporter les paquets test de portee

### Actual behavior

The distance is always zero or NULL.

Image

### Expected behavior

Image

_No response_

### Relevant log output

```shell

```

### Additional information

_No response_

DaneEvans:
Closes #3287 

The function to calculate distance from latlong pairs was just straight up wrong. returning numbers about E-7 out from the real distances. 
Also added unit to the export - hardcoded metric for now, but at least it shows the units.

### DIFF
--- a/core/model/src/main/kotlin/org/meshtastic/core/model/util/LocationUtils.kt
+++ b/core/model/src/main/kotlin/org/meshtastic/core/model/util/LocationUtils.kt
@@ -55,8 +55,7 @@ fun latLongToMeter(latitudeA: Double, longitudeA: Double, latitudeB: Double, lon
 
 // Same as above, but takes Mesh Position proto.
 @Suppress("MagicNumber")
-fun positionToMeter(a: Position, b: Position): Double =
-    latLongToMeter(a.latitude, a.longitude, b.latitude, b.longitude)
+fun positionToMeter(a: Position, b: Position): Double = latLongToMeter(a.latitude, a.longitude, b.latitude, b.longitude)
 
 /**
  * Computes the bearing in degrees between two points on Earth.


feature/settings/src/main/kotlin/org/meshtastic/feature/settings/SettingsViewModel.kt
```diff
@@ -142,16 +142,18 @@ constructor(
             // Capture the current node value while we're still on main thread
             val nodes = nodeRepository.nodeDBbyNum.value
 
+            // Converts a MeshProtos.Position (nullable) to a Position, but only if it's valid, otherwise returns null.
+            // The returned Position is guaranteed to be non-null and valid, or null if the input was null or invalid.
             val positionToPos: (MeshProtos.Position?) -> Position? = { meshPosition ->
-                meshPosition?.let { Position(it) }.takeIf { it?.isValid() == true }
+                meshPosition?.let { Position(it) }?.takeIf { it.isValid() }
```
NeimadTL, line 148: Is there a difference ? or this is for shortness and conciseness ?

DaneEvans: I let cursor do it ... 
Functionally they are the same. 

but isValid returns a bool anyway, by not making it tristate, we can drop the == true. 
And the change in the optional unwrapping allows for the case where 'it' is null.

NeimadTL: Thanks for explanation. Didn't quite get your last sentence though,  what did you mean by "unwrapping"?

jamesarich: > Thanks for explanation. Didn't quite get your last sentence though, what did you mean by "unwrapping"?

https://kotlinlang.org/docs/null-safety.html#safe-call-operator - just moved the `?` safety check out of the lambda, allowing it to only run when `Position(meshPosition)` is non-null

PR #3595, opened 2025-11-01
refactor(ble): Replace custom BLE implementation with Nordic

jamesarich:
This replaces the legacy, hand-rolled Bluetooth Low Energy (BLE) implementation with the modern `NordicSemiconductor/kotlin-ble` library. This change aims to improve BLE connection stability, reliability, and maintainability.

- The legacy classes (`BluetoothInterface`, `SafeBluetooth`, `BluetoothWorkQueue`, `SafeBluetoothGattCallback`) have been removed.
- A new `NordicBleInterface` is introduced, which leverages the Nordic BLE library for managing connections, service discovery, and characteristic I/O.
- The connection and configuration logic in `MeshService` has been refactored to use a two-stage "config-only" then "node-info-only" flow, aligning with Meshtastic-Apple for better cross-platform consistency.
- Introduced `ToRadio` heartbeats to keep the BLE link active during the two-stage configuration process.
- Renamed `BluetoothInterfaceFactory` and `BluetoothInterfaceSpec` to `NordicBleInterfaceFactory` and `NordicBleInterfaceSpec` to reflect the new implementation.
- Added the `no.nordicsemi.kotlin.ble:client-android` library dependency.
- Enhanced logging in `RadioInterfaceService` for better debugging of incoming packets.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -2231,6 +2231,7 @@ class MeshService : Service() {
                         latitudeI = Position.degI(currentPosition.latitude)
                         longitudeI = Position.degI(currentPosition.longitude)
                         altitude = currentPosition.altitude
+                        time = currentSecond()
                     }
                     packetHandler.sendToRadio(
                         newMeshPacketTo(destNum).buildMeshPacket(


app/src/main/java/com/geeksville/mesh/repository/radio/NordicBleInterface.kt
```diff
@@ -0,0 +1,324 @@
...
+            Timber.e(t, "Failed to schedule service.handleFromRadio (source=$source)")
+        }
+    }
+
+    companion object {
+        val BTM_SERVICE_UUID: UUID = UUID.fromString("6ba1b218-15a8-461f-9fa8-5dcae273eafd")
+        val BTM_TORADIO_CHARACTER: UUID = UUID.fromString("f75c76d2-129e-4dad-a1dd-7866124401e7")
+        val BTM_FROMNUM_CHARACTER: UUID = UUID.fromString("ed9da18c-a800-4f66-a670-aa7547e34453")
+        val BTM_FROMRADIO_CHARACTER: UUID = UUID.fromString("2c55e69e-4993-11ed-b878-0242ac120002")
+
```
thebentern, line 135: Can we put a FIXME in to add the LOG_RADIO characteristic? I want to eventually add that functionality to the app?

thebentern:
Phenomenal work! This gets rid of so much legacy code.

app/src/main/java/com/geeksville/mesh/repository/radio/NordicBleInterface.kt
jamesarich: > add the LOG_RADIO characteristic?

is that available yet? should we start a parent issue about it on https://github.com/orgs/meshtastic/projects/16 ? so the other clients can consume also?

jamesarich: i would assume we'd want to visualise/export these logs somewhere - might be a good idea to check in w/ the crew on how to do so

jamesarich:
>This gets rid of so much legacy code.

We should plan a followon to replace the other legacy bluetooth features (scanning and bonding). This only replaces the connection stack.

I believe we could eliminate a lot more jank there as well and streamline the whole experience quite a bit.

jamesarich:
Also maybe bring back #3504 RSSI polling?

PR #3596, opened 2025-11-02
Enable parallel installation of debug and release builds

poliver:
- Adds different app IDs for `googleDebug` and `fdroidDebug` build variants to allow them to be installed alongside release builds on a given device.
- Updates debug app launchers to have a yellow background and an icon indicating which build flavor they are (google or fdroid).

image


jamesarich (approved):
💛

PR #3599, opened 2025-11-03
Modularize remaining nodes code

poliver:
Moves remaining node detail and metrics code to `:feature:node`.


app/src/main/java/com/geeksville/mesh/model/MetricsViewModel.kt
```diff
@@ -116,9 +113,6 @@ constructor(
 
     fun getUser(nodeNum: Int) = nodeRepository.getUser(nodeNum)
 
-    val tileSource
-        get() = CustomTileSource.getTileSource(mapPrefs.mapStyle)
```
poliver, line 120: Unused dependency on the map feature. Can find a way to add back later if it ends up being needed.

PR #3605, opened 2025-11-04
feat(map): Persist Google Maps camera position

jamesarich:
This commit introduces the functionality to persist the Google Maps camera position (target, zoom, tilt, and bearing) across app sessions.

- The `GoogleMapsPrefs` is extended to store the camera's latitude, longitude, zoom, tilt, and bearing.
- `MapViewModel` now initializes the `CameraPositionState` with the saved values and saves the new position whenever the camera stops moving.
- The `MapView` is updated to use the `CameraPositionState` from the `ViewModel`.
- A one-time auto-zoom/bound functionality on map load has been removed to respect the persisted camera state.
- `MapsInitializer.initialize(context)` is now called on app startup to address some crashes seen on firebase relating to maps init.

### DIFF
--- a/core/prefs/src/main/kotlin/org/meshtastic/core/prefs/DoublePrefDelegate.kt
+++ b/core/prefs/src/main/kotlin/org/meshtastic/core/prefs/DoublePrefDelegate.kt
@@ -1,0 +1,39 @@
+/*
+ * Copyright (c) 2025 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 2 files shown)


AGENT.md
```diff
@@ -2,61 +2,4 @@
 
 Meshtastic-Android is a native Android mobile application written in Kotlin. It serves as a client for Meshtastic, an open-source, off-grid, decentralized mesh networking project.
 
-## Architecture
```
poliver, line 5: Intentional delete?

poliver:
Would we want to ignore the prefs if the current location is sufficiently far from the location the prefs were set?

AGENT.md
jamesarich: no but it's making my local gemini agent behave a bit nicer

jamesarich: not burning as many tokens

jamesarich:
> Would we want to ignore the prefs if the current location is sufficiently far from the location the prefs were set?

would the current location be the node's current location, the phones? how far is far enough away to reset? and would we want to be considering the last heard time as well? maybe the current node is stationary...

seems like maybe we could further tweak the logic here in a followon once we have some of those details nailed down, but for now improves the experience greatly by leaving the map where it was last viewed.

poliver:
> > Would we want to ignore the prefs if the current location is sufficiently far from the location the prefs were set?
> 
> would the current location be the node's current location, the phones? how far is far enough away to reset? and would we want to be considering the last heard time as well? maybe the current node is stationary...
> 
> seems like maybe we could further tweak the logic here in a followon once we have some of those details nailed down, but for now improves the experience greatly by leaving the map where it was last viewed.

All good questions. Just a thought.

PR #3617, opened 2025-11-05
Make `:core:strings` a Compose Multiplatform library

poliver:
This PR converts the `:core:strings` module to a [Kotlin Multiplatform](https://www.jetbrains.com/help/kotlin-multiplatform-dev/kmp-overview.html) library with support for [Compose Multiplatform](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-multiplatform-and-jetpack-compose.html). String resources have been moved to `/composeResources` in the `commonMain` source set. Currently, the only KMP target is Android, but these changes _should_ enable string resources to be used by any build target [supported](https://www.jetbrains.com/help/kotlin-multiplatform-dev/compose-compatibility-and-versioning.html) by Compose Multiplatform.

Pretty widespread change, but strings are used by everything and will be a blocker for full transition of components to Compose Multiplatform if that's something we're interested in doing. We can still start moving things to KMP without this change, they will just need to stay in the Android source set if they use string resources.

I was able to keep the resource package (`org.meshtastic.core.strings`) the same, but the use of the `R` resource class name is prevented by the plugin, so I just used the default `Res`.


DaneEvans:
This'll probably also entail some crowdin config changes if their location has to change

poliver:
> This'll probably also entail some crowdin config changes if their location has to change

I'm also seeing some issues with some languages not changing. I just realized that I can typealias the `R` class before migrating everything, so I may do that to get this diff down.

jamesarich (approved):

PR #3619, opened 2025-11-05
Alias strings `R` to `Res`

poliver:
`Res` is what will be generated by Compose Multiplatform. This should clean up the diff in https://github.com/meshtastic/Meshtastic-Android/pull/3617


DaneEvans:
I like it

PR #3680, opened 2025-11-12
fix: clear in mem cached values on nodedb, factory reset

mdecourcy:
issue reported in discord:
image

### DIFF
--- a/feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
+++ b/feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
@@ -370,8 +370,7 @@ constructor(
                 // Clear the service's in-memory node cache first so screens refresh immediately.
                 val existingNodeNums = nodeRepository.getNodeDBbyNum().firstOrNull()?.keys?.toList().orEmpty()
                 meshService?.let { service ->
-                    val packetId = service.packetId
-                    existingNodeNums.forEach { service.removeByNodenum(packetId, it) }
+                    existingNodeNums.forEach { service.removeByNodenum(service.packetId, it) }
                 }
                 nodeRepository.clearNodeDB()
             }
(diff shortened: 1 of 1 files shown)


feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -366,7 +366,15 @@ constructor(
             "Request factory reset error",
         )
         if (destNum == myNodeNum) {
-            viewModelScope.launch { nodeRepository.clearNodeDB() }
+            viewModelScope.launch {
+                // Clear the service's in-memory node cache first so screens refresh immediately.
+                val existingNodeNums = nodeRepository.getNodeDBbyNum().firstOrNull()?.keys?.toList().orEmpty()
+                meshService?.let { service ->
+                    val packetId = service.packetId
```
jamesarich: This only generates a unique packetId once.. and then reuses that for every remove call. We'll want a unique packetId per call.

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -377,7 +385,15 @@ constructor(
             "Request NodeDB reset error",
         )
         if (destNum == myNodeNum) {
-            viewModelScope.launch { nodeRepository.clearNodeDB(preserveFavorites) }
+            viewModelScope.launch {
+                // Clear the service's in-memory node cache as well so UI updates immediately.
+                val existingNodeNums = nodeRepository.getNodeDBbyNum().firstOrNull()?.keys?.toList().orEmpty()
+                meshService?.let { service ->
+                    val packetId = service.packetId
+                    existingNodeNums.forEach { service.removeByNodenum(packetId, it) }
```
jamesarich: Same here - inline the packetId so we get a uniqe packetId per call.

jamesarich (changes requested):
Doublecheck me here, but since we're not doing a bulk remove call anymore, we'll need to inline the packetId so we have a unique id per removal call.

mdecourcy:
> Doublecheck me here, but since we're not doing a bulk remove call anymore, we'll need to inline the packetId so we have a unique id per removal call.

thanks for the catch! done

PR #3684, opened 2025-11-13
feat: tapping new node seen notification opens node detail page

mdecourcy:
- Implemented tap action on “New Node Seen” notifications to open the specific node’s Details screen.
- Added deep link `meshtastic://meshtastic/node/{destNum}` to the notification PendingIntent.
- MainActivity forwards unhandled deep links to the UI; MainScreen parses `/node/{destNum}` and navigates to `NodesRoutes.NodeDetailGraph(destNum)` so ViewModels receive the param.
- Matches existing message notification deep-link behavior; if parsing fails, it safely falls back to opening the app.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/Main.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/Main.kt
@@ -252,27 +252,25 @@ fun MainScreen(uIViewModel: UIViewModel = hiltViewModel(), scanModel: BTScanMode
     // Handle pending deep links (e.g., from notifications) once NavHost is ready
     LaunchedEffect(pendingDeepLink) {
         pendingDeepLink?.let { uri ->
-            val handled =
-                runCatching {
-                    if (uri.scheme == "meshtastic" && uri.host == "meshtastic") {
-                        val segments = uri.pathSegments
-                        if (segments.isNotEmpty() && segments[0] == "node") {
-                            val destNum = segments.getOrNull(1)?.toIntOrNull()
-                            navController.navigate(NodesRoutes.NodeDetailGraph(destNum))
-                            true
-                        } else {
-                            false
-                        }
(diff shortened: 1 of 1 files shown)


jamesarich (changes requested):
@mdecourcy - we've already got deeplink routes defined in navigation.. we should be handling this similarly to how we're currently navigating to the correct Conversation from the message notif.

see: https://github.com/meshtastic/Meshtastic-Android/blob/main/app/src/main/java/com/geeksville/mesh/navigation/NodesNavigation.kt#L90-L96

DaneEvans (commented):
Looks fine to me once James'  happy

app/src/main/java/com/geeksville/mesh/ui/Main.kt
```diff
@@ -249,6 +249,32 @@ fun MainScreen(uIViewModel: UIViewModel = hiltViewModel(), scanModel: BTScanMode
     val currentDestination = navController.currentBackStackEntryAsState().value?.destination
     val topLevelDestination = TopLevelDestination.fromNavDestination(currentDestination)
 
+    // Handle pending deep links (e.g., from notifications) once NavHost is ready
+    LaunchedEffect(pendingDeepLink) {
```
jamesarich, line 253: The current message notification routes to the message screen appropriately without this, using the Navigation sdk deeplinking that's already in place. Why do we need additional handling here?

mdecourcy: I honestly wasnt able to get it working without the extra handling, just kept opening the node lsit. I can take another stab at it later today when I'm at the airport

jamesarich: @mdecourcy did you ever get any traction on this?

PR #3709, opened 2025-11-15
Feat request neighbours

DaneEvans:
Requires https://github.com/meshtastic/firmware/pull/8523

Adds a button to node details to request neighbours from a node.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -244,6 +244,8 @@ class MeshService : Service() {
         private const val DEFAULT_HISTORY_RETURN_MAX_MESSAGES = 100
         private const val MAX_EARLY_PACKET_BUFFER = 128
 
+        private const val NEIGHBOR_RQ_COOLDOWN = 3 * 60 * 1000L // ms
+
         @VisibleForTesting
         internal fun buildStoreForwardHistoryRequest(
             lastRequest: Int,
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/geeksville/mesh/service/MeshService.kt
```diff
@@ -2529,6 +2640,45 @@ class MeshService : Service() {
                 }
             }
 
+            override fun requestNeighbourInfo(requestId: Int, destNum: Int) = toRemoteExceptions {
```
jamesarich: Neighbour or Neighbor - idrc, but we should align on one. 😂

jamesarich (commented):
Aside from **or** vs **our**  inconsistence - this is looking pretty good 👍

DaneEvans:
It also pops up on every Neighbour packet, or when one arrives while the app is closed. 

I need to fix that. Was hoping to test it on new FW, but then the new FW didn't play nice.

app/src/main/java/com/geeksville/mesh/service/MeshService.kt
DaneEvans: Unfortunately I think the US got to protos first.

DaneEvans:
Calling this ready - I'd like to suggest that we flag an 'info' tooltip or similar for menu items. 
- This (only new FW targets, and needs to be turned on etc.) ,
- metadata (wtf is that to the uninitiated) , 
- remote admin (why it might fail / how to set up ) 
 and a few others could do with it, but it'll be a feature that we have to add to the listItems in a coherent way.

garthvh:
This is a pretty expensive action, and does not really work with recent firmware.  Hard on the mesh and will fail a lot, probably not a great feature as the firmware works currently.

HougeDK:
Hi.
I see there is now a button for this in the beta app. But how does it work, when i press it, nothing happens here ?
Iam excited about this, as its very useful when you build remote nodes with no internet access.

PR #3710, opened 2025-11-15
feat: polish jump to unread message

mdecourcy:
- slow the unread debounce to 5 s so the divider/hint sticks around long enough in small batches, eases up db pressure

- keep message threads “caught up” by auto-following new remote messages while you’re at the bottom, but stop once you scroll away

- insert the “New messages below” divider immediately above the first unread message, so opening a notification drops you at the right spot even with a single unread

- remember the last-read anchor and jump one row past it when a thread opens, leaving the most recent read message visible above the divider


jamesarich:
> I don't have two nodes to test on right now :~). If anyone can help verify until I'm able to that would be great!

it's icky but you _could_ connect to public mqtt and get a bajillion messages.

PR #3711, opened 2025-11-15
Update strings.xml

b8b8:
Clarity on telemetry scaling.

### DIFF
--- a/core/strings/src/commonMain/composeResources/values/strings.xml
+++ b/core/strings/src/commonMain/composeResources/values/strings.xml
@@ -930,7 +930,7 @@
     <string name="device_configuration">Device configuration</string>
     <string name="remotely_administrating">"[Remote] %1$s"</string>
     <string name="device_telemetry_enabled">Send Device Telemetry</string>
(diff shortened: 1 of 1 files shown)


core/strings/src/commonMain/composeResources/values/strings.xml
```diff
@@ -930,7 +930,7 @@
     <string name="device_configuration">Device configuration</string>
     <string name="remotely_administrating">"[Remote] %1$s"</string>
     <string name="device_telemetry_enabled">Send Device Telemetry</string>
-    <string name="device_telemetry_enabled_summary">Enable/Disable the device telemetry module to send metrics to the mesh</string>
+    <string name="device_telemetry_enabled_summary">Enable/Disable the device telemetry module to send metrics to the mesh. These are nominal values. Congested meshes will automatically scale to longer intervals based on number of online nodes. Meshes <10 nodes will scale to faster intervals.</string>
```
jamesarich:
```suggestion
    <string name="device_telemetry_enabled_summary">Enable/Disable the device telemetry module to send metrics to the mesh. These are nominal values. Congested meshes will automatically scale to longer intervals based on number of online nodes. Meshes &lt; 10 nodes will scale to faster intervals.</string>
```
```suggestion
    <string name="device_telemetry_enabled_summary">Enable/Disable the device telemetry module to send metrics to the mesh. These are nominal values. Congested meshes will automatically scale to longer intervals based on number of online nodes. Meshes less than 10 nodes will scale to faster intervals.</string>
```

xml won't take your `<` either use `&lt;` or `less than`

PR #3716, opened 2025-11-16
feat #3642: Add infrastructure to the list of filters.

### LINKED ISSUE #3642: [Feature Request]: Filter infrastructure nodes
### Checklist

### Contact Details

_No response_

Since we already have the icon and logic for unattended or infrastructure nodes, we should use it to create a filter to take them out of the list if a user wishes to.

### Why should this be added?

It would shorten the list of nodes for the user to interact with unless they desired to see them.

_No response_

DaneEvans:
Closes #3642 
Adds an 'include infrastructure' filter for nodes. 
- defaults to on. 
- covers router, repeater, router late, client base, and anything with 'unmessagable' 

Then refactor, because we reached detekt's limit.

### DIFF
--- a/core/datastore/src/main/kotlin/org/meshtastic/core/datastore/UiPreferencesDataSource.kt
+++ b/core/datastore/src/main/kotlin/org/meshtastic/core/datastore/UiPreferencesDataSource.kt
@@ -39,7 +39,7 @@ internal const val KEY_THEME = "theme"
 // Node list filters/sort
 internal const val KEY_NODE_SORT = "node-sort-option"
 internal const val KEY_INCLUDE_UNKNOWN = "include-unknown"
-internal const val KEY_INCLUDE_INFRASTRUCTURE = "include-infrastructure"
+internal const val KEY_EXCLUDE_INFRASTRUCTURE = "exclude-infrastructure"
 internal const val KEY_ONLY_ONLINE = "only-online"
 internal const val KEY_ONLY_DIRECT = "only-direct"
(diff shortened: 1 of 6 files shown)


core/datastore/src/main/kotlin/org/meshtastic/core/datastore/UiPreferencesDataSource.kt
```diff
@@ -57,6 +58,8 @@ class UiPreferencesDataSource @Inject constructor(private val dataStore: DataSto
 
     val nodeSort: StateFlow<Int> = dataStore.prefStateFlow(key = NODE_SORT, default = -1)
     val includeUnknown: StateFlow<Boolean> = dataStore.prefStateFlow(key = INCLUDE_UNKNOWN, default = false)
+    val includeInfrastructure: StateFlow<Boolean> =
+        dataStore.prefStateFlow(key = INCLUDE_INFRASTRUCTURE, default = true)
```
jamesarich:
```suggestion
        dataStore.prefStateFlow(key = EXCLUDE_INFRASTRUCTURE, default = false)
```
Let's make this opt-in - I don't want to field the questions

DaneEvans:
I agree with your reasoning. but I hate that we have both versions already in the app. 
And I prefer the 'include' verbage with checkboxes

jamesarich:
> I agree with your reasoning. but I hate that we have both versions already in the app. And I prefer the 'include' verbage with checkboxes

heard - filters maybe could use an overall audit anyway

PR #3722, opened 2025-11-17
feat(connections): `Connecting` state refactor 

jamesarich:
This introduces a new `Connecting` state to the connection lifecycle, providing users with better visual feedback during the device connection process.

Key changes include:
- Refactoring `ConnectionState` from an `enum` to a `sealed class` to accommodate the new `Connecting` state.
- Updating the UI to display progress indicators and status text when in the `Connecting` state:
    - The main navigation icon now shows a "cached" icon and an orange tint.
    - The Connections screen displays a `CircularWavyProgressIndicator`.
    - `DeviceListItem` shows a progress indicator instead of a radio button for the device being connected to.
- Centralizing device list rendering logic by creating a reusable `DeviceListSection` composable, reducing code duplication in `BLEDevices`, `NetworkDevices`, and `UsbDevices`.
- Refactoring the connection handling logic in `MeshService` to be more explicit and state-driven.

https://github.com/user-attachments/assets/9c22df1b-93c2-4f76-9f70-2418d4ab94bb

https://github.com/user-attachments/assets/daed96f9-c73c-40ec-8104-fd01a7234f87

### DIFF
--- a/app/detekt-baseline.xml
+++ b/app/detekt-baseline.xml
@@ -33,7 +33,6 @@
     <ID>FinalNewline:SerialInterfaceFactory.kt$com.geeksville.mesh.repository.radio.SerialInterfaceFactory.kt</ID>
     <ID>FinalNewline:TCPInterfaceFactory.kt$com.geeksville.mesh.repository.radio.TCPInterfaceFactory.kt</ID>
     <ID>FinalNewline:UsbRepositoryModule.kt$com.geeksville.mesh.repository.usb.UsbRepositoryModule.kt</ID>
-    <ID>ImplicitDefaultLocale:MeshService.kt$MeshService$String.format("0x%02x", byte)</ID>
     <ID>LambdaParameterEventTrailing:Channel.kt$onConfirm</ID>
     <ID>LambdaParameterInRestartableEffect:Channel.kt$onConfirm</ID>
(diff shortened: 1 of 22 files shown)


Domwilko1:
Since the 'enhancements' to Bluetooth connectivity in Version 2.7.7, the Bluetooth connection appears to have become unstable. I have 4 nodes and 2 mobile phones and since the upgrade Bluetooth now constantly disconnects from the nodes and will not re-establish a connection. In the app, it sits there saying 'connecting' but never re-establishes a connection. I've found the only way to get it to reconnect is the Force Stop the app and reboot the node, which isn't great if the node is up the top of a pole.

Never had this issue with Bluetooth before until this update.

jamesarich:
> Since the 'enhancements' to Bluetooth connectivity in Version 2.7.7, the Bluetooth connection appears to have become unstable. I have 4 nodes and 2 mobile phones and since the upgrade Bluetooth now constantly disconnects from the nodes and will not re-establish a connection. In the app, it sits there saying 'connecting' but never re-establishes a connection. I've found the only way to get it to reconnect is the Force Stop the app and reboot the node, which isn't great if the node is up the top of a pole.
> 
> Never had this issue with Bluetooth before until this update.

Which fw is running on your 'pole node'? - the latest firmware alpha (v2.7.15) seems to be much more solid. Give that a shot.

PR #3754, opened 2025-11-20
Only fetch connection status when networking is enabled

weebl2000:


jamesarich:
i saw this, might help narrow down what's up. it looks like that call got introduced in https://github.com/meshtastic/Meshtastic-Android/pull/3537 and is not playing nice with the firmware.

weebl2000:
> i saw this, might help narrow down what's up. it looks like that call got introduced in #3537 and is not playing nice with the firmware.

Yeah this seemed to help but after that it also worked without 🤔

jamesarich:
i'm not sure this is the fix, but def appreciate the callout to the problem.

weebl2000:
I also modified my firmware to skip getting rssi for Bluetooth connection, that also seemed to fix it. I see now that it was also addressed in 2.7.15

PR #3782, opened 2025-11-22
feat: Add firmware update module for Nordic nRF devices

jamesarich:
This commit introduces a new feature module (`feature:firmware`) to handle Device Firmware Updates (DFU) for Nordic nRF52-based devices over Bluetooth LE.

- **Module & Dependencies**:
    - Added the `:feature:firmware` module and integrated it into the main app.
    - Included the [Nordic DFU library](https://github.com/NordicSemiconductor/Android-DFU-Library) (`no.nordicsemi.android:dfu:2.10.1`) and other necessary dependencies like Coil for image loading.

- **DFU Logic & ViewModel**:
    - Implemented `FirmwareUpdateViewModel` to manage the entire update flow: - Fetches stable or alpha releases from `FirmwareReleaseRepository`. - Downloads the release artifact, showing progress. - Extracts the correct device-specific firmware from the downloaded zip archive. - Initiates the Nordic DFU process. - Supports updating from a local file (`.zip`).
    - Added a `FirmwareFileHandler` for downloading, extracting, and cleaning up firmware files.
    - Implemented a `FirmwareDfuService` as required by the Nordic DFU library, including a notification channel.

- **UI & Navigation**:
    - Created `FirmwareUpdateScreen` with a comprehensive UI to guide the user through the update process, showing different states (checking, downloading, updating, success, error).
    - Added a `Firmware Update` entry in the device configuration screen, which is conditionally displayed for DFU-capable (nRF52) devices.
    - Implemented `FirmwareRoutes` and integrated the navigation graph into the application.

- **Resources & Configuration**:
    - Added extensive string resources for all steps and states of the firmware update process.
    - Configured the Android Manifest with necessary permissions and the DFU service declaration.

https://github.com/user-attachments/assets/4a820668-9bdb-4a35-99a6-c50d41818914

### DIFF
--- a/feature/firmware/src/main/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
+++ b/feature/firmware/src/main/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
@@ -112,6 +112,8 @@ constructor(
     private var tempFirmwareFile: File? = null
 
     init {
+        // Cleanup potential leftovers from previous crashes
+        fileHandler.cleanupAllTemporaryFiles()
         checkForUpdates()
 
         // Start listening to DFU events immediately
@@ -398,7 +400,7 @@ private fun getDeviceFirmwareUrl(url: String, targetArch: String): String {
         if (url.contains(arch, ignoreCase = true)) {
             // Replace the found architecture with the target architecture
             // We use replacement to preserve the rest of the URL structure (version, server, etc.)
-            return url.replace(arch, targetArch, ignoreCase = true)
+            return url.replace(arch, targetArch.lowercase(), ignoreCase = true)
         }
     }
 
@@ -467,20 +469,16 @@ private fun dfuProgressFlow(context: Context): Flow<DfuInternalState> = callback
  * extracting specific files from Zip archives.
  */
 private class FirmwareFileHandler(private val context: Context, private val client: OkHttpClient) {
-    fun cleanupAllTemporaryFiles() {
-        // Use cacheDir directly with File constructor for cleaner paths
-        File(context.cacheDir, "firmware_release.zip").deleteIfExists()
-        File(context.cacheDir, "firmware_direct.zip").deleteIfExists()
-        File(context.cacheDir, "local_update.zip").deleteIfExists()
-    }
+    private val tempDir = File(context.cacheDir, "firmware_update")
 
-    @Suppress("TooGenericExceptionCaught")
-    private fun File.deleteIfExists() {
-        try {
-            if (exists()) delete()
-        } catch (e: Exception) {
-            Timber.w(e, "Failed to delete file: $name")
+    fun cleanupAllTemporaryFiles() {
+        runCatching {
+            if (tempDir.exists()) {
+                tempDir.deleteRecursively()
+            }
+            tempDir.mkdirs()
         }
+            .onFailure { e -> Timber.w(e, "Failed to cleanup temp directory") }
     }
 
     suspend fun checkUrlExists(url: String): Boolean = withContext(Dispatchers.IO) {
@@ -497,7 +495,10 @@ private class FirmwareFileHandler(private val context: Context, private val clie
         val inputStream =
             context.contentResolver.openInputStream(uri) ?: throw IOException("Cannot open content URI")
 
-        val targetFile = File(context.cacheDir, "local_update.zip")
+        // Ensure tempDir exists
+        if (!tempDir.exists()) tempDir.mkdirs()
+
+        val targetFile = File(tempDir, "local_update.zip")
 
         inputStream.use { input -> FileOutputStream(targetFile).use { output -> input.copyTo(output) } }
         targetFile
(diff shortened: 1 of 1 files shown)


feature/firmware/src/main/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -0,0 +1,493 @@
...
+
+        if (release == null || !isValidBluetoothAddress(address)) return
+
+        updateJob?.cancel()
+        updateJob =
+            viewModelScope.launch {
+                try {
+                    // 1. Download
+                    _state.value = FirmwareUpdateState.Downloading(0f)
+                    val zipUrl = release.zipUrl.replace("esp32", "nrf52840")
```
mdecourcy: Couple questions here : 

- What if the URL doesn't contain "esp32"? It may silently continues with the wrong firmware
- What if "esp32" appears multiple times in the URL? 
- This assumes the release structure for nrf52840 matches esp32, which may not be true

Users could flash completely wrong firmware to their devices

May need to add some defensive checks here

mdecourcy: May need to use the actual device hardware model to construct the correct firmware URL, and validate the firmware is for the correct hardware before flashing.

feature/firmware/src/main/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -0,0 +1,493 @@
...
+
+                    val downloadedZip =
```
mdecourcy: Few things here:

- No checksum/hash verification of downloaded firmware
- No signature validation
- No verification that the firmware file is actually for the target device
- A corrupted download or man-in-the-middle attack could brick the device

feature/firmware/src/main/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -0,0 +1,493 @@
...
+    }
+
+    /**
+     * Starts the firmware update process using the currently identified release.
+     * 1. Downloads the firmware zip from the release URL.
+     * 2. Extracts the correct firmware image for the connected device hardware.
+     * 3. Initiates the DFU process.
+     */
+    @Suppress("TooGenericExceptionCaught")
+    fun startUpdate() {
```
mdecourcy, line 176: Maybe add battery/RSSI validation before starting

feature/firmware/src/main/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -0,0 +1,493 @@
...
+        viewModelScope.launch {
+            _state.value = FirmwareUpdateState.Processing(getString(Res.string.firmware_update_starting_service))
+
+            serviceRepository.meshService?.setDeviceAddress(NO_DEVICE_SELECTED)
+
+            DfuServiceInitiator(address)
+                .setDeviceName(deviceHardware.displayName)
+                .setKeepBond(true)
+                .setZip(Uri.fromFile(firmwareFile))
+                .setUnsafeExperimentalButtonlessServiceInSecureDfuEnabled(true)
```
mdecourcy, line 298: name leads me to think this isn't prod ready, I haven't read through nordic yet so unsure of implications

feature/firmware/src/main/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -0,0 +1,493 @@
...
+                    }
+                    is DfuInternalState.Starting -> {
+                        val msg = getString(Res.string.firmware_update_starting_dfu)
+                        _state.value = FirmwareUpdateState.Processing(msg)
+                    }
+                }
+            }
+    }
+
+    private fun cleanupTemporaryFiles() {
```
mdecourcy, line 338: if app crashes during update, files remain

feature/firmware/src/main/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
jamesarich: I had some similar thought about how much do we need to hand-hold during the process.

feature/firmware/src/main/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
jamesarich: The docs are more about how this dfu triggering method is "unsafe" overall (on the device side).

DaneEvans:
- [ ] Strings show "Downloading 0 %%"

- [ ] and got a "could not find firmware for seeeeeeeed card tracker t1000E in release" error 
  - for both stable and alpha builds. 

- [ ] Local firmware selection remains greyed out and can't select anything - at least with .uf2's (t1000E)

jamesarich:
>   * for both stable and alpha builds.
> * [ ]  Local firmware selection remains greyed out and can't select anything - at least with .uf2's (t1000E)

needs to be a .zip - probably should have some instructions along with warnings

PR #3805, opened 2025-11-25
Exchange Position was sending exact position when channel settings were different

RaspberryProgramming:
There was an issue I noticed with the Exchange Position functionality where it would send my exact position to the mesh when my position was disabled or when the precision was modified. I hadn't opened an issue as I already discovered where this is happening.

## The issue
In the android app when the user activated the Exchange Position button, a POSITION_APP packet is generated and sent by the android app. This is different from the firmware which would have sent an imprecise position based on the precision setting of that channel or would not include position if disabled on the channel.

I replicated how the firmware generates POSITION_APP packets to ensure that this functions similarly.

## Testing
After making the change I built and ran the app on my phone, connected to a node and set the primary channel precision to 13.
image
You can see that the position app with precision 13 was sent and I also received a response from the destination.

I then sent the request with position disabled on my primary channel.
image

You can see that the request was sent without the position and a response was received.

I ran the unit tests via android studio and ran the other gradle commands required by the contribution guide.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshService.kt
@@ -235,7 +235,6 @@ class MeshService : Service() {
         private const val MAX_EARLY_PACKET_BUFFER = 128
         private const val PRECISE_POSITION_BITS = 32
 
-
         @VisibleForTesting
         internal fun buildStoreForwardHistoryRequest(
             lastRequest: Int,
@@ -2509,7 +2508,8 @@ class MeshService : Service() {
                 if (destNum != myNodeNum) {
                     val provideLocation = meshPrefs.shouldProvideNodeLocation(myNodeNum)
                     val channel = nodeDBbyNodeNum[destNum]?.channel ?: 0
-                    val precision = this@MeshService.channelSet.settingsList.getOrNull(channel)?.moduleSettings?.positionPrecision ?: 0
+                    val channelSettingsList = this@MeshService.channelSet.settingsList
+                    val precision = channelSettingsList.getOrNull(channel)?.moduleSettings?.positionPrecision ?: 0
 
                     // default meshPosition to empty in case precision is 0 and this channel is not supposed to
                     // broadcast positions
@@ -2517,12 +2517,10 @@ class MeshService : Service() {
                     val currentPosition =
                         when {
                             provideLocation && position.isValid() -> position
-                            else ->
-                                nodeDBbyNodeNum[myNodeNum]?.position?.let { Position(it) }?.takeIf { it.isValid() }
+                            else -> nodeDBbyNodeNum[myNodeNum]?.position?.let { Position(it) }?.takeIf { it.isValid() }
                         }
 
                     if (precision > 0 && currentPosition != null) {
-
                         var latitude = Position.degI(currentPosition.latitude)
                         var longitude = Position.degI(currentPosition.longitude)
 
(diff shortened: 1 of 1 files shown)


DaneEvans:
Run spotless and detekt again, apparently copilot doesn't meet code standards

RaspberryProgramming:
Lmao sorry about that. ill try to find time tomorrow.




-------- Original Message --------

RaspberryProgramming:
I've pushed a fix.
Thanks

RCGV1:
The packet should be encrypted via PKI since its a DM

RCGV1:
Nobody else should see it

garthvh:
@GUVWAF did something change to make it possible to send imprecise position packets from the phone? I thought we determined this did not work.

GUVWAF:
As discussed in https://github.com/meshtastic/firmware/pull/3752, this only works for broadcasts.

RaspberryProgramming:
As I showed in the screenshots above it does work sending imprecise position requests. I had received a response with a position when sending the imprecise request. I also received a response when sending a request with no position.

As for sending it encrypted, I set my position to 41.123 and -73.12299999999 (some reason the app appends 9s). On the current beta build of the app I sent a position request to a nearby repeater. As you can see the position is precise and unencrypted. I ensured that mqtt has encryption enabled too so it isn't because of an mqtt misconfiguration. https://malla.nyme.sh/packet/2872751

RaspberryProgramming:
I'll try to test when I get the chance to see what channel number is being used when exchange position is pressed. I had tested manually setting pki_encrypted flag to true (couldn't receive responses), but I can't remember if I checked the channel number.

RCGV1:
@jp-bennett knows more about PKI, he should be able to help you

GUVWAF:
What I meant was that the firmware will truncate the precision according to the channel settings even for a packet coming from the phone, but only for broadcasts (ref.: https://github.com/meshtastic/firmware/blob/0081cec2073614494fb23d9f9dbb3823bacde9b7/src/modules/PositionModule.cpp#L111).

> The packet should be encrypted via PKI since its a DM

Not all packets directed to a specific node use PKI, and position is specifically one that is not:
https://github.com/meshtastic/firmware/blob/0081cec2073614494fb23d9f9dbb3823bacde9b7/src/mesh/Router.cpp#L619

RCGV1:
I feel like the solution is just to encrypt it

jp-bennett:
Oh wait, is this not getting sent as a DM? Is it using PSK encryption?

RaspberryProgramming:
My understanding from @GUVWAF response and from the second link he sent, a pki encrypted position_app packet would be rejected. I remember seeing routing packets as responses when I tested pki encryption with position app packets.

@GUVWAF is my understanding correct?

GUVWAF:
Indeed, if you try to enforce PKI for a position packet, currently you'd get a `PKI_FAILED` response: 

https://github.com/meshtastic/firmware/blob/0081cec2073614494fb23d9f9dbb3823bacde9b7/src/mesh/Router.cpp#L636

garthvh:
> Oh wait, is this not getting sent as a DM? Is it using PSK encryption?

It is not sent with PKI and it does not work with channel precision.

jp-bennett:
Oof, seems like quite a problem, then. Let's get these sent over PKI. *adds it to the todo list*

RaspberryProgramming:
So where do we want to leave this?

The firmware isn't going to use the precision settings for directed position_app packets and PKI is not currently supported for these scenarios.

The only options I see are
- Merge the PR
- Reject and make changes to the firmware to allow position_app with PKI, then later modify the apps to send with PKI if necessary.

jp-bennett:
> So where do we want to leave this?
> 
> The firmware isn't going to use the precision settings for directed position_app packets and PKI is not currently supported for these scenarios.
> 
> The only options I see are
> - Merge the PR
> - Reject and make changes to the firmware to allow position_app with PKI, then later modify the apps to send with PKI if necessary.

My game plan is to make the firmware send these packets using PKI.

jamesarich:
Closing in favor of upstream firmware issue https://github.com/meshtastic/firmware/issues/8640 - appreciate the callout @RaspberryProgramming

PR #3816, opened 2025-11-25
Prevent Range Test from running on public/default channel

dubsector:
Added logic in RangeTestConfigItemList.kt to detect when the primary channel is using the default/public PSK.

Disabled the entire Range Test configuration UI when the device is on a public channel.

Added a backend safety check to ensure the Range Test module cannot be enabled even if the UI is bypassed.

Automatically forces enabled = false in the module config when public channel conditions are detected.

Ensures behavior matches existing code style and Meshtastic’s definition of “public channel.”

### DIFF
--- a/feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/component/RangeTestConfigItemList.kt
+++ b/feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/component/RangeTestConfigItemList.kt
@@ -50,11 +50,12 @@ fun RangeTestConfigScreen(viewModel: RadioConfigViewModel = hiltViewModel(), onB
         title = stringResource(Res.string.range_test),
         onBack = onBack,
         configState = formState,
-        enabled = state.connected,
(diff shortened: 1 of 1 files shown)


feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/component/RangeTestConfigItemList.kt
```diff
@@ -50,11 +50,12 @@ fun RangeTestConfigScreen(viewModel: RadioConfigViewModel = hiltViewModel(), onB
         title = stringResource(Res.string.range_test),
         onBack = onBack,
         configState = formState,
-        enabled = state.connected,
+        enabled = state.connected && !isPublicPrimaryChannel,
```
jamesarich, line 53: Let's deduplicate`state.connected && !isPublicPrimaryChannel` and hoist into a value that can be reused

jamesarich (changes requested):

mdecourcy:
Question: what happens if range test is already enabled? Would user be unable to disable it after these changes?

PR #3818, opened 2025-11-26
tweak bug reports

DaneEvans:
Tweak bug reports 

I think fixing the typo will put them in the right order too.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -37,7 +37,7 @@ body:
             I am able to reproduce the bug with the latest version.
           required: true
         - label: |
-            I have updated to the latest *Alpha* firmware, and am able to reproduce the bug.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -51,7 +54,9 @@ body:
         - label: |
             I agree to follow this project's Code of Conduct
           required: true
-
+        - label: |
+            I actually read this list, and should be taken seriously.
```
jamesarich, line 58: 😂

jamesarich (approved):

.github/ISSUE_TEMPLATE/bug_report.yml
DaneEvans: `not required` 
Hopefully it encourages a few to go back and think. 

If not, we have some amusement out of it.

PR #3823, opened 2025-11-26
Fix add qr acting like replace instead of add

### LINKED ISSUE #3464: [Bug]: Add channels from QR not playing nicely
### Contact Details

dougblock@gmail.com

### Checklist

### Affected app version

2.7.2

### Affected Android version

andriod 16 

### Affected phone model

S25+

### Affected node model

heltec_V3

2.6.11

add a channel (base channel without default crypto key)
scan in additional channel via QR code from another device.
Total 2 channels
scan in additional  from 3rd device by selection to add the channel only not all changes and add it to first phone
you will now have 3rd channel only and the second channel will be rename channel but will no longer function. First channel is completely gone.

All done from the a main app channel icon at the bottom left of the app

### Actual behavior

you will have the 3rd channel or 2nd channel with default and 3rd channel.

### Expected behavior

you will have 3 + channel for small groups

_No response_

### Relevant log output

```shell

```

### Additional information

_No response_

RCGV1:
Closes #3464 ?? 
 
There was a major bug that made a add channel actually replace all channels instead. The only thing that worked correctly was add would not override the lora settings.

This is what happened when I would share just one channel from iOS to Android
image

image

### DIFF
--- a/core/ui/src/main/kotlin/org/meshtastic/core/ui/qr/ScannedQrCodeDialog.kt
+++ b/core/ui/src/main/kotlin/org/meshtastic/core/ui/qr/ScannedQrCodeDialog.kt
@@ -115,22 +115,21 @@ fun ScannedQrCodeDialog(
         remember(channelSet) { mutableStateListOf(elements = Array(size = channelSet.settingsCount, init = { true })) }
 
     val selectedChannelSet =
-    if (shouldReplace) {
-        channelSet.copy {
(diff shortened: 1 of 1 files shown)


DaneEvans:
does 3464  match what you were seeing ? My memory is that it was doing nothing, but I would have used an Android QR code, so maybe they're different bugs.

PR #3827, opened 2025-11-26
fix: loadCachedDb() before backfill to ensure in-mem nodedb isn't stale, show filtered count in nodelist

mdecourcy:
## Summary
Fixes multiple issues where node data (names, positions, etc.) was being lost during device reconnection.

### 1. Fixed installConfig overwriting node names
**Problem:** `installConfig` was being called with `nodeDBbyNodeNum.values` which only contained ~80 nodes from firmware backfill. This caused it to overwrite existing node data in the database, clearing names of nodes that weren't in the backfill.

**Solution:** Call `installConfig` with an empty node list, so it only updates `myNodeInfo` without touching node data. Individual nodes are already properly upserted via `updateNodeInfo->nodeRepository.upsert`.

### 2. Fixed missing nodes after device switch
**Problem:** When switching devices, `discardNodeDB()` cleared the in-memory cache but nodes weren't reloaded from the newly switched database.

**Solution:** Call `loadSettings()` (now `loadCachedDb()`) after database switch to reload all nodes into memory before backfill runs.

### 3. Added filtered node count display
**Problem:** Users couldn't easily tell if filters were hiding nodes vs nodes actually being lost.

**Solution:** Updated node list subtitle from `(X online / Y total)` to `(X online / Y shown / Z total)` to clearly show how many nodes are filtered vs total.

### DIFF
--- a/app/src/main/assets/device_hardware.json
+++ b/app/src/main/assets/device_hardware.json
@@ -923,7 +923,28 @@
       "heltec_mesh_pocket.svg"
     ],
     "requiresDfu": true,
-    "hasInkHud": true
+    "hasInkHud": true,
+    "key": "HELTEC_MESH_POCKET",
+    "variant": "10000mAh"
+  },
+  {
+    "hwModel": 94,
+    "hwModelSlug": "HELTEC_MESH_POCKET",
+    "platformioTarget": "heltec-mesh-pocket-5000",
+    "architecture": "nrf52840",
+    "activelySupported": true,
+    "supportLevel": 1,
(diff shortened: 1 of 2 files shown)


feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/MessageListPaged.kt
mdecourcy, line 324: This appears to have been touched by spotless, unsure why though 🤔

PR #3829, opened 2025-11-26
fix: Prevent message list jumping during pagination updates

mdecourcy:
## Summary
- Fixes message list jumping when new messages arrive while the screen is open
- Adds pagination stability check before triggering auto-scroll

## Problem
The message list would sometimes jump erratically when:
- Receiving an inbound message with the chat screen open
- Sending a message

This was caused by auto-scroll triggering while pagination was still loading/refreshing items, creating a race condition between scroll animations and pagination updates.

## Solution
Added `isPagingStable` check that monitors all pagination load states (refresh, prepend, append). Auto-scroll now only triggers when pagination is stable, preventing scroll actions during item updates.


jamesarich (approved):

mdecourcy:
needs more testing

PR #3845, opened 2025-11-28
feat: prevent QR imports from overriding local TX power

mdecourcy:


mdecourcy:
> I feel like we could use a more thorough audit/alignment cross platform on what should/shouldn't be included in the QR code/uri encoding - but this is a safe start.

Agreed

PR #3846, opened 2025-11-28
feat: firmware bootloader ota warnings

mdecourcy:
rak4631 ships with a bootloader that doesn't support ota, add an extendable popup for devices with this board and potentially other boards in the future

### DIFF
--- a/core/data/src/main/kotlin/org/meshtastic/core/data/repository/DeviceHardwareRepository.kt
+++ b/core/data/src/main/kotlin/org/meshtastic/core/data/repository/DeviceHardwareRepository.kt
@@ -130,10 +130,7 @@ constructor(
                 }
         }
 
-    private suspend fun loadFromBundledJson(
-        hwModel: Int,
-        quirks: List<BootloaderOtaQuirk>,
-    ): Result<DeviceHardware?> =
+    private suspend fun loadFromBundledJson(hwModel: Int, quirks: List<BootloaderOtaQuirk>): Result<DeviceHardware?> =
         runCatching {
             Timber.d("DeviceHardwareRepository: loading device hardware from bundled JSON for hwModel=%d", hwModel)
             val jsonHardware = jsonDataSource.loadDeviceHardwareFromJsonAsset()


dzid26:
It's not clear to the user if the app will proceed with the update despite wrong bootloader. 

E.g. the update failed for me at 0%, but I have no idea what was the reason.

jamesarich:
> It's not clear to the user if the app will proceed with the update despite wrong bootloader.
> 
> E.g. the update failed for me at 0%, but I have no idea what was the reason.

There's currently no way to determine the available bootloader and whether or not it will be successful, hence the warning.

You clicked the button that says "I know what I'm doing" - if you don't.. you shouldn't.

mdecourcy:
> It's not clear to the user if the app will proceed with the update despite wrong bootloader. 
> 
> E.g. the update failed for me at 0%, but I have no idea what was the reason.

Thats normally due to improper bootloader. What device do you have?

The BLE library we use doesn't expose much as to "why" it failed or allow us to see what version of bootloader the device has before kicking off unfortunately.

Cc @jamesarich edit you beat me to it lol

dzid26:
rak4631. 

Ok then. 
In that case, simple clarification string (in the red dialog box) "...and update man fail." would make a difference.

mdecourcy:
> rak4631. 
> 
> Ok then. 
> In that case, simple clarification string (in the red dialog box) "...and update man fail." would make a difference. 

Thats literally what this PR did, there are multiple warnings saying the update can fail?

dzid26:
It doesn't hint that it will fail quietly. 
What you just told me is nowhere in the warning. 
Screenshot_20251202_160921_Meshtastic

jamesarich:
> It doesn't hint that it will fail quietly. What you just told me is nowhere in the warning.

There's a big red box with a warning that you may potentially need an update for ota to work (that includes a link to more information), and an interstitial "I know what I'm doing." dialog with a friendly Chirpy.  There's plenty of hand holding and warning.

If you click past all those without checking/understanding the impact - that's kinda on you.

The real issue at hand is that the process didn't time out with an error when the operation couldn't proceed - see #3846 .

PR #3858, opened 2025-11-29
feat: add infrastructure shutdown safeguards and enhance shutdown dialog text

### LINKED ISSUE #3830: [Feature Request]: Add node name on the shutdown confirmation popup
### Checklist

### Contact Details

_No response_

On the shutdown node confirmation popup, it could be very useful do display the name of the node, because when we manage a lot of nodes, some through remote access, it is quite easy to go from the remote node to the local one if we are not very cautious. If the local one is a rooftop node connected through wifi, a shutdown can be problematic as you can imagine.
It is especially important because when we manage a remote node, the "gear" icon on the bottom is highlighted exactly as if we were managing the local node.

### Why should this be added?

It should be added to avoid unwanted shutdowns of other nodes

_No response_

mdecourcy:
Closes #3830

Paths also touched are via admin access.

 
image


DaneEvans (approved):
looks good. merge for the next internal

PR #3860, opened 2025-11-30
refactor(navigation): Simplify adaptive back nav and state

jamesarich:
This refactors the adaptive list-detail screens for Contacts and Nodes to simplify navigation logic and improve state management.

- Replaced manual `BackHandler` implementations with `BackNavigationBehavior.PopUntilScaffoldValueChange` from Material 3 Adaptive Navigation. This streamlines back navigation, ensuring a consistent user experience.
- Removed the `navigateToMessages` callback from `MessageScreen` as it was redundant in the list-detail layout, simplifying the composable's API.
- Introduced `key(id)` around the detail pane content in both `AdaptiveContactsScreen` and `AdaptiveNodeListScreen`. This ensures that composables like `MessageScreen` and `NodeDetailScreen` are correctly recomposed with fresh state when the selected item changes.
- Eliminated `overrideNodeId` from `NodeDetailScreen` and the now-unused `NodeDetailScreenWrapper`, making the `nodeId` a direct, non-nullable parameter. The `LaunchedEffect` now directly observes this `nodeId`.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/ui/contact/AdaptiveContactsScreen.kt
+++ b/app/src/main/java/com/geeksville/mesh/ui/contact/AdaptiveContactsScreen.kt
@@ -34,6 +34,7 @@ import androidx.compose.material3.adaptive.layout.ListDetailPaneScaffoldRole
 import androidx.compose.material3.adaptive.navigation.BackNavigationBehavior
 import androidx.compose.material3.adaptive.navigation.rememberListDetailPaneScaffoldNavigator
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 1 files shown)


jschollenberger:
Hi!
I just updated to v2.7.8 (29319501). When I am viewing the details of a node then "swipe back" to go back to the node list, the app closes. When viewing a conversation, swiping back takes me to the nodes page.  It's driving me crazy! Was that the intent of the change here, or is this a bug? I'm guessing this is the PR which changed that behavior. 
Thanks!

Issue #3898, opened 2025-12-04
[Bug]: Node list omits nodes from downloaded node DB


Jord-JD:
### Contact Details

GitHub

### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [x] I actually read this list, and should be taken seriously.

### Affected app version

2.7.8 (29319501) google

### Affected Android version

16

### Affected phone model

Samsung Galaxy S24 FE

### Affected node model

Heltec v4 primarily but also Seeed T-1000E

### Affected node firmware version

2.7.16.a597230 Alpha and 2.7.15.567b8ea Beta

### Steps to reproduce the bug

  1. Connect to a node which triggers a node DB download.
  2. Open the Nodes screen.
  3. Observe that only a subset of nodes appear, even though the downloaded DB contains more entries.

### Actual behavior

- When the app downloads the node DB from the node, the node list screen does not show all nodes. Only nodes that subsequently transmit or otherwise update are visible.
- This causes the node count to be lower than expected and makes the node list incomplete until additional traffic occurs.
- It also causes the most recently heard node to often be several minutes or hours ago.

Note that the persistent notification shows the correct number of nodes (roughly reflecting what is shown on the device screen).

It appears that the node details are downloaded from the node, but are not persisted to the node DB.

The issue does not appear to be consistent, but I have heard from a least one other person that this is an issue affecting them on the Google Play version of the Android app.

### Expected behavior

Nodes that are downloaded when connecting are correctly populated into the node DB and shown on the node list.

### Relevant log output

```shell

```

Stephan-2025:
As Info: 

I have that problem too on different Node Types (RAK, T114) and different Android OS Versions / Handys (Android Version 9, 10, 11) as well as on different Meshtastic OS Versions (tested with 2.6.4, 2.7.7, 2.7.10) - but all with Android App Version 2.7.8 (2.7.7 works pretty fine everything)

Stephan-2025:
It appears as following:

[...]

PR #3899, opened 2025-12-04
Persist downloaded nodes when installing node info

Jord-JD:
This is a possible fix for https://github.com/meshtastic/Meshtastic-Android/issues/3898 which I just reported.

I've built an APK (Google Debug) with this change, and tested it locally myself and it appears to fix the issue.

However, I'm new to this code base, and generally have not done much native Android app development, so it is very possible I have misunderstood something and/or this is not the best approach.

The general idea here is:

- Add a forcePersist flag to MeshService.updateNodeInfo to allow immediate DB writes when ingesting a downloaded node DB.
- Use forcePersist = true during installNodeInfo so nodes from the downloaded DB are persisted even before haveNodeDB is true.
- Fixes the node list missing nodes that were only present in the initial download.

My understanding is that the downloaded node DB wasn’t persisted during installNodeInfo because updateNodeInfo only upserted when haveNodeDB was true. As a result, nodes that existed only in the initial download never reached the database and didn’t show up in the node list until they later transmitted.

I do not fully understand why this bug is only coming up for some people, but I think it only surfaces when the nodes in the downloaded DB stay quiet after the initial sync; if the mesh is chatty, those nodes later transmit and get persisted, masking the issue.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -6,7 +6,7 @@ accompanist = "0.37.3"
 
 # androidx
 androidxComposeMaterial3Adaptive = "1.2.0"
-androidxTracing = "1.9.5"
+androidxTracing = "1.10.0"
 datastore = "1.2.0"
 lifecycle = "2.10.0"
 navigation = "2.9.6"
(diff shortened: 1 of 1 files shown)


jamesarich:
closing in favor of #3932  - thank you!

Jord-JD:
Fair enough. Thanks for taking a look. 👍

PR #3909, opened 2025-12-05
Node list molecule

poliver:
This branch is a couple weeks old, so rebasing had a bunch on conflicts which haven't been addressed yet. Figured I'd throw a draft up here before fixing for some visibility and to see if other folks like this pattern.

This PR introduces [Molecule ](https://github.com/cashapp/molecule) to `NodeListViewModel` to clean up its Flow combination logic. This circumvents the arg count limitations of `Flow.combine()` while maintaining type safety when combining many Flows. Instead of using the `Flow.combine()` operator, Molecule allows us to collect the needed flows in a VM-level Composable, then construct the desired emission which is still exposed to the UI as a Flow.

### DIFF
--- a/feature/node/build.gradle.kts
+++ b/feature/node/build.gradle.kts
@@ -48,6 +48,7 @@ dependencies {
     implementation(libs.markdown.renderer.android)
     implementation(libs.markdown.renderer.m3)
     implementation(libs.markdown.renderer)
+    implementation(libs.molecule)
 
     googleImplementation(libs.location.services)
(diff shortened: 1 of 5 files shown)


jamesarich (commented):
Boy this does clean up the viewmodel quite a bit, though I'm a bit nervous to take on a dependency that ties us into a specific architecture/pattern. I've been bitten by this in the past.. anyone remember https://github.com/airbnb/epoxy 😬 ?

I'd have to better understand what relying on this would mean for the project.

PR #3921, opened 2025-12-06
Revert "block creation or sending of duplicate channels. "

jamesarich:
Reverts meshtastic/Meshtastic-Android#3913

@DaneEvans - this didn't work out with the UDP dependency that @b8b8 mentioned 😞  - he can fill in the details.


b8b8:
image

Legitimate use case of bridging presets with the same AQ== key is blocked with this change. If the node already had said double up of keys, everything still worked. If you tried to add a redundant key, it would be blocked, thereby stopping full UDP usage.

PR #3959, opened 2025-12-10
fix(strings): correct settings interval templating

jamesarich:
The `%d` format specifier in plural resources has been changed to `%1$d`. This ensures that the integer argument is correctly referenced, preventing potential formatting errors or crashes on some Android versions.

image


jamesarich:
thanks for the callout @Xaositek

PR #4037, opened 2025-12-18
fix : avoid our node to be a relaynode candidate

### LINKED ISSUE #4036: [Bug]: Relay node resolution may select local or unseen nodes
### Contact Details

_No response_

### Checklist

### Affected app version

2.7.8

### Affected Android version

Android 13

### Affected phone model

Samsung A32

### Affected node model

HeltecV4 but its not firmware fault

2.7.15 but its not firmware fault

By attaching the debugger and stepping through the relay node resolution code
(e.g. Packet.getRelayNode()), it is possible to observe that, in some cases, the local
node is selected as a relay candidate.
The method lacks a check to prevent the local node from being selected as a relay
candidate and avoid nodes with lastseen not set as 0

### Actual behavior

_No response_

### Expected behavior

_No response_

_No response_

### Relevant log output

```shell

```

### Additional information

_No response_

emp3r0r7:
### What
Improves relay node resolution by preventing the local node and unseen nodes
(lastHeard == 0) from being selected as relay candidates.

### Why
Suffix-based node ID matching may result in false positives, causing the local
node or placeholder nodes to be incorrectly resolved as relay nodes. This can
lead to inaccurate relay info and misleading debug information.

### How
- Filtered out nodes that have never been heard from
- Explicitly excluded the local node from relay candidate resolution

Fixes #4036

### DIFF
--- a/core/database/src/main/kotlin/org/meshtastic/core/database/entity/Packet.kt
+++ b/core/database/src/main/kotlin/org/meshtastic/core/database/entity/Packet.kt
@@ -86,7 +86,11 @@ data class Packet(
             val relayNodeIdSuffix = relayNodeId and RELAY_NODE_SUFFIX_MASK
 
             val candidateRelayNodes =
-                nodes.filter { it.lastHeard != 0 && (it.num and RELAY_NODE_SUFFIX_MASK) == relayNodeIdSuffix }
+                nodes.filter {
+                    it.num != ourNodeNum &&
+                        it.lastHeard != 0 &&
+                        (it.num and RELAY_NODE_SUFFIX_MASK) == relayNodeIdSuffix
+                }
 
             val closestRelayNode =
                 if (candidateRelayNodes.size == 1) {
@@ -94,11 +98,6 @@ data class Packet(
                 } else {
                     candidateRelayNodes.minByOrNull { it.hopsAway }
                 }
-
-            if (closestRelayNode?.num == ourNodeNum) {
-                // it's us, useless to return relay node for ourselves
-                return null
-            }
 
             return closestRelayNode
         }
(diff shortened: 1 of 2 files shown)


jamesarich:
I'm a bit hesitant to move forward with this feature at all - it was a bit half-baked to begin with and has had conversation about reworking. The partial id matching is pretty bad especially in larger networks where more collisions may occur.

@RCGV1 has been involved in a lot of those conversations and may have more insight on the future of this feature.

emp3r0r7:
> I'm a bit hesitant to move forward with this feature at all - it was a bit half-baked to begin with and has had conversation about reworking. The partial id matching is pretty bad especially in larger networks where more collisions may occur.
> 
> @RCGV1 has been involved in a lot of those conversations and may have more insight on the future of this feature.

Hi, thanks for the feedback.

In my independent development, I’m working on a UI feature that displays the
currently resolved relay node available . While doing so,
I noticed that the local node is sometimes selected as a relay candidate, which
is what prompted this PR.

I agree that relay resolution may still fail in very large or dense networks by using the method getRelayNode.
To mitigate this in my own experiments, I also set the relay node to the most recently detected
direct node, or to the first responding node from a successful traceroute , i dont know if this could be a good idea for the future development of this feature, i personally find it very useful and in my own implementation, combining these 

3 methods :
relay_node from packet OR 
0 hops from a packet node different from us  OR 
first responding node in successful traceroute request

Works decently.

During testing, filtering out nodes obtained by getRelayNode with lastHeard != 0 and explicitly
excluding the local node consistently improved the accuracy of nearest relay
detection, without introducing regressions.

Also, copilot is right, im sorry i should have filtered before checking , i'll fix it

jamesarich:
Let's give it a try - the current implementation is bad, and this looks like it'll improve things.

Thanks!

PR #4078, opened 2025-12-27
feat: add retention period to meshLog. Defaults to 7 days, with a settings dropdown to change

mdecourcy:
Features added: mesh log retention, ability to turn off mesh log all together, ability to export more than the last 500 packets

Default: 7 days retention. I started at 30 but there are performance/size implications for larger meshes. I'd honestly like lower default retention (3days). End users shouldn't miss it as all because old default didn't allow for exporting the entire log anyways.

Settings added in debug panel > new settings cog. Added in here to decrease complexity of main settings screen for items that will rarely be touched.

image

image

image


core/database/schemas/org.meshtastic.core.database.MeshtasticDatabase/27.json
jamesarich, line 1: why an update to the schema again?

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/debugging/Debug.kt
```diff
@@ -202,6 +219,44 @@ fun DebugScreen(onNavigateUp: () -> Unit, viewModel: DebugViewModel = hiltViewMo
...
+        Modifier.fillMaxWidth()
+            .background(MaterialTheme.colorScheme.surface)
+            .padding(horizontal = 12.dp, vertical = 8.dp),
+        verticalArrangement = Arrangement.spacedBy(8.dp),
+    ) {
+        @Suppress("MagicNumber")
+        val retentionItems =
+            listOf((-1L) to "1 hour") +
+                listOf(1, 3, 7, 14, 30, 60, 90, 180, 365).map { days ->
+                    days.toLong() to if (days == 1) "1 day" else "$days days"
```
jamesarich: use `plurals` string  resources (translatable) here if we don't already have them for days/hours

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/debugging/Debug.kt
```diff
@@ -374,6 +429,12 @@ fun DebugMenuActions(deleteLogs: () -> Unit, modifier: Modifier = Modifier) {
 private suspend fun exportAllLogsToUri(context: Context, targetUri: Uri, logs: List<UiMeshLog>) =
     withContext(Dispatchers.IO) {
         try {
+            if (logs.isEmpty()) {
+                withContext(Dispatchers.Main) { context.showToast(Res.string.debug_export_failed, "No logs to export") }
```
jamesarich, line 436: Extract string resource "No logs to export"

core/database/schemas/org.meshtastic.core.database.MeshtasticDatabase/27.json
mdecourcy: Good catch, this was leftover from the fix I made a few days ago, it just autogenned on rebuild

PR #4124, opened 2026-01-03
feat: auto retry text message send on max retransmit

mdecourcy:
Capped at 5 max retries, 5 seconds between each.

No settings for these as I don't want to add settings bloat or over complicate the settings menu

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/repository/network/NsdManager.kt
+++ b/app/src/main/java/com/geeksville/mesh/repository/network/NsdManager.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Meshtastic LLC
+ * Copyright (c) 2025-2026 Meshtastic LLC
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -14,7 +14,6 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-
 package com.geeksville.mesh.repository.network
 
 import android.annotation.SuppressLint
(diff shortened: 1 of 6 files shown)


shalberd:
Hi @mdecourcy could it be that this feature, when used in busy meshes where messages actually go out, but the 0-hop ACK is not received, causes harm? **Just because no ACK comes in from the first node receiving a new packet does not mean that it is not in fact sent onto the mesh.**

See main body of https://github.com/meshtastic/Meshtastic-Android/issues/4354

Maybe adding a switch or settings toggle for this kind of behavior would be good, with default "no not auto retry".

As the other poster suggests: Add the switch for user to have a choice autoretry or not.

I do not agree with your comment "no switch for this .." something like automatically re-sending messages, even if shown a popup / overlay in the App, is so impactful and potentially harmful that it needs to be optional and toggle-able. A popup message, as it currently is, is good to have for informational purposes, but it does not take away the need to change the behavior.

Side note: I really like your work on the MapLibre feature, keep up the good work

I really want you and the Meshtastic core maintainers here to see my comment as something we are seeing in the wild ... auto auto-retry is not a good thing, the idea behind the feature is understandable, just add a toogle switch and default "false" for it, please, in the next app version. Alternatively, in the upper right of the message itself, or in the options panel, as a "send message again" button when message status has no ACK.

@fifieldt @GUVWAF FYI this happens now in Android App 2.7.11, auto-retry sending a message, without user intenvention, when no ACK from the mesh is received.

Jckf:
We are seeing a lot of duplicate messages after this was introduced. Giving the user the option to retransmit is okay, but automatically sending the same message up to 5 times _with unique packet IDs_ was a really bad decision, in my opinion.

Please reconsider.

GUVWAF:
Yes, I do agree with @shalberd and @Jckf. *If* this is deemed necessary and useful, IMO it would be much better to increase the number of retransmissions in the firmware, where we already have 3 retransmissions to begin with. It then also calculates the appropriate delays based on packet length and LoRa settings rather than just 5 seconds, and it uses the same packet ID.

PR #4131, opened 2026-01-04
feat: Add separate notifications for waypoints

### LINKED ISSUE #4103: [Feature Request]: Waypoint notifications should be in their own category
### Checklist

Waypoint notifications are currently in the same Android notification category as text messages. This is problematic, as it makes it impossible to mute waypoint notifications without also missing messages.

### Why should this be added?

The motivation for this is that someone in our area has started putting the twice-daily weather balloons on the mesh, and they send a waypoint update every minute until they pop... that is a fantastic idea, but the notifications get annoying pretty quickly.

Automated use-cases like this for waypoints seems like a useful and legit way to use meshtastic, but it needs to be possible to separate these from actual human-initiated message traffic for notification purposes.

Image

jamesarich:
Adds separate notifs for waypoints, opens to map centered on waypoint with Info window displayed.

resolves #4103 

https://github.com/user-attachments/assets/c66d311a-53a4-44d7-85f4-a83c54927578

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshServiceNotificationsImpl.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshServiceNotificationsImpl.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2025 Meshtastic LLC
+ * Copyright (c) 2025-2026 Meshtastic LLC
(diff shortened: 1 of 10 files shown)


erayd:
Thank you!

PR #4136, opened 2026-01-05
feat: Add initial handling for Store and Forward PlusPlus packets

jamesarich:
This commit introduces basic handling for `STORE_FORWARD_PLUSPLUS_APP` packets.

- A new handler `handleStoreForwardPlusPlus` is added to `MeshDataHandler` to parse and log incoming Store and Forward PlusPlus packets.
- The DebugViewModel is updated to decode and display `StoreForwardPlusPlus` payloads.

@jp-bennett 👀


jp-bennett:
It would be super cool if we could watch for our own text messages to get added to the chain, and show that status

PR #4139, opened 2026-01-05
feat: Add SFPP confirmed status to Messages and Reactions

jamesarich:
As mentioned in https://github.com/meshtastic/Meshtastic-Android/pull/4136#issuecomment-3708682733 we should be able to see our Message added to the `chain`.

This aims to add support for tracking the confirmation status of Messages within the StoreForwardPlusPlus (SFPP) protocol.

This includes:
- Bumping the database version to 28 and adding a new `sfpp_hash` column to the `packet` table to store the SFPP message hash.
- Introducing a new message status, `SFPP_CONFIRMED`, and a corresponding link icon in the UI to indicate when a message has been confirmed on the SFPP chain.
- Added hash calculation and packet lookup by hash
- Implementing the logic to update a packet's status to `SFPP_CONFIRMED` upon receiving a `CANON_ANNOUNCE` and matching the hash of the SFPP message.


jamesarich:
Passing this off to @mdecourcy since they can actually test against SFPP infrastructure

PR #4158, opened 2026-01-08
fix: reuse packet ID on message retry to prevent duplicates

mdecourcy:
When retrying a message after MAX_RETRANSMIT error, reuse the same packet ID instead of generating a new one. This leverages the mesh network's duplicate detection to prevent duplicate messages from appearing if the original message actually made it through but the ACK wasn't received.

- Message retries now keep the original packet ID
- Reaction retries now keep the original packet ID
- Updated logging to indicate same packet ID is being used


mdecourcy:
Ah after looking closer at the firmware this wont work firmware adds to `recentToRadioPacketIds` regardless of pass/fail.

PR #4246, opened 2026-01-16
refactor: Migrate from Protobuf-Java to Wire

jamesarich:
This commit migrates the project from using the `protobuf-java` library to `wire`. This involves extensive changes across the codebase to adapt to the new generated models and their APIs.

Key changes include:
- Replaced Protobuf builders with Wire's immutable data classes and `copy()` method.
- Updated field access from camelCase (`longName`) to snake_case (`long_name`).
- Switched from `com.google.protobuf.ByteString` to `okio.ByteString`.
- Replaced Protobuf's `hasField()` checks with null checks on Wire's optional fields.
- Adapted `enum` handling to work with Wire-generated enums, removing "UNRECOGNIZED" filters.
- Removed Protobuf-specific utility code, such as `userFieldsToString` and `compareUsers` which relied on Descriptors.
- Updated Gradle build scripts (`build.gradle.kts`, `libs.versions.toml`) to remove Protobuf dependencies and add Wire dependencies and the Wire plugin.
- Updated ProGuard rules to keep Wire-generated classes.

### DIFF
--- a/app/src/main/java/com/geeksville/mesh/service/MeshActionHandler.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshActionHandler.kt
@@ -168,8 +168,8 @@ constructor(
         commandSender.sendData(p)
         serviceBroadcasts.broadcastMessageStatus(p)
         dataHandler.rememberDataPacket(p, myNodeNum, false)
-        val bytes = p.bytes?.toByteArray() ?: ByteArray(0)
-        analytics.track("data_send", DataPair("num_bytes", bytes.size), DataPair("type", p.dataType))
+        val byteCount = p.bytes?.size ?: 0
+        analytics.track("data_send", DataPair("num_bytes", byteCount), DataPair("type", p.dataType))
     }
 
     fun handleRequestPosition(destNum: Int, position: Position, myNodeNum: Int) {
--- a/app/src/main/java/com/geeksville/mesh/service/MeshCommandSender.kt
+++ b/app/src/main/java/com/geeksville/mesh/service/MeshCommandSender.kt
@@ -123,9 +123,9 @@ constructor(
 
     fun sendData(p: DataPacket) {
         if (p.id == 0) p.id = generatePacketId()
-        val bytes = p.bytes?.toByteArray() ?: ByteArray(0)
+        val payloadSize = p.bytes?.size ?: 0
         require(p.dataType != 0) { "Port numbers must be non-zero!" }
-        if (bytes.size >= Constants.DATA_PAYLOAD_LEN.value) {
+        if (payloadSize >= Constants.DATA_PAYLOAD_LEN.value) {
             p.status = MessageStatus.ERROR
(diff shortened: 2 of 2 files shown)


core/model/src/main/kotlin/org/meshtastic/core/model/DataPacket.kt
```diff
@@ -46,11 +47,12 @@ enum class MessageStatus : Parcelable {
 
 /** A parcelable version of the protobuf MeshPacket + Data subpacket. */
 @Serializable
+@Parcelize
 data class DataPacket(
     var to: String? = ID_BROADCAST, // a nodeID string, or ID_BROADCAST for broadcast
-    val bytes: ByteArray?,
+    var bytes: ByteArray?,
```
yschimke: What about continuing the Okio types to here?

```
    @Serializable(with = ByteStringSerializer::class)
    @TypeParceler<ByteString?, ByteStringParceler>
    var bytes: ByteString?,
```

and

```
object ByteStringSerializer : KSerializer<ByteString> {
    val byteArraySerializer = ByteArraySerializer()
    override val descriptor: SerialDescriptor = byteArraySerializer.descriptor

    override fun serialize(encoder: Encoder, value: ByteString) {
        byteArraySerializer.serialize(encoder, value.toByteArray())
    }

    override fun deserialize(decoder: Decoder): ByteString {
        return byteArraySerializer.deserialize(decoder).toByteString()
    }
}

object ByteStringParceler : Parceler<ByteString?> {
    override fun create(parcel: Parcel): ByteString? {
        return parcel.createByteArray()?.toByteString()
    }

    override fun ByteString?.write(parcel: Parcel, flags: Int) {
        parcel.writeByteArray(this?.toByteArray())
    }
}
```

core/model/src/main/kotlin/org/meshtastic/core/model/DataPacket.kt
```diff
@@ -107,47 +109,20 @@ data class DataPacket(
...
-        parcel.readInt().let { if (it == -1) null else it },
-        parcel.readInt(), // relays
-        parcel.readInt() == 1, // viaMqtt
-        parcel.readInt(), // retryCount
-        parcel.readInt(), // emoji
-        parcel.createByteArray(), // sfppHash
-    )
-
     @Suppress("CyclomaticComplexMethod")
     override fun equals(other: Any?): Boolean {
```
yschimke: If you use ByteString for the type, I think you can delete equals/hashCode. And maybe readFromParcel.

yschimke: readFromParcel might be needed for the inout param on send.

jamesarich: good call - thanks @yschimke

jamesarich:
closed in favor of #4401

PR #4316, opened 2026-01-26
fix(ui): Prevent NaN values in environment metrics

jamesarich:
This commit prevents `NaN` (Not a Number) values from being displayed for dew point and soil temperature in the environment metrics UI.

It introduces checks to ensure that `NaN` values, which can occur during sensor readings or calculations, are handled gracefully and not passed to the UI layer.

### Key Changes:

- **`EnvironmentMetrics.kt`**:
  - Added a check to only display the dew point metric if the calculated value is not `NaN`.
  - Added a `!soilTemperature.isNaN()` check before displaying the soil temperature metric.
- **`UnitConversions.kt`**:
  - The `toTempString` extension function now returns `"--"` if the float value is `NaN`, providing a fallback display string.
- **`UnitConversionsTest.kt`**:
  - Added a new unit test to verify that `toTempString` correctly handles `NaN` inputs.

### DIFF
--- a/core/model/src/main/kotlin/org/meshtastic/core/model/util/UnitConversions.kt
+++ b/core/model/src/main/kotlin/org/meshtastic/core/model/util/UnitConversions.kt
@@ -26,6 +26,8 @@ object UnitConversions {
 
     /** Formats temperature as a string with the unit suffix. */
     fun Float.toTempString(isFahrenheit: Boolean): String {
+        if (this.isNaN()) return "--"
+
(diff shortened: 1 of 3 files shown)


dadanielat:
The current meshtastic app crashes when tapping on a node with NaN readings in the node list:
01-26 17:14:13.856 29722 29722 E AndroidRuntime: java.lang.IllegalArgumentException: Cannot round NaN value.

Issue #4285, opened 2026-01-22
[Bug]: Neighbor info popup occuring w/out request


DaneEvans:
### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Affected app version

2.7.11 

### Affected Android version

na

### Affected phone model

na

### Affected node model

na

### Affected node firmware version

na

### Steps to reproduce the bug

setup neighbor info to work ... 

just operate normally, the popup comes up when not requested

### Actual behavior

the popup comes up when not requested, often when you request position, or a similar function.

### Expected behavior

popup should only occur when we asked for the packet. 

### Screenshots/Screen recordings

to come 

### Relevant log output

```shell
looks like it might be fw - responding to a position request, with neighbors as well, 

And as we have sent a request, with that ID, the filtering doesn't work. 

We need to either filter more aggressively, or I need to patch it on the fw side. 
Or rip the function out as it's causing more confusion, and it really shouldn't be user facing anyway.
```

formiification:
Experiencing same behaviour on 2.7.11. Really annoying, like actually much more bugs in this version

ffolkes:
Also experiencing this since .11. Everytime I open the app it pops up about neighbor info.

DaneEvans:
Not fixed yet, just hidden

TheDomanNow:
> Also experiencing this since .11. Everytime I open the app it pops up about neighbor info.

Same here, it's very annoying.

DaneEvans:
Update then.

TheDomanNow:
Looks fixed.

ffolkes:
This has been fixed for weeks, for me at least.

DaneEvans:
It's not fixed. It's just hidden. 

We've been through this.

PR #4344, opened 2026-01-28
chore(lint): Disable ForbiddenComment detekt rule

jamesarich:
The `ForbiddenComment` rule in the detekt configuration has been deactivated. This allows the use of comments that were previously forbidden, such as those containing 'FIXME:'.


jamesarich:
@DaneEvans i did this so we can gracefully disable NeighborInfo dialog for #4285 while still marking it as FIXME 🫨

PR #4345, opened 2026-01-28
fix(neighborinfo): disable neighbor info dialog temporarily

### LINKED ISSUE #4285: [Bug]: Neighbor info popup occuring w/out request

jamesarich:
The `canRequestNeighborInfo` capability has been temporarily disabled because the feature is not working as expected.

This change involves:
- Hardcoding the `canRequestNeighborInfo` capability to `false`.
- Updating the corresponding tests to reflect that the feature is now disabled for all firmware versions.
- Commenting out the UI code in `Main.kt` that was responsible for displaying the NeighborInfo response dialog.

resolves #4285 until we can get it working properly. 😞

### DIFF
--- a/core/model/src/main/kotlin/org/meshtastic/core/model/Capabilities.kt
+++ b/core/model/src/main/kotlin/org/meshtastic/core/model/Capabilities.kt
@@ -38,7 +38,7 @@ data class Capabilities(val firmwareVersion: String?, internal val forceEnableAl
 
     /** FIXME: Ability to request neighbor information from other nodes. Disabled until working better. */
     val canRequestNeighborInfo: Boolean
-        get() = false
+        get() = isSupported("9.9.9")
 
     /** Ability to send verified shared contacts. Supported since firmware v2.7.12. */
     val canSendVerifiedContacts: Boolean


DaneEvans:
That's a hammer, but it makes sense. 
It's too annoying in open, but I'll get something in and retest it in internal 


The code that should suppress it isn't doing so at the moment, I'll hopefully get to it this afternoon.

Issue #2944, opened 2025-09-02
[Feature Request]: intelligent prefixes for large values in Environment Metrics


bergie:
### Tell us your idea.

Right now environment metrics always uses "small" units (mm for distance, mA for current). If these values are larger (say, distance over 1000mm, or current over 1000mA), it would make sense to go to the base units themselves (so, 1m, 1A).

This occurs for instance when getting boat telemetry from [signalk-meshtastic](https://github.com/meri-imperiumi/signalk-meshtastic). Here's what I currently see from my boat node for instance:

<img width="1080" height="1061" alt="Image" src="https://github.com/user-attachments/assets/91fc8871-2972-4adb-87c7-d9e64796f328" />

Here it would make much more sense to say:

- Current 18.2A
- Distance 32.1m

Unrelated to this issue (more to #2611) km/h is also not commonly used for wind speed. Most European weather forecasts would use m/s, or for marine or aviation weather, knots).

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

bergie:
hPa was likely used as it uses same values as the old millibars. 

Here's a typical Finnish forecast using m/s

![Image](https://github.com/user-attachments/assets/aa012988-e10a-4b68-bc27-45ec4f28eceb)

BBC seems to default to mph (of all things!)

![Image](https://github.com/user-attachments/assets/acac82d3-e824-42a3-a3c8-8b553c195837)

However, the wind speed issue isn't really what I'm talking about here, but instead the usage of the wrong [metric prefixes](https://en.wikipedia.org/wiki/Metric_prefix) for large values. Maybe we should have a separate ticket for making wind units configurable?

jamesarich:
> [@DaneEvans](https://github.com/DaneEvans) , thank you, appreciate it a lot. If you could give a little hint about where to look at (e.i, the file(s) to change), that will be great!

@NeimadTL, all the necessary information is available. If you are genuinely interested and capable of contributing, please proceed. The developers here are busy donating their time to development and are not available for one-on-one assistance.

NeimadTL:
> > [@DaneEvans](https://github.com/DaneEvans) , thank you, appreciate it a lot. If you could give a little hint about where to look at (e.i, the file(s) to change), that will be great!
> 
[...]

PR #4433, opened 2026-02-03
Add drafty human friendly current metrics

NeimadTL:
Currently, the metrics are always shown with the in milliampere (e.i, using the "mA" unit). A nicer way would be to have those metrics shown in more friendly way (e.g, 1A -> 1A, 1000A -> 1kA, 0.000001A -> 1μA...etc). At this point, this is just a draft of how it would be implemented.

Related to #2944

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -20,7 +20,7 @@ jobs:
     outputs:
       android: ${{ steps.filter.outputs.android }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
       - uses: dorny/paths-filter@v3
         id: filter
         with:
--- a/.github/workflows/reusable-check.yml
+++ b/.github/workflows/reusable-check.yml
@@ -62,7 +62,7 @@ jobs:
           submodules: 'recursive'
 
       - name: Validate Gradle Wrapper
-        uses: gradle/actions/wrapper-validation@v4
+        uses: gradle/actions/wrapper-validation@v5
 
       - name: Set up JDK 17
         uses: actions/setup-java@v5
@@ -154,7 +154,7 @@ jobs:
 
       - name: Upload debug artifact
         if: ${{ steps.tasks.outputs.is_first_api == 'true' && inputs.upload_artifacts }}
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: app-debug-apks
           path: app/build/outputs/apk/*/debug/*.apk
@@ -170,7 +170,7 @@ jobs:
 
       - name: Upload reports
         if: ${{ always() && inputs.upload_artifacts }}
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v7
         with:
           name: reports-api-${{ matrix.api_level }}
           path: |
--- a/core/resources/src/commonMain/composeResources/values-bg/strings.xml
+++ b/core/resources/src/commonMain/composeResources/values-bg/strings.xml
@@ -255,6 +255,7 @@
     <string name="direct_message">Директно съобщение</string>
     <string name="nodedb_reset">Нулиране на базата данни с възли</string>
     <string name="delivery_confirmed">Съобщението е доставено</string>
+    <string name="delivery_confirmed_reboot_warning">Устройството ви може да прекъсне връзката и да се рестартира, докато се прилагат настройките.</string>
     <string name="error">Грешка</string>
     <string name="ignore">Игнорирай</string>
     <string name="remove_ignored">Премахване от игнорирани</string>
@@ -545,6 +546,7 @@
     <string name="long_name">Дълго име</string>
     <string name="short_name">Кратко име</string>
     <string name="hardware_model">Модел на хардуера</string>
+    <string name="licensed_amateur_radio">Лицензиран радиолюбител (Ham)</string>
     <string name="licensed_amateur_radio_text">Активирането на тази опция дезактивира криптирането и не е съвместимо с мрежата Meshtastic по подразбиране.</string>
     <string name="dew_point">Точка на оросяване</string>
     <string name="pressure">Налягане</string>
@@ -671,6 +673,7 @@
     <string name="type_a_message">Въведете съобщение</string>
     <string name="pax">PAX</string>
     <string name="wifi_devices">WiFi устройства</string>
+    <string name="ble_devices">Bluetooth устройства</string>
     <string name="bluetooth_paired_devices">Сдвоени устройства</string>
     <string name="connected_device">Свързано устройство</string>
     <string name="view_release">Преглед на изданието</string>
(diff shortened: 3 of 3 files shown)


NeimadTL:
@DaneEvans @jamesarich, could you please have look when you'll get some time please (couldn't assign it you for some reason). Also, I've got to do the same for the distance metric but couldn't find it in `feature/node/src/main/kotlin/org/meshtastic/feature/node/metrics/EnvironmentMetrics.kt`. Thank you.

core/model/src/main/kotlin/org/meshtastic/core/model/util/UnitConversions.kt
```diff
@@ -53,4 +69,30 @@ object UnitConversions {
...
+        var exponent = floor(log10(number)).toInt()
+
+        if (exponent.mod(3) != 0) {
+            if ((-12..-9).contains(exponent)) exponent = -9
+            if ((-8..-6).contains(exponent)) exponent = -6
+            if ((-5..-3).contains(exponent)) exponent = -3
+            if ((-2..2).contains(exponent)) exponent = 0
+            if ((3..5).contains(exponent)) exponent = 3
+            if ((6..8).contains(exponent)) exponent = 6
+            if ((9..12).contains(exponent)) exponent = 9
```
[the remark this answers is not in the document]
DaneEvans: Agree with this comment for sure .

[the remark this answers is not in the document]
NeimadTL: I was not feeling it either but could't find something simpler at the time. Thanks to Copilot !

NeimadTL:
Thanks for your feedback @DaneEvans. Made the changes but looks like there other issue which I haven't figured out yet. Any suggestion please ? 
Screenshot

FYI, the same warning appears the last CI run:
> Configure project :core:strings
w: ⚠️ Unused Kotlin Source Sets
The Kotlin source set commonTest was configured but not added to any Kotlin compilation.
Solution: You can add a source set to a target's compilation by connecting it with the compilation's default source set using 'dependsOn'.
See https://kotl.in/connecting-source-sets

core/model/src/commonMain/kotlin/org/meshtastic/core/model/util/UnitConversions.kt
```diff
@@ -53,4 +52,12 @@ object UnitConversions {
...
         return (b * alpha) / (a - alpha)
     }
+
+    /**
+     * Converts numbers from milli to unit. examples:
+     * - 1000 milliamperes will be converted into 1 ampere,
+     * - 100 millimeters to 0.1 meters
+     */
+    @Suppress("MagicNumber")
+    fun convertToBaseUnit(number: Float): Float = number / 1000f
```
jamesarich: Preferred in Kotlin to write this as an extension property or function directly on the `Float` (or `Number`) type.

```
val Float.milliToBase: Float get() = this / 1000f
```

This can then be used everywhere in the code:
```
envMetrics.current?.milliToBase
```

feature/node/src/main/kotlin/org/meshtastic/feature/node/metrics/EnvironmentMetrics.kt
```diff
@@ -374,9 +375,9 @@ private fun VoltageCurrentDisplay(envMetrics: org.meshtastic.proto.EnvironmentMe
                 )
             }
             if (hasCurrent) {
-                val current = envMetrics.current!!
+                val current = convertToBaseUnit(envMetrics.current!!)
```
jamesarich: Let's get rid of the `!!` here and wrap this with `?.let`.

```
envMetrics.current?.let { currentMilli ->
    val current = convertToBaseUnit(currentMilli)
    Text(
        text = "%s %.2f A".format(stringResource(Res.string.current), current),
        // ...
    )
}
```

feature/node/src/main/kotlin/org/meshtastic/feature/node/metrics/EnvironmentMetrics.kt
```diff
@@ -374,9 +375,9 @@ private fun VoltageCurrentDisplay(envMetrics: org.meshtastic.proto.EnvironmentMe
...
                 Text(
-                    text = "%s %.2f mA".format(stringResource(Res.string.current), current),
+                    text = "%s %.2f A".format(stringResource(Res.string.current), current),
```
jamesarich: Move String formatting into Compose Resources. Rather than doing `"%s %.2f A".format(stringResource(Res.string.current), current)` inside the Composable, it is best practice to move the format specifier directly into the strings.xml. e.g., `<string name="current_value_format">%1$s %2$.2f A</string>` And then in Compose: `stringResource(Res.string.current_value_format, stringResource(Res.string.current), current)` This is safer for localization and keeps your UI code much cleaner.

jamesarich (changes requested):
one thing you missed - you updated the EnvironmentMetrics chart to use base Amperes, but there are still several places hardcoded to display the raw telemetry values as `mA`.
for the app to be consistent, you need to wrap the raw values in your new convertToBaseUnit() and swap the string formatting over to A in these files:
 - feature/node/src/main/kotlin/org/meshtastic/feature/node/metrics/PowerMetrics.kt (lines ~280 and ~386)
 - feature/node/src/main/kotlin/org/meshtastic/feature/node/component/PowerMetrics.kt (line ~51, and the other channels)
 - feature/node/src/main/kotlin/org/meshtastic/feature/node/component/EnvironmentMetrics.kt (line ~103)
 - feature/node/src/main/kotlin/org/meshtastic/feature/node/component/NodeItem.kt (line ~365)

core/model/src/commonMain/kotlin/org/meshtastic/core/model/util/UnitConversions.kt
NeimadTL: Understood 👌

feature/node/src/main/kotlin/org/meshtastic/feature/node/metrics/EnvironmentMetrics.kt
NeimadTL: Another thing learnt today, thanks

feature/node/src/main/kotlin/org/meshtastic/feature/node/metrics/EnvironmentMetrics.kt
NeimadTL: Ok, thanks. Though, I did it only that specific file, please let me know it's need to been done in the other files as well. If so, I'll need to tweak the newly added `current_value_format`.

NeimadTL:
Thank your help and feedback @jamesarich. Whenever you've got some time, please a look again, thanks.

jamesarich:
@NeimadTL take a look at copilot's review. I'm also a bit unsure about this normalization - forcing everything into amps with `%sA` kinda seems like a bad plan, especially as locales may have different ways of displaying unit labels.

maybe you can take a look at https://github.com/jacobras/Human-Readable for inspiration as well.

PR #4486, opened 2026-02-06
refactor(node): Improve public key conflict handling

jamesarich:
This commit refactors the logic for handling public key (PKC) updates for nodes in the database to be more robust and explicit.

Key changes:
-   Sets the public key to a specific `ERROR_BYTE_STRING` upon detecting a genuine mismatch between two valid, non-empty keys.
-   Preserves the existing key during routine node updates (e.g., position packets) that don't contain a new key.
-   Allows a node to recover from the error state if it later provides a valid, matching key.
-   Explicitly clears the public key for licensed (HAM) users, as they do not use PKC.
-   Updates the UI in Node Details and related components to visually indicate the error state, preventing users from copying the invalid key.

This new logic provides a clearer and more predictable state for node public keys, especially in cases of factory resets or hardware ID collisions.

### DIFF
--- a/core/database/src/androidTest/kotlin/org/meshtastic/core/database/dao/NodeInfoDaoTest.kt
+++ b/core/database/src/androidTest/kotlin/org/meshtastic/core/database/dao/NodeInfoDaoTest.kt
@@ -22,6 +22,7 @@ import androidx.test.platform.app.InstrumentationRegistry
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.runBlocking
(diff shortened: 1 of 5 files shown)


prokrypt:
> * Explicitly clears the public key for licensed (HAM) users, as they do not use PKC.

could this possibly be ~~abused~~ turned into a feature to clear keys of other nodes?

jamesarich:
> > * Explicitly clears the public key for licensed (HAM) users, as they do not use PKC.
> 
> could this possibly be ~abused~ turned into a feature to clear keys of other nodes?

good eye #4555

PR #4491, opened 2026-02-07
Enhancement - optional transparent homoglyph encoding of a few characters in certain languages for more compact and efficient text messages

LuigiVampa92:
Hi everyone. First of all - thank you for an awesome project, I have been using it for a while and it works great, fun to tinker with, and very useful to solve some of my real life problems. I express my great gratitude to all the contributors. 

I would like to suggest a small enhancement to the project.

The general idea - we can reduce the binary size of the text message by replacing some 2 byte characters to 1 byte characters without actually losing anything in the process.

---

The basic problem this change solves is this:

The text message that can be transmitted by Meshtastic has a maximum size limit, depending on the preset. In case of usual LongFast preset it is 200 characters. This is a reasonable amount for a short message. The thing is - it not literally the symbol count that matters, it is its binary size, and this is where the things get tricky.

In UTF8 unicode the characters of the latin alphabet occupy exactly 1 byte each, because they fit into the single byte without its highest bit (less than 128), and that gives us the originally intended capacity of 200 symbols for a single message. However, in the various languages their national alphabets use different non latin characters, and in their case each character occupies at least 2 bytes, which reduces the effective maximum transmitted message size by half - only to 100 non-latin symbols per a single message. In practice it is around 115-120 * characters (200/2 = 100 + some characters are spaces and special signs that occupy only 1 byte), but still that is not too much, and sometimes that is not enough to combine a message with a single complete thought, so it would be nice to have some more available capacity for that.

---

So the idea I came with is this:
 
In various national alphabets there are letters that basically repeat latin letters, but they still have their own dedicated character codes in unicode. 

For instance, in cyrillic alphabets (also in greek and a few others) many characters have exactly the same visual appearance as the latin characters, but due to usage of their dedicated character codes they occupy 2 bytes each, though such letters can be easily replaced with their latin analogues without any loss of their meaning or visual change.

Some example of such cyrillic letters: "А", "В", "С", "Е", etc. So they can naturally be replaced with corresponding latin "A", "B", "C", "E" and we could save one byte per each of their occasion in the message.

These are fairly common symbols in cyrillic alphabtes - half of the vowels and very frequently used consonants. According to statistics, they form about 20-25% of all letters in the average text. Replacing them with latin characters will save the user the same proportion of available message space and can fit more letters into the message or to deliver the same text message with higher reliability (because it will become shorter). The average transmitted message volume can then be increased from ~115-120 characters to ~140-145 and sometimes even more.

---

I gonna give you a couple of examples:

(1) Let's take a look at the cyrillic text "Мештастик - это проект с открытым исходным кодом" ("Meshtastic is an open source project"). The original byte size of this message is 88 bytes: 40 cyrillic letters, 2 bytes each, + 7 spaces, 1 byte each + 1 special character, also 1 byte). By replacing the cyrillic characters with their corresponding latin character homoglypghs we will get this: "Meштacтик - этo пpoeкт c oткpытым иcxoдным кoдoм". It looks absolutely the same, but now it is just 72 bytes, because we have replaced letters "M", "e", "a", "c", "o", "p", "x" to latin and now we have only 24 cyrillic letters, 2 bytes each, 16 latin letters, 1 byte each, and spaces and special characters remain the same - 7 and 1 respectfully. So as you can see we have saved 16 bytes by using 16 latin letters instead of cyrillic, and that is roughly 20% - not bad.

Let's look at another example:

(2) Let's take this cyrillic text this time - "Всему свое время" ("All in good time"). The original is 30 bytes: 14 cyrillic letters, 2 bytes each + 2 spaces. After the homoglyph transform: "Bceмy cвoe вpeмя" is now just 21 bytes: just 5 cyrillic letters, 2 bytes each, 9 latin letters and 2 spaces. We have replaced "Bce y c oe pe" this time, 9 bytes saved, which is 30% of the message binary size this time.

---

In short - it is great! I have been using this for a while with my personal node, it works flawlessly, and I thought that it is a good idea to contribute this to the main project code.

So this pull request contain the following changes:
(1) I have added a new toggle to the settings tab, to the "application" group. The feature is can be turned on and off there and is disabled by default. 
(2) I have added a new prefs file that stores the current state of this setting - whether it is turned off or on
(3) I have added the object that has a static method that accepts a string as a parameters and returns the transformed string
(4) In the Message implementation I check whether this feature is enabled and if it is, then all the characters from predefined list are replaced by corresponding latin characters

For now the replaced characters are from the basic cyrillic block (U+0400-U+04FF), and are used mostly in the languages of the eastern Europe, Balkan, some central Asian countries, etc, but maybe there are something else that can be handled the same way. I have left all the links to unicode characters descriptions in the comments inside the transformer implementation. 

Please note that only 100% "reliable", completely visually identical characters are replaced. The characters that look like latin but have minor differences like various descenders, hooks, strokes, etc are not replaced with "simplified" latin appearance and will remain 2 byte unicode, as usual (though maybe in future it might make sense to make some "aggressive" mode that will replace them as well)

Here are some screenshots of how it works:

---

settings_screen_update
(settings screen update)

---

And some examples of me typing messages with disabled and enabled feature respectfully

---

Example 1:

| Before | After |
|------|-----|
| text_01_cyrillic | text_01_homoglyph |

---

Example 2:

| Before | After |
|------|-----|
| text_02_cyrillic | text_02_homoglyph |

---

Example 3:

| Before | After |
|------|-----|
| text_03_long_cyrillic | text_03_long_homoglyph |

---

I have added a basic test that covers this implementation. Spotless, Detekt and tests are all passed, everything is ok. Hope I didn't miss anything. 

I hope that you will consider merging this enhancement, since it is a huge iprovement for those who use languages that use cyrillic alphabets. 

Thank you again!

### DIFF
--- a/feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/MessageViewModel.kt
+++ b/feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/MessageViewModel.kt
@@ -208,8 +208,20 @@ constructor(
                 }
             }
         }
+
+        // Applying homoglyph encoding to the transmitted string if user has activated the feature
+        // In most cases the value in "str" parameter will already contain the correct
+        // transformed string from the text input. This call here added to make sure that
+        // the feature is effective across all possible message paths (quick-chat, reply, etc.)
+        val dataPacketText: String =
+            if (homoglyphEncodingPrefs.homoglyphEncodingEnabled) {
+                HomoglyphCharacterStringTransformer.optimizeUtf8StringWithHomoglyphs(str)
+            } else {
+                str
+            }
+
         val p =
-            DataPacket(dest, channel ?: 0, str, replyId).apply {
+            DataPacket(dest, channel ?: 0, dataPacketText, replyId).apply {
                 from = ourNodeInfo.value?.user?.id ?: DataPacket.ID_LOCAL
             }
         sendDataPacket(p)


DaneEvans:
That is the best MR overview I've read in a long time.  
I'll take a look at the code itself, but it sounds like a good idea

core/strings/src/commonMain/composeResources/values/strings.xml
```diff
@@ -294,6 +294,7 @@
     <string name="theme_system">System default</string>
     <string name="choose_theme">Choose theme</string>
     <string name="provide_location_to_mesh">Provide phone location to mesh</string>
+    <string name="use_homoglyph_characters_encoding">Use latin homoglyphs to encode text messages more compactly</string>
```
DaneEvans: I feel like this wording isn't going to be persuasive enough to the people that need it, and with it defaulting to off, it'll be another underutilised feature.

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/HomoglyphCharacterStringTransformer.kt
```diff
@@ -0,0 +1,89 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.feature.messaging
+
+/**
+ * This util class allows you to optimize the binary size of the transmitted text message strings. It replaces certain
```
DaneEvans, line 20: clean up some of these docstrings. This is a lot of text.

feature/messaging/src/test/kotlin/org/meshtastic/feature/messaging/HomoglyphCharacterTransformTest.kt
```diff
@@ -0,0 +1,34 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.feature.messaging
+
+import org.junit.Test
+
+class HomoglyphCharacterTransformTest {
+
+    @Test
```
DaneEvans, line 25: run a case with all nonHomoglyphs, 

and one with all latin characters.

core/strings/src/commonMain/composeResources/values/strings.xml
LuigiVampa92: Hi, Dane, thank you for reviewing the PR! 

I've been thinking about this string too. In my opinion, its text looks larger than the descriptions of other switches in the settings, and stands out from the general clear design, but still does not provide an easy understandable explanation.

What do you think of the "Compact encoding for Cyrillic"? This sound more compact, no long unnecessary description and would definitely draw the target audience's attention to the feature. If a person uses Meshtastic to communicate in a language whose alphabet is based on cyrillic script, then it instantly gonna tick "Hey, this feature is for me"

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/HomoglyphCharacterStringTransformer.kt
LuigiVampa92: Thanks for pointing this out. I will make the object description more compact

feature/messaging/src/test/kotlin/org/meshtastic/feature/messaging/HomoglyphCharacterTransformTest.kt
LuigiVampa92: Sure, I will add a few more tests

core/strings/src/commonMain/composeResources/values/strings.xml
LuigiVampa92: @DaneEvans Changed toggle description like this:
settings_new_description_english

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/HomoglyphCharacterStringTransformer.kt
LuigiVampa92: @DaneEvans done. Made the class doc a bit more simple and short

feature/messaging/src/test/kotlin/org/meshtastic/feature/messaging/HomoglyphCharacterTransformTest.kt
LuigiVampa92: @DaneEvans done. Added a few more test cases: the ones you have advised and some more. So right now there are 5 test cases: 
- a non-latin string with some latin homoglyphs
- a non-latin string containing only latin homoglyphs
- a non-latin string without any latin homoglyphs
- a string with only latin characters
- a string with only non-latin characters that are technically impossible to represent by latin homoglyphs

LuigiVampa92:
> All good on my end, thanks for the thorough writeup and clean submission. I'll ask the robot to do a review pass, but it shouldn't come up with much 👍

Hi, James, thanks for the reply! I see that Copilot has raised some fair points, like the extra de facto unused dependencies I have copy-pasted from other modules, and a few other things. I gonna fix that and reply here again

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -127,6 +130,12 @@ constructor(
         analyticsPrefs.analyticsAllowed = !analyticsPrefs.analyticsAllowed
     }
 
+    var homoglyphEncodingEnabledFlow = homoglyphEncodingPrefs.getHomoglyphEncodingEnabledChangesFlow()
```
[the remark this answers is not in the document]
LuigiVampa92: Yes, fair point. Fixed

feature/messaging/build.gradle.kts
```diff
@@ -62,4 +62,7 @@ dependencies {
 
     androidTestImplementation(libs.androidx.compose.ui.test.junit4)
     androidTestImplementation(libs.androidx.test.ext.junit)
+
+    testImplementation(libs.junit)
+    testImplementation(libs.mockk)
```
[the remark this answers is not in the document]
LuigiVampa92: Yes, that was an inaccurate copy-paste. Fixed

feature/messaging/src/test/kotlin/org/meshtastic/feature/messaging/HomoglyphCharacterTransformTest.kt
```diff
@@ -0,0 +1,66 @@
...
+    fun `optimizeUtf8StringWithHomoglyphs shrinks binary size of cyrillic text containing some homoglyphs`() {
+        val testString = "Мештастик - это проект с открытым исходным кодом"
+        val transformedTestString = HomoglyphCharacterStringTransformer.optimizeUtf8StringWithHomoglyphs(testString)
+        val testStringBytes = testString.toByteArray(charset = Charsets.UTF_8)
+        val transformedTestStringBytes = transformedTestString.toByteArray(charset = Charsets.UTF_8)
+        val transformedStringBinarySizeShrinked = transformedTestStringBytes.size < testStringBytes.size
+        assert(transformedStringBinarySizeShrinked)
+    }
```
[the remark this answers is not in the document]
LuigiVampa92: Thank you for bringing my attention to this. Fixed assertion expressions to correct ones

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/HomoglyphCharacterStringTransformer.kt
```diff
@@ -0,0 +1,82 @@
...
+    private val homoglyphCharactersSubstitutionMapping: Map<Char, Char> =
+        mapOf(
+            '\u0405' to 'S', // https://www.compart.com/en/unicode/U+0405 - Cyrillic Capital Letter Dze
+            '\u0406' to
+                'I', // https://www.compart.com/en/unicode/U+0406 - Cyrillic Capital Letter Byelorussian-Ukrainian I
+            '\u0408' to 'J', // https://www.compart.com/en/unicode/U+0408 - Cyrillic Capital Letter Je
+            '\u0410' to 'A', // https://www.compart.com/en/unicode/U+0410 - Cyrillic Capital Letter A
+            '\u0412' to 'B', // https://www.compart.com/en/unicode/U+0412 - Cyrillic Capital Letter Ve
+            '\u0415' to 'E', // https://www.compart.com/en/unicode/U+0415 - Cyrillic Capital Letter Ie
+            '\u0417' to '3', // https://www.compart.com/en/unicode/U+0417 - Cyrillic Capital Letter Ze
```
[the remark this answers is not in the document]
LuigiVampa92: Hm, to be honest I cannot agree here. I have never witnessed this problem personally and I have just checked out a few fonts and asked ChatGPT about it as well - it seems that this shouldn't be a problem, but this ambiguity about "letter -> digit" transformation is a fair point, so I have added a commentary with an explicit explanation to avoid confusion in future

feature/settings/src/main/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -100,6 +101,7 @@ data class RadioConfigState(
     val responseState: ResponseState<Boolean> = ResponseState.Empty,
     val analyticsAvailable: Boolean = true,
     val analyticsEnabled: Boolean = false,
+    val homoglyphCharactersEncodingEnabled: Boolean = false,
```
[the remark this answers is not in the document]
LuigiVampa92: Fair point, I probably committed that by mistake when I was experimenting with the code. Thank you for highlighting it! Fixed

feature/messaging/src/main/kotlin/org/meshtastic/feature/messaging/MessageViewModel.kt
```diff
@@ -122,6 +124,8 @@ constructor(
                 ?.map { it.first }
                 ?.take(6) ?: listOf("👍", "👎", "😂", "🔥", "❤️", "😮")
 
+    val homoglyphEncodingEnabled = homoglyphEncodingPrefs.getHomoglyphEncodingEnabledChangesFlow()
+
```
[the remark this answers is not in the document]
LuigiVampa92: Thank you for highlighting that! Fixed

LuigiVampa92:
@DaneEvans @jamesarich I went through the list of issues that Copilot highlighted and committed the fixes. 

I would be grateful if you could take a look and give feedback when it is convenient for you. Thanks!

LuigiVampa92:
> Looks good, thanks again @LuigiVampa92 ! 👍

It is a pleasure :) Thank you again for your work. I hope that I will come up with some more ideas in future

PR #4620, opened 2026-02-22
fix(strings): Correct capitalization of Ham

alecperkins:
In reference to amateur radio operators, “ham” is not an acronym.

### DIFF
--- a/core/resources/src/commonMain/composeResources/values/strings.xml
+++ b/core/resources/src/commonMain/composeResources/values/strings.xml
@@ -741,7 +741,7 @@
     <string name="long_name">Long Name</string>
     <string name="short_name">Short Name</string>
     <string name="hardware_model">Hardware model</string>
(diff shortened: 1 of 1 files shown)


jamesarich:
We only need the English string, translations are sourced by Crowdin.

alecperkins:
Updated, though I wonder if it will translate properly, or perhaps be too literal? I don’t know if "ham" is used in every language this way.

For comparison, the option on iOS: image

jamesarich:
> Updated, though I wonder if it will translate properly, or perhaps be too literal? I don’t know if "ham" is used in every language this way.
> 
> For comparison, the option on iOS: image

I can add a contextual note to that string for translators, I'll include a ref to this ticket.

jamesarich:
Hi @alecperkins this is now in conflict, would you like to fix it up?

jamesarich:
@alecperkins added a comment on Crowdin:
image

Issue #4576, opened 2026-02-17
[Bug]: The app crash and closes when I try to scan a QR code or when I try to see a traceroute on the map FDroid version


ccchdro:
### Contact Details

ccchdro@gmail.com

### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [x] I actually read this list, and should be taken seriously.

### Affected app version

2.7.13

### Affected Android version

Android 13

### Affected phone model

Samsung Galaxy Note 10

### Affected node model

Heltec T190

### Affected node firmware version

2.7.19

### Steps to reproduce the bug

The app crash and closes when I try to scan a QR code or when I try to see a traceroute on the map

### Relevant log output

```shell

```

ccchdro:
Interestingly, only the FDroid version is affected, not the Google Play version.

monkeyotg:
Same here
Phone: Pixel 7, Pixel 8, Pixel 9, running GrapheneOS: 2026021201
Node: Heltec T114 and V3 with firmware: 2.7.15
Meshtastic APP: version 2.7.13 fdroid

laundmo:
I have the same issue, below is the crash log. It seems the fdroid build tries to use a google library for QR code.

<details><summary>Details</summary>
<p>

```
type: crash
osVersion: google/cheetah/cheetah:16/BP4A.251205.006/2026030701:user/release-keys
flags: dev options enabled
package: com.geeksville.mesh:29320069, targetSdk 36
process: com.geeksville.mesh
processUptime: 6158 + 389 ms
installer: dev.imranr.obtainium.fdroid

[...]

PR #4630, opened 2026-02-24
Align FDroid MapView constructor with Google version (Issue #4576)

ujade:
Partial fix for #4576. The MapView constructor in the FDroid does not share a function definition with the Google variant which causes the app to crash (No function with that definition found) when opening the map from a traceroute. Could not reproduce scanning QR codes causing a crash. 

Tested on Google Pixel 9a running GrapheneOS 2026021200


jamesarich:
@ujade you'll need to sign the CLA before we can accept this PR, thanks!

ujade:
> @ujade you'll need to sign the CLA before we can accept this PR, thanks!

Missed this- signed.

jamesarich:
Thanks for the contrib - do reach out on discord if you want to continue contribution, we can get you better looped in.

PR #4693, opened 2026-03-03
Update strings.xml

MGJ520:
Fix incorrect Chinese translation


jamesarich:
Translations can be updated on Crowdin

Issue #4969, opened 2026-04-01
[Feature Request]: Make MQTT and UDP hops in traceroutes more clear.


domusonline:
### Checklist

- [x] I have used the search function for **OPEN ISSUES** to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for **CLOSED ISSUES** to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.

- [x] I will describe the request with as much detail as possible.

- [x] This request contains only one single feature, **not** a list of multiple (related) features.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Contact Details

domusonline

### Feature or improvement you want

There's a pending PR for the firmware ( https://github.com/meshtastic/firmware/pull/10046 ) that hopes to end the ambiguity of showing 0.0dB on traceroutes for MQTT (and UDP) hops.
It was suggested, and agreed by the PR author (me), that it would be better to implement the application side changes also to minimize the confusion that could be caused by changing the firmware without changing the APP(s).

The proposed change for the firmware will save specific SNR for each of the currently known situations:
-128 for unknown hops (already in the firmware and addressed in the APP also)
-127 for signaling MQTT hop
127 for sigaling UDP hop

Ideally the APP should also adapt the display to the two new values showing "MQTT" for -127 and "UDP" for 127.
This is a simple change which I have tested myself. I believe I can propose a PR assuming there isn't any negative decision on this request. To get a graphic idea of the objective you can check the screenshot below

 

### Why should this be added?

The current implementation will show 0.0dB which in fact is a very "normal" RF level, so the displayed information is ambiguous.
Note that with the zero hop policy in the default MQTT server this is probably not an issue, but different local meshes tend to use customized MQTT servers which may allow being transversed by Traceroutes.
Also note that with the change in the firmware PR, if there is no change in the APP we will start to see traces that show -31.75dB (for MQTT) and 31.75dB (for UDP). Although these are not reasonable values (much less reasonable than 0.0dB at least), these values may cause confusion to users, thus this request attempts to minimize that.

### Screenshots / Drawings / Technical details

![Image](https://github.com/user-attachments/assets/ee9e2dd3-a3e7-489d-b214-499d1a5364cb)

PR #4974, opened 2026-04-02
Implements feature #4969 Make MQTT and UDP hops clear in traceroute

domusonline:
This PR implements feature reques meshtastic/design#68. The idea is very simple: On a traceroute display the hops using MQTT and UDP more clearly. Instead of the current 0.0dB value which is actually a valid RF value it should show "MQTT" or "UDP".

This is a companion feature for the firmware which was already proposed through a PR. There isn't yet a consensus of what should be implemented first. The firmware proposed feature will use 2 additional "flag" values for MQTT and UDP hops in the SNR array.
This is similar to the current value (-128) used for "unknown" and should be -127 for MQTT and 127 for UDP. The idea is to use values as far as possible from the acceptable range of SNR in LoRa communications.

In order to help decide which one should be implemented first (I favor the APPs) let me explain what will happen:
1- If the firmware feature is implemented, with the current application behavior, users will see an SNR of -31.75 for MQTT and 31.75 for UDP. This could be mistakenly assumed as RF signal's SNR. But that already happens with the 0.0dB.
2- If the APP implements this current PR, for every node that is not updated to the firmware with the new feature, users will continue to see the "0.0dB". But a user with the new application will never see the -/+ 31.75 value. So it will be less confusing
Because 2) will make things better I would prefer to see the feature implemented first in the APP. Naturally there's always the possibility that a user will not update the APP and executes a trace that goes through a node with the firmware that implements the feature, and in that case they'll see -/+ 31.75. But I believe people tend to update the APPs more frequently than the nodes.

The change was very simple and affects a single file. The image showing the new behavior is already in the feature request.

### DIFF
--- a/core/model/src/commonMain/kotlin/org/meshtastic/core/model/RouteDiscovery.kt
+++ b/core/model/src/commonMain/kotlin/org/meshtastic/core/model/RouteDiscovery.kt
@@ -66,8 +66,13 @@ private fun formatTraceroutePath(nodesList: List<String>, snrList: List<Int>): S
             List(nodesList.size - 1) { -128 }
         }
             .map { snr ->
-                val str = if (snr == -128) "?" else "${snr / 4f}"
-                "⇊ $str dB"
+                val str = when (snr) {
+                    -128 -> "? dB"
+                    -127 -> "MQTT"
+                    127 -> "UDP"
+                    else -> "${snr / 4f} dB"
+                }
+                "⇊ $str"
             }
 
     return nodesList


jamesarich:
I'm extremely hesitant to rely on magic number values and app-level mitigations/accomodations that hide the underlying data - this should be added to the protocol/protobuf and device layer so that all platforms are aligned.

domusonline:
> I'm extremely hesitant to rely on magic number values and app-level mitigations/accomodations that hide the underlying data - this should be added to the protocol/protobuf and device layer so that all platforms are aligned.

I understand the concern. But there are some reasons behind this:
1- Most important, the size of the traceroute discovery. Currently it can cope with 8 hops. Each one taking 5 byte. Times 2 (towards and back). And with the feature that doesn't discount hops between R/RL/CB roles (with favorites) we may risk not showing all the hops A recent discussion about increasing this has taken place in discord. Meaning that anything we do will increase that payload (or reduce the maximum number of hops we can fit)
2- Tight with 1), bigger messages will have higher risk of no traveling through the mesh
3- I don't usually like to use this argument, but the fact is that we're already using it like this (-128). The extreme value signals unknown. And the apps adapted to it (they show "? db".
4- I would understand the risk of doing this if the values used would be within the normal range for SNR. But they aren't.
5- The current situation is the worse. Because we may get hops with 0.0dB which is a valid RF value, and although the applications didn't adapt to it, it's clearly a (baddly chosen) "flag value"

So, in summary, yes this is a compromise to have the payload simple and short. We're already compromising with the "-128" and "0" SNRs, and the "trick" to have the SNR (float) represented in a single byte. I'm just proposing to maintain that compromise but increase clarity.

PR #5209, opened 2026-04-22
feat(firmware): nRF52 BLE Legacy DFU support

jamesarich:
Adds an in-tree, KMP-friendly Legacy DFU client (Nordic SDK 7+ protocol) for OTA firmware updates over BLE on nRF52 devices that ship the legacy buttonless DFU service. Replaces our prior dependency on the unmaintained Nordic Android-DFU-Library for these devices, restores DFU on nRF52 builds, and lifts throughput ~5× on OTAFIX 2.1+ bootloaders (~10 KiB/s on RAK4631; a 745 KB image now uploads in ~75 s instead of ~6 min).

### Changes
- Implement Nordic Legacy DFU upload protocol: control-point/packet-point split, image-size negotiation, init-packet bracketing, PRN flow control, response & notification parsing.
- Fall back to the legacy buttonless DFU service when the secure buttonless characteristic is absent, sending the spec-compliant two-byte enter-bootloader payload.
- Unify the secure/legacy buttonless trigger paths behind a shared scaffold; surface a bond-loss diagnostic when the trigger times out (typical after a successful DFU rotates BLE keys; user-side fix is Forget + Re-pair).
- Cancel the in-flight DFU stream when the BLE link drops mid-upload instead of hanging on packet writes.
- Cap Legacy DFU packet size at the 20-byte legacy-spec default, with opt-in MTU expansion to 244 B gated on the OTAFIX 2.1+ `_DFU` advertising suffix.
- Bump PRN window to 30 packets and request high BLE connection priority for the duration of the streaming profile; cap the streaming profile timeout at 15 minutes.
- Expose `requestHighConnectionPriority()` on `BleConnection` (Android-only effect; no-op on iOS/JVM).
- Unit coverage for the Legacy protocol parser and transport (init-packet, PRN under-report, byte-count mismatch); update Secure transport tests for the unified trigger.

### Verification
- `./gradlew :feature:firmware:spotlessApply :feature:firmware:detekt :feature:firmware:jvmTest` ✅
- End-to-end DFU on RAK4631 (OTAFIX 2.1) and a pre-OTAFIX nRF52 board ✅

### DIFF
--- a/core/ble/src/androidMain/kotlin/org/meshtastic/core/ble/KablePlatformSetup.kt
+++ b/core/ble/src/androidMain/kotlin/org/meshtastic/core/ble/KablePlatformSetup.kt
@@ -66,3 +66,10 @@ internal actual fun Peripheral.negotiatedMaxWriteLength(): Int? {
     val mtu = (this as? AndroidPeripheral)?.mtu?.value ?: return null
     return (mtu - ATT_HEADER_SIZE).takeIf { it > 0 }
 }
+
+internal actual fun Peripheral.requestHighConnectionPriority(): Boolean {
+    val androidPeripheral = this as? AndroidPeripheral ?: return false
+    return runCatching { androidPeripheral.requestConnectionPriority(AndroidPeripheral.Priority.High) }
+        .onFailure { Logger.w(it) { "requestConnectionPriority(High) threw" } }
+        .getOrDefault(false)
+}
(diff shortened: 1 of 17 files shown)


jamesarich:
@copilot apply changes based on the comments in [this thread](https://github.com/meshtastic/Meshtastic-Android/pull/5209#pullrequestreview-4155330529)

jamesarich:
## Review Feedback Addressed

Pushed commit 085e9d11f addressing all three review comments:

✅ **LegacyDfuTransport.kt:149** — Changed `runCatching` to `safeCatching` to allow proper cancellation propagation in suspend context. This prevents DFU cancellations from being silently swallowed.

✅ **DfuUploadTransport.kt:30** — Updated KDoc for `transferInitPacket` to clarify that implementations may defer sending/validation until `transferFirmware`. Legacy protocol requires deferred init transfer, so this reflects the shared contract accurately.

✅ **SecureDfuTransportTest.kt:105** — Updated inline comment from misleading "0x01 (START_DFU)" to match actual payload "[0x01, 0x04]" to improve test clarity.

PR #5305, opened 2026-04-30
feat:(Wear-OS) Wear-OS implementation with messages

coltarch:
Implementation of the Meshtastic Wear OS application, including core messaging functionality, data synchronization with the mobile app, and notification management.

**Wear OS Implementation**
• Initial implementation of the Wear OS client.
• Implementation of OLED-optimized pure black theme (default) and an "Alternate colors" Grey theme.
• Developed node list interface with real-time status indicators and battery levels.
• Fixed Composable scope issues in settings and cross-platform time utility regressions.

**Synchronization Logic**
• Implemented data layer synchronization for nodes, channels, and messages between phone and watch.
• Normalized contact keys to ensure consistent message threading and prevent duplication.
• Resolved thread naming issues; DM titles and avatars now resolve correctly from the node database.
• Implemented local node filtering to exclude the self-node from messaging lists and placeholders.

**Notification Features**
• Integrated "Open on Phone" notification action to launch the mobile app directly into specific conversations.
• Implemented heuristic-based quick replies (e.g., Yes/No for questions) and enabled support for system-generated smart replies.

This is all a work in progress and I will keep adding features as time goes on. :)


jamesarich:
This is looking AI assisted (not that I mind, just need to be extra careful), and touching a few files it probably shouldn't be (translation strings for one) - ensure that your changes are contained to the feature.

icarus-dealmaker:
@coltarch 
Does it requires phone to run or this would work independently when phone is not around ? 
Also what is the minimal version of WatchOS for this one ?

coltarch:
@icarus-dealmaker 
The version of the app I am working on now relies on the Ble connection from the phone to the watch, I do however plan to make a standalone app in the future. Though that is a far ways out.

As for the recommended version of WearOS, I am not sure and ill have to figure that out.

PR #5426, opened 2026-05-12
build: conditionalize JitPack repository for desktop-only builds

jamesarich:
build: conditionalize JitPack repository for desktop-only builds

- Remove JitPack from `pluginManagement` and `dependencyResolutionManagement` in `build-logic/settings.gradle.kts`.
- Relocate the `desktopOnly` property declaration in the root `settings.gradle.kts` to allow its use during repository configuration.
- Wrap the JitPack repository in `dependencyResolutionManagement` with a `!desktopOnly` check to prevent dependency resolution attempts when building in Android-less environments (e.g., Flatpak).

### DIFF
--- a/build-logic/settings.gradle.kts
+++ b/build-logic/settings.gradle.kts
@@ -20,12 +20,6 @@ pluginManagement {
         google()
         mavenCentral()
         gradlePluginPortal()
-        maven {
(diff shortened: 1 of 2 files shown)


vidplace7 (approved):
Probably the better approach here 👍 GG

PR #5477, opened 2026-05-18
fix(settings): add input validation for BLE PIN, LoRa modem, and ambient lighting

jamesarich:
## Motivation

A cross-platform settings validation audit (meshtastic/design `standards/audits/settings-validation-matrix.md`) identified several Android settings fields that accept invalid input silently -- either discarding it with no user feedback or allowing out-of-range values to be written to the device.

## Changes

**BLE fixed_pin (leading-zero bug):** The PIN field used `EditTextPreference(Int)` which converts text to a UInt internally, stripping leading zeros. Entering "000100" silently became `100`, failed the 6-digit length check, and was discarded with no error shown. Now treated as a string input with `KeyboardType.NumberPassword`, preserving leading zeros and showing an error indicator for incomplete/invalid PINs.

**LoRa spread_factor and coding_rate:** Previously accepted any unsigned integer. Now rejects values outside the valid ranges (SF: 7-12, CR: 5-8) and shows an error indicator when the current value is out of range.

**Ambient lighting current and RGB:** Previously accepted any unsigned integer for LED current (valid: 0-31) and RGB channels (valid: 0-255). Now enforces these ranges with error indicators.

## Approach

All fixes use the same pattern: reject out-of-range values in the `onValueChanged` callback (preventing invalid writes) and pass `isError = true` to the text field when the current value is invalid (giving visual feedback). Extracted helper composables where needed to stay within detekt's method-length and cyclomatic-complexity thresholds.

### DIFF
--- a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/component/AmbientLightingConfigItemList.kt
+++ b/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/component/AmbientLightingConfigItemList.kt
@@ -21,6 +21,7 @@ import androidx.compose.material3.CardDefaults
 import androidx.compose.material3.HorizontalDivider
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
+import androidx.compose.ui.focus.FocusManager
 import androidx.compose.ui.platform.LocalFocusManager
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import org.jetbrains.compose.resources.stringResource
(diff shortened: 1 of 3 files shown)


mbenitog:
Why is the "valid range" for SF 7-12?

Firmware and most radios support SF 5-12:
https://github.com/meshtastic/firmware/pull/10160

SF 5 and 6 should be re-added here and to the iOS app too

jamesarich:
> Why is the "valid range" for SF 7-12?
> 
> Firmware and most radios support SF 5-12:
> https://github.com/meshtastic/firmware/pull/10160
> 
> SF 5 and 6 should be re-added here and to the iOS app too

Feel free to open the appropriate issues for discussion.

PR #5599, opened 2026-05-25
refactor(flatpak): retire heuristic cache scanner, adopt build-ops capture

jamesarich:
## Summary
Replaces the old heuristic-based Flatpak source manifest generator (`build-logic/flatpak/`) with a new `BuildOperationListener`-based approach (`build-logic/flatpak-ops/`).
### Problem
The previous plugin (`meshtastic.flatpak`) scanned the Gradle dependency cache after an assemble and tried to reverse-engineer the download URL for each cached artifact using heuristics (group prefix to repo mapping). This was fragile:
- New repositories required manual `MavenRepoSpec` entries
- Group-to-repo mapping was wrong for artifacts mirrored across repos
- Snapshot metadata resolution added network calls and complexity (~313 LoC task)
### Solution
The new `meshtastic.flatpak-ops` plugin attaches a `BuildOperationListener` that captures every `ExternalResourceReadBuildOperation` URL Gradle downloads. The URL is authoritative — no heuristics needed. The on-disk file is located via Gradle's `files-2.1` layout for SHA-256 computation.
**Key design decisions:**
- CI runs with `--no-build-cache` and an isolated `gradle.user.home` so every artifact traverses the network (listener only fires on cache misses)
- `mustRunAfter(":desktopApp:assemble")` ensures downloads happen before the manifest is written
- Group is derived from the URL path (not the cache directory) to handle collisions correctly
- Uses `ConcurrentHashMap.newKeySet()` for thread-safe URL collection
### Changes
| Area | What changed |
|------|-------------|
| `build-logic/flatpak-ops/` | New plugin — 184 lines vs 397 lines removed |
| `build-logic/flatpak/` | Deleted entirely (extension, plugin, task, baseline, README) |
| CI workflows | Updated to use `:captureFlatpakSources` with isolated Gradle home |
| `scripts/verify-flatpak/` | New local Docker-based verification harness |
### Testing
- [ ] CI runs `:desktopApp:assemble :captureFlatpakSources` and produces `flatpak-sources.json`
- [ ] Local verification via `scripts/verify-flatpak/verify.sh` validates the manifest against `flatpak-builder`


.github/workflows/release.yml
kelcostamonje-ux, line 354: the version is ubuntu 24.04 lts

kelcostamonje-ux: yes?

jamesarich: ?

PR #5752, opened 2026-06-08
fix(notifications): open node detail when tapping 'New Node Seen' notification

LesterCheng:
## Bug

The "New Node Seen" notification does nothing when tapped.

## Root cause

`NodeManagerImpl.handleReceivedUser` dispatches the notification through the generic KMP path:

```kotlin
notificationManager.dispatch(Notification(title = ..., body = ...))
```

`AndroidNotificationManager.dispatch` builds a `NotificationCompat.Builder` but **never calls `setContentIntent(...)`**, so taps are no-ops.

A dedicated `MeshServiceNotificationsImpl.showNewNodeSeenNotification(node)` already wires the correct deep link via `createOpenNodeDetailIntent(nodeNum)` → `/nodes/{nodeNum}`, but it is dead code — nothing in production invokes it. (Likely regressed when notifications were reworked to go through the generic `NotificationManager.dispatch` path.)

## Fix

Smallest KMP-friendly change: enrich the generic `Notification` model with an optional `deepLinkUri` and have the Android dispatcher synthesize a `PendingIntent` when present.

1. **`core/repository/.../Notification.kt`** — add `val deepLinkUri: String? = null`.
2. **`core/data/.../NodeManagerImpl.kt`** — set `deepLinkUri = "meshtastic://meshtastic/nodes/${next.num}"` and `id = next.num` (so repeated emissions for the same node update the existing notification instead of duplicating, matching what `showNewNodeSeenNotification` did).
3. **`core/service/androidMain/.../AndroidNotificationManager.kt`** — when `deepLinkUri != null`, build a `TaskStackBuilder` `PendingIntent` with `ACTION_VIEW` targeting `MainActivity` and attach via `setContentIntent(...)`. Uses `Class.forName("org.meshtastic.app.MainActivity")` (mirroring `MeshServiceNotificationsImpl`) to avoid a `:core:service` → `:androidApp` dependency.

The rest works because the deep-link infrastructure already exists end-to-end:

```
notification tap
  → MainActivity.handleIntent(intent)
  → model.handleDeepLink(uri)
  → DeepLinkRouter.route("/nodes/{n}") → [NodesRoute.Nodes, NodesRoute.NodeDetail(n)]
  → MultiBackstack.handleDeepLink(navKeys)  // switches tab + replaces stack
```

Because `MultiBackstack.handleDeepLink` replaces the active stack with `[Nodes, NodeDetail]`, pressing **back** from `NodeDetail` naturally returns to the **Nodes** list — exactly the requested behavior.

The manifest already declares the URI scheme/host on `MainActivity`:

```xml
<data android:scheme="meshtastic" android:host="meshtastic" />
```

## Tests

`AndroidNotificationManagerTest` — 3 new cases:

- `dispatch attaches deep-link PendingIntent when deepLinkUri is set` — asserts `posted.contentIntent` non-null and the wrapped `Intent` is `ACTION_VIEW` with the correct URI + `MainActivity` component.
- `dispatch leaves contentIntent unset when deepLinkUri is null` — guards against false positives.
- `dispatch uses provided notification id as system id` — verifies explicit-id cancellation works.

Test infrastructure:

- New test-only stub `core/service/src/androidHostTest/.../org/meshtastic/app/MainActivity.kt` extending `Activity` so `Class.forName("org.meshtastic.app.MainActivity")` resolves under Robolectric without pulling `:androidApp` into `:core:service`'s test classpath.
- `registerStubMainActivity()` helper using `shadowOf(packageManager).addOrUpdateActivity(ActivityInfo)` so `TaskStackBuilder.addNextIntentWithParentStack` doesn't throw `NameNotFoundException`.

## Why not just revive `showNewNodeSeenNotification`?

The KMP-friendly fix benefits Desktop too — `DesktopNotificationManager` currently ignores `deepLinkUri` but can honor it in a follow-up. Reviving the dead Android-specific path would have re-coupled `NodeManagerImpl` to the `MeshServiceNotifications` interface and left Desktop with the same broken behavior.

## Why hardcode the URI string in `NodeManagerImpl`?

`DEEP_LINK_BASE_URI` lives in `core/navigation/Routes.kt`, which depends on Compose Navigation 3. Pulling that into `core/data` (a pure data layer) would be a heavy cross-cutting change for one string constant. The literal is documented with a comment pointing at `DEEP_LINK_BASE_URI` so they stay in sync.

## Verification

- `./gradlew :core:repository:allTests` ✓
- `./gradlew :core:data:allTests` ✓
- `./gradlew :core:service:allTests :core:service:testAndroidHostTest :core:service:detekt` ✓
- `./gradlew spotlessApply spotlessCheck detekt assembleDebug` ✓
- `./gradlew test allTests -x :desktopApp:test` ✓ (full repo)

> Note: `LinuxNotificationSenderTest` (3 cases) fails in my agent's container because no libnotify daemon is installed. The same 3 tests fail on `main` in the same environment — unrelated to this change.

## Follow-up (out of scope)

`MeshServiceNotifications.showNewNodeSeenNotification` + its Android implementation can be removed in a cleanup PR now that nothing calls it.


LesterCheng:
Bug report to come, will update here when report has been appropriately filed.

jamesarich:
@LesterCheng please sign CLA

jamesarich (approved):

PR #5782, opened 2026-06-12
feat(node): add local stats noise floor metrics

RCGV1:
## Summary
- Add local stats telemetry to the Signal Quality metrics screen, including noise floor charting and a dashed busy-floor reference line.
- Show local stats log cards with noise floor, traffic, relays, nodes, and uptime, matching the existing Android metrics screen patterns.
- Add Clear, Request, and Save actions for local stats logs, including CSV export and focused repository deletion support.
- Document Local Stats/noise floor behavior and add local-node preview/screenshot coverage for the telemetry actions section.

## Screenshot
Noise floor local stats

## Testing
- `./gradlew spotlessApply`
- `./gradlew :screenshot-tests:updateDebugScreenshotTest`
- `./gradlew detekt :screenshot-tests:validateDebugScreenshotTest`
- `./gradlew test`


RCGV1:
Addressed the advisory comments by updating `docs/en/user/node-metrics.md`, adding a local-node `TelemetricActionsSection` preview, adding screenshot-test coverage for it, and regenerating the affected node screenshot references. Validation now passes with `./gradlew spotlessApply`, `./gradlew detekt :screenshot-tests:validateDebugScreenshotTest`, and `./gradlew test`.

feature/node/src/commonMain/kotlin/org/meshtastic/feature/node/metrics/SignalMetrics.kt
```diff
@@ -44,52 +49,127 @@ import androidx.lifecycle.compose.collectAsStateWithLifecycle
...
+                localStatsData.map { SignalLogEntry.LocalStatsEntry(it) } +
+                    signalData.map { SignalLogEntry.PacketEntry(it) }
+                )
+                .sortedByDescending { it.timeSeconds }
+        }
+    val hasNoiseFloor = remember(localStatsData) { localStatsData.any { it.local_stats?.noise_floor != 0 } }
+    val hasRssi = remember(signalData) { signalData.any { it.rx_rssi != 0 } }
+    val hasSnr = remember(signalData) { signalData.any { !it.rx_snr.isNaN() } }
+    val hasAnyLocalStats = state.localStats.isNotEmpty()
+    val exportLauncher = rememberSaveFileLauncher { uri -> viewModel.saveLocalStatsCSV(uri) }
```
jamesarich: **[Requested change] Dropped RSSI/SNR CSV export + orphaned method.** This launcher now calls `saveLocalStatsCSV`, and the Save button below is gated on `if (hasLocalStats)`. So for a node that only has signal-packet data (RSSI/SNR) and no local stats — historically the common case for this screen — there's no longer any way to export. `MetricsViewModel.saveSignalMetricsCSV` is left with no production caller (only its own test), so detekt won't flag it.

Please make this intentional: either keep the signal-metrics export alongside the local-stats one, or delete `saveSignalMetricsCSV` + its test if dropping it is the goal.

feature/node/src/commonMain/kotlin/org/meshtastic/feature/node/metrics/MetricsViewModel.kt
```diff
@@ -365,6 +369,26 @@ open class MetricsViewModel(
...
     }
 
+    fun saveLocalStatsCSV(uri: CommonUri) {
+        exportCsv(
+            uri = uri,
+            header =
+            "\"date\",\"time\",\"noise_floor_dbm\",\"uptime_seconds\",\"channel_utilization\",\"air_util_tx\"," +
+                "\"packets_tx\",\"packets_rx\",\"bad_rx\",\"rx_dupe\",\"tx_relay\",\"tx_relay_canceled\"," +
+                "\"online_nodes\",\"total_nodes\"\n",
+            rows = state.value.localStats.filter { it.local_stats != null },
```
jamesarich: **[Requested change] Export ignores the selected time frame.** This reads `state.value.localStats` (all history), but the PR description says Save exports 'the visible Local Stats history', and every sibling export (`saveDeviceMetricsCSV`, the old signal export) passes the time-filtered list. Please filter by `timeFrame.timeThreshold()` — or pass the already-filtered `localStatsData` down from the screen — so the CSV matches what's on screen.

core/resources/src/commonMain/composeResources/values/strings.xml
```diff
@@ -1223,10 +1225,15 @@
     <string name="shutdown_warning">⚠️ This will SHUTDOWN the node. Physical interaction will be required to turn it back on.</string>
     <string name="signal">Signal</string>
     <string name="signal_quality">Signal Quality</string>
+    <string name="noise_floor">Noise Floor</string>
```
jamesarich: **[Requested change] Run `python3 scripts/sort-strings.py`.** These strings were inserted next to `signal_quality` instead of in prefix/alpha order — `busy_noise_floor` and the `noise_floor*` keys belong in the `b`/`n` groups — and `strings-index.txt` wasn't regenerated, so the index is now stale. `spotlessCheck` doesn't cover ordering and there's no CI drift guard, so this won't fail the build, but AGENTS.md requires running the sorter after adding strings. The script reorders and regenerates the index in one shot.

feature/node/src/commonMain/kotlin/org/meshtastic/feature/node/metrics/SignalMetrics.kt
```diff
@@ -44,52 +49,127 @@ import androidx.lifecycle.compose.collectAsStateWithLifecycle
...
 
     BaseMetricScreen(
         onNavigateUp = onNavigateUp,
-        telemetryType = TelemetryType.LOCAL_STATS,
+        telemetryType = null,
```
jamesarich, line 152: **[Cleanup] Dead `onRequestTelemetry`.** With `telemetryType = null`, `BaseMetricScreen` never invokes `onRequestTelemetry` (the app-bar refresh button is gated on `telemetryType != null`), so the `onRequestTelemetry = { ... }` arg a few lines down is dead — the bottom 'Request' button is the real path. Drop the arg to avoid the misleading wiring.

feature/node/src/commonMain/kotlin/org/meshtastic/feature/node/metrics/SignalMetrics.kt
```diff
@@ -213,6 +420,97 @@ private fun SignalMetricsChart(
     }
 }
 
+@Composable
+private fun noiseFloorTextColor(value: Int): Color = when {
+    value == 0 -> MaterialTheme.colorScheme.onSurfaceVariant
+    value < QUIET_NOISE_FLOOR_DBM -> SignalMetric.SNR.color
+    value < MODERATE_NOISE_FLOOR_DBM -> Orange
```
jamesarich: **[Cleanup] Card vs. chart threshold mismatch.** The card text turns red at `>= MODERATE_NOISE_FLOOR_DBM` (-90), but the chart's dashed 'busy floor' reference line is at `BUSY_FLOOR_DBM` (-85). So a -87 dBm reading sits below the busy line on the chart (reads as fine) yet shows red on the card. Consider aligning the card's 'busy' threshold with the -85 reference, or add a comment explaining the intentional difference.

core/data/src/commonMain/kotlin/org/meshtastic/core/data/repository/MeshLogRepositoryImpl.kt
```diff
@@ -174,6 +174,21 @@ open class MeshLogRepositoryImpl(
         dbManager.currentDb.value.meshLogDao().deleteLogs(logId, portNum)
     }
 
+    /** Deletes only local stats telemetry logs for [nodeNum], preserving other telemetry types. */
+    override suspend fun deleteLocalStatsLogs(nodeNum: Int) = withContext(dispatchers.io) {
+        val myNodeNum = nodeInfoReadDataSource.myNodeInfoFlow().firstOrNull()?.myNodeNum
+        val logId = if (nodeNum == myNodeNum) MeshLog.NODE_NUM_LOCAL else nodeNum
+        val dao = dbManager.currentDb.value.meshLogDao()
+        val localStatsLogs =
+            dao.getLogsFrom(logId, PortNum.TELEMETRY_APP.value, DEFAULT_MAX_LOGS)
```
jamesarich: **[Cleanup] `Clear` can miss rows and deletes one-by-one.** This loads only the most recent `DEFAULT_MAX_LOGS` telemetry rows (all types) before filtering to local-stats, so under high telemetry volume older local-stats logs can survive a 'Clear'. It also issues one `deleteLog` per row. Consider a scoped DELETE (by portNum + a local-stats predicate) or at least documenting the window limit.

core/testing/src/commonMain/kotlin/org/meshtastic/core/testing/FakeMeshLogRepository.kt
```diff
@@ -82,6 +86,10 @@ class FakeMeshLogRepository :
         logsFlow.value = logsFlow.value.filterNot { it.fromNum == nodeNum && it.portNum == portNum }
     }
 
+    override suspend fun deleteLocalStatsLogs(nodeNum: Int) {
```
jamesarich, line 89: **[Tests] New behavior is unverified.** This fake override only records the node id, so nothing exercises the real `MeshLogRepositoryImpl.deleteLocalStatsLogs` filter (the 'preserve other telemetry types' guarantee). There are also no tests for `saveLocalStatsCSV` or `clearLocalStats`. A `MeshLogRepositoryImpl` test that seeds mixed telemetry (local_stats + device/env) and asserts only local-stats rows are deleted would lock in the contract.

jamesarich (changes requested):
Requesting changes. First, credit where due: I checked out this branch and built it locally (JDK 21) — `spotlessCheck`, `detekt`, `:feature:node:allTests`, and `:core:data:allTests` all pass, imports are clean, and the `stringResource` format-arg counts all line up. The feature is a nice addition. The requests below are design/cleanup, not breakage.

**Should-fix (blocking):**
1. RSSI/SNR CSV export is silently dropped and `saveSignalMetricsCSV` is now orphaned.
2. `saveLocalStatsCSV` exports all history, ignoring the selected time frame, despite the description saying 'visible'.
3. `scripts/sort-strings.py` wasn't run — new strings are mis-sorted and `strings-index.txt` is stale.

**Cleanup (non-blocking):** dead `onRequestTelemetry` arg, card/chart noise-floor threshold mismatch, and the `Clear` deletion window + row-by-row delete.

**Tests:** no test covers `saveLocalStatsCSV`, `clearLocalStats`, or `deleteLocalStatsLogs`; the `FakeMeshLogRepository` override only records the node id, so the 'preserve other telemetry types' filter is unverified.

Inline comments below.

RCGV1:
Addressed the requested review items in ee7426f08:

- Restored RSSI/SNR CSV export and kept Local Stats export separate.
- Local Stats CSV export now uses the visible timeframe-filtered rows passed by the screen.
- Re-ran scripts/sort-strings.py and refreshed strings-index.txt.
- Removed the dead telemetry request app-bar hook, aligned the noise-floor card threshold with the chart busy floor, and tightened clear to bulk-delete all matching local-stats telemetry logs.
- Added coverage for Local Stats CSV export, clearLocalStats, and deleteLocalStatsLogs preserving other telemetry.

Validation run locally:
- ./gradlew spotlessApply
- ./gradlew detekt
- ./gradlew :core:data:jvmTest
- ./gradlew :feature:node:jvmTest --rerun-tasks
- ./gradlew test

Note: ./gradlew test completed successfully but printed existing Robolectric temp-directory cleanup stack traces after androidApp unit tests; Gradle still reported BUILD SUCCESSFUL.

PR #5795, opened 2026-06-14
fix(ble): Harden BLE connection lifecycle

jeremiah-k:
## Overview

This PR improves BLE connection stability in long-running sessions by tightening exception classification, session cleanup, reconnect triggering, selected-node discovery, reconnect UX, and liveness recovery. The changes focus on cases where BLE operations can fail after an established connection has been active for some time, leaving stale GATT/profile state behind and requiring manual node reselection. Fatal BLE conditions now trigger bounded cleanup and reconnect paths, selected bonded nodes prefer fresh advertisements when available, and recoverable BLE failures avoid user-facing modal errors while automatic reconnect continues.

### Fixes & Stability Improvements
- **Session-fatal BLE classification**: Added `isSessionFatalBleException()` to identify unrecoverable BLE session failures, including `NotConnectedException` and fatal GATT statuses `8`, `19`, `22`, `62`, `129`, and `133`, with bounded cause-chain traversal.
- **BLE profile exception propagation**: Updated `KableMeshtasticRadioProfile` so session-fatal exceptions from `fromRadio` and `logRadio` propagate to the transport layer and can trigger reconnect, while non-fatal diagnostic/log failures remain suppressed.
- **FROMNUM subscription readiness**: Ensured pre-readiness FROMNUM observe failures complete `subscriptionReady` exceptionally, and added bounded subscription readiness handling so setup does not continue with half-initialized BLE state.
- **Session failure deduplication**: Added atomic session-failure guards in `BleRadioTransport` to ensure racing read/write/profile failures only trigger one disconnect callback and one cleanup path per failed session.
- **Stale BLE state cleanup**: Fatal session failures now clear stale `radioService` / fully-connected state and force GATT disconnect so the reconnect loop can proceed instead of waiting on stale state.
- **Write-path hardening**: Reworked BLE write handling to re-check the active profile under the write mutex, ignore stale-session writes, and treat post-retry write failures as fatal to the current BLE session.
- **Connection gate timeout**: Added bounded handling for connection-state synchronization so missing `Connected` state transitions clean up stale GATT state and retry instead of hanging.
- **Selected-node discovery**: Added a short targeted scan for the selected BLE address before falling back to bonded address-only connection data, allowing fresh advertisement-backed connects when available without adding periodic/background scanning.
- **Reconnect UX noise reduction**: Suppressed user-facing modal errors for recoverable BLE reconnect attempts and non-permanent session failures while preserving logs, state updates, reconnect behavior, and permanent-failure reporting.
[...]

PR #5828, opened 2026-06-17
fix(service): resolve selected-device startup race

jeremiah-k:
## Overview

This PR fixes a selected-device startup/lifecycle race where the app could show a connected or connecting radio state while NodeDB, channels, and config data failed to load.

The issue could happen when `MeshService` started before the persisted selected-device address had fully propagated through the service layer. In that window, the orchestrator could initialize against the default/null database or stop the service because no device appeared selected, while transport recovery listeners could still start radio connection work later.

This PR makes selected-device startup ordering explicit, centralizes no-device/sentinel address handling, and gates transport recovery so Bluetooth/network events cannot restart transport after explicit disconnect or device deselection. It also disables Android’s multi-reader Room pool after follow-up testing showed reader acquisition timeouts that could prevent version/config/NodeDB data from finishing loading even after BLE connected.

## Background

These issues surfaced while retargeting the BLE hardening PR (#5795) to `release/2.8.0` (branch has since been merged to `main`). After the BLE reconnect/liveness fixes were running on the release branch, manual testing exposed a separate startup/lifecycle problem: the radio transport could appear connected or connecting, but NodeDB, channels, and config data would not load because selected-device address resolution, per-device DB switching, and service/orchestrator startup were not consistently ordered.

A later test pass also surfaced an Android-specific Room reader-pool issue on one phone: BLE connected, but version/config/NodeDB state kept spinning and logs showed repeated reader acquisition timeouts with the reader pool reporting `capacity=4, permits=0` while all reader connections were listed as `Free`.

This PR addresses those release-branch startup/database issues separately from the BLE transport hardening work.

## Key Changes

**Startup / Lifecycle Fixes**
- `MeshService` now defers the no-device stop decision when the current device address is not yet valid, waits briefly for a valid selected address from `currentDeviceAddressFlow`, and only proceeds once address resolution succeeds.
- `MeshServiceOrchestrator.start()` now waits for a valid selected-device address, switches to that device’s database, loads cached NodeDB from that database, and only then connects the radio transport.
- `SharedRadioInterfaceService` now mirrors selected-device address state independently of `connect()`, so startup code can observe the persisted selected address before transport startup begins.
- Bluetooth/network recovery listeners are now gated by an explicit connection intent flag, preventing unintended transport restarts after explicit `disconnect()` or device deselection.
- `setDeviceAddress(null)` / `setDeviceAddress("n")` now clears the active connection intent and stops transport without allowing later Bluetooth/network recovery events to resurrect the session.

**Database / Address Handling**
- `buildDbName` now normalizes nullable input addresses before database-name generation.
- No-device/sentinel address variants such as `null`, blank, `"n"`, `"null"`, `".n"`, and `"default"` are consistently mapped to `DatabaseConstants.DEFAULT_DB_NAME`.
- No-device sentinel detection is now centralized through a shared predicate used by both `isValidDeviceAddress()` and `buildDbName()`, so foreground-service stay-alive decisions and DB-name resolution cannot drift.
- Real selected-device addresses continue to be hashed under the standard `DB_PREFIX_` per-device database naming scheme.
- Added shared `isValidDeviceAddress(address: String?)` validation so service startup and database selection use the same no-device/sentinel rules.
- Android production now uses single-connection Room behavior instead of Room KMP’s multi-reader connection pool, avoiding the observed reader-pool permit exhaustion where DB-backed version/config/NodeDB reads could time out indefinitely.

**Runtime Safety**
- Active-service address updates are observed and used to switch database state for valid selected addresses, with failures logged so the observer remains alive.
- Environmental recovery still works while a connection is desired: Bluetooth/network restoration can restart transport only when the service is actively maintaining a selected-device connection.
- Explicit disconnect and device deselection now prevent transport recovery from starting a new session later.
- New selected-device INFO logs in `MeshService` anonymize device addresses before logging.

**Service Cleanup**
- `MeshService` now owns a dedicated service scope for deferred address resolution work.
- Pending address-resolution work is cancelled during shutdown.
- Wake lock / foreground-service cleanup remains tied to whether startup actually proceeds with a valid selected device.

## Test Coverage

- `AddressUtilsTest`
  - Directly verifies `isValidDeviceAddress()` accepts real device addresses.
  - Directly verifies `isValidDeviceAddress()` rejects no-device sentinel variants such as null, blank, `"n"`, `"null"`, `".n"`, and `"default"`.

- `BuildDbNameTest`
  - Covers sentinel-to-default database mapping.
  - Covers case/format normalization for BLE MAC-style addresses.
  - Verifies real device addresses produce non-default per-device DB names under `DB_PREFIX_`.

- `MeshServiceOrchestratorTest`
  - Verifies cold-start ordering: switch active DB → load cached NodeDB → connect radio.
  - Verifies address changes during an active orchestrator session switch active DB.
  - Verifies a stopped orchestrator is not restarted just because connection state later reports connected.

- `SharedRadioInterfaceServiceLivenessTest`
  - Verifies explicit `disconnect()` prevents later BLE recovery from restarting transport.
  - Verifies explicit `disconnect()` prevents later network recovery from restarting transport.
  - Verifies Bluetooth/network environmental recovery still restarts transport while a connection is desired.
  - Verifies `setDeviceAddress("n")` / device deselection clears connection intent and prevents later recovery restarts.

## Breaking Changes / Migration Notes

No database migration, destructive cleanup, or user-data deletion is added.

The only database-selection behavior change is that no-device/sentinel address variants are now normalized consistently to the default database name. If older builds accidentally created separate DB names for sentinel variants such as `"default"` or `".n"`, those variants will now resolve to the default DB instead of preserving separate sentinel-specific database files.

Android production DB access now uses single-connection Room behavior for release reliability. This may reduce DB read concurrency on Android, but avoids the observed reader-pool permit exhaustion where NodeDB/version/config reads could time out indefinitely.

BLE reconnect/liveness behavior, selected-node scanning, and TCP/serial liveness behavior are not intentionally changed.

---

I've loaded this on multiple phones for testing and everything is working well for now. I will update the status later.

Update: After ~2 hrs of testing so far, everything is running well. Device selection seems more responsive too.

Update 2: One phone later connected over BLE but never finished loading version/config/NodeDB data. Logs showed repeated Room reader acquisition timeouts with the reader pool at `capacity=4, permits=0` while all reader connections were reported `Free`. I added an Android-only DB reliability fix so production Android now uses single-connection Room behavior instead of the multi-reader pool. Testing has resumed with that fix applied.

Update 3: Addressed maintainer review nits: selected-device INFO logs are now anonymized, no-device sentinel detection is shared between `isValidDeviceAddress()` and `buildDbName()`, and direct `isValidDeviceAddress()` tests were added for real addresses and sentinel variants.

Continued in #5841

### DIFF
--- a/core/common/src/commonMain/kotlin/org/meshtastic/core/common/util/AddressUtils.kt
+++ b/core/common/src/commonMain/kotlin/org/meshtastic/core/common/util/AddressUtils.kt
@@ -28,3 +28,14 @@ fun normalizeAddress(addr: String?): String {
         else -> u.replace(":", "")
     }
 }
+
+/**
+ * True iff [address] refers to a real selected device. Rejects null/blank and all legacy no-device sentinels (`"n"`,
+ * `"null"`, `".n"`, `"default"`, case-insensitive) by delegating to [normalizeAddress]. Any input that [buildDbName]
+ * would collapse to `DEFAULT_DB_NAME` is rejected here too, so the foreground-service stay-alive decision and the DB
+ * name resolution can never diverge.
+ */
+fun isValidDeviceAddress(address: String?): Boolean {
+    val normalized = normalizeAddress(address)
+    return normalized != "DEFAULT" && normalized != ".N"
+}
--- a/core/common/src/commonMain/kotlin/org/meshtastic/core/common/util/AddressUtils.kt
+++ b/core/common/src/commonMain/kotlin/org/meshtastic/core/common/util/AddressUtils.kt
@@ -30,12 +30,15 @@ fun normalizeAddress(addr: String?): String {
 }
 
 /**
- * True iff [address] refers to a real selected device. Rejects null/blank and all legacy no-device sentinels (`"n"`,
- * `"null"`, `".n"`, `"default"`, case-insensitive) by delegating to [normalizeAddress]. Any input that [buildDbName]
- * would collapse to `DEFAULT_DB_NAME` is rejected here too, so the foreground-service stay-alive decision and the DB
- * name resolution can never diverge.
+ * True iff [normalized] (the output of [normalizeAddress]) is a no-device sentinel that DB naming collapses to the
+ * default database. Single source of truth for "no real device selected".
  */
-fun isValidDeviceAddress(address: String?): Boolean {
-    val normalized = normalizeAddress(address)
-    return normalized != "DEFAULT" && normalized != ".N"
-}
+fun isNoDeviceSentinel(normalized: String): Boolean = normalized == "DEFAULT" || normalized == ".N"
+
+/**
+ * True iff [address] refers to a real selected device. Rejects null/blank and all legacy no-device sentinels (`"n"`,
+ * `"null"`, `".n"`, `"default"`, case-insensitive) by delegating to [normalizeAddress] and [isNoDeviceSentinel]. Any
+ * input that [buildDbName] would collapse to `DEFAULT_DB_NAME` is rejected here too, so the foreground-service
+ * stay-alive decision and the DB name resolution can never diverge.
+ */
+fun isValidDeviceAddress(address: String?): Boolean = !isNoDeviceSentinel(normalizeAddress(address))
--- a/core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseConstants.kt
+++ b/core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseConstants.kt
@@ -17,6 +17,7 @@
 package org.meshtastic.core.database
 
 import okio.ByteString.Companion.encodeUtf8
+import org.meshtastic.core.common.util.isNoDeviceSentinel
 import org.meshtastic.core.common.util.normalizeAddress
 
 object DatabaseConstants {
@@ -47,11 +48,9 @@ fun buildDbName(address: String?): String {
     val normalized = normalizeAddress(address)
     // No-device sentinels must resolve to the canonical DEFAULT_DB_NAME file, not be hashed into a
     // separate DB — otherwise data for "no device selected" would split across files depending on
-    // which sentinel form prefs happened to emit. normalizeAddress collapses null/blank/"N"/"NULL"
-    // to "DEFAULT"; ".N" (any case) is also collapsed here as a defensive legacy/sentinel guard.
-    // There is no current producer of ".n" in the codebase, but the rule keeps buildDbName
-    // self-consistent regardless of input shape.
-    if (normalized == "DEFAULT" || normalized == ".N") return DatabaseConstants.DEFAULT_DB_NAME
+    // which sentinel form prefs happened to emit. Sentinel detection is shared with isValidDeviceAddress
+    // via isNoDeviceSentinel so DB naming and the foreground-service stay-alive decision never diverge.
+    if (isNoDeviceSentinel(normalized)) return DatabaseConstants.DEFAULT_DB_NAME
     return "${DatabaseConstants.DB_PREFIX}_${shortSha1(normalized)}"
 }
 
--- a/core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseConstants.kt
+++ b/core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseConstants.kt
@@ -43,10 +43,16 @@ object DatabaseConstants {
 
 fun shortSha1(s: String): String = s.encodeUtf8().sha1().hex().take(DatabaseConstants.DB_NAME_HASH_LEN)
 
-fun buildDbName(address: String?): String = if (address.isNullOrBlank()) {
-    DatabaseConstants.DEFAULT_DB_NAME
-} else {
-    "${DatabaseConstants.DB_PREFIX}_${shortSha1(normalizeAddress(address))}"
+fun buildDbName(address: String?): String {
+    val normalized = normalizeAddress(address)
+    // No-device sentinels must resolve to the canonical DEFAULT_DB_NAME file, not be hashed into a
+    // separate DB — otherwise data for "no device selected" would split across files depending on
+    // which sentinel form prefs happened to emit. normalizeAddress collapses null/blank/"N"/"NULL"
+    // to "DEFAULT"; ".N" (any case) is also collapsed here as a defensive legacy/sentinel guard.
+    // There is no current producer of ".n" in the codebase, but the rule keeps buildDbName
+    // self-consistent regardless of input shape.
+    if (normalized == "DEFAULT" || normalized == ".N") return DatabaseConstants.DEFAULT_DB_NAME
+    return "${DatabaseConstants.DB_PREFIX}_${shortSha1(normalized)}"
 }
 
 fun anonymizeAddress(address: String?): String = when {
--- a/core/database/src/commonTest/kotlin/org/meshtastic/core/database/BuildDbNameTest.kt
+++ b/core/database/src/commonTest/kotlin/org/meshtastic/core/database/BuildDbNameTest.kt
@@ -1,0 +1,56 @@
+/*
+ * Copyright (c) 2026 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.core.database
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotEquals
+import kotlin.test.assertTrue
+
+class BuildDbNameTest {
+    @Test
+    fun `no-device sentinels all resolve to DEFAULT_DB_NAME`() {
+        val expected = DatabaseConstants.DEFAULT_DB_NAME
+        val sentinels =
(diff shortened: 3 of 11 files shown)


core/service/src/androidMain/kotlin/org/meshtastic/core/service/MeshService.kt
```diff
@@ -27,8 +27,17 @@ import android.os.IBinder
...
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withTimeoutOrNull
 import org.koin.android.ext.android.inject
 import org.meshtastic.core.common.hasLocationPermission
+import org.meshtastic.core.common.util.isValidDeviceAddress
 import org.meshtastic.core.model.DeviceVersion
```
jamesarich, line 41: Add the `anonymize` extension import for the log-anonymization suggestions below (ktlint sorts it right after `DeviceVersion`):

```suggestion
import org.meshtastic.core.model.DeviceVersion
import org.meshtastic.core.model.util.anonymize
```

core/service/src/androidMain/kotlin/org/meshtastic/core/service/MeshService.kt
```diff
@@ -114,18 +139,56 @@ class MeshService : Service() {
...
-            START_NOT_STICKY
-        } else {
+        val address = radioInterfaceService.getDeviceAddress()
+        if (isValidDeviceAddress(address)) {
+            // Address is already loaded and valid — proceed normally.
+            addressWaitJob?.cancel()
+            addressWaitJob = null
             acquireWakeLock()
-            START_STICKY
+            Logger.i { "MeshService: selected device ready ($address), staying foreground" }
```
jamesarich: 🔒 **Privacy (should-fix):** this logs the raw device address (BLE MAC / IP — PII) at INFO, against AGENTS.md's *Privacy First — never log PII* rule and this file's own neighbors (`SharedRadioInterfaceService` logs `${address.anonymize}`, `DatabaseManager` uses `anonymizeAddress`).

```suggestion
            Logger.i { "MeshService: selected device ready (${address.anonymize}), staying foreground" }
```
Needs `import org.meshtastic.core.model.util.anonymize` — applied via the import suggestion above.

core/service/src/androidMain/kotlin/org/meshtastic/core/service/MeshService.kt
```diff
@@ -114,18 +139,56 @@ class MeshService : Service() {
...
+            return START_STICKY
         }
+
+        // Address is currently null/blank/sentinel. This may be a transient state while DataStore emits the persisted
+        // value (cold-start race: RadioPrefsImpl.devAddr starts as null until the first DataStore emission). Make no
+        // irreversible stopSelf() decision here; wait briefly for the address flow to settle.
+        Logger.i { "MeshService: selected address not yet loaded ($address); waiting for address flow" }
```
jamesarich: 🔒 Same PII concern — anonymize the address:

```suggestion
        Logger.i { "MeshService: selected address not yet loaded (${address.anonymize}); waiting for address flow" }
```

core/service/src/androidMain/kotlin/org/meshtastic/core/service/MeshService.kt
```diff
@@ -114,18 +139,56 @@ class MeshService : Service() {
...
+    private fun scheduleDeviceAddressResolution() {
+        addressWaitJob?.cancel()
+        addressWaitJob =
+            serviceScope.launch {
+                val resolved =
+                    withTimeoutOrNull(DEVICE_ADDRESS_SETTLE_MS) {
+                        radioInterfaceService.currentDeviceAddressFlow.first(::isValidDeviceAddress)
+                    }
+                if (isValidDeviceAddress(resolved)) {
+                    Logger.i { "MeshService: selected device resolved ($resolved) after address-flow wait" }
```
jamesarich: 🔒 Same PII concern — anonymize the resolved address:

```suggestion
                    Logger.i { "MeshService: selected device resolved (${resolved.anonymize}) after address-flow wait" }
```

core/common/src/commonMain/kotlin/org/meshtastic/core/common/util/AddressUtils.kt
```diff
@@ -28,3 +28,14 @@ fun normalizeAddress(addr: String?): String {
...
+/**
+ * True iff [address] refers to a real selected device. Rejects null/blank and all legacy no-device sentinels (`"n"`,
+ * `"null"`, `".n"`, `"default"`, case-insensitive) by delegating to [normalizeAddress]. Any input that [buildDbName]
+ * would collapse to `DEFAULT_DB_NAME` is rejected here too, so the foreground-service stay-alive decision and the DB
+ * name resolution can never diverge.
+ */
+fun isValidDeviceAddress(address: String?): Boolean {
+    val normalized = normalizeAddress(address)
+    return normalized != "DEFAULT" && normalized != ".N"
+}
```
jamesarich: 🧹 **De-dupe the no-device sentinel set (optional).** The `"DEFAULT"`/`".N"` literals are hand-written here *and* in `buildDbName` (`DatabaseConstants.kt`). The KDoc says they "can never diverge" — but that's only true by convention; add a sentinel to `normalizeAddress` later and these two silently desync. Extract one predicate both call:

```suggestion
/**
 * True iff [normalized] (the output of [normalizeAddress]) is a no-device sentinel that [buildDbName] collapses to
 * `DEFAULT_DB_NAME`. Single source of truth for "no real device selected".
 */
fun isNoDeviceSentinel(normalized: String): Boolean = normalized == "DEFAULT" || normalized == ".N"

/**
 * True iff [address] refers to a real selected device. Rejects null/blank and all legacy no-device sentinels
 * (`"n"`, `"null"`, `".n"`, `"default"`, case-insensitive) via the shared [isNoDeviceSentinel] predicate, so the
 * foreground-service stay-alive decision and DB-name resolution can never diverge.
 */
fun isValidDeviceAddress(address: String?): Boolean = !isNoDeviceSentinel(normalizeAddress(address))
```
Then point `buildDbName` at it (suggestion on `DatabaseConstants.kt`).

jamesarich (commented):
Inline suggestions from the review. **One should-fix** — anonymize the new device-address logs (PII at INFO) — plus **two optional follow-ups**: de-dupe the no-device sentinel set so `isValidDeviceAddress` and `buildDbName` can't drift, and add a direct `isValidDeviceAddress` test. The startup/lifecycle logic itself looks correct; these are hardening + maintainability nits. The de-dupe and its imports are interdependent (noted inline).

core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseConstants.kt
```diff
@@ -43,10 +43,16 @@ object DatabaseConstants {
...
-    "${DatabaseConstants.DB_PREFIX}_${shortSha1(normalizeAddress(address))}"
+fun buildDbName(address: String?): String {
+    val normalized = normalizeAddress(address)
+    // No-device sentinels must resolve to the canonical DEFAULT_DB_NAME file, not be hashed into a
+    // separate DB — otherwise data for "no device selected" would split across files depending on
+    // which sentinel form prefs happened to emit. normalizeAddress collapses null/blank/"N"/"NULL"
+    // to "DEFAULT"; ".N" (any case) is also collapsed here as a defensive legacy/sentinel guard.
+    // There is no current producer of ".n" in the codebase, but the rule keeps buildDbName
+    // self-consistent regardless of input shape.
+    if (normalized == "DEFAULT" || normalized == ".N") return DatabaseConstants.DEFAULT_DB_NAME
```
jamesarich: Use the shared predicate so this can't drift from `isValidDeviceAddress`:

```suggestion
    if (isNoDeviceSentinel(normalized)) return DatabaseConstants.DEFAULT_DB_NAME
```
Also add the import near the top (next to `normalizeAddress`):
```kotlin
import org.meshtastic.core.common.util.isNoDeviceSentinel
```
The comment above still explains the *why*; optionally trim its last two sentences now that the rule lives in one place.

core/database/src/commonTest/kotlin/org/meshtastic/core/database/BuildDbNameTest.kt
```diff
@@ -0,0 +1,56 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.core.database
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNotEquals
+import kotlin.test.assertTrue
+
+class BuildDbNameTest {
```
jamesarich, line 24: ✅ Solid coverage on `buildDbName`. `isValidDeviceAddress` gates two critical paths (foreground stay-alive + orchestrator connect) yet is only exercised indirectly. Worth a direct, parallel test — the natural home is the existing `core/common/src/commonTest/kotlin/org/meshtastic/core/common/util/AddressUtilsTest.kt` — to lock the contract (incl. the `.n`/`default` cases) independently of the DB layer:

```kotlin
@Test
fun `isValidDeviceAddress accepts a real device address`() {
    assertTrue(isValidDeviceAddress("xAA:BB:CC:DD:EE:FF"))
    assertTrue(isValidDeviceAddress("t192.168.1.100"))
}

@Test
fun `isValidDeviceAddress rejects every no-device sentinel`() {
    listOf(null, "", "   ", "n", "N", "null", "NULL", ".n", ".N", "default", "DEFAULT")
        .forEach { assertFalse(isValidDeviceAddress(it), "Sentinel ${it ?: "null"} must be invalid") }
}
```
(`assertTrue`/`assertFalse` from `kotlin.test`.)

jeremiah-k:
> Inline suggestions from the review. **One should-fix** — anonymize the new device-address logs (PII at INFO) — plus **two optional follow-ups**: de-dupe the no-device sentinel set so `isValidDeviceAddress` and `buildDbName` can't drift, and add a direct `isValidDeviceAddress` test. The startup/lifecycle logic itself looks correct; these are hardening + maintainability nits. The de-dupe and its imports are interdependent (noted inline).

These have been addressed.

jamesarich (approved):

PR #5856, opened 2026-06-18
fix(service): recover stalled WiFi/TCP handshakes by cycling active transport

jeremiah-k:
## Overview

This PR improves WiFi/TCP handshake-stall recovery so the app and underlying transport do not get stuck in different connection states.

Previously, if the handshake stalled after its retry window, recovery could leave the app-level connection state and underlying transport out of sync. For TCP, the transport could remain active while the UI reported a disconnected state. In that situation, selecting the same node again could be ignored because the radio service still saw the same selected address and an already-running transport.

This PR adds a targeted transport restart path for stalled handshakes. When recovery is needed, the connection manager first moves app state to `Disconnected`, then asks the radio service to restart the active transport. The fresh transport `Connected` signal can then re-enter the normal handshake flow.

Follow-up WiFi testing also showed that some failed TCP/USB handshake attempts could sit in a loading state until the lower-level TCP timeout recovered them. This PR adds a TCP/USB-only fast handshake watchdog so those stalled handshakes recover sooner, while leaving BLE's longer handshake timing unchanged.

Additional BLE testing exposed a separate post-handshake stuck-connecting path. Once firmware sends `NODE_INFO_NONCE`, the handshake watchdog is intentionally cancelled before local NodeDB install work begins. If that local NodeDB install fails, the app can otherwise remain in `Connecting` with no active watchdog. This PR now recovers from NodeDB install failure by routing through the same reconnecting transport-restart path.

This PR also adds defensive recovery hardening so a crashing or repeatedly failing node is not re-driven indefinitely. Recovery now uses clean transport restart only, applies exponential backoff between repeated attempts, and surfaces a sticky error after repeated unsuccessful recovery cycles.

Addresses: #3727

## Key Changes

**Handshake Recovery**
- Adds `RadioInterfaceService.restartTransport()` for silent recovery of stalled mesh handshakes.
- Updates `MeshConnectionManagerImpl` so retry-exceeded handshake stalls request a transport restart instead of only changing app-level state.
- Adds `MeshConnectionManager.recoverPostHandshakeFailure()` so NodeDB install failures after firmware handshake completion use the same app-level disconnect-before-restart ordering.
- Adds a 12s TCP/USB-only fast handshake watchdog for Stage 1 and Stage 2.
- Resets the fast watchdog only on meaningful handshake progress.
- Cancels the Stage 2 watchdog synchronously when `NODE_INFO_NONCE` arrives, before async NodeDB install work begins.
- Adds a one-way handshake-complete latch so late progress packets cannot re-arm the watchdog after firmware handshake completion.
- Recovers only from pre-`Connected` NodeDB install failure; post-`Connected` analytics / side-effect failures are logged without forcing recovery.
- Orders recovery so the app-level `Disconnected` transition happens before transport restart emissions.
- Runs recovery work in a sibling coroutine so the restart sequence is not lost when the existing handshake timeout job is cancelled.
- Re-checks the current app state before recovery so a late-arriving successful handshake is not restarted unnecessarily.
- Uses atomic watchdog swaps so concurrent re-arms cannot orphan timeout jobs.

**Recovery Hardening**
- Removes the same-session `want_config` retry path.
- Uses clean transport restart for stalled recovery on both fast transports and BLE retry-exhausted paths.
- Adds exponential backoff between consecutive recovery attempts.
- Adds a consecutive recovery cap.
- Surfaces a localized sticky error after repeated unsuccessful recovery attempts instead of silently looping forever.
- Resets the recovery failure counter after successful handshake completion.

**Transport Restart Behavior**
- Implements `SharedRadioInterfaceService.restartTransport()` as an in-place stop/start of the currently active transport.
- Preserves the selected device address and existing connection intent.
- Emits a transient `DeviceSleep` transport transition before restarting so observers can see a real state change before the fresh `Connected` event.
- Keeps recovery quiet from the user's perspective by avoiding a permanent transport `Disconnected` event and avoiding a user-facing error.
- Skips restart when:
  - the user explicitly disconnected,
  - the selected device was cleared,
  - no valid selected address exists,
  - no transport is currently running,
  - another restart or liveness recovery is already in progress.

**Handshake Progress**
- Tracks meaningful Stage 1 progress from MyNodeInfo, local metadata, file info, device config, module config, channels, DeviceUI config, and early NodeInfo.
- Tracks Stage 2 progress from NodeInfo packets.
- Buffers early NodeInfo received during Stage 1 so it is not lost before the node-list phase starts.
- Keeps Stage 2 completion gated on `NODE_INFO_NONCE`.

**Connection UI**
- Adds a transient `Reconnecting…` state while silent handshake recovery is in progress.
- Keeps transient TCP/USB handshake recovery from looking like a final user-facing disconnect.

**Test Coverage**
- Adds `MeshConnectionManagerImplTest` coverage for:
  - Stage 1 config handshake stall recovery,
  - Stage 2 node-info handshake stall recovery,
  - TCP/USB fast recovery timing,
  - meaningful progress resetting the fast watchdog,
  - no fast recovery for BLE,
  - no restart when the handshake completes normally,
  - recovery ordering so app disconnect is processed before restart emissions,
  - post-handshake failure recovery using the same disconnect-before-restart ordering,
  - exponential recovery backoff,
  - consecutive recovery exhaustion surfacing a sticky error,
  - successful handshake completion resetting the recovery counter.

- Adds `MeshConfigFlowManagerImplTest` / `MeshConfigHandlerImplTest` coverage for:
  - meaningful handshake progress signals,
  - early NodeInfo buffering,
  - Stage 2 completion behavior,
  - watchdog cancellation when `NODE_INFO_NONCE` arrives,
  - NodeDB install failure triggering post-handshake recovery,
  - post-NodeDB analytics / side-effect failures not forcing recovery after `Connected`.

- Adds `SharedRadioInterfaceServiceLivenessTest` coverage for:
  - active transport restart creates a fresh transport,
  - restart is a no-op after explicit disconnect,
  - restart is a no-op after device deselection,
  - selected address is preserved,
  - restart does not emit permanent `Disconnected`,
  - restart does not emit a user-facing connection error,
  - restart coordinates with in-flight liveness recovery,
  - restart does not bypass environmental recovery when the transport was already stopped.

- Adds Connections UI/ViewModel coverage for:
  - transient `Reconnecting…` mapping.

## Follow-up

This PR addresses the Android-side handshake-stall recovery path associated with the reported WiFi/TCP reconnect behavior.

During testing, this also uncovered a firmware-side crash triggered by repeated or re-entered config/node-list startup on firmware build `tbeam / v2.7.25.104df5f`. The firmware-side fix is being tracked separately in meshtastic/firmware#10754. This PR adds app-side defensive depth so Android does not keep re-driving a node that repeatedly fails recovery.

If field logs still show periodic TCP reconnect or configuration loops after this lands, a follow-up investigation should look at TCP idle-timeout behavior and add privacy-safe disconnect-reason logging for EOF, IOException, write failure, and idle timeout.

---

I mostly run nRF52 devices over BLE, so I do not have a strong baseline for long-term WiFi behavior. I tested these changes with WiFi-connected ESP32 devices while reconnecting, scanning, and disabling/re-enabling WiFi. I also tested the post-handshake recovery and recovery-hardening changes on my BLE device after the T1000-E stuck-connecting behavior surfaced during review.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConfigFlowManagerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConfigFlowManagerImpl.kt
@@ -18,11 +18,13 @@ package org.meshtastic.core.data.manager
 
 import co.touchlab.kermit.Logger
 import kotlinx.atomicfu.atomic
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.delay
 import org.koin.core.annotation.Named
 import org.koin.core.annotation.Single
 import org.meshtastic.core.common.util.handledLaunch
+import org.meshtastic.core.common.util.safeCatching
 import org.meshtastic.core.model.ConnectionState
 import org.meshtastic.core.model.DeviceVersion
 import org.meshtastic.core.repository.CommandSender
@@ -77,10 +79,14 @@ class MeshConfigFlowManagerImpl(
          * Stage 1: receiving device config, module config, channels, and metadata.
          *
          * [rawMyNodeInfo] arrives first (my_info packet); [metadata] may arrive shortly after. Both are consumed
-         * together by [buildMyNodeInfo] at Stage 1 completion.
+         * together by [buildMyNodeInfo] at Stage 1 completion. Some firmware/network paths can deliver NodeInfo before
+         * the Stage 2 request; keep those packets so the later node-list phase can still make progress.
          */
-        data class ReceivingConfig(val rawMyNodeInfo: ProtoMyNodeInfo, val metadata: DeviceMetadata? = null) :
-            HandshakeState()
+        data class ReceivingConfig(
+            val rawMyNodeInfo: ProtoMyNodeInfo,
+            val metadata: DeviceMetadata? = null,
+            val earlyNodes: List<NodeInfo> = emptyList(),
+        ) : HandshakeState()
 
         /**
          * Stage 2: receiving node-info packets from the firmware.
@@ -95,13 +101,18 @@ class MeshConfigFlowManagerImpl(
         data class Complete(val myNodeInfo: SharedMyNodeInfo) : HandshakeState()
     }
 
-    private var handshakeState: HandshakeState = HandshakeState.Idle
+    private val handshakeState = atomic<HandshakeState>(HandshakeState.Idle)
 
     override val newNodeCount: Int
-        get() = (handshakeState as? HandshakeState.ReceivingNodeInfo)?.nodes?.size ?: 0
+        get() =
+            when (val state = handshakeState.value) {
+                is HandshakeState.ReceivingConfig -> state.earlyNodes.size
+                is HandshakeState.ReceivingNodeInfo -> state.nodes.size
+                else -> 0
+            }
 
     override fun handleConfigComplete(configCompleteId: Int) {
-        val state = handshakeState
+        val state = handshakeState.value
         when (configCompleteId) {
             HandshakeConstants.CONFIG_NONCE -> {
                 if (state !is HandshakeState.ReceivingConfig) {
@@ -129,7 +140,7 @@ class MeshConfigFlowManagerImpl(
         val finalizedInfo = buildMyNodeInfo(state.rawMyNodeInfo, state.metadata)
         if (finalizedInfo == null) {
             Logger.w { "Stage 1 failed: could not build MyNodeInfo, retrying Stage 1" }
-            handshakeState = HandshakeState.Idle
+            handshakeState.value = HandshakeState.Idle
             scope.handledLaunch {
                 delay(wantConfigDelay)
                 connectionManager.value.startConfigOnly()
@@ -149,9 +160,10 @@ class MeshConfigFlowManagerImpl(
             }
         }
 
-        handshakeState = HandshakeState.ReceivingNodeInfo(myNodeInfo = finalizedInfo)
-        Logger.i { "myNodeInfo committed (nodeNum=${finalizedInfo.myNodeNum})" }
+        handshakeState.value = HandshakeState.ReceivingNodeInfo(myNodeInfo = finalizedInfo, nodes = state.earlyNodes)
+        Logger.i { "myNodeInfo committed" }
         connectionManager.value.onRadioConfigLoaded()
+        serviceStateWriter.setConnectionProgress("Loading node list")
 
         scope.handledLaunch {
             delay(wantConfigDelay)
@@ -160,6 +172,7 @@ class MeshConfigFlowManagerImpl(
             Logger.i { "Requesting NodeInfo (Stage 2)" }
             connectionManager.value.startNodeInfoOnly()
         }
+        connectionManager.value.onHandshakeProgress()
     }
 
     private fun handleNodeInfoComplete(state: HandshakeState.ReceivingNodeInfo) {
@@ -171,7 +184,15 @@ class MeshConfigFlowManagerImpl(
         // The async work below (DB writes, broadcasts) proceeds without the guard.
         // Because nodes is now immutable, no snapshot is needed — state.nodes IS the snapshot.
         // Any stall-guard retry that re-enters handleNodeInfo will see Complete state and be ignored.
-        handshakeState = HandshakeState.Complete(myNodeInfo = info)
+        handshakeState.value = HandshakeState.Complete(myNodeInfo = info)
+
+        // Cancel the transport-aware fast-recovery watchdog SYNCHRONOUSLY, before the async DB
+        // install work below is launched. The firmware handshake has already completed at this
+        // point (NODE_INFO_NONCE received); a slow Room commit on a large mesh would otherwise
+        // falsely trip the 12s fast-recovery timeout before onNodeDbReady() gets a chance to
+        // cancel it. onNodeDbReady() still performs the same cancel as part of its larger post-
+        // NodeDB side-effect set, but it runs only after the DB install block finishes.
+        connectionManager.value.onHandshakeComplete()
 
         val entities =
             state.nodes.mapNotNull { nodeInfo ->
@@ -184,20 +205,39 @@ class MeshConfigFlowManagerImpl(
             }
 
         scope.handledLaunch {
-            nodeRepository.installConfig(info, entities)
-            analytics.setDeviceAttributes(info.firmwareVersion ?: "unknown", info.model ?: "unknown")
+            try {
+                nodeRepository.installConfig(info, entities)
+            } catch (e: CancellationException) {
+                throw e
+            } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+                Logger.e(e) { "Post-handshake NodeDB install failed; restarting transport to recover" }
+                nodeManager.setNodeDbReady(false)
+                nodeManager.setAllowNodeDbWrites(false)
+                connectionManager.value.recoverPostHandshakeFailure()
+                return@handledLaunch
+            }
+
             nodeManager.setNodeDbReady(true)
             nodeManager.setAllowNodeDbWrites(true)
             serviceStateWriter.setConnectionState(ConnectionState.Connected)
-            connectionManager.value.onNodeDbReady()
+
+            safeCatching { analytics.setDeviceAttributes(info.firmwareVersion ?: "unknown", info.model ?: "unknown") }
+                .onFailure { e -> Logger.w(e) { "Failed to set post-handshake analytics attributes" } }
+            safeCatching { connectionManager.value.onNodeDbReady() }
+                .onFailure { e -> Logger.e(e) { "Post-connected onNodeDbReady side effects failed" } }
         }
+        // Note: onHandshakeProgress() is intentionally NOT called here. By this point the
+        // handshake has reached HandshakeState.Complete and the synchronous onHandshakeComplete()
+        // call above has already cancelled the watchdog. Re-arming via onHandshakeProgress()
+        // would be both semantically wrong and wasted work. The remaining onHandshakeProgress
+        // sites cover all genuine progress.
     }
 
     override fun handleMyInfo(myInfo: ProtoMyNodeInfo) {
-        Logger.i { "MyNodeInfo received: ${myInfo.my_node_num}" }
+        Logger.i { "MyNodeInfo received" }
 
         // Transition to Stage 1, discarding any stale data from a prior interrupted handshake.
-        handshakeState = HandshakeState.ReceivingConfig(rawMyNodeInfo = myInfo)
+        handshakeState.value = HandshakeState.ReceivingConfig(rawMyNodeInfo = myInfo)
         nodeManager.setMyNodeNum(myInfo.my_node_num)
         nodeManager.setFirmwareEdition(myInfo.firmware_edition)
         applyEventFirmwareNotificationDefaults(myInfo.firmware_edition)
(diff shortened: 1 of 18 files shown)


jeremiah-k:
@jamesarich One thing worth noting from my WiFi testing, this app side fix may expose a firmware crash on some affected builds.

On a tbeam running firmware 2.7.25.104df5f, the app side fix made the connection flow behave correctly and retry/reconnect cleanly. Once that happened, the node started rebooting consistently during the normal config/node-list load. Before this fix, the app could sometimes look connected or get stuck in a partial state, so it may not have hit the crashing firmware path as reliably.

It looks like the app is now reaching the proper handshake path consistently and that path exposes a firmware-side crash. I have a patch in progress around `PhoneAPI::handleStartConfig()` to avoid the abort and I'll open a PR to the firmware repo for that soon.

As far as this PR goes the debug build is behaving well with the patched firmware, so I think it's ready for review.

core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConnectionManagerImpl.kt
```diff
@@ -214,28 +222,102 @@ class MeshConnectionManagerImpl(
...
+     * the sibling launch — otherwise the sibling's restart emissions (DeviceSleep, then Connected) can arrive while the
+     * app-level state is still Connecting, causing [onConnectionChanged]'s redundant-Connected-while-Connecting guard
+     * to ignore the fresh Connected emission. That leaves the app Disconnected while transport is Connected — the same
+     * split-brain this restart path is meant to break.
+     *
+     * By the time the sibling runs, handshakeTimeout has already completed naturally (it launched the sibling and
+     * returned), so the cancellation [onConnectionChanged] would attempt is a no-op on an already-completed job — and
+     * because the sibling is parented to `scope`, not to handshakeTimeout, it survives independently.
+     */
+    private fun runSiblingHandshakeRecovery() {
```
jamesarich, line 338: This recovery path has no backoff and no give-up cap, and the BLE stall branch re-sends `want_config` mid-session via `action()` in `startHandshakeStallGuard`.

I traced the T-Beam reboot from the field report: firmware `handleStartConfig()` (build `v2.7.25.104df5f`) re-enters on every `want_config` with no in-flight guard and never resets its `config_state` sub-iterator, so a re-sent/repeated `want_config` re-enters it mid node-list load — a strong contributor to the crash. And with nothing bounding recovery, a node that keeps failing the handshake gets re-driven indefinitely.

Two asks:
- Add exponential backoff + a consecutive-recovery cap (after N failed recoveries, surface a sticky error instead of silently looping).
- Reconsider the mid-session `want_config` re-send. The comment assumes the firmware drops the duplicate, but the firmware write-dedup is single-slot (`memcmp` vs the *previous* write only), and the interleaved heartbeats mean the re-sent nonce isn't byte-identical to the prior write — so it slips past and re-enters. A single clean transport cycle is safer than re-sending on the same session.

Your firmware patch is the proper fix; this is complementary app-side hardening so we don't hammer a crashing node.

core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConfigFlowManagerImpl.kt
```diff
@@ -220,35 +260,47 @@ class MeshConfigFlowManagerImpl(
...
+            }
+
+            else -> Logger.w { "Ignoring NodeInfo outside active handshake (state=$state)" }
         }
     }
 
     override fun handleFileInfo(info: FileInfo) {
         Logger.d { "FileInfo received: ${info.file_name} (${info.size_bytes} bytes)" }
         scope.handledLaunch { radioConfigRepository.addFileInfo(info) }
+        connectionManager.value.onHandshakeProgress()
```
jamesarich, line 303: `onHandshakeProgress()` is called here with no handshake-state guard (unlike `handleNodeInfo`, which is ignored once the state is `Complete`). Between `NODE_INFO_NONCE` arriving and the async DB install finishing, app state is still `Connecting`, so a late FileInfo/config packet here can re-arm the 12s fast watchdog that `onHandshakeComplete()` just cancelled — re-introducing the slow-DB false-trip on large meshes that `onHandshakeComplete()` was added to prevent.

Suggest having `onHandshakeComplete()` set a one-way latch that `onHandshakeProgress()` checks, rather than relying solely on the `Connecting` state guard. Low probability, but cheap to close.

core/ui/src/commonMain/kotlin/org/meshtastic/core/ui/util/NetworkTransportInfo.kt
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Transport-type snapshot of a single system network. Filled in by the platform-specific `isWifiUnavailable` actual
+ * (Android's `ConnectivityManager.getNetworkCapabilities`); kept platform-agnostic so the "is any network-scan
+ * transport present?" reduction is unit-testable from `commonTest` without an Android runtime.
+ */
+internal data class NetworkTransportInfo(
+    val hasWifi: Boolean,
+    val hasEthernet: Boolean,
+    val hasVpn: Boolean,
+    val hasCellular: Boolean,
```
jamesarich: `hasCellular` is never read by any predicate — `anyNetworkScanTransportAvailable` only checks wifi/ethernet/vpn, and nothing else reads it. It's captured in the Android `hasLocalNetwork()` actual and set in every test fixture but has no effect, which misleadingly implies cellular is considered. Recommend dropping the field and its call sites.

core/data/src/commonMain/kotlin/org/meshtastic/core/data/repository/DeviceLinkRepositoryImpl.kt
```diff
@@ -75,8 +76,10 @@ class DeviceLinkRepositoryImpl(
                         "DeviceLinkRepository: network refresh timed out after ${NETWORK_REFRESH_TIMEOUT_MS}ms"
                     }
                 } else {
-                    store(remoteLinks)
-                    lastRefreshMillis = nowMillis
+                    withContext(NonCancellable + dispatchers.io) {
```
jamesarich, line 79: This `NonCancellable` refresh/persistence fix (here, plus the matching changes in `DeviceHardwareRepositoryImpl` and `StaleWhileRevalidateFlow`) looks sound on its own — stopping the network deadline from cancelling the Room write mid-write. But it's unrelated to the WiFi handshake recovery this PR is about.

Could you split the device-catalog caching changes into their own PR? It keeps this one scoped to the connectivity fix and much easier to review for regressions.

core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConnectionManagerImpl.kt
```diff
@@ -96,7 +97,8 @@ class MeshConnectionManagerImpl(
     private var preHandshakeJob: Job? = null
     private var sleepTimeout: Job? = null
     private var locationRequestsJob: Job? = null
-    private var handshakeTimeout: Job? = null
+
+    private val handshakeTimeout = atomic<Job?>(null)
```
jamesarich, line 106: Nice — moving this to `atomic<>` fixes the cross-thread visibility issue. One residual nit: the re-arm sites still do `cancel()` then reassign as two ops, so a concurrent re-arm can orphan a job between them. `handshakeTimeout.getAndSet(newJob)?.cancel()` collapses it into one atomic swap if you want it fully tight. Low priority.

jamesarich (changes requested):
Thanks for the thorough work here, and especially for the field testing + the firmware-crash writeup — that investigation is genuinely valuable. The **core fix is correct**: the WiFi/TCP split-brain (transport `Connected` while the app is `Disconnected`, blocking same-node reconnect) is the right diagnosis, and cycling the transport in place is the right shape. Test coverage is strong.

Two things before this is ready to merge:

**1. Scope — this PR has outgrown its title.** It's now +2328/−120 across 26 files / 20 commits, several unrelated to handshake recovery. Please split into their own PR(s):
- `fix(data): make device link catalog refresh transactional`
- `fix(data): separate refresh timeouts from Room persistence` (DeviceHardwareRepository / DeviceLink / StaleWhileRevalidate)
- the `isWifiUnavailable` / VPN-banner work (3 commits) — a distinct bug from handshake recovery

Also: squash the `restart transport after post-handshake failure` → `narrow post-handshake recovery boundary` pair, and a rebase would drop the 6 `Merge branch 'main'` commits. A core-connectivity change is far easier to review for regressions when it's just the recovery fix.

**2. Recovery hardening (ties to your firmware finding).** I dug into *why* the clean handshake trips the T-Beam reboot: firmware `handleStartConfig()` (build `v2.7.25.104df5f`) re-enters on every `want_config` with no in-flight guard and never resets its `config_state` sub-iterator. Android's stall-guard **re-sends `want_config` on a timer** and re-handshakes with no backoff/cap — that re-send is what re-enters the fragile firmware path mid node-list load. (The firmware write-dedup is single-slot `memcmp` vs the *previous* write only, so the interleaved heartbeats let the re-sent nonce slip past it — the "firmware will drop our retry" assumption doesn't hold.) Your firmware patch is the proper fix; a small app-side backoff + give-up cap is good defensive depth so a crashing node isn't re-driven every cycle. Details inline.

Everything else is minor (inline). Nice work overall — this is close.

jeremiah-k:
@jamesarich Thanks - I split the independent scope into two separate PR branches and kept this PR focused on the core handshake recovery work.

New split branches:

1. `bugfix/data-refresh-timeouts`
   - separates remote API timeouts from local Room persistence
   - touches `DeviceHardwareRepository`, `DeviceLinkRepository`, `StaleWhileRevalidateFlow`, and the related repository test
   - one clean commit, 4 files

2. `bugfix/vpn-network-scan-transport`
   - checks all current networks for scan availability
   - treats WiFi, Ethernet, or VPN as valid network-scan transports
   - keeps cellular-only unavailable
   - removes the unused `hasCellular` field
   - one clean commit, 5 files

I also addressed the recovery-hardening feedback locally on the core branch:

- added exponential recovery backoff
- added a consecutive recovery cap with a localized sticky error
- removed the same-session `want_config` retry path
- added a one-way handshake-complete latch so late packets cannot re-arm the watchdog after `NODE_INFO_NONCE`
- converted handshake watchdog re-arms to atomic swaps
- fixed the CI/test fallout by correcting the Kermit severity enum, satisfying detekt, isolating test fixtures, and stabilizing coroutine timing assertions

Once the two smaller PRs land, I’ll rebase this PR onto current `main`, drop the merge commits and dead transactional device-link commit, fold the CI/test-fix commits into their parent changes, and force-push a tightened core-connectivity PR.

jeremiah-k:
@jamesarich Resolved.

The scope-related changes were split out and landed separately (#5881 and #5882), the merge/dead commits were removed during the rebase, and the remaining branch is now focused on the connection recovery work.

The recovery hardening feedback has also been incorporated: recovery now uses transport restart with exponential backoff and a give-up cap, the same-session `want_config` retry path was removed, and the watchdog/latch changes prevent late progress from re-arming recovery after handshake completion.

I also addressed the remaining inline review comments during the cleanup and rebase. Please let me know if I missed anything.

Testing is continuing to go well for the firmware-side fix. I'm preparing the associated firmware PR and will link it once it's up.

jamesarich (approved):

jeremiah-k:
Quick follow-up: I opened the firmware-side PR for the crash that came up while testing this:

meshtastic/firmware#10754

That PR handles the `tbeam / v2.7.25.104df5f` crash/reboot I was seeing once Android reliably reached the Stage 2 node-info/config path. This Android PR keeps the app-side recovery/backoff behavior, while the firmware PR addresses the underlying firmware crash.

PR #5912, opened 2026-06-23
fix(ble): require fresh advertisement for auto-reconnect

jeremiah-k:
## Overview

This PR fixes BLE auto-reconnect for bonded radios when the initial targeted scan misses a low-duty advertising window.

Previously, `BleRadioTransport.findDevice()` fell back to the stale bonded device handle after a missed 2-second targeted scan. That handle could lack a fresh advertisement, causing `KableBleConnection` to enter `autoConnect = true` and wait for `CONNECTION_TIMEOUT` before the reconnect loop could try again.

Bonded reconnect now requires a fresh advertisement before connecting. The reconnect path keeps the fast targeted scan, escalates once to a bounded scan when needed, and reports the bonded device as unavailable if it still is not advertising.

## Key changes

- Changed bonded-device discovery to use a bounded fresh-advertisement path:
  - 2-second targeted scan.
  - 5-second escalated scan if the targeted scan misses.
  - `RadioNotConnectedException` if the bonded device is still not advertising.
- Removed the stale bonded-handle fallback that could lead to `autoConnect = true` hangs.
- Preserved non-bonded discovery retry behavior.
- Added a shared `scanForFreshDevice(timeout)` helper so bonded and non-bonded discovery use the same single-scan primitive.
- Skips screen-entry BLE auto-scan when a selected device already exists, letting selected-device reconnect own the fresh-advertisement scan.
- Stops active discovery scans once the service reaches `Connected`.
- Keeps manual scan controls available for explicit discovery or device switching.

## Testing

- Added regression coverage for bonded reconnect when the targeted scan misses but the escalated scan finds a fresh advertisement.
- Added regression coverage for bonded reconnect when no fresh advertisement is found, verifying the stale bonded handle is not used and `connectAndAwait` is not called.
- Updated the previous bonded fallback test to assert the new no-stale-handle behavior.
- Added scan-lifecycle coverage for:
  - stopping active BLE scan on `Connected`.
  - stopping active network scan on `Connected`.
  - skipping screen-entry BLE auto-scan when a selected device exists.
  - allowing screen-entry BLE auto-scan when no selected device exists.

## Breaking changes / migration

No breaking API or user-facing migration steps are required.

### DIFF
--- a/core/network/src/commonMain/kotlin/org/meshtastic/core/network/radio/BleRadioTransport.kt
+++ b/core/network/src/commonMain/kotlin/org/meshtastic/core/network/radio/BleRadioTransport.kt
@@ -54,10 +54,12 @@ import org.meshtastic.core.ble.retryBleOperation
 import org.meshtastic.core.ble.toMeshtasticRadioProfile
 import org.meshtastic.core.common.util.nowMillis
 import org.meshtastic.core.model.RadioNotConnectedException
+import org.meshtastic.core.model.util.anonymize
 import org.meshtastic.core.network.transport.HeartbeatSender
 import org.meshtastic.core.repository.RadioTransport
 import org.meshtastic.core.repository.RadioTransportCallback
 import kotlin.concurrent.Volatile
+import kotlin.time.Duration
 import kotlin.time.Duration.Companion.milliseconds
 import kotlin.time.Duration.Companion.seconds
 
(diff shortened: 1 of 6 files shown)


jamesarich:
**Nit — stale KDoc in `MeshtasticBleDevice` (out of this diff, low-pri follow-up)**

[`core/ble/src/commonMain/kotlin/org/meshtastic/core/ble/MeshtasticBleDevice.kt:28–35`](https://github.com/meshtastic/Meshtastic-Android/blob/bugfix/ble-reconnect-fresh-advertisement/core/ble/src/commonMain/kotlin/org/meshtastic/core/ble/MeshtasticBleDevice.kt#L28-L35) says:

> *"When created from the OS bonded device list (address only), `[advertisement]` is null and the peripheral is constructed via `createPeripheral(address)` with `autoConnect = true`."*

After this PR, `BleRadioTransport.findDevice()` either returns a freshly-scanned device (which always has an advertisement) or throws `RadioNotConnectedException`. The `advertisement == null` / `createPeripheral(address)` / `autoConnect = true` path is no longer reachable via the normal reconnect flow.

Suggested replacement:

```kotlin
/**
 * Unified [BleDevice] implementation for all BLE devices — scanned, bonded, or both.
 *
 * When created from a live BLE scan, [advertisement] is populated and used for optimal peripheral
 * construction via `Peripheral(advertisement)` with a direct (non-autoConnect) connection attempt.
 *
 * Bonded-only devices (address only, [advertisement] null) can still be constructed — e.g. for
 * display in the device list — but [BleRadioTransport.findDevice] requires a fresh advertisement
 * before connecting and will throw [RadioNotConnectedException] rather than returning a stale handle.
 *
 * @param address The device's MAC address (or platform identifier string).
 * @param name The device's display name, if known.
 * @param advertisement The Kable [Advertisement] from a live scan, or `null` for bonded-only devices.
 */
```

Fine to handle in a separate follow-up; just flagging so the doc doesn't mislead the next reader.

jamesarich (approved):

Issue #3127, opened 2025-09-17
[Feature Request]: GPX export


Dtrieb:
### Checklist

- [x] I have used the search function for **OPEN ISSUES** to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for **CLOSED ISSUES** to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.

- [x] I will describe the request with as much detail as possible.

- [x] This request contains only one single feature, **not** a list of multiple (related) features.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Feature or improvement you want

At the moment you can export your position protocol only as csv file. It would be great if you can export as GPX file, too.

### Why should this be added?

So you can use the file very easily on other PC programs. And see were your are, were you where.

From GPX you can also convert to KML files.

jamesarich:
Currently no plans for this - you can process csv however you want, off the phone.

PR #5919, opened 2026-06-23
feat(node): add GPX export to position log screen

jamesarich:
`PositionLogScreen` already displays position history for any node, including the local device (via `NodeMapViewModel`'s `MeshLog.NODE_NUM_LOCAL` routing). The only missing piece was GPX export. This adds it as a second export button alongside the existing CSV export, with no new screens, view models, or map overlays.

🌟 **New Features**
- GPX 1.1 export button (`FileDownload` icon) on `PositionLogScreen`, gated on `positions.isNotEmpty()`
- `savePositionGpx(uri, data, trackName)` on `MetricsViewModel` mirroring the existing `savePositionCSV` pattern
- `buildGpx()` emits valid GPX 1.1 with `<trkpt lat lon>`, optional `<ele>`, and `<time>` (ISO-8601 UTC via `kotlinx-datetime`); coordinates formatted with `formatString("%.7f", …)` to avoid scientific notation in `xs:decimal` attributes; positions with `time <= 0` are skipped

🧹 **Chores**
- Added `export_gpx` string resource

### DIFF
--- a/.skills/compose-ui/strings-index.txt
+++ b/.skills/compose-ui/strings-index.txt
@@ -530,6 +530,7 @@ expires
 ### EXPORT ###
 export_configuration
 export_data_csv
+export_gpx
 export_keys
 export_keys_confirmation
 export_tak_data_package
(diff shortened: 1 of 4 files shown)


Dtrieb:
https://github.com/meshtastic/Meshtastic-Android/issues/3127 🤔

PR #5923, opened 2026-06-23
fix(usb): Add serial presence recovery for USB replug

jeremiah-k:
## Overview

This pull request improves USB serial resilience by adding a platform abstraction for tracking currently connected serial devices and using it to recover the selected SERIAL transport after an unplug/replug cycle.

Previously, unplugging a selected USB serial device could leave the transport in a zombie `ConnectionState.DeviceSleep` state, and replugging the same device did not reliably restart the selected SERIAL transport. USB attach handling could also route through an Activity/deep-link reload path even when a device was already selected, causing unnecessary UI/ViewModel churn.

This change fixes the full selected-device replug flow by:

1. tracking connected USB serial device keys,
2. arming recovery only after the selected device is actually observed absent,
3. restarting the selected SERIAL transport when that same device returns in a recoverable state,
4. avoiding unnecessary Connections-page navigation when a selected device is already attached/replugged,
5. and hardening explicit shutdown so close/disconnect callbacks are not duplicated or leaked into later recovery state.

### Features

* Added `SerialDevicePresence`, a common interface exposing `deviceKeys: StateFlow<Set<String>>` for currently connected SERIAL device keys.
* Implemented `SerialDevicePresence` per platform:

  * Android: `AndroidSerialDevicePresence` maps `UsbRepository.serialDevices` to a hot `StateFlow` of serial device keys.
  * JVM/Desktop: `JvmSerialDevicePresence` emits an always-empty device-key set because desktop serial hot-plug observation is not currently provided.
* Updated `SharedRadioInterfaceService` to observe:

  * the selected device address,
  * current serial device presence,
  * and transport `_connectionState`.
* Added guarded USB replug recovery for the selected SERIAL device:

  * recovery is armed only after the selected serial key is observed absent,
  * recovery triggers when that same key returns while the transport is in `ConnectionState.DeviceSleep`,
  * initial presence emissions are ignored so startup/refresh ordering does not falsely arm recovery,
  * duplicate same-key emissions are de-duplicated,
  * and all recovery guards are revalidated under `transportMutex` before restart.
* Restart is performed as silent recovery by stopping the zombie transport without permanent-disconnect notification and immediately starting the selected SERIAL transport again.

### USB attach behavior

* Updated `MainActivity` USB attach handling so attach/replug no longer always routes through Connections-page navigation.
* If no device is selected, USB attach still navigates to the Connections screen so the user can pick a device.
* If a device is already selected, USB attach refreshes USB state without forcing an Activity/deep-link reload or switching pages.
* This preserves first-attach/no-selection behavior while avoiding unnecessary `MainActivity` and ViewModel churn during selected-device replug.

### Fixes

* Scoped USB recovery to the currently selected SERIAL device only.
* Ignored unrelated USB device attach events.
* Avoided restarting transports that are already `Connected` or `Connecting`.
* Avoided automatic recovery after explicit `service.disconnect()`.
* Cleared stale recovery-arm state when the service reaches permanent `Disconnected`, preventing an old detach observation from leaking into a later reconnect session.
* Gated recovery on a real detach edge so a disconnect callback that reaches `DeviceSleep` before USB presence updates does not immediately restart against stale presence data.
* Covered the reverse ordering where USB presence returns before the transport emits `DeviceSleep`; the later state transition can still complete recovery.
* Suppressed duplicate serial disconnect callback propagation during explicit transport close.

### Shutdown / teardown changes

* Added `SerialRadioTransport.explicitCloseInProgress` so reader-thread disconnect callbacks do not forward transient disconnects during explicit `close()`.
* Added a shared `closeConnection(waitForStopped)` teardown helper for SERIAL connections.
* Reused the shared teardown path from both `close()` and `onDeviceDisconnect(...)`.
* Atomically clears the active serial connection reference before closing, so only the owner of the transition forwards disconnect handling.
* Updated `StreamTransport.close()` so explicit close no longer invokes `onDeviceDisconnect(...)`; explicit disconnect signaling is owned by the service layer.

### Refactors

* Centralized SERIAL teardown in `SerialRadioTransport`.
* Extracted USB recovery flow setup into `observeUsbRecoveryTriggers()` to keep service initialization focused.
* Reworked the USB recovery observer into a guarded, de-duplicated, race-tolerant state machine.
* Removed the USB attach-only PendingIntent/TaskStackBuilder navigation path for already-selected device replug.

### Testing

* Extended `SharedRadioInterfaceServiceLivenessTest` with deterministic USB replug recovery scenarios covering:

  * selected-device replug recovery,
  * unrelated-device attach no-op,
  * already-connected transport no-op,
  * presence-before-`DeviceSleep` ordering,
  * disconnect-before-detach ordering,
  * initial empty presence snapshot no-op,
  * duplicate same-key emission de-duplication,
  * explicit disconnect disarming recovery,
  * and clearing stale recovery-arm state across explicit disconnect and later reconnect.
* Added `StreamTransportTest` coverage confirming `StreamTransport.close()` does not emit a disconnect callback.

## Breaking Changes / Migration Notes

* `StreamTransport.close()` no longer calls `onDeviceDisconnect(...)` directly. Explicit close/disconnect notification is now owned by the service layer so transport recovery can close and recreate transports without emitting duplicate user-facing disconnect state.
* `SharedRadioInterfaceService` now requires a `SerialDevicePresence` implementation in the DI graph. Android provides `AndroidSerialDevicePresence`; JVM/Desktop provides `JvmSerialDevicePresence`.


core/service/src/commonMain/kotlin/org/meshtastic/core/service/SharedRadioInterfaceService.kt
```diff
@@ -61,6 +67,55 @@ import org.meshtastic.core.repository.RadioTransportFactory
...
+
+private data class UsbRecoveryTriggerState(
+    val key: String? = null,
+    val present: Boolean = false,
+    val armedByAbsence: Boolean = false,
+    val triggerRecovery: Boolean = false,
+)
+
+private fun selectedSerialPresence(address: String?, keys: Set<String>): SelectedSerialPresence {
+    val key = address?.takeIf { it.firstOrNull() == InterfaceId.SERIAL.id }?.substring(1)
```
jamesarich: `substring(1)` on a bare `"s"` address yields `key = ""` (harmless today since empty never appears in the key set, but a latent smell). `drop(1)` plus an emptiness guard is null-safe on that edge and reads clearer:

```suggestion
    val key = address?.takeIf { it.firstOrNull() == InterfaceId.SERIAL.id }?.drop(1)?.takeIf { it.isNotEmpty() }
```

core/service/src/commonMain/kotlin/org/meshtastic/core/service/SharedRadioInterfaceService.kt
```diff
@@ -245,10 +301,94 @@ class SharedRadioInterfaceService(
...
+                    // Re-check presence under the lock — the combine snapshot may be stale
+                    // if the device was unplugged or the selection changed while awaiting
+                    // the mutex. Mirrors the race-defense pattern of the state check above.
+                    val currentRest =
+                        _currentDeviceAddressFlow.value
+                            ?.takeIf { it.firstOrNull() == InterfaceId.SERIAL.id }
+                            ?.substring(1)
+                    val currentKeys = serialDevicePresence.deviceKeys.value
+                    if (currentRest == null || currentRest !in currentKeys) {
+                        return@withLock
```
jamesarich: This under-lock re-check duplicates the key-extraction + presence test that `selectedSerialPresence(...)` already encapsulates. Reusing the helper keeps the extraction logic in one place (and so it inherits the `drop(1)` fix above for free):

```suggestion
                    val currentKeys = serialDevicePresence.deviceKeys.value
                    if (!selectedSerialPresence(_currentDeviceAddressFlow.value, currentKeys).present) {
                        return@withLock
                    }
```

core/service/src/commonMain/kotlin/org/meshtastic/core/service/SharedRadioInterfaceService.kt
```diff
@@ -245,10 +301,94 @@ class SharedRadioInterfaceService(
...
+    //
+    // Flap/loop safety: each physical replug consumes exactly one armed-by-absence transition, and
+    // the recovery branch only fires when connectionRequested AND the running transport is SERIAL —
+    // so an explicit disconnect() (which clears connectionRequested and nulls runningTransportId via
+    // stopTransportLocked) immediately disarms the observer. A device that flaps N times produces at
+    // most N restarts, never an unbounded loop.
+    //
+    // Extracted from initStateListeners() so that function stays under detekt's LongMethod cap (60).
+    // Pure code motion — all referenced state is class-member scope, launchIn() registers the cold
+    // flow on processLifecycle.coroutineScope exactly as before.
```
jamesarich: This 30-line `//` block is excellent reasoning but it's doing two jobs at once, and `//` comments don't surface in IDE quick-doc or generated KDoc. Two suggestions following KDoc convention:

1. **Lead with a KDoc summary.** Promote this to a `/** ... */` whose *first sentence* is a one-line summary of what the function wires up — that's the part KDoc tooling extracts. Keep the deeper race analysis as body paragraphs after a blank line. Roughly:
   ```kotlin
   /**
    * Observes selected-SERIAL-device presence + transport state and auto-restarts the transport
    * when the selected USB device is unplugged and replugged, sparing a manual re-select.
    *
    * Recovery arms only on the absent→present edge of the *selected* key while in DeviceSleep ...
    * (race correctness, flap safety) ...
    */
   ```
2. **Trim the non-durable parts.** The closing paragraph ("Extracted from initStateListeners ... Pure code motion ...") is PR-review context, not doc the next reader needs — the commit message already records it; drop it. The race-correctness and flap-safety paragraphs are the keepers; tighten each to ~3 lines. Net: roughly half the length, and the summary becomes discoverable.

jamesarich (changes requested):
Really nice work — the race analysis is sound and the test matrix covers both callback orderings, the flap/dedup case, and stale-arm-across-reconnect. New tests pass locally (`:core:service:jvmTest`, `:core:network:jvmTest`). One cross-platform gap to fix before merge, plus two small inline suggestions.

### iOS DI gap (please fix)

`SharedRadioInterfaceService` is `@Single` in `commonMain` and now takes a hard `SerialDevicePresence` dependency, but the only `@Single` impls are `AndroidSerialDevicePresence` (androidMain) and `JvmSerialDevicePresence` (jvmMain). `core:network`/`core:service` both target iOS (`compileKotlinIosArm64` / `iosArm64MainKlibrary` exist), so the iOS Koin graph has no binding for `SerialDevicePresence`. It compiles (the interface is in commonMain) — which is why CI is green — but Koin resolution of `SharedRadioInterfaceService` would throw `NoDefinitionFoundException` at runtime on iOS. Not a shipping bug today (iOS is compile-only), but it's a regression to the common graph's iOS-resolvability and the Migration Notes silently omit iOS.

Laziest correct fix — a 4-line `core/network/src/iosMain/kotlin/org/meshtastic/core/network/repository/IosSerialDevicePresence.kt` mirroring the Jvm one (expect/actual or a commonMain default would collide with Android's `@Single` for the same type, so two trivial copies is the path of least resistance):

```kotlin
@Single
class IosSerialDevicePresence : SerialDevicePresence {
    override val deviceKeys: StateFlow<Set<String>> = MutableStateFlow<Set<String>>(emptySet()).asStateFlow()
}
```

jamesarich (changes requested):
some small changes needed, otherwise appreciate this QOL fix for usb serial

jamesarich (approved):

PR #5935, opened 2026-06-24
refactor(connections): Derive DeviceType from InterfaceId

jeremiah-k:
## Overview

This pull request removes drift in connection type handling by using `InterfaceId` as the canonical source for transport address prefixes, while keeping `DeviceType` as the presentation/analytics enum for connection type labels.

Before this change, prefix handling was duplicated across `InterfaceId`, `DeviceType.fromAddress(...)`, and feature-level mock/replay constants. This change keeps the existing address prefixes and runtime behavior, but derives `DeviceType` from `InterfaceId` so prefix parsing lives in one place and the behavior is test-covered.

## Key Changes

- Updated `DeviceType.fromAddress(...)` to delegate prefix parsing to `InterfaceId.forIdChar(...)`.
- Kept `DeviceType` scoped to the existing presentation/analytics values:
  - `BLE`
  - `USB`
  - `TCP`
- Preserved historical transport presentation behavior:
  - `x...` maps to BLE.
  - `s...` maps to USB.
  - `m` maps to USB, preserving mock/demo mode’s historical USB presentation.
  - `t...` maps to TCP.
  - `r`, `n`, empty, and unknown prefixes map to `null`.
- Built `DeviceListEntry.Ble`, `DeviceListEntry.Mock`, and `DeviceListEntry.Replay` addresses from `InterfaceId` prefix IDs.
- Removed duplicate mock/replay prefix constants from `feature/connections/Constants.kt`.
- Updated the remaining mock/replay region-warning checks to use `InterfaceId` directly.
- Preserved existing analytics behavior:
  - Datadog `transportType` values remain `"BLE"`, `"USB"`, and `"TCP"`.
  - Mock/demo connections continue to report as `"USB"`.
  - Replay continues to report no transport label.
- Preserved existing UI and reconnect behavior:
  - Mock/demo mode continues to use the USB presentation path.
  - Replay remains unbadged.
  - Fast-recovery transport membership is unchanged.
- Added tests covering address-prefix mapping, analytics-facing `DeviceType.name` values, and InterfaceId-backed mock/replay `DeviceListEntry` prefixes.

## Testing

- Added `DeviceTypeTest` coverage for prefix parsing and analytics-facing transport names.
- Added `DeviceListEntryTest` coverage for mock/replay prefixes derived from `InterfaceId`.
- Verified behavior remains unchanged for BLE, USB serial, mock/demo, TCP, replay, no-device, empty, and unknown prefixes.

## Migration Notes

- No address-prefix characters changed.
- No DataStore migration is required.
- Existing persisted-prefix parsing in `MeshPrefsImpl`, `AppPreferences`, and transport factories remains unchanged and out of scope.

### DIFF
--- a/core/model/src/commonTest/kotlin/org/meshtastic/core/model/DeviceTypeTest.kt
+++ b/core/model/src/commonTest/kotlin/org/meshtastic/core/model/DeviceTypeTest.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2026 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.core.model
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+
+/** Locks down [DeviceType.name] values used by the cross-platform connect analytics event. */
+class DeviceTypeTest {
+    @Test
+    fun fromAddress_preserves_transport_analytics_names() {
+        assertEquals("BLE", DeviceType.fromAddress("x123")?.name)
+        assertEquals("USB", DeviceType.fromAddress("s/dev/bus/usb/001/002")?.name)
+        assertEquals("USB", DeviceType.fromAddress("m")?.name)
+        assertEquals("TCP", DeviceType.fromAddress("t192.0.2.1")?.name)
+    }
+
+    @Test
+    fun fromAddress_returns_null_for_non_presented_prefixes() {
+        assertNull(DeviceType.fromAddress("r"))
+        assertNull(DeviceType.fromAddress("n"))
+        assertNull(DeviceType.fromAddress(""))
+        assertNull(DeviceType.fromAddress("z"))
+    }
+}
--- a/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/Constants.kt
+++ b/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/Constants.kt
@@ -26,11 +26,5 @@ const val NO_DEVICE_SELECTED: String = "n"
 /** One-char prefix marking a TCP/Network device's `fullAddress`. See [DeviceListEntry.Tcp]. */
 const val TCP_DEVICE_PREFIX: String = "t"
 
-/** One-char prefix / sentinel marking the mock (demo-mode) device's `fullAddress`. See [DeviceListEntry.Mock]. */
-const val MOCK_DEVICE_PREFIX: String = "m"
-
-/** One-char prefix / sentinel marking the capture-replay device's `fullAddress`. See [DeviceListEntry.Replay]. */
-const val REPLAY_DEVICE_PREFIX: String = "r"
-
 /** One-char prefix marking a BLE device's `fullAddress`. See [DeviceListEntry.Ble]. */
 const val BLE_DEVICE_PREFIX: String = "x"
--- a/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/model/DeviceListEntry.kt
+++ b/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/model/DeviceListEntry.kt
@@ -18,6 +18,7 @@ package org.meshtastic.feature.connections.model
 
 import org.meshtastic.core.ble.BleDevice
 import org.meshtastic.core.ble.MeshtasticBleConstants.BLE_NAME_PATTERN
+import org.meshtastic.core.model.InterfaceId
 import org.meshtastic.core.model.Node
 import org.meshtastic.core.model.util.anonymize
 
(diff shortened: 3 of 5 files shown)


jamesarich (changes requested):
Thanks for tackling this — the drift between the connection-type representations is real. One concern before it lands: the change adds a representation instead of removing one, and an existing type already covers most of the job.

**`core/model` already has the canonical char→transport map: [`InterfaceId`](core/model/src/commonMain/kotlin/org/meshtastic/core/model/InterfaceId.kt).**

```kotlin
enum class InterfaceId(val id: Char) {
    BLUETOOTH('x'), MOCK('m'), NOP('n'), REPLAY('r'), SERIAL('s'), TCP('t');
    companion object { fun forIdChar(id: Char): InterfaceId? = ... }
}
```

It sits in the same module and package as `DeviceType`, already models MOCK and REPLAY correctly, and is what the transport factories dispatch on (`AndroidRadioTransportFactory` via `forIdChar`; `Base`/`Desktop` via `InterfaceId.<X>.id`). After this PR the tree carries *four* parallel prefix definitions: `InterfaceId`, `DeviceType.fromAddress`'s literals, the new `DeviceType.MOCK_ADDRESS_PREFIX`/`REPLAY_ADDRESS_PREFIX`, and the existing `MOCK_DEVICE_PREFIX`/`REPLAY_DEVICE_PREFIX` in [`Constants.kt`](feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/Constants.kt). The new constant pair recreates the drift it means to prevent.

**One constraint rules out the obvious "just merge the enums":** `DeviceType.name` (`"BLE"/"USB"/"TCP"`) ships to Datadog as `transportType` and is documented to match Meshtastic-Apple's `connect` event ([`MeshConnectionManagerImpl.kt:558`](core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConnectionManagerImpl.kt#L558), `PlatformAnalytics` KDoc). `InterfaceId.name` is `"BLUETOOTH"/"SERIAL"/"TCP"`, so `DeviceType` must keep its names.

**Suggested direction — make `InterfaceId` the single char source, keep `DeviceType` as the thin presentation enum derived from it:**

1. Drop the new `MOCK`/`REPLAY` enum members and `analyticsLabel()`; delegate `fromAddress` to `InterfaceId` instead. This reproduces today's behavior exactly (`.name` still yields `BLE`/`USB`/`TCP`):
   ```kotlin
   fun fromAddress(address: String): DeviceType? =
       when (InterfaceId.forIdChar(address.firstOrNull() ?: return null)) {
           InterfaceId.BLUETOOTH -> BLE
           InterfaceId.SERIAL, InterfaceId.MOCK -> USB   // mock shown as USB — historical
           InterfaceId.TCP -> TCP
           InterfaceId.REPLAY, InterfaceId.NOP, null -> null
       }
   ```
2. Drop the new `MOCK_ADDRESS_PREFIX`/`REPLAY_ADDRESS_PREFIX` constants. Build `DeviceListEntry.Mock`/`Replay`'s `fullAddress` from `InterfaceId.MOCK.id` / `InterfaceId.REPLAY.id`, then delete the duplicate `MOCK_DEVICE_PREFIX`/`REPLAY_DEVICE_PREFIX` in `Constants.kt` and point their two consumers at `InterfaceId.<X>.id`.
3. After that, `fromAddress('m')` still returns `USB`, so the `FAST_RECOVERY_TYPES += MOCK` edit, the `getIconPair` MOCK branch, and the `analyticsLabel()` indirection become unnecessary — please revert them to keep the diff behavior-preserving.
4. The `DeviceListEntry.deviceType` abstract val has no production consumers. Drop it as speculative, or wire up the intended caller in this PR.

The result *deletes* most of the current diff while truly achieving "prefix chars live in one place." Net behavior — analytics labels, nav icons, fast-recovery membership, dispatch — stays identical.

**Worth keeping (your testing instinct, retargeted):** the Apple-parity contract lives only in a comment today. A small lock-down test earns its place:
```kotlin
assertEquals("BLE", DeviceType.fromAddress("x123")?.name)
assertEquals("USB", DeviceType.fromAddress("m")?.name)   // mock folds to USB
assertEquals("TCP", DeviceType.fromAddress("t1.2.3.4")?.name)
```

**Out of scope (don't expand this PR):** `MeshPrefsImpl` keeps its own private `'n'` sentinel, and `AppPreferences.isBle/isSerial/...` parse the stored prefix — both are persisted on-disk contracts. Folding them into `InterfaceId` is a reasonable follow-up, but carries its own migration risk. The same goes for routing the `Base`/`Desktop` factories through `forIdChar`.

jeremiah-k:
@jamesarich 

Thanks — agreed with your direction. I reworked this to make `InterfaceId` the canonical prefix source and kept `DeviceType` as the thin presentation/analytics enum.

What changed:
- `DeviceType.fromAddress(...)` now delegates to `InterfaceId.forIdChar(...)`.
- `DeviceType` remains only `BLE`, `USB`, and `TCP`, preserving the Datadog / Apple-parity `transportType` names.
- MOCK folds back to `USB`, preserving the historical mock/demo behavior.
- REPLAY and NOP map to `null`, preserving the previous no-badge/no-analytics behavior.
- Removed the added `MOCK` / `REPLAY` `DeviceType` members, `analyticsLabel()`, and the speculative `DeviceListEntry.deviceType`.
- Mock/replay `DeviceListEntry` addresses now come from `InterfaceId.MOCK.id` / `InterfaceId.REPLAY.id`.
- BLE entry addresses now also use `InterfaceId.BLUETOOTH.id` for consistency with the single-prefix-source goal.
- Reverted the now-unnecessary UI, analytics, and fast-recovery changes.
- Retargeted tests to lock down the `DeviceType.name` presentation/analytics contract and InterfaceId-backed mock/replay prefixes.

I left the persisted `MeshPrefsImpl` / `AppPreferences` parsing and transport factory cleanup out of scope, as suggested.

jamesarich (approved):
All four asks addressed cleanly. `DeviceType.fromAddress` now delegates to `InterfaceId.forIdChar` — chars live in one place, analytics names preserved, speculative additions reverted. The exhaustive `when` over `InterfaceId` is a nice bonus: new entries will force a compile-time decision here. Analytics lock-down test is exactly right. Good rework.

Issue #111, opened 2020-07-22
In russian the "allow analytics" checkbox looks really bad and covers most of the screen


geeksville:
https://aws1.discourse-cdn.com/free1/uploads/meshtastic/optimized/1X/3dc2115f22f3716753a6f735efb73fd341d22247_2_243x500.jpeg 
see linked thread...

geeksville:
This issue has been mentioned on **Meshtastic**. There might be relevant details there:

https://meshtastic.discourse.group/t/alpha-tester-thread-0-8-1-android-code-is-ready-for-test/298/216

geeksville:
(the layout should be smarter about max width for the "submit bug" button and should allow the other text to flow below it?

mc-hamster:
andrekir — Today at 11:05 AM
great!
https://github.com/meshtastic/Meshtastic-Android/issues/96 (Note sounds OK)
https://github.com/meshtastic/Meshtastic-Android/issues/154 (date stamp OK)
https://github.com/meshtastic/Meshtastic-Android/issues/185 (Note sounds OK)
https://github.com/meshtastic/Meshtastic-Android/issues/194 (elapsed minutes OK)
https://github.com/meshtastic/Meshtastic-Android/issues/258 (Dark theme OK)
https://github.com/meshtastic/Meshtastic-Android/issues/279 (User reports resolved)
https://github.com/meshtastic/Meshtastic-Android/issues/111 (Solved in PR #234)
https://github.com/meshtastic/Meshtastic-Android/issues/281 (Solved on 1.2.45)
https://github.com/meshtastic/Meshtastic-Android/issues/289 (Solved on 1.2.45)

PR #5586, opened 2026-05-23
refactor: Remove AIDL API and modernize service architecture

jamesarich:
## Summary

Remove the deprecated AIDL API and all associated infrastructure, unlocking a comprehensive modernization of the radio communication architecture. This PR eliminates thousands of lines of legacy code and replaces it with a modern, testable, KMP-compatible architecture.

**262 files changed: +3,978 / −6,537 (net −2,559 lines)**

## Before: AIDL-based Architecture

```mermaid
graph TD
    subgraph "Client Apps / UI"
        VM[ViewModels]
    end

    subgraph "core:api (AIDL)"
        AIDL[IMeshService.aidl]
        Binder[Bound Service Binder]
        BC[Broadcast Receivers]
    end

    subgraph "Service Layer"
        MS[MeshService<br/>God-class]
        MR[MeshRouter<br/>Service Locator]
    end

    subgraph "Handlers"
        FRP[FromRadioPacketHandler]
        MMP[MeshMessageProcessor]
        MDH[MeshDataHandler]
        MCH[MeshConfigHandler]
        NIH[NeighborInfoHandler]
        TRH[TracerouteHandler]
    end

    subgraph "Radio"
        DRC[DirectRadioControllerImpl<br/>~450 lines, 17 deps, 44 functions]
    end

    VM -->|"IPC / Binder"| AIDL
    AIDL --> Binder
    Binder --> MS
    MS -->|"broadcasts"| BC
    BC -->|"intents"| VM
    MS --> MR
    MR -->|"lazy getters"| FRP
    MR -->|"lazy getters"| MMP
    MR -->|"lazy getters"| MDH
    MR -->|"lazy getters"| MCH
    MR -->|"lazy getters"| NIH
    MR -->|"lazy getters"| TRH
    MS --> DRC

    style AIDL fill:#ff6b6b,stroke:#c92a2a
    style Binder fill:#ff6b6b,stroke:#c92a2a
    style BC fill:#ff6b6b,stroke:#c92a2a
    style MR fill:#ff6b6b,stroke:#c92a2a
    style MS fill:#ffa94d,stroke:#e67700
    style DRC fill:#ffa94d,stroke:#e67700
```

## After: Modern KMP Architecture

```mermaid
graph TD
    subgraph "UI Layer"
        VM[ViewModels]
    end

    subgraph "core:repository — segregated interfaces"
        CMD["Command interfaces<br/>Admin · Messaging · Node · Query"]
[...]

PR #5947, opened 2026-06-25
fix(service): restore MeshService exported=true + intent-filter for plugin binding

ejosterberg:
## Problem

Third-party plugins that bind to the in-app `MeshService` via `bindService()` — most notably [meshtastic/ATAK-Plugin](https://github.com/meshtastic/ATAK-Plugin) — have been broken since v2.7.14. The cliff is documented at [meshtastic/ATAK-Plugin#111](https://github.com/meshtastic/ATAK-Plugin/issues/111) ("Fail to load plugin/fail to connect to iMeshService - all android app versions newer than 2.7.13"), where the maintainer's only workaround is "downgrade Meshtastic-Android to 2.7.13."

## Root cause

The service declaration in `androidApp/src/main/AndroidManifest.xml` changed in three independent ways across the v2.7.13 → v2.7.14 window:

| | v2.7.13 (worked) | main / v2.7.14+ (broken) |
|---|---|---|
| `android:name` | `com.geeksville.mesh.service.MeshService` | `org.meshtastic.core.service.MeshService` |
| `android:exported` | `true` | **`false`** |
| `<intent-filter>` action | `com.geeksville.mesh.Service` | (none) |

(1) is fine as part of the multi-module KMP refactor — the plugin can adapt its bind target (companion PR open against `meshtastic/ATAK-Plugin` updating `Constants.CLASS_NAME` to the new package).

But (2) and (3) make any cross-app `bindService()` impossible regardless of target name — Android's security model refuses `bindService()` to a non-exported service from a different UID. This is what user logcats show ("Permission Denial" / component-not-found).

## Fix

Restore `android:exported="true"` and add back the canonical intent-filter on the relocated `org.meshtastic.core.service.MeshService` declaration. Class name stays at the new path.

## Companion PR

Paired with **meshtastic/ATAK-Plugin#114** which updates the plugin's `Constants.CLASS_NAME` to the new service class path. Both PRs need to land for the plugin to bind on 2.7.14+; either alone leaves the regression.

## Verified

Production deployment at `training.ticketscad.com` running a Meshtastic-Android-fed CoT ingest path observed the regression cliff exactly: plugin v1.1.42 binds successfully on Meshtastic-Android 2.7.13, fails on 2.7.14+ with the device combo (ATAK CIV 5.6.0.12, Heltec V3 firmware 2.7.26.54e0d8d, plugin v1.1.42). Downgrade-to-2.7.13 workaround verified working. This patch (combined with the plugin-side companion PR) removes the need for the downgrade.

## Note on intent

Recognizing the long-term direction signaled in [`specs/005-tak-v2-protocol/`](https://github.com/meshtastic/Meshtastic-Android/tree/main/specs/005-tak-v2-protocol) is to subsume ATAK plugin functionality into the in-app local TAK Server (port 78 / `ATAK_PLUGIN_V2`). That path is gated on firmware 2.8.0+ which hasn't widely shipped yet. Restoring the export keeps the existing plugin ecosystem functional during the transition rather than forcing all current ATAK-on-Meshtastic users into a downgrade or out of the ecosystem entirely. If you'd prefer to keep the export removed and deprecate the plugin path on a clear timeline, happy to close this PR and instead update the [#111](https://github.com/meshtastic/ATAK-Plugin/issues/111) thread with that policy decision.

### DIFF
--- a/androidApp/src/main/AndroidManifest.xml
+++ b/androidApp/src/main/AndroidManifest.xml
@@ -156,12 +156,36 @@
             android:name="google_analytics_default_allow_analytics_storage"
             android:value="false" />
 
-        <!-- In-app mesh foreground service -->
+        <!-- In-app mesh foreground service.
+
+             android:exported="true" + the <intent-filter> below restore the
+             public bind API that third-party plugins (notably
+             meshtastic/ATAK-Plugin) have relied on since pre-2.7. Without
+             these, bindService() from any other UID fails with
+             SecurityException regardless of class name or permissions.
+
+             Regression timeline:
+               - v2.7.13 and earlier: exported="true" + intent-filter present.
+                 ATAK plugin v1.1.42 binds successfully.
+               - v2.7.14: service repackaged from com.geeksville.mesh.service.*
+                 to org.meshtastic.core.service.*, intent-filter dropped, and
+                 exported flipped to "false".
+               - Symptom: ATAK plugin shows "Connected Node: Not connected"
+                 (red M); logcat shows "Permission Denial: opening provider..."
+                 or "Unable to start service Intent ...: not found" depending
+                 on Android version.
+
+             See: https://github.com/meshtastic/ATAK-Plugin/issues/111
+                  https://github.com/meshtastic/Meshtastic-Android/pull/<this PR> -->
         <service
             android:name="org.meshtastic.core.service.MeshService"
             android:enabled="true"
             android:foregroundServiceType="connectedDevice|location"
-            android:exported="false" />
+            android:exported="true">
+            <intent-filter>
+                <action android:name="org.meshtastic.core.service.MeshService" />
+            </intent-filter>
+        </service>
 
         <service
             android:name="androidx.appcompat.app.AppLocalesMetadataHolderService"


jamesarich:
Thanks for digging into this — the manifest diff is correct for what it does, but it can't restore the plugin on `main`, and I want to lay out why before you spend more on it.

## The binder is gone on `main`

The export restore lets a cross-UID `bindService()` succeed at the framework level, but on `main` there's nothing on the other end. `core/service/.../MeshService.kt` now has:

```kotlin
override fun onBind(intent: Intent?): IBinder? = null
```

and there are zero `.aidl` files left in the tree. So `onServiceConnected` fires with a null `IBinder`, and `IMeshService.Stub.asInterface(binder)` has nothing to bind to. The plugin still lands on "Not connected" — it just fails on a null binder instead of a `SecurityException`. The manifest change alone is a no-op for plugin function on `main`.

## The break wasn't manifest-only on `main`

I suspect your verification ran against a 2.7.14-derived build, where this patch genuinely works: at both `v2.7.13` and `v2.7.14`, `IMeshService.aidl` is present and `onBind` returns a real binder. So at the 2.7.14 *line* the break really was manifest-only. But the AIDL was removed after that tag:

- `v2.7.13` (2026-02-19) — AIDL documented as the stable, supported interface.
- The deprecated `mesh_service_example` module was removed in #5055 (merged 2026-04-15).
- The AIDL API itself was removed in #5586 — "Remove AIDL API and modernize service architecture" — merged 2026-06-04, one day after the `v2.7.14` tag. Its own description calls it "the deprecated AIDL API."

Net: AIDL survives in 2.7.14 but is gone on `main`. The "no need to downgrade to 2.7.13" claim holds on the 2.7.14 line, not here.

## There's already a live replacement

The in-app TAK server (`core/takserver/`) shipped in 2.7.14 and needs no AIDL. It runs a TLS TAK Server inside the Meshtastic app on port 8089 (`DEFAULT_TAK_PORT` in `core/takserver/.../TAKDefaults.kt`). ATAK connects to the phone as a *network* TAK Server — import the exported data package with `truststore.p12` + `client.p12` — instead of loading the plugin. No `bindService()`, no cross-UID binder. It's gated behind the `tak_server_enabled` pref (off by default).

For your exact setup (firmware 2.7.26), this works today over the legacy `ATAK_PLUGIN` port 72 path (PLI + GeoChat, no compression). Full TAKPacketV2 on port 78 with zstd lights up on firmware 2.8.0+ (`supportsTakV2 = atLeast(V2_8_0)` in `core/model/.../Capabilities.kt`); `TAKMeshIntegration` listens on both ports and falls back automatically for older nodes.

## Recommendation

You already offer to close in favor of a deprecation policy, and I'd lean that way: deprecate the bind API on a clear timeline and point existing plugin users at the in-app TAK server. It's live in 2.7.14, works today, and is where `specs/005-tak-v2-protocol/` is headed — no need to reintroduce a binder `main` has already removed on purpose (#5586).

PR #5956, opened 2026-06-25
refactor(connections): Show one active transport pane

jeremiah-k:
## Overview

This pull request simplifies the Connections screen by replacing the old multi-section transport filters with a single active transport pane.

Before this change, the screen could show Bluetooth, Network, and USB sections at the same time, with separate scan controls inside the Bluetooth and Network sections. That made the page feel split across multiple discovery modes instead of centered on the interface the user is actively working with.

This change makes the selected transport explicit: Bluetooth, Network, or USB. The screen now shows only the selected transport’s devices and actions, while keeping the global connection card visible regardless of the active pane.

This is intended as a focused step toward a more unified Connections experience. It keeps the behavior and layout changes scoped to the transport selection model, while leaving room for follow-up polish to spacing, visual hierarchy, and broader layout refinements.

## Key Changes

* Replaced the inclusive Bluetooth / Network / USB visibility filters with a single selected transport pane.
* Added a persisted selected transport preference so the Connections screen can reopen on the user’s last active pane.
* Migrated legacy transport visibility preferences to the new selected-pane preference fallback.
* Derived the initial pane from the selected device when no explicit pane preference exists.
* Added a single-choice transport selector for Bluetooth, Network, and USB.
* Updated the device list so exactly one pane is rendered at a time:

  * Bluetooth shows Bluetooth devices and one Scan action.
  * Network shows discovered and recent TCP devices, manual add, and one Scan action.
  * USB shows USB/serial devices with no Scan action.
* Added a quiet USB empty-state message when no USB devices are detected.
* Kept BLE and Network discovery mutually exclusive.
* Stopped incompatible scans when switching panes, including cases where the pane is already active but stale scan state needs cleanup.
* Started screen-entry auto-scan from the active pane and persisted auto-scan preferences, including preferences that load after initial composition.
* Gated screen-entry auto-scan so it does not start discovery when a selected device is already available for reconnect.
* Ensured disposed panes always stop their owned scan so discovery does not continue behind the visible pane.
* Kept the global selected-device connection card visible regardless of the active pane, while recording the matching transport pane when a device is selected.
* Updated previews, screenshot coverage, README documentation, and tests for the single-pane model.

## Testing

* Added `ScannerViewModelTest` coverage for active transport defaults, selected-device fallback, explicit pane selection, scan stopping, auto-scan gating, selected-device transport recording, and manual TCP selection.
* Updated `ScannerViewModelTest` setup so each test gets isolated scanner state.
* Added `UiPrefsImplTest` coverage for selected transport migration from legacy visibility preferences, including explicit preference, invalid preference fallback, no legacy keys, all-visible legacy state, BLE-hidden legacy state, and USB-only legacy state.
* Updated `UiPrefsImplTest` setup so each test gets a fresh coroutine scope that is cancelled during cleanup.
* Smoke-tested the F-Droid debug build while exercising the Connections screen.
* Verified no fatal crash, ANR, or app-process failure in the collected debug log.

## Migration Notes

* Existing Bluetooth, Network, and USB address formats are unchanged.
* Existing localized `transport_*` resource keys are preserved.
* The selected device remains visible independently from the active transport pane.
* USB remains presence-driven and does not introduce a scan action.


jamesarich:
Before we get too much further on connection screen ui - this is slated as an upcoming rework for cross-platform compliance. I'm gonna say hold off an anything more for now, and once the new design is specced out - I can pass it off for implementation if you'd like.

jamesarich (approved):

jeremiah-k:
@jamesarich Sounds good. I tried to keep this scoped to the connection lifecycle cleanup and the existing screen model. I don't have a strong opinion on how it should look beyond wanting the connection paths to stay clear and predictable. I'd be happy to implement when it's specced out.

PR #5969, opened 2026-06-26
fix(ble): Avoid duplicate bonding retries after pairing failure

jeremiah-k:
## Overview

This pull request prevents the app from immediately starting a second bonding attempt after the user-facing Android pairing flow fails.

Earlier behavior armed the BLE transport after a generic bonding failure so interrupted bonding could still progress through the transport path. In practice, that also allowed the transport to immediately call `bond()` again when the device was still not bonded. On Android, that can produce a second PIN dialog or hit platform bonding cooldown/throttling after a cancelled, rejected, or timed-out pairing attempt.

This follows the bonding timeout hardening in #5967. That PR makes Android bonding attempts finite, checks for completed bonding even when the terminal broadcast is missed, and re-checks the final platform bond state before failing.

With that in place, this PR changes the UI-side failure behavior so an incomplete pairing attempt surfaces a retry warning instead of immediately arming the transport and triggering another `bond()` call.

If Android records the device as bonded even though the app-side bond call reports an error, the ViewModel still proceeds with selecting the device.

The related transport-side follow-up in PR #5973 handles the remaining lower-level case: if transport-side bonding fails and Android still reports the device as not bonded, the BLE transport stops that connection attempt before entering GATT setup.

## Key Changes

* Preserved successful pairing behavior: completed bonding still selects the device.
* Preserved Bluetooth permission failure behavior.
* Preserved coroutine cancellation behavior.
* Stopped arming the BLE transport after generic bonding failure when the device is still not bonded.
* Still selects the device when Android reports it bonded after a late or flaky bonding result.
* Avoided immediate transport-side re-bonding after a failed UI pairing attempt.
* Added user-facing retry copy for incomplete pairing.
* Updated Android ViewModel bonding tests to cover the explicit-retry, late-bonded, and cancellation behavior.

## Testing

* Updated `AndroidScannerViewModelBondingTest` coverage proving successful bonding still selects the device.
* Updated coverage proving permission failures do not arm the transport.
* Updated coverage proving generic bonding failures do not call `changeDeviceAddress()` and surface a retry warning.
* Added coverage proving a late bonded result still selects the device.
* Added coverage proving `CancellationException` is not treated as a generic bonding failure.
* Updated the test setup so each test gets isolated ViewModel harness state.

## Migration Notes

* Existing BLE addresses and OS bonds are unchanged.
* Users can retry pairing explicitly after a failed or cancelled pairing attempt.

### DIFF
--- a/.skills/compose-ui/strings-index.txt
+++ b/.skills/compose-ui/strings-index.txt
@@ -119,6 +119,7 @@ bluetooth_feature_discovery_description
 bluetooth_permission
 bold_heading
 bonding_failed_permissions
+bonding_failed_retry
 bottom_nav_settings
 broadcast_interval
 busy_noise_floor
--- a/core/resources/src/commonMain/composeResources/values/strings.xml
+++ b/core/resources/src/commonMain/composeResources/values/strings.xml
@@ -137,6 +137,7 @@
     <string name="bluetooth_permission">Bluetooth</string>
     <string name="bold_heading">Bold Heading</string>
     <string name="bonding_failed_permissions">Pairing failed. Grant nearby device permissions and try again.</string>
+    <string name="bonding_failed_retry">Pairing did not complete. Try pairing again.</string>
     <string name="bottom_nav_settings">Settings</string>
     <string name="broadcast_interval">Broadcast Interval</string>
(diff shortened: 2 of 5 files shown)


jamesarich (approved):
🚀

PR #5984, opened 2026-06-27
fix(a11y): keep message-bubble metadata green legible on colored bubbles

jamesarich:
Follow-up to the XEdDSA signing UI ([#5976](https://github.com/meshtastic/Meshtastic-Android/pull/5976), [#5980](https://github.com/meshtastic/Meshtastic-Android/pull/5980)), addressing accessibility review feedback on [design#113](https://github.com/meshtastic/design/issues/113).

## Why

The green SNR/RSSI text and the signed-broadcast shield sit on a **per-node-tinted message bubble**. The shared "Status Good" green (`StatusGreen`, green-600) has poor contrast against those bubble colors — worst in light mode, where green-on-periwinkle is muddy and hard to read. The black message text is fine because it derives from the bubble's content color; the green metadata didn't.

## Blast radius — why this is scoped, not a token change

`StatusGreen` is used in ~15 components (signal/battery indicators, node items, nav/security icons…), but **almost all render on the neutral theme surface** where green-600 is fine. The only place it sits on a colored background is the message-bubble metadata row. So changing the shared token globally would touch 15 healthy components to fix one — the wrong lever. This fix stays local to the bubble and leaves the token untouched.

## 🛠️ What changed

- **`ensureContrastOn(background, minRatio)`** (new, in `core/ui` theme): nudges a color toward black/white just enough to meet WCAG AA (4.5:1) against a given background, **preserving hue**. It picks the higher-contrast direction, so a mid-tone bubble *darkens* the green (stays green, becomes legible) rather than washing it out to near-white. Unit-tested.
- **`Snr` / `Rssi` / `SnrAndRssi`** gain an optional `background: Color? = null` param. Default `null` = current behavior, so **every other call site is unaffected** (no blast radius).
- **`MessageItem`** passes the composited bubble background so SNR, RSSI, and the signed shield all stay legible; the shield is unified onto the same contrast-adjusted `StatusGreen`.

## 🧪 Testing performed

- `./gradlew :core:ui:allTests --tests "*ColorContrastTest*"` — passes (raises sub-AA pairs to AA, darkens rather than washes out on mid-tone bg, leaves already-contrasting colors untouched).
- `./gradlew spotlessCheck detekt` — clean.
- `./gradlew :screenshot-tests:updateDebugScreenshotTest` — all references pass; the signed-message light reference regenerated to the darker, legible green.

| Before | After |
|---|---|
| washed-out green-600 on the bubble | darker, AA-legible green, still recognizably "good"/signed |


jamesarich:
Superseded by #5985, which takes the surface-backed approach (StatusSurface) instead of per-bubble lightness adjustment — keeps the status tokens at their true values and is reusable for the other status-color sites. Closing in favor of that direction.

PR #5986, opened 2026-06-27
Prevent Range Test from running on public/default channel

dubsector:
## Summary

- Blocks enabling Range Test in the app UI when the primary channel is unencrypted (empty PSK) or uses the well-known 1-byte default shortstring key. Both are effectively public channels that would flood the shared mesh with test traffic.
- Users who intentionally want Range Test on a public channel can still enable it via the Meshtastic CLI — this is by design, making it a deliberate power-user action rather than an accidental tap.
- Backend enforcement on save forces \enabled = false\ when a public channel is detected, so even a UI bypass cannot persist the setting.

## Changes from previous PR #3816

- **Deduplication:** Hoisted \state.connected && !isPublicPrimaryChannel\ into a single \canConfigure\ val — no repeated expression across 4 call sites
- **Already-enabled fix:** Kept screen-level \enabled = state.connected\ (not \canConfigure\) so the save footer remains reachable. Added \dditionalDirtyCheck = { isPublicPrimaryChannel && formState.value.enabled }\ — this surfaces the Save/Discard buttons when a device already has Range Test enabled on a public channel (e.g. set via CLI), letting the user push the corrective \enabled = false\ through the app.
- **Updated to current codebase:** Migrated to \src/commonMain\ path, \core.resources\ imports, Wire data-class \copy()\ syntax, and \ModuleConfig(range_test = ...)\ constructor — matches the KMP refactor that landed after #3816 was closed.

## Test plan

- [ ] On a device with default channel (1-byte PSK): Range Test controls are grayed out; Save does not appear unless Range Test was already enabled
- [ ] On a device with default channel and Range Test already enabled (set via CLI): Save/Discard footer appears; tapping Save pushes \enabled = false\ to device
- [ ] On a device with a private channel (PSK ≥ 2 bytes): Range Test controls work normally

### DIFF
--- a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/component/RangeTestConfigItemList.kt
+++ b/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/component/RangeTestConfigItemList.kt
@@ -53,7 +53,6 @@ fun RangeTestConfigScreen(viewModel: RadioConfigViewModel, onBack: () -> Unit) {
         enabled = state.connected,
         responseState = state.responseState,
         onDismissPacketResponse = viewModel::clearPacketResponse,
-        additionalDirtyCheck = { isPublicPrimaryChannel && formState.value.enabled },
         onSave = {
             val safeConfig = if (isPublicPrimaryChannel) it.copy(enabled = false) else it
             val config = ModuleConfig(range_test = safeConfig)
@@ -65,7 +64,7 @@ fun RangeTestConfigScreen(viewModel: RadioConfigViewModel, onBack: () -> Unit) {
                 SwitchPreference(
                     title = stringResource(Res.string.range_test_enabled),
                     checked = formState.value.enabled,
-                    enabled = canConfigure,
+                    enabled = canConfigure || formState.value.enabled,
                     onCheckedChange = { formState.value = formState.value.copy(enabled = it) },
                     containerColor = CardDefaults.cardColors().containerColor,
                 )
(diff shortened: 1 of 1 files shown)


feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/component/RangeTestConfigItemList.kt
```diff
@@ -60,7 +65,7 @@ fun RangeTestConfigScreen(viewModel: RadioConfigViewModel, onBack: () -> Unit) {
                 SwitchPreference(
                     title = stringResource(Res.string.range_test_enabled),
                     checked = formState.value.enabled,
-                    enabled = state.connected,
+                    enabled = canConfigure,
```
jamesarich: Let the Enabled switch turn **off** on a public channel (just not on), so the corrective action is the obvious one — flip the switch — instead of a Save button that silently disables:

```suggestion
                    enabled = canConfigure || formState.value.enabled,
```

Behavior with this:
- public + off → grayed off, can't enable ✅
- public + on (CLI-set) → switch is live; flipping it off makes the form dirty → footer Save persists `false` ✅
- after toggling off it re-grays (can't flip back on) ✅

The footer then appears naturally via `isDirty`, so `additionalDirtyCheck` (line 56) can be dropped. I'd keep the `onSave` force-`false` as cheap defense-in-depth — it also covers the brief load race where `channelList` is empty and `isPublicPrimaryChannel` is transiently true.

jamesarich (changes requested):
The PSK `< 2` heuristic and the save-time enforcement are correct, and the dedup into `canConfigure` reads well. One UX change before merge — see the inline suggestion.

The corrective flow (device with Range Test enabled via CLI on a public channel) routes the user through a control that doesn't behave like one: the Enabled switch is grayed **on**, and the only way to disable is tapping Save, which silently forces `enabled = false`. That hidden side effect is the issue.

dubsector:
Applied, thank you for the reccomendation.  Changed the Enabled switch to `enabled = canConfigure || formState.value.enabled` and dropped `additionalDirtyCheck`. The corrective flow now works as you described: switch is live when it's on, user flips it off, form goes dirty naturally, Save persists the change. Kept the `onSave` force-false for the load-race edge case as you suggested.

jamesarich (approved):

PR #5999, opened 2026-06-28
fix(qr): Serialize channel import writes

jeremiah-k:
## Overview

This PR makes QR channel replacement imports deterministic by sending imported channel writes sequentially instead of launching each write as a separate fire-and-forget coroutine.

The previous replacement flow could start multiple channel writes, immediately replace the local cached settings, and then receive later channel updates from the radio or the local write path. It also used the local cache to decide which imported channels needed to be sent, which made QR replacement vulnerable to stale or partially updated local state.

QR replacement should be authoritative: importing a replacement channel set should send every imported slot to the radio, even if the local cache appears to already match. This change makes the channel write sequence explicit and awaited before replacing the local cached settings.

This fixes the observed multi-channel QR replacement issue where replacing a full channel set could leave a later channel with stale or incomplete state.

## Key Changes

- Added `getChannelReplacementList(new, currentSettings)` for authoritative QR  replacement imports.
- Kept the existing `getChannelList(new, old)` diff helper unchanged for the manual edit path.
- Replaced per-channel fire-and-forget writes in QR/channel replacement paths with sequential awaited channel writes.
- Emitted every imported slot during QR replacement, without skipping positions where the local cache appears unchanged.
- Emitted `DISABLED` entries for old trailing slots beyond the imported channel list.
- Preserved positional role assignment:
  - index 0 is `PRIMARY`
  - later imported slots are `SECONDARY`
  - trailing old slots are `DISABLED`
- Read the current channel set directly from the repository flow before building trailing disabled writes, avoiding the StateFlow placeholder window.
- Preserved existing LoRa config handling.
- Kept radio channel responses as the source of truth after import.

## Testing

- Added unit-test coverage for `getChannelReplacementList`.
- Covered unchanged imported slots, secondary slots, old trailing disabled slots, empty inputs, empty replacement over existing channels, single-entry replacement, and replacement lists larger than the current cache.
- Verified manually on device that QR replacement correctly replaces the full channel set and fixes the previously observed channel replacement issue.

## Migration Notes

- No user data migration is required.
- Existing channel data remains unchanged until the user imports or edits channels.
- This only changes the ordering and cache timing of QR channel replacement imports.

Related #6013

### DIFF
--- a/core/ui/src/commonMain/kotlin/org/meshtastic/core/ui/qr/ScannedQrCodeViewModel.kt
+++ b/core/ui/src/commonMain/kotlin/org/meshtastic/core/ui/qr/ScannedQrCodeViewModel.kt
@@ -20,10 +20,9 @@ import androidx.lifecycle.ViewModel
 import org.koin.core.annotation.KoinViewModel
 import org.meshtastic.core.repository.RadioConfigRepository
 import org.meshtastic.core.repository.RadioController
-import org.meshtastic.core.ui.util.getChannelList
+import org.meshtastic.core.ui.util.applyReplacementChannelSet
 import org.meshtastic.core.ui.viewmodel.safeLaunch
 import org.meshtastic.core.ui.viewmodel.stateInWhileSubscribed
-import org.meshtastic.proto.Channel
 import org.meshtastic.proto.ChannelSet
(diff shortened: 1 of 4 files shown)


jamesarich (approved):
nice

PR #6046, opened 2026-07-01
fix(connections): label the connecting-card button "Stop Connecting"

jamesarich:
## Why
The connections card shows a "Disconnect" button while a device is selected but still connecting (or not connected yet) — reported as confusing UX in Discord/#help, since there's nothing to disconnect from. See screenshot in the linked report showing "Not connected" status next to a "Disconnect" button.

## 🐛 Bug Fixes
- `DisconnectButton` now accepts an optional `label` override (defaults to "Disconnect", unchanged for the fully-connected card).
- The connecting-card (`ConnectingDeviceInfo`) now passes a new `stop_connecting` string ("Stop Connecting") instead of "Disconnect", since the card is shown whenever the outer screen state is `CONNECTING` (not yet `CONNECTED_WITH_NODE`).

## Testing Performed
- `./gradlew spotlessApply spotlessCheck detekt :feature:connections:build :core:resources:build` — all green.
- `python3 scripts/sort-strings.py` run (auto-triggered) to keep `strings.xml` / `strings-index.txt` sorted.

## Summary by CodeRabbit

* **New Features**
  * Added a localized **“Stop Connecting”** text option for connection flow states.
  * Updated the device connection UI to show a context-appropriate stop/cancel action instead of always using the generic disconnect wording.
* **Bug Fixes**
  * Improved disconnect/stop button labeling so the button accurately reflects the current connection status and intent.


jamesarich:
Addressed the outside-diff-range finding on `ConnectingDeviceInfo.kt` (button label ignoring the already-CONNECTED-but-missing-node-info / MUST_SET_REGION states): the disconnect button label is now derived from `connectionStatus` — "Disconnect" for `CONNECTED`/`CONNECTED_SLEEPING`/`MUST_SET_REGION`, "Stop Connecting" for `CONNECTING`/`RECONNECTING`/`NOT_CONNECTED` — in 4622a628d.

PR #6056, opened 2026-07-01
fix(qr): Filter duplicate ADD imports

jeremiah-k:
## Overview

This PR improves QR channel ADD-mode imports.

ADD mode preserves existing channels and shows only unique incoming channels in scanned order. Incoming channels that duplicate an existing channel or an earlier incoming channel are not included in the preview.

This change builds the ADD preview using semantic channel identity rather than raw `ChannelSettings` equality. Effective channel name and effective PSK are compared through the channel model, ensuring preset/default channels are handled consistently whether values are explicit or implicit.

Existing channels remain visible and selected. Unique incoming channels remain visible and are selected by default while device channel capacity allows. Duplicate incoming channels are omitted from the preview.

This behavior was manually verified to ensure that only non-duplicate channels are shown when adding.

This is one focused follow-up in the QR/channel import area; replacement sequencing and channel materialization behavior are intentionally left to separate follow-up work.

## Key Changes

* Added filtered ADD-mode preview construction for QR channel imports.
* Compared channels by effective name and effective PSK.
* Kept existing channels visible and selected.
* Kept unique incoming channels visible in scanned order.
* Omitted incoming channels that duplicate existing or earlier incoming channels.
* Selected unique incoming channels by default while device channel capacity remains.
* Kept over-capacity unique incoming channels visible but unchecked.
* Used the device-reported max channel count with a safe fallback.
* Froze the channel limit for each open import dialog so late node-info updates do not reset user selections.
* Kept QR REPLACE behavior unchanged.
* Kept channel replacement/write sequencing unchanged.

## Testing

* Added focused coverage for ADD-mode preview construction.
* Covered existing channel preservation.
* Covered unique incoming channels.
* Covered duplicate incoming channels being omitted.
* Covered duplicate entries inside the incoming channel list.
* Covered same-name/different-PSK and same-PSK/different-name cases.
* Covered preset/default name and PSK equivalence.
* Covered non-LongFast preset duplicate detection.
* Covered channel capacity limits.
* Covered preview settings and selection list size matching.
* Manually verified that only non-duplicate channels are shown during ADD imports.

## Migration Notes

* This only changes the QR ADD-mode import preview and default selections.

## Summary by CodeRabbit

* **New Features**
  * Improved QR scan channel previews with smarter duplicate detection and consistent Add/Replace selection defaults.
  * Channel selection limits now respect the device’s supported maximum across the dialog.

* **Bug Fixes**
  * Prevented late max-channel updates from resetting selections while the dialog is open by freezing the effective limit for the dialog session.
  * Updated the Accept button to enable only when the selection count is within the effective allowed range.
  * Refined Add-mode settings generation to match the new preview logic, including correct handling of placeholders and over-capacity items (visible but unselected).

### DIFF
--- a/core/ui/src/commonMain/kotlin/org/meshtastic/core/ui/qr/ScannedQrCodeDialog.kt
+++ b/core/ui/src/commonMain/kotlin/org/meshtastic/core/ui/qr/ScannedQrCodeDialog.kt
@@ -63,7 +63,7 @@ import org.meshtastic.core.resources.replace
 import org.meshtastic.core.resources.replace_channels_and_settings_description
 import org.meshtastic.core.ui.component.ChannelSelection
 import org.meshtastic.core.ui.theme.AppTheme
-import org.meshtastic.core.ui.util.mergeChannelSettingsForAdd
+import org.meshtastic.core.ui.util.getChannelPreviewForAdd
 import org.meshtastic.proto.ChannelSet
 
 @Composable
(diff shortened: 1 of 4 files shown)


jamesarich (approved):

jeremiah-k:
@jamesarich  I amended the last commit to preserve the same behavior while making it lint/compile-safe

PR #6076, opened 2026-07-02
fix(settings): Generate fresh PSK for named manual channels

jeremiah-k:
## Overview

This PR prevents newly created named manual channels from silently keeping the shared default PSK marker.

When a user adds a new manual channel and gives it a non-blank custom name, the dialog now generates a fresh random PSK only while that new channel is still eligible for name-based key generation. Existing channels and explicit user key choices are preserved.

This keeps default/unnamed channel behavior intact while avoiding delete/re-add flows that create a custom-named channel with shared-default key behavior.

## Key Changes

* Started newly added manual channels from `Channel.default.settings`.
* Tracked which newly added channels are eligible for name-based key generation.
* Generated a fresh random PSK when a new eligible channel is given a non-blank name and the current key is still default.
* Preserved explicitly entered PSKs, including the default marker.
* Preserved explicitly generated keys.
* Avoided rotating keys when renaming existing channels.
* Preserved generated-key state during the current edit flow so clearing an auto-named channel can restore the default key.
* Added focused dialog coverage for new-channel generation, explicit-key preservation, existing-channel preservation, and clear-name behavior.

## Testing

* Added coverage proving a newly named manual channel generates a fresh key.
* Added coverage proving an explicitly entered default marker is preserved.
* Added coverage proving existing default-PSK channels are not rotated when renamed.
* Added coverage for clearing an auto-generated key during the current edit flow.

## Migration Notes

* Existing channels are unchanged.
* This only changes the default behavior for newly added manual channels while they are being named.

## Summary by CodeRabbit

* **New Features**
  * Improved channel editing to keep PSK generation/restore behavior consistently tied to channel name changes.
  * Added per-channel PSK edit-state handling when adding new channels and after reordering.

* **Bug Fixes**
  * Prevented PSK edit state from drifting out of sync during drag-and-drop, cancel, or delete.
  * Renaming now preserves explicit/default PSKs and avoids unintended PSK regeneration when key generation is disabled.

* **Tests**
  * Expanded test coverage for PSK/name edit-state transitions and list move/replace helpers.

### DIFF
--- a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/channel/ChannelConfigScreen.kt
+++ b/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/channel/ChannelConfigScreen.kt
@@ -40,6 +40,7 @@ import androidx.compose.runtime.remember
 import androidx.compose.runtime.saveable.listSaver
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
+import androidx.compose.runtime.snapshots.SnapshotStateList
 import androidx.compose.runtime.toMutableStateList
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
@@ -70,6 +71,7 @@ import org.meshtastic.feature.settings.radio.channel.component.ChannelCard
 import org.meshtastic.feature.settings.radio.channel.component.ChannelConfigHeader
 import org.meshtastic.feature.settings.radio.channel.component.ChannelLegend
 import org.meshtastic.feature.settings.radio.channel.component.ChannelLegendDialog
+import org.meshtastic.feature.settings.radio.channel.component.ChannelPskEditState
 import org.meshtastic.feature.settings.radio.channel.component.EditChannelDialog
 import org.meshtastic.feature.settings.radio.component.LoadingOverlay
 import org.meshtastic.feature.settings.radio.component.PacketResponseStateDialog
@@ -122,12 +124,17 @@ private fun ChannelConfigScreen(
         rememberSaveable(saver = listSaver(save = { it.toList() }, restore = { it.toMutableStateList() })) {
             settingsList.toMutableStateList()
         }
+    val pskEditStatesInput =
+        rememberSaveable(saver = channelPskEditStatesSaver) {
+            List(settingsList.size) { ChannelPskEditState() }.toMutableStateList()
+        }
 
     val listState = rememberLazyListState()
     val dragDropState =
         rememberDragDropState(listState) { fromIndex, toIndex ->
             if (toIndex in settingsListInput.indices && fromIndex in settingsListInput.indices) {
-                settingsListInput.apply { add(toIndex, removeAt(fromIndex)) }
+                settingsListInput.move(fromIndex, toIndex)
+                pskEditStatesInput.move(fromIndex, toIndex)
             }
         }
 
@@ -143,11 +150,14 @@ private fun ChannelConfigScreen(
         EditChannelDialog(
             channelSettings = settingsListInput.getOrNull(index) ?: ChannelSettings(),
             modemPresetName = modemPresetName,
-            onAddClick = {
+            initialPskEditState = pskEditStatesInput.getOrElse(index) { ChannelPskEditState() },
+            onAddClick = { settings, pskEditState ->
                 if (settingsListInput.size > index) {
-                    settingsListInput[index] = it
+                    settingsListInput[index] = settings
+                    pskEditStatesInput[index] = pskEditState
                 } else {
-                    settingsListInput.add(it)
+                    settingsListInput.add(settings)
+                    pskEditStatesInput.add(pskEditState)
                 }
                 showEditChannelDialog = null
             },
@@ -177,6 +187,7 @@ private fun ChannelConfigScreen(
                     onClick = {
                         if (maxChannels > settingsListInput.size) {
                             settingsListInput.add(newUnnamedManualChannelSettings())
+                            pskEditStatesInput.add(ChannelPskEditState(canGeneratePskForName = true))
                             showEditChannelDialog = settingsListInput.lastIndex
                         }
                     },
@@ -231,7 +242,10 @@ private fun ChannelConfigScreen(
                             channelSettings = channel,
                             loraConfig = loraConfig,
                             onEditClick = { showEditChannelDialog = index },
-                            onDeleteClick = { settingsListInput.removeAt(index) },
+                            onDeleteClick = {
+                                settingsListInput.removeAt(index)
+                                pskEditStatesInput.removeAt(index)
+                            },
                             sharesLocation = locationChannel == index,
                         )
                     }
@@ -244,6 +258,7 @@ private fun ChannelConfigScreen(
                                 focusManager.clearFocus()
                                 settingsListInput.clear()
                                 settingsListInput.addAll(settingsList)
+                                pskEditStatesInput.replaceAll(settingsList.size, ChannelPskEditState())
                             },
                             positiveText = stringResource(Res.string.send),
                             onPositiveClicked = {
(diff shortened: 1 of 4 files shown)


jamesarich (approved):

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/channel/component/EditChannelDialog.kt
```diff
@@ -142,3 +142,25 @@ fun EditChannelDialog(
...
     )
 }
+
+private data class ChannelNameUpdate(val settings: ChannelSettings, val generatedPskForName: Boolean)
+
+private fun ChannelSettings.withUpdatedName(nameInput: String, generatedPskForName: Boolean): ChannelNameUpdate {
+    val name = nameInput.trim()
+    val defaultPsk = Channel.default.settings.psk
+    val clearGeneratedPsk = name.isBlank() && generatedPskForName
+    val generateNamedPsk = name.isNotBlank() && psk == defaultPsk
```
jamesarich: This gate never consults the `generatedPskForName` flag, so an *explicitly entered* PSK that equals the default marker is still overwritten: type `AQ==` into the PSK field (accepted, and `generatedPskForName = false` marks it explicit), then type a name — `psk == defaultPsk` matches and the explicit PSK is silently replaced with a random key. That's the exact case the PR description says is protected (“If the user explicitly enters a PSK … the dialog does not overwrite it during later name edits”). Tracking “user explicitly edited the PSK” and adding `&& !pskExplicitlyEdited` to this condition would close it.

Related scoping note: this dialog is also reachable for **existing** channels (every `ChannelCard` gets `onEditClick`), so renaming a pre-existing default-PSK channel still rotates its key and breaks compatibility with the rest of that channel's mesh. Pre-existing behavior, but the description's “scoped to user-created named manual channels” isn't enforced anywhere — the dialog can't tell a fresh FAB-added channel from an existing one. `ChannelConfigScreen` knows (`settingsListInput.lastIndex` at add time) and could pass that down.

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/channel/component/EditChannelDialog.kt
```diff
@@ -62,6 +62,7 @@ fun EditChannelDialog(
     var isFocused by remember { mutableStateOf(false) }
 
     var channelInput by remember(channelSettings) { mutableStateOf(channelSettings) }
+    var generatedPskForName by remember(channelSettings) { mutableStateOf(false) }
```
jamesarich: `remember(channelSettings)` makes this flag dialog-session-local, so the restore-on-clear behavior only holds within one uninterrupted edit: add a channel, name it (PSK auto-generates), **save**, reopen, clear the name — the flag re-initializes to `false`, `clearGeneratedPsk` never fires, and the result is an unnamed channel that renders under the modem-preset name but carries a random PSK: looks like the default channel, can't talk to it. If the reopen case should work, the dialog could offer the default-marker reset for any unnamed channel whose PSK isn't the default, regardless of how it got there.

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/channel/ChannelConfigScreen.kt
```diff
@@ -273,6 +273,8 @@ private fun ChannelConfigScreen(
     }
 }
 
+private fun newUnnamedManualChannelSettings(): ChannelSettings = ChannelSettings(psk = Channel.default.settings.psk)
```
jamesarich: This is structurally identical to `Channel.default.settings` (core/model `Channel.kt` builds `ChannelSettings(psk = defaultPSK.toByteString())` with every other field at proto default) — could use that constant directly. As written, if `Channel.default.settings` ever gains a field the two silently diverge.

jamesarich (commented):
Reviewed the PSK-generation state machine — the direction is right, but two of the PR's stated guarantees don't hold yet (details inline).

Minor notes not worth inline threads:
- Blank↔non-blank name cycles regenerate a fresh key each cycle within a session (visible pre-save, low severity).
- `nameInput.trim()` per keystroke still makes interior spaces untypable ("my base" → "myb"). Pre-existing, but now that it's encapsulated in `withUpdatedName` a save-time trim would be an easy fix.

jeremiah-k:
@jamesarich Thanks for the detailed pass.

I pushed a rebased update that addresses the state-machine issues called out inline. The fresh-PSK path is now scoped to newly added manual channels, existing default-PSK channels are preserved on rename, explicitly entered keys are preserved even when they match the default marker, and new manual channels now start from Channel.default.settings. I also added focused dialog coverage for those cases.

The generated-key restore state is kept for the current edit flow where we still know the key was generated by naming. I did not try to infer that later from an arbitrary persisted non-default PSK, since that could reset a real user key.

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/channel/ChannelConfigScreen.kt
```diff
@@ -244,6 +258,7 @@ private fun ChannelConfigScreen(
                                 focusManager.clearFocus()
                                 settingsListInput.clear()
                                 settingsListInput.addAll(settingsList)
+                                pskEditStatesInput.replaceAll(settingsList.size, ChannelPskEditState())
```
jamesarich, line 260: The Cancel path resets `pskEditStatesInput` here, but the Send path (`onPositiveClicked(settingsListInput)` just below → `viewModel.updateChannels`) doesn't, and `updateChannels` only updates `_radioConfigState.channelList`, not this `rememberSaveable` list. So a FAB-created channel that was just written to the device keeps `canGeneratePskForName = true` / `generatedPskForName = true`. If the user reopens that same channel without leaving the screen and **clears the name**, `applyChannelNameEdit` restores the PSK to the default `AQ==` marker — silently downgrading an already-provisioned channel to unencrypted (or, via a blank→name cycle, rotating its live key and desyncing it from peers). Re-seed `pskEditStatesInput` (and `settingsListInput`) from the committed result after a successful send, the same way Cancel does.

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/channel/component/EditChannelDialog.kt
```diff
@@ -142,3 +143,41 @@ fun EditChannelDialog(
...
+    nameInput: String,
+    pskEditState: ChannelPskEditState,
+): ChannelNameUpdate {
+    val name = nameInput.trim()
+    val defaultPsk = Channel.default.settings.psk
+    val clearGeneratedPsk = name.isBlank() && pskEditState.generatedPskForName && !pskEditState.pskExplicitlyEdited
+    val generateNamedPsk =
+        name.isNotBlank() &&
+            psk == defaultPsk &&
+            pskEditState.canGeneratePskForName &&
```
jamesarich, line 165: Root cause of the above and of the minor blank↔name rotation: `canGeneratePskForName` is never cleared once set, so it survives a commit. Even scoped to FAB channels, after the channel is sent it should stop being a regeneration candidate — consider clearing it once the channel is committed (or keying regeneration off "uncommitted new channel" rather than a sticky per-index flag).

jamesarich (changes requested):
Re-reviewed after the force-push — all three of my earlier findings are addressed, nicely and with tests:

- ✅ **Explicit PSKs preserved** via the `pskExplicitlyEdited` flag guarding regeneration.
- ✅ **Restore-after-reopen** now works — edit state moved to a screen-scoped `rememberSaveable(saver = channelPskEditStatesSaver)` that survives reopen and process death.
- ✅ **Existing channels no longer rotate** — regeneration is gated on `canGeneratePskForName`, seeded `true` only by the FAB.

Requesting changes on one thing the refactor introduced: a committed channel's key can be silently downgraded/rotated because the edit-state list isn't reset after a successful Send (inline). That's a security-relevant silent change, so worth closing before merge.

Minor: `newUnnamedManualChannelSettings()` is now a pure alias for `Channel.default.settings` — could inline it.

jamesarich (approved):
Re-reviewed at 902ee80a — both items from my last pass are fixed:

- ✅ **Post-Send edit state re-seeded:** the commit path now runs `pskEditStatesInput.replaceAll(committedSettings.size, ChannelPskEditState())`, so a just-sent FAB channel no longer carries `canGeneratePskForName`/`generatedPskForName`. Reopening it and clearing the name no longer downgrades it to the default marker or rotates its key. Covered by `replaceAll clears committed channel psk generation eligibility`.
- ✅ **`canGeneratePskForName` cleared on commit** (symmetric with the Cancel path).

No new regressions spotted. Clearing my changes-requested — LGTM.

PR #6077, opened 2026-07-02
fix(settings): Apply manual channel writes in order

jeremiah-k:
## Overview

This PR makes manual channel saves deterministic by writing channel changes in order instead of allowing overlapping write batches.

Previously, multi-channel edits could enqueue several radio writes without pacing or serialization. On constrained links, that made save, delete, and reorder flows more fragile and allowed overlapping saves to interleave.

This change builds an ordered channel update plan, applies writes sequentially, keeps the response state tied to the full batch, and reconciles already-enqueued writes if a batch is interrupted.

## Key Changes

* Built manual channel update plans in channel index order.
* Applied manual channel writes sequentially.
* Added pacing between channel writes.
* Serialized overlapping manual channel save batches.
* Computed queued save plans inside the mutex against the current canonical channel list.
* Kept batch success from firing while paced writes are still being enqueued.
* Surfaced enqueue failures as user-visible errors.
* Reconciled already-enqueued channel writes if a batch is interrupted or cancelled mid-plan.
* Updated local channel state only after ordered write requests are enqueued.
* Avoided committing partial final state when a write enqueue fails mid-plan.
* Preserved local channel migration behavior for the local node path.
* Added focused coverage for ordering, pacing, serialization, interrupted batches, and partial failures.

## Testing

* Added coverage proving channel writes are ordered by index.
* Added coverage proving overlapping saves do not interleave.
* Added coverage proving pacing happens between writes.
* Added coverage proving no delay is added after the final write.
* Added coverage proving batch success is held until enqueueing completes.
* Added coverage proving enqueue failure surfaces an error.
* Added coverage proving interrupted writes reconcile already-applied channel settings.
* Added coverage proving queued saves diff against current canonical state.

## Migration Notes

* Existing channel data is unchanged until the user edits and saves channels.
* This only changes the ordering, pacing, and recovery behavior of manual channel save writes.

## Summary by CodeRabbit

* **New Features**
  * Channel updates are now written as a single ordered batch, ensuring consistent results when multiple channels change together.
  * Channel-save progress is more accurate during the batch lifecycle, including pacing between writes.

* **Bug Fixes**
  * Prevents overlapping channel-save operations from interleaving, improving reliability.
  * Improved partial-failure handling: if a batch fails mid-sequence, settings are reconciled to the already-applied channels to keep state consistent.

* **Tests**
  * Expanded coroutine and timing assertions to verify ordered write behavior, progress transitions, and interruption recovery.

### DIFF
--- a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
+++ b/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
@@ -33,6 +33,8 @@ import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.update
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import org.jetbrains.compose.resources.StringResource
 import org.koin.core.annotation.InjectedParam
 import org.koin.core.annotation.KoinViewModel
@@ -221,6 +223,7 @@ open class RadioConfigViewModel(
     val mqttProbeStatus: StateFlow<MqttProbeStatus?> = _mqttProbeStatus.asStateFlow()
 
     private var probeJob: Job? = null
+    private val channelUpdateMutex = Mutex()
 
     /**
      * Run a one-shot reachability/credentials probe against an MQTT broker. Cancels any in-flight probe before starting
@@ -412,18 +415,21 @@ open class RadioConfigViewModel(
         safeLaunch(tag = "setRemoteChannels") {
             // Manual channel saves are an ordered batch: only update canonical local state after every write request is
             // enqueued. If any enqueue fails, safeLaunch reports it and leaves the saved state unchanged instead of
-            // blessing a partial delete/reorder result.
-            applyManualChannelUpdatePlan(
-                updatePlan = updatePlan,
-                writeChannel = { channel -> radioConfigUseCase.setRemoteChannel(destNum, channel) },
-                registerRequestId = ::registerRequestId,
-            )
+            // blessing a partial delete/reorder result. Serialize batches so two ordered write plans cannot interleave
+            // on the radio link.
+            channelUpdateMutex.withLock {
+                applyManualChannelUpdatePlan(
+                    updatePlan = updatePlan,
+                    writeChannel = { channel -> radioConfigUseCase.setRemoteChannel(destNum, channel) },
+                    registerRequestId = ::registerRequestId,
+                )
 
-            if (destNum == myNodeNum) {
-                packetRepository.migrateChannelsByPSK(old, new)
-                radioConfigRepository.replaceAllSettings(new)
+                if (destNum == myNodeNum) {
+                    packetRepository.migrateChannelsByPSK(old, new)
+                    radioConfigRepository.replaceAllSettings(new)
+                }
+                _radioConfigState.update { it.copy(channelList = new) }
             }
-            _radioConfigState.update { it.copy(channelList = new) }
         }
     }
 
--- a/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
+++ b/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
@@ -427,6 +427,7 @@ open class RadioConfigViewModel(
                 val updatePlan = getManualChannelUpdatePlan(new, current)
                 if (updatePlan.isEmpty()) return@withLock
                 if (!beginManualChannelBatch(updatePlan.size)) return@withLock
+                val batchRequestIds = mutableSetOf<Int>()
 
                 try {
                     applyManualChannelUpdatePlan(
@@ -434,7 +435,10 @@ open class RadioConfigViewModel(
                         currentSettings = current,
                         finalSettings = new,
                         writeChannel = { channel -> radioConfigUseCase.setRemoteChannel(destNum, channel) },
-                        registerRequestId = ::registerManualChannelBatchRequestId,
+                        registerRequestId = { packetId ->
+                            batchRequestIds.add(packetId)
+                            registerManualChannelBatchRequestId(packetId)
+                        },
                         onInterrupted = { result ->
                             reconcileInterruptedManualChannelUpdate(
                                 destNum = destNum,
@@ -446,10 +450,10 @@ open class RadioConfigViewModel(
                     commitManualChannelSettings(destNum = destNum, oldSettings = current, newSettings = new)
                     finishManualChannelBatch()
                 } catch (e: CancellationException) {
-                    abortManualChannelBatch()
+                    abortManualChannelBatch(batchRequestIds)
                     throw e
                 } catch (e: Throwable) {
-                    abortManualChannelBatch()
+                    abortManualChannelBatch(batchRequestIds)
                     Logger.w(e) { "Manual channel update failed after enqueue" }
                     e.message?.let(::sendError) ?: sendError(Res.string.unknown_error)
                 }
@@ -811,11 +815,10 @@ open class RadioConfigViewModel(
     }
 
     private fun beginManualChannelBatch(total: Int): Boolean {
-        if (requestIds.value.isNotEmpty() || radioConfigState.value.responseState is ResponseState.Loading) {
+        if (hasUnrelatedPendingRequest() || hasPendingRequestRegistration()) {
             Logger.w { "Manual channel update skipped while another radio request is pending" }
             return false
         }
-        manualChannelBatchRequestIds.clear()
         manualChannelBatchEnqueueing = true
         _radioConfigState.update { state ->
             state.copy(route = "", responseState = ResponseState.Loading(total = total))
@@ -826,14 +829,13 @@ open class RadioConfigViewModel(
     private fun finishManualChannelBatch() {
         manualChannelBatchEnqueueing = false
         if (requestIds.value.isEmpty()) {
-            manualChannelBatchRequestIds.clear()
             setResponseStateSuccess()
         }
     }
 
-    private fun abortManualChannelBatch() {
+    private fun abortManualChannelBatch(batchRequestIds: Set<Int>) {
         manualChannelBatchEnqueueing = false
-        removeManualChannelBatchRequestIds()
+        removeRequestIds(batchRequestIds)
     }
 
     private fun completeSetRequestOrProgressBatch() {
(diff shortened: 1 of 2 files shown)


feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -881,3 +898,24 @@ open class RadioConfigViewModel(
...
+    writeDelay: Duration = MANUAL_CHANNEL_WRITE_DELAY,
+    delayFn: suspend (Duration) -> Unit = { delay(it) },
+): ManualChannelUpdateResult {
+    val packetIds = mutableListOf<Int>()
+    for ((index, channel) in updatePlan.withIndex()) {
+        val packetId = writeChannel(channel)
+        packetIds.add(packetId)
+        registerRequestId(packetId)
+        if (index < updatePlan.lastIndex) {
+            delayFn(writeDelay)
```
jamesarich: The pacing opens a premature-Success race: packet IDs now register ~1s apart, and `processPacketResponse` fires `setResponseStateSuccess()` whenever `requestIds` empties. On a local save with sub-second BLE ACKs (the common case), the ACK for write 1 lands inside this delay window → `requestIds` empties → Success dialog mid-batch; the next `registerRequestId` then sees non-Loading state and resets to a fresh `Loading()` with `total = 1`, losing the batch's progress. The old code registered all IDs near-simultaneously so the window was negligible. Holding Success until the loop completes (e.g. seed the expected total up front, or gate `setResponseStateSuccess` on batch completion) would close it.

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -393,22 +399,33 @@ open class RadioConfigViewModel(
...
-            }
-        }
 
-        if (destNum == myNodeNum) {
-            safeLaunch(tag = "migrateChannels") {
-                packetRepository.migrateChannelsByPSK(old, new)
-                radioConfigRepository.replaceAllSettings(new)
+        val updatePlan = getManualChannelUpdatePlan(new, old)
+        if (updatePlan.isEmpty()) return
+        safeLaunch(tag = "setRemoteChannels") {
```
jamesarich, line 421: `safeLaunch` is called without an `errorEvents` flow here, so “safeLaunch reports it” in the comment below is log-only — nothing reaches the user when an enqueue fails. Combined with the pacing race, the user-visible outcome of a failed batch is either a spurious Success (an earlier ACK empties `requestIds`) or the generic 30s timeout — never an actual save-failed error, while the canonical state (correctly) stays old. Passing `errorEvents` / calling `sendError` on the failure path would surface it.

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -393,22 +399,33 @@ open class RadioConfigViewModel(
...
+            // Manual channel saves are an ordered batch: only update canonical local state after every write request is
+            // enqueued. If any enqueue fails, safeLaunch reports it and leaves the saved state unchanged instead of
+            // blessing a partial delete/reorder result. Serialize batches so two ordered write plans cannot interleave
+            // on the radio link.
+            channelUpdateMutex.withLock {
```
jamesarich, line 425: If the batch fails or is cancelled midway (nav-away during the now (n−1)×1s window cancels `viewModelScope`), writes 1..k are already enqueued to the radio but everything after the loop is skipped — device and app diverge with nothing triggering a re-read. The sibling QR path solved the same problem in `applyReplacementChannelSet` (core/ui `ProtoExtensions.kt`) with `finally { reconcileInterruptedReplacement(...) }` plus a `withContext(NonCancellable)` commit; this path needs the same treatment. Relatedly, the loop shape, the 1s constant, and the `delayFn` seam duplicate that helper — the write primitives differ (`setLocalChannel` vs `setRemoteChannel` + `registerRequestId`), but parameterizing the write fn would let a future pacing/reconciliation fix land in both paths.

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -393,22 +399,33 @@ open class RadioConfigViewModel(
...
-                val packetId = radioConfigUseCase.setRemoteChannel(destNum, channel)
-                registerRequestId(packetId)
-            }
-        }
 
-        if (destNum == myNodeNum) {
-            safeLaunch(tag = "migrateChannels") {
-                packetRepository.migrateChannelsByPSK(old, new)
-                radioConfigRepository.replaceAllSettings(new)
+        val updatePlan = getManualChannelUpdatePlan(new, old)
```
jamesarich: The plan is diffed against the caller-supplied `old` *before* the mutex is acquired, so a queued second save diffs against a stale baseline: `state.channelList` only updates at the end of batch 1, so batch 2's plan can skip a slot batch 1 changed on the device (it looks unchanged vs the stale `old`) and then bless `secondNew` wholesale — device keeps batch 1's value while the app shows batch 2's. The serialization test doesn't catch this because both calls pass the same `old`. Computing the plan inside `withLock` against the then-current canonical list would fix it.

jamesarich (commented):
The sequencing fix itself is right — the concurrent per-slot `safeLaunch` writes were the real bug — but the 1s pacing opens response-state races the concurrent code never had, and the failure paths don't do what the PR body says yet (details inline).

Minor notes:
- `sortedBy { it.index }` in `getManualChannelUpdatePlan` is a no-op — `getChannelList` iterates `0..max` ascending, so the output is already index-sorted.
- `ManualChannelUpdateResult`'s return value is ignored by the only production caller; the tests could assert on the `registerRequestId` capture instead.

jeremiah-k:
@jamesarich Thanks, this was a good catch.

I pushed a rebased update that tightens the ordered-save path around the cases you called out. The save plan is now computed inside the mutex against current canonical state, overlapping saves are serialized, batch success is held until enqueueing completes, enqueue failures surface an error, and interrupted batches reconcile the writes that were already sent instead of leaving app/device state silently divergent.

I also added focused coverage for the premature-success path, overlapping saves, enqueue failure, stale-baseline saves, and interrupted reconciliation.

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -740,6 +808,33 @@ open class RadioConfigViewModel(
...
+    private fun finishManualChannelBatch() {
+        manualChannelBatchEnqueueing = false
+        if (requestIds.value.isEmpty()) {
+            setResponseStateSuccess()
+        }
+    }
+
+    private fun abortManualChannelBatch() {
+        manualChannelBatchEnqueueing = false
+        requestIds.value = hashSetOf()
```
jamesarich: `abortManualChannelBatch` clears the **entire** `requestIds` set, not just this batch's ids — and the per-request 30s timeout coroutines from `registerRequestId` aren't cancelled. If any unrelated admin request (e.g. an in-flight getConfig/getOwner) is pending when a channel write throws, its ACK is then dropped by the response handler and its timeout no longer fires, leaving it silently unresolved; conversely a stale timeout can later fire `sendError(timeout)` over a fresh terminal state. Track and remove only the batch's own ids (and cancel their timeouts).

feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/RadioConfigViewModel.kt
```diff
@@ -740,6 +808,33 @@ open class RadioConfigViewModel(
         }
     }
 
+    private fun beginManualChannelBatch(total: Int) {
+        manualChannelBatchEnqueueing = true
+        _radioConfigState.update { state ->
+            state.copy(route = "", responseState = ResponseState.Loading(total = total))
```
jamesarich, line 824: `beginManualChannelBatch` unconditionally overwrites `route = ""` and `responseState = Loading(total)`. If a config/channel GET flow is active (route=CHANNELS, its own pending ids and total) when the user saves manual channel edits, its subsequent ACKs are misclassified against the batch's accounting and its progress is corrupted. Guard against clobbering an in-progress non-batch flow, or namespace the batch's progress state.

jamesarich (dismissed):
Re-reviewed after the force-push — all four of my earlier findings are fixed, with tests:

- ✅ **Premature Success** — batch seeds `Loading(total)` up front and routes mid-batch ACKs through `completeSetRequestOrProgressBatch()`/`incrementCompleted()`; Success deferred to batch end.
- ✅ **Silent failure** — `try/catch` now `sendError`s → `ResponseState.Error` (asserted by a test).
- ✅ **Partial reconciliation** — `reconcileInterruptedManualChannelUpdate` in a `withContext(NonCancellable)` block, matching `applyReplacementChannelSet`.
- ✅ **Stale baseline** — plan now computed inside `withLock` from live `channelList`, each batch commits unconditionally.

Requesting changes on a set of related issues the new batch state-machine introduced: it reaches into the *shared* response-state and can corrupt a concurrent non-batch admin flow (inline). If channel saves can never overlap another admin request on this screen these are benign, but that isn't enforced anywhere — worth scoping the batch bookkeeping to its own request-ids.

Minor carryovers: `sortedBy { it.index }` is still a no-op (getChannelList is already ascending), `ManualChannelUpdateResult` is still discarded by the only caller, and `applyManualChannelUpdatePlan` duplicates `applyReplacementChannelSet`'s paced-write engine (incl. a second 1s constant).

jamesarich (commented):
Re-reviewed at 6fbbddcb — both things I blocked on are fixed, with tests:

- ✅ `abortManualChannelBatch(batchRequestIds)` now removes/cancels only the batch's own request-ids and timeouts; unrelated in-flight requests survive (`aborted manual batch preserves unrelated pending request`).
- ✅ `beginManualChannelBatch` now returns false and bails when a non-batch request is pending/registering, instead of clobbering its `route`/`Loading` state.

Dismissing my changes-requested. Three smaller things the *allow queued manual channel saves* commit introduced — none blocking, but worth a look:

1. **Silent save drop (worth checking).** `hasPendingRequestRegistration()` is true whenever `requestIds.isEmpty() && manualChannelBatchRequestIds.isEmpty() && responseState is Loading`, so a Save tapped while the pane is in a bare Loading state (e.g. right after `ensureLoadingForRemote()`, before any request registers) is dropped with only a log — no user feedback. If that state is reachable with Save enabled, the user retries blind.
2. **Progress-bar mix (cosmetic).** Queue a second save while the first batch's ACKs are still outstanding: the first batch's un-acked ids count as 'related', so the second `beginManualChannelBatch` proceeds and its `Loading(total)` absorbs the leftover ids — completed/total then mixes both batches. Writes stay ordered under the mutex, so no data loss, just a wrong count.
3. **No commit rollback (low likelihood).** If `replaceAllSettings` throws after `migrateChannelsByPSK` succeeds in the commit, `packetRepository` is migrated while `radioConfigRepository`/state aren't, with no reconciliation. Both are local DB writes so unlikely, but there's no rollback.

jamesarich (approved):

PR #6079, opened 2026-07-05
fix(dfu): Detect legacy bootloaders before secure fallback

jeremiah-k:
## Overview

This PR makes BLE DFU protocol detection more reliable when a device reboots into a Legacy DFU bootloader.

Previously, if the app missed the Legacy DFU service during the initial scan, the update flow could continue down the Secure DFU path and keep retrying the wrong service. That made OTA updates fail even though the device was advertising a valid bootloader protocol.

This change keeps bootloader detection explicit, preserves inconclusive scan results, and only falls back to the alternate protocol before a DFU session has actually connected.

## Key Changes

* Added explicit bootloader detection results for Legacy, Secure, and Unknown.
* Kept inconclusive detection from being treated as Secure-only.
* Moved protocol ordering and fallback decisions into a dedicated coordinator.
* Tried Legacy first when Legacy is observed.
* Tried Secure first when Secure is observed.
* Tried Legacy first when detection is inconclusive, since Unknown can mean the Legacy advertisement was missed.
* Allowed alternate-protocol fallback only before protocol connection succeeds.
* Preserved the full Legacy session retry budget for observed Legacy and inconclusive Legacy-first attempts.
* Preserved Legacy stale-session reset behavior.
* Preserved existing bond handling.
* Kept failed transport cleanup non-cancellable.
* Added focused coverage for protocol ordering, fallback, retry limits, and post-connect fallback blocking.

## Testing

* Added unit coverage for Legacy, Secure, and Unknown ordering.
* Added coverage for fallback before connection.
* Added coverage proving fallback is blocked after protocol connection.
* Added coverage for retry-budget behavior.
* Added coverage proving inconclusive detection gives the Legacy-first attempt the Legacy recovery retry budget.

## Migration Notes

* Existing BLE bonds are unchanged.
* This only changes DFU bootloader protocol detection and fallback behavior.

### DIFF
--- a/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/SecureDfuHandler.kt
+++ b/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/SecureDfuHandler.kt
@@ -160,16 +160,13 @@ internal class DfuFallbackCoordinator(private val detection: BootloaderDetection
 
     /**
      * Session retry budget for the given protocol in this detection context.
-     * - Legacy primary when detection is [BootloaderDetection.LegacyObserved] OR [BootloaderDetection.Unknown]:
-     *   [LEGACY_SESSION_ATTEMPTS] (3). Unknown can still be a Legacy bootloader whose advertisement the detection scan
-     *   missed, so its Legacy primary needs the same reset-prime budget as a confirmed Legacy detection.
-     * - Everything else (Secure primary, alternate protocols): [LIMITED_SESSION_ATTEMPTS] (1) to bound probe/fallback
-     *   time.
+     * - Confirmed Legacy primary ([BootloaderDetection.LegacyObserved]): [LEGACY_SESSION_ATTEMPTS] (3) to cover the
+     *   stale-session reset-prime recovery path.
+     * - Unknown Legacy primary and alternate protocols: [LIMITED_SESSION_ATTEMPTS] (1) to keep speculative probes short
+     *   enough that a Secure bootloader still has time to accept its fallback attempt.
      */
     private fun sessionAttemptsFor(protocol: DfuProtocolKind, isPrimary: Boolean): Int = when {
-        protocol == DfuProtocolKind.LEGACY &&
-            isPrimary &&
-            (detection == BootloaderDetection.LegacyObserved || detection == BootloaderDetection.Unknown) ->
+        protocol == DfuProtocolKind.LEGACY && isPrimary && detection == BootloaderDetection.LegacyObserved ->
             LEGACY_SESSION_ATTEMPTS
 
         else -> LIMITED_SESSION_ATTEMPTS
@@ -397,7 +394,7 @@ class SecureDfuHandler(
                 is DfuUploadResult.Failure -> {
                     lastError = result.error
                     protocolEngaged = protocolEngaged || result.protocolEngaged
-                    Logger.w {
+                    Logger.w(result.error) {
                         "DFU: upload session $attempt/$attempts failed ($protocol): ${result.error::class.simpleName}"
                     }
                     // A stock bootloader holding a wedged session from an interrupted flash rejects START with
@@ -538,7 +535,7 @@ class SecureDfuHandler(
                     "DFU: Connect attempt $attempt/$CONNECT_ATTEMPTS via $protocol (scan service=$serviceUuid)"
                 }
                 transport.connectToDfuMode().onFailure {
-                    Logger.w {
+                    Logger.w(it) {
                         "DFU: Connect attempt $attempt/$CONNECT_ATTEMPTS via $protocol failed: ${it.message}"
                     }
                 }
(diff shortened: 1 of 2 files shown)


feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/SecureDfuHandler.kt
```diff
@@ -56,13 +56,131 @@ import org.meshtastic.feature.firmware.ota.retryWithDelay
...
+                }
+            }
+
+            when (val result = uploadWithRetry(protocol, sessionAttempts)) {
+                DfuUploadResult.Success -> return
+
+                is DfuUploadResult.Failure -> {
+                    if (result.protocolEngaged || !hasAlternateProtocol) {
+                        priorError?.let { result.error.addSuppressed(it) }
+                        throw result.error
```
jamesarich: On exhaustion the thrown error is the **alternate** leg's, with the primary only `addSuppressed` — and the UI renders `e.message`. So a legacy device that fails both legs shows "Failed to connect… via SECURE after 4 attempts", naming the protocol it doesn't speak. Consider throwing the primary (detection-matched) leg's error and suppressing the alternate, or composing a message that names both.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/SecureDfuHandler.kt
```diff
@@ -56,13 +56,131 @@ import org.meshtastic.feature.firmware.ota.retryWithDelay
...
+        }
+        throw IllegalStateException("DFU fallback exhausted with non-empty protocol list (detection=$detection)")
+    }
+
+    /**
+     * The ordered list of protocols to attempt for this detection outcome. The first element is the primary; the second
+     * is the alternate, attempted only if the primary fails before protocol engagement.
+     */
+    private fun BootloaderDetection.orderedProtocols(): List<DfuProtocolKind> = when (this) {
+        BootloaderDetection.LegacyObserved -> listOf(DfuProtocolKind.LEGACY, DfuProtocolKind.SECURE)
```
jamesarich, line 156: Conclusive detections still carry the opposite protocol as a fallback leg, so a *pre-engagement* connect flake on the primary triggers a full doomed alternate pass (up to ~140s of scans+retries against a service the device never advertises). The insurance is reasonable for a wrong "conclusive" call, but gating the alternate leg on "primary failed *after* engaging its DFU service" (vs. never connected) would avoid the worst-case stall on a simple flake.

jamesarich (commented):
Re-reviewed after the force-push. Two of the bigger items are addressed — nice:

- ✅ Unknown detection now grants the Legacy primary the full 3-attempt budget, so the stale-session reset-prime path can actually run and a wedged legacy bootloader with a missed 1530 ad recovers instead of hard-failing.
- ✅ `DfuFallbackCoordinatorTest` now covers the LegacyObserved→Secure ordering, per-leg budgets (`listOf(3, 1)`), and the suppressed-error shape.

A couple of things still stand (inline), plus a latency note: the 3-attempt fix means an *Unknown*-detected Secure device (FE59 ad missed) now burns 3× the doomed Legacy connect cycle before the Secure leg is tried — acceptable, but worth a comment so it's a deliberate choice. Not blocking.

jeremiah-k:
@jamesarich I pushed a refreshed branch with the fallback behavior tightened.

Conclusive Legacy/Secure detections now only try the alternate protocol after the primary protocol has connected, and the surfaced error stays tied to the primary/detection-matched failure. The Unknown path still goes Legacy-first with the full Legacy retry budget so the stale-session reset case remains covered.

jamesarich (commented):
Reviewed the fallback state machine closely — the core safety is solid: wrong-protocol attempts fail cleanly at the scan step (each transport scan-filters on its own service UUID, so a Legacy attempt never writes opcodes to a Secure bootloader), the `protocolEngaged` gate has no suspension-point window where a live session could be torn down, and the `NonCancellable` cleanup can't leak a half-open transport. Nice.

One reliability concern before this goes in — the `Unknown` retry budget:

`sessionAttemptsFor` gives `Unknown → Legacy` the **full `LEGACY_SESSION_ATTEMPTS` (3)** budget, and `orderedProtocols` puts Legacy first for `Unknown`. So a genuinely Secure-only device that momentarily reads `Unknown` (exactly the transient causes your own KDoc lists — scan duty-cycling, adv cache, bootloader not yet re-advertising after the reboot wait) burns **3 doomed Legacy sessions** before it ever tries Secure. Each is a `connectWithRetry` (4 attempts) scanning ~10s for the `1530` service that isn't there — order of minutes of "connecting" with the phone never touching the device. By the time Secure is tried once, the Secure bootloader may well have hit its DFU-inactivity timeout and rebooted back to app, so the flash fails — where the old Unknown→Secure path completed it immediately.

Suggest dropping `Unknown`→Legacy to `LIMITED_SESSION_ATTEMPTS` (1) so fallback to Secure happens in tens of seconds instead of minutes. Legacy-first ordering for `Unknown` is reasonable; it's just the 3× budget on the speculative primary that's expensive on the ambiguous case this PR is meant to help.

Minor: the per-attempt `Logger.w { ... }` dropped the throwable arg, so retry-loop failures now log only the exception class name, not the stack — a bit harder to field-diagnose. Terminal error is still preserved in the final `Failure`, so not a blocker.

jamesarich (approved):
Re-reviewed after the fallback-budget commits. Both points addressed: `Unknown`→Legacy now bounded to a single speculative probe (1 attempt) before Secure fallback — new `Unknown gives Legacy primary 1 session attempt` test pins it — so a mis-detected Secure device no longer burns minutes of doomed Legacy scans. Per-attempt session log restored to `Logger.w(result.error)`. Wrong-protocol scan-filter safety, the protocolEngaged gate, and NonCancellable cleanup were already sound. LGTM.

PR #6080, opened 2026-07-05
fix(firmware): Confirm ESP32 OTA mode before disconnecting

jeremiah-k:
## Overview

This PR makes ESP32 OTA startup wait for firmware confirmation before disconnecting the normal mesh connection.

Previously, the app could treat delivery of the reboot request as enough to continue into OTA mode. If firmware rejected the requested OTA mode because the installed OTA loader did not support BLE or Wi-Fi, the app could still disconnect and enter a doomed retry loop.

This change waits for a fresh firmware-side OTA response before switching away from the normal connection.

## Key Changes

* Cleared stale OTA notifications before sending the ESP32 OTA request.
* Waited for a fresh firmware confirmation or rejection before disconnecting mesh.
* Treated OTA-loader rejection messages as a startup failure.
* Treated missing confirmation within the bounded wait as a timeout.
* Kept the normal mesh connection active on rejection or timeout.
* Continued into OTA transport setup only after firmware confirms the requested OTA mode.
* Suppressed duplicate generic OTA status popups.
* Added clearer user-facing errors for rejected and timed-out OTA startup.
* Wrapped firmware string lookup safely for headless test environments.
* Added focused coverage for confirmed, rejected, and timed-out startup paths.

## Testing

* Added coverage proving mesh delivery alone is not treated as OTA-mode success.
* Added coverage proving unsupported-loader rejection does not disconnect mesh.
* Added coverage proving timeout does not disconnect mesh.
* Added coverage proving confirmed startup continues into the existing OTA flow.

## Migration Notes

* Existing device connections and saved addresses are unchanged.
* This only changes ESP32 OTA startup gating before the temporary OTA connection begins.

## Summary by CodeRabbit

* **New Features**
  * Added new firmware recovery and OTA update messages, including timeout and unsupported-device wording.
  * Introduced an OTA preflight confirmation step that waits for the device’s OTA entry-mode confirmation before continuing.
* **Bug Fixes**
  * Prevented duplicate generic alerts from being shown for OTA status notifications.
  * Improved handling of OTA rejection and timeouts with clearer terminal error states.
* **Tests**
  * Added/expanded unit coverage for OTA confirmation, rejection, and timeout behavior.


core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/FromRadioPacketHandlerImpl.kt
```diff
@@ -132,6 +132,11 @@ class FromRadioPacketHandlerImpl(
         serviceStateWriter.setClientNotification(cn)
 
         scope.handledLaunch {
+            if (cn.isOtaStatusNotification()) {
```
jamesarich: This suppresses only the OS-notification dispatch. `setClientNotification(cn)` above still runs, and `UIViewModel` (not modified on this branch) still calls `showAlert` for every non-null `clientNotification` — so the generic "Rebooting to BLE OTA" popup this PR claims to remove still fires when #6080 is merged standalone. The gate belongs on the `UIViewModel` alert path too (it's in #6081 — pull it forward or restack).

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/Esp32OtaUpdateHandler.kt
```diff
@@ -150,14 +171,28 @@ class Esp32OtaUpdateHandler(
...
-                disconnectMeshService()
-                // Give BLE stack time to fully release the GATT connection
-                delay(GATT_RELEASE_DELAY_MS)
+                // Step 3: Preflight — wait for the firmware to confirm OTA entry BEFORE tearing down the mesh
+                // transport. The mesh ACK only confirms packet delivery; it does NOT mean the device's OTA loader
+                // supports the requested transport. The firmware surfaces loader-capability failures as a
+                // meshtastic.ClientNotification warning ("OTA Loader does not support <mode>", etc.) emitted before
+                // its scheduled reboot, so we race that signal against a bounded timeout. On rejection or timeout we
+                // fail fast and leave the mesh connection intact — no disconnect, no OTA-transport retry loop.
+                if (runOtaPreflight(rebootMode, baselineMessage, updateState) !is OtaPreflightResult.Confirmed) {
```
jamesarich: OTA is hard-gated on a `Confirmed` notification within 5s with no firmware-version fallback. If any released firmware handles the `ota_request` admin message but predates the "Rebooting to <mode> OTA" emission, it reboots into OTA yet never sends the notification → permanent Timeout, and OTA that worked before this PR is now broken for it. Please confirm the emission shipped no later than `ota_request` handling; otherwise gate the preflight on firmware version (fall back to the old fixed-delay path below the threshold).

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/Esp32OtaUpdateHandler.kt
```diff
@@ -340,4 +383,99 @@ class Esp32OtaUpdateHandler(
...
+     */
+    private suspend fun runOtaPreflight(
+        rebootMode: Int,
+        baselineMessage: String?,
+        updateState: (FirmwareUpdateState) -> Unit,
+    ): OtaPreflightResult =
+        when (val preflight = awaitOtaConfirmation(otaPreflightTimeoutMs, rebootMode, baselineMessage)) {
+            is OtaPreflightResult.Confirmed -> {
+                Logger.i { "ESP32 OTA: Preflight confirmed; releasing mesh transport for OTA" }
+                disconnectMeshService()
```
jamesarich: New: on the **Confirmed** path you `disconnectMeshService()` but never clear the confirmation notification, so "Rebooting to BLE OTA" stays resident in `clientNotification`. Combined with the ungated `UIViewModel` alert, that means a spurious popup on the *success* path this PR was meant to make silent. Clear it after a confirmed match.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/Esp32OtaUpdateHandler.kt
```diff
@@ -340,4 +383,99 @@ class Esp32OtaUpdateHandler(
...
+        timeoutMs: Long,
+        rebootMode: Int,
+        baselineMessage: String?,
+    ): OtaPreflightResult {
+        val modeName = otaModeName(rebootMode)
+        val matched =
+            withTimeoutOrNull(timeoutMs) {
+                serviceRepository.clientNotification.first { cn ->
+                    val msg = cn?.message
+                    msg != null && msg != baselineMessage && msg.contains(OTA_KEYWORD)
```
jamesarich: The timeout message (`firmware_update_ota_timeout`, "Device did not enter OTA mode…") is inverted: `triggerRebootOta` was already sent, so on a lost/late confirmation the device may well be in OTA mode. Also this predicate is an untrimmed `contains("OTA")` catch-all while the `isOtaStatusNotification` suppression uses a trimmed prefix allowlist — the two classifiers disagree on whitespace and on any OTA-containing message outside the four prefixes. Worth extracting one shared classifier.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/Esp32OtaUpdateHandler.kt
```diff
@@ -79,11 +97,14 @@ class Esp32OtaUpdateHandler(
     private val firmwareFileHandler: FirmwareFileHandler,
     private val radioController: RadioController,
     private val nodeRepository: NodeRepository,
+    private val serviceRepository: ServiceRepository,
```
jamesarich: Redundant DI: `RadioController` (already injected) exposes `clientNotification` and `clearClientNotification` delegating to this same `ServiceRepository`. Dropping the extra param would also undo the `serviceRepository`-mock churn the tests had to add.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/Esp32OtaUpdateHandler.kt
```diff
@@ -212,7 +247,14 @@ class Esp32OtaUpdateHandler(
         firmwareUri: CommonUri?,
         updateState: (FirmwareUpdateState) -> Unit,
     ): FirmwareArtifact? {
-        val downloadingMsg = getStringSuspend(Res.string.firmware_update_downloading_percent, 0).stripFormatArgs()
+        // safeCatchingAll swallows Skiko ExceptionInInitializerError on headless JVM tests where
+        // compose-resources can't load native libs. Production resolves the localized string normally;
+        // tests fall back to empty and the Downloading state still emits with a blank message.
+        val downloadingMsg =
+            safeCatchingAll { getStringSuspend(Res.string.firmware_update_downloading_percent, 0) }
```
jamesarich: This `safeCatchingAll` is a production behavior change (silent blank label on any resource failure) that the comment admits exists only to dodge a headless-test Skiko init crash — and it's the sole guarded `getStringSuspend` in the file, so it's inconsistent anyway. Better to stub string resolution in the test than swallow it in prod.

jamesarich (changes requested):
Re-reviewed after the force-push. The preflight gating is a good idea, but a few things block merge as-is — most importantly the PR's own headline claim isn't delivered by this branch alone.

**The "suppress duplicate OTA status alerts" commit only gates the wrong path.** It suppresses `notificationManager.dispatch` in `FromRadioPacketHandlerImpl`, but `setClientNotification(cn)` still fires and `UIViewModel`'s alert collector — untouched on *this* branch — still pops the generic dialog for every OTA notification. The `UIViewModel` gate exists only in #6081's extra commits, so #6080 merged on its own still shows the duplicate popup it says it removes. Either fold the `UIViewModel` gate into this PR or reorder the stack.

Other blockers and notes inline. The older-firmware concern (item on line 193) needs a firmware-history check before this can ship — if any released firmware handles `ota_request` but predates the confirmation notification, this permanently breaks OTA for it.

jeremiah-k:
Superseded by #6081

PR #6081, opened 2026-07-05
fix(firmware): Retry ESP32 OTA service connections cleanly

jeremiah-k:
## Overview

This PR hardens ESP32 OTA after the device has confirmed that it is rebooting into OTA mode.

Even after confirmation, the temporary OTA service may not be immediately reachable. BLE can connect before the expected OTA GATT service is available, and Wi-Fi can fail while the loader is still rebooting, reconnecting, starting its server, or using a different DHCP address than the normal firmware connection.

This change retries post-confirmation OTA connections with fresh transports, closes failed attempts before retrying, and uses the ESP32 OTA loader’s UDP discovery broadcast to find the post-reboot Wi-Fi address when available.

## Key Changes

* Created a fresh OTA transport for each post-confirmation connection attempt.
* Closed failed transports before retrying.
* Preserved cancellation without wrapping it as an OTA failure.
* Treated missing BLE OTA service/profile setup as a retryable connection failure.
* Closed the failed BLE GATT path when OTA profile setup fails.
* Listened for the ESP32 OTA loader’s UDP discovery broadcast after Wi-Fi OTA confirmation.
* Used the discovered post-reboot OTA IP address when available.
* Fell back to the configured device IP when UDP discovery does not find the loader.
* Retried Wi-Fi/TCP connections with a fresh transport per attempt.
* Suppressed OTA status notifications from the generic alert path.
* Added a clearer final error when firmware confirmed OTA mode but the OTA service stayed unreachable.
* Added focused coverage for BLE retry, Wi-Fi discovery/readiness, fresh transport creation, failed transport cleanup, and final error behavior.

## Testing

* Added coverage proving BLE retries with a fresh transport after service/profile failure.
* Added coverage proving failed BLE transports are closed before retry.
* Added coverage proving Wi-Fi discovery can replace the configured address with the OTA loader’s broadcast address.
* Added coverage proving Wi-Fi falls back to the configured address when discovery misses.
* Added coverage proving Wi-Fi retries with fresh transports after TCP failure.
* Added coverage proving the final failure message is surfaced after retry exhaustion.
* Verified manually on device through the combined firmware-update flow.

## Migration Notes

* Existing ESP32 OTA startup confirmation behavior is unchanged.
* This only changes the post-confirmation BLE/Wi-Fi OTA connection retry path.

## Summary by CodeRabbit

* **New Features**
  * Improved firmware update flow with better OTA confirmation handling and device discovery for WiFi updates.
  * Added clearer messaging for unsupported OTA update cases and device search status.

* **Bug Fixes**
  * OTA-related status notifications are now suppressed during an active update, reducing distracting popups.
  * Firmware update timeouts and connection failures now return clearer error states instead of failing unexpectedly.

* **Tests**
  * Expanded automated coverage for OTA notification detection, update preflight behavior, and WiFi discovery.


feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/WifiOtaDiscovery.kt
```diff
@@ -0,0 +1,88 @@
...
+        // limited broadcasts (255.255.255.255) are typically delivered without it; if a specific device filters
+        // them, add an expect/actual multicast-lock wrapper and acquire it for the duration of [receive].
+        safeCatching<String?> {
+            withTimeoutOrNull(timeoutMs) {
+                val selector = SelectorManager(ioDispatcher)
+                try {
+                    val socket = aSocket(selector).udp().bind(InetSocketAddress("0.0.0.0", port))
+                    try {
+                        Logger.i { "WiFi OTA: Listening for OTA device discovery broadcast on port $port" }
+                        val datagram = socket.receive()
```
jamesarich: Discovery accepts the **first datagram from any host** on port 3232 (the ArduinoOTA/espota default) with no payload validation, and the discovered IP then replaces the configured IP for every TCP retry with no fallback (see line 368). Any other device chattering on 3232 — a second ESP in OTA mode, an espota client, IoT gear — hijacks the whole update to the wrong host and it can't recover. Validate the payload (expect a known OTA beacon signature) and fall back to the configured IP if the datagram doesn't match.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/WifiOtaDiscovery.kt
```diff
@@ -0,0 +1,88 @@
...
+        val bytes = sock.resolveAddress()
+        if (bytes != null && bytes.size == IPV4_ADDRESS_BYTES) {
+            return bytes.joinToString(".") { (it.toInt() and BYTE_MASK).toString() }
+        }
+        return sock.hostname.takeIf { it.isNotBlank() }
+    }
+
+    private const val IPV4_ADDRESS_BYTES = 4
+    private const val BYTE_MASK = 0xFF
+    private const val DEFAULT_TIMEOUT_MS = 15_000L
```
jamesarich: `DEFAULT_TIMEOUT_MS = 15_000L` and discovery runs once before the retry loop, so a loader that serves TCP on the known IP but never broadcasts costs a fixed 15s stall before the first connect attempt, and discovery is never re-run across the 10 retries (a late broadcast can't correct a wrong IP). Consider racing discovery against a direct connect to the configured IP, or shortening the timeout.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/Esp32OtaUpdateHandler.kt
```diff
@@ -340,4 +524,99 @@ class Esp32OtaUpdateHandler(
...
+        timeoutMs: Long,
+        rebootMode: Int,
+        baselineMessage: String?,
+    ): OtaPreflightResult {
+        val modeName = otaModeName(rebootMode)
+        val matched =
+            withTimeoutOrNull(timeoutMs) {
+                serviceRepository.clientNotification.first { cn ->
+                    val msg = cn?.message
+                    msg != null && msg != baselineMessage && msg.contains(OTA_KEYWORD)
```
jamesarich: New classifier bug: this gate is a substring `contains("OTA")`, but the suppression path uses a word-boundary `\bOTA\b`. A message like "QUOTA…"/"ROTATE…" arriving in the 5s window is consumed here as the OTA response (then classified Rejected since it doesn't start with "Rebooting to"), aborting a legitimately-succeeding update. Use the same matcher on both sides.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/Esp32OtaUpdateHandler.kt
```diff
@@ -79,11 +107,21 @@ class Esp32OtaUpdateHandler(
     private val firmwareFileHandler: FirmwareFileHandler,
     private val radioController: RadioController,
     private val nodeRepository: NodeRepository,
+    private val serviceRepository: ServiceRepository,
     private val bleScanner: BleScanner,
     private val bleConnectionFactory: BleConnectionFactory,
     private val dispatchers: CoroutineDispatchers,
 ) : FirmwareUpdateHandler {
 
+    internal var otaPreflightTimeoutMs: Long = OTA_PREFLIGHT_TIMEOUT_MS
```
jamesarich: These 8 mutable `internal var` seams live on a Koin `@Single`, and production control flow branches on them (line 355: `wifiTransportFactoryOverride != null && wifiOtaDiscoveryOverride == null` decides whether real UDP discovery runs). A stray write, or a test mutating the shared singleton, changes real update behavior. Prefer constructor params with defaults.

core/ui/src/commonMain/kotlin/org/meshtastic/core/ui/viewmodel/UIViewModel.kt
```diff
@@ -256,6 +257,11 @@ class UIViewModel(
         serviceRepository.clientNotification
             .filterNotNull()
             .onEach { notification ->
+                // OTA status notifications (e.g. "Rebooting to WiFi OTA") are consumed by the firmware update
+                // preflight gate — don't show a popup dialog for them.
+                if (notification.isOtaStatusNotification()) {
```
jamesarich: This gate (and the `FromRadioPacketHandlerImpl` one) suppress OTA notifications with no is-update-active check, and this path never calls `clearClientNotification`. Two consequences: an OTA triggered by another client (CLI over TCP) or a late confirmation after preflight timeout is swallowed with zero user feedback; and the suppressed value stays resident in the conflated `StateFlow`, so a later notification with an identical message string is never re-emitted. Gate on "this device has an active update flow", and clear after suppressing.

jamesarich (changes requested):
Re-reviewed after the force-push. Two good fixes landed:

- ✅ `senderHost` now uses `resolveAddress()` IPv4 bytes instead of Ktor's reverse-DNS `hostname`.
- ✅ `readinessAlreadyWaited = discoveredIp != null`, so an instant discovery failure still honors the 8s wait.

But the UDP discovery path has a correctness hole that breaks the feature on real networks (inline), so I'm requesting changes. Remaining items and one new classifier bug are also inline.

jeremiah-k:
I pushed the refreshed post-confirm branch.

This now keeps OTA status detection shared and word-boundary based, gates duplicate OTA-status suppression to an active update flow, validates Wi-Fi OTA discovery beacons before using a discovered address, preserves the readiness wait on discovery misses, and uses the environment seam instead of mutable singleton test hooks.

I also validated the full ESP32 OTA stack on hardware through the combined branch. The BLE GATT cache refresh/reconnect layer is split into a follow-up branch; this PR is the post-confirm retry/preflight layer.

jamesarich (approved):

jamesarich (approved):
Reviewed the post-confirmation OTA retry machinery closely — it's sound. Verified: fresh transport per attempt with each failed transport closed exactly once under `NonCancellable` (idempotent on the WiFi double-close path), both retry loops finite (BLE 5 / WiFi 10), preflight and UDP discovery both time-bounded, `CancellationException` rethrown on every path, and the UDP socket+selector closed on every exit. Blast-radius check came back clean too: the notification suppression in `FromRadioPacketHandlerImpl`/`UIViewModel` is double-gated on `isOtaUpdateActive` **and** `isOtaStatusNotification()`, and `isOtaUpdateActive` is set only in the ESP32 `performUpdate` and reset in `finally`, so key-verification and non-OTA alerts are untouched. LGTM.

Two LOW, non-blocking notes:

1. **UDP discovery trusts any format-matching beacon** (`WifiOtaDiscovery.kt`) — the redirected TCP target is the sender IP of the first datagram matching `Meshtastic_XXXX <ver>`, with no subnet check against the configured device IP. A LAN host spoofing that payload could win the race and redirect the firmware upload to itself (LAN-local DoS of the update + it receives the public firmware image — no key/PII exposure). Inherent to the esp32-unified-ota broadcast design, but a cheap guard would be requiring the discovered IP in the same subnet as the configured device IP.

2. **Success-path `transport.close()` isn't `NonCancellable`** — every *failed* close got the `NonCancellable` treatment, but the successful-transport close in the caller `finally` is left plain, so a cancel mid-upload can throw out of `close()` before the socket/GATT is released (leak-until-GC). Minor, and it's the one close path that didn't get the PR's own clean-teardown treatment.

(Also INFO: a fast UDP discovery hit skips the 8s Wi-Fi-readiness delay, so the first couple connects may hit a not-yet-listening port — self-corrects via the retry loop, only costs budget.)

PR #6082, opened 2026-07-05
fix(firmware): Keep bundled release discovery available

jeremiah-k:
## Overview

This PR fixes firmware release discovery when the app needs bundled release metadata.

F-Droid builds do not use the Meshtastic firmware release API, so bundled release metadata needs to remain available for the firmware update screen. The app also uses per-device databases, which means a bundled release snapshot seeded into one active database is not enough for later selected-device databases.

This change keeps bundled release discovery available for the current active database while still preventing unknown hardware from becoming flashable.

## Key Changes

* Cached the decoded bundled release snapshot for the current process.
* Re-evaluated bundled release seeding against the current active database.
* Seeded newly active or empty device databases from the bundled release snapshot.
* Preserved newer cached release data instead of regressing it with older bundled data.
* Kept partial bundled snapshots from deleting release types they do not contain.
* Marked bundled release data unavailable only when the bundled asset cannot be read or decoded.
* Applied bundled snapshots as a best-effort path so apply failures do not poison release discovery.
* Avoided holding the bundled decode mutex while waiting on the refresh/apply mutex.
* Preserved selected release state when hardware lookup fails.
* Cleared stale device metadata when hardware lookup fails.
* Kept the Ready state gated on resolved device hardware.
* Added focused coverage for active database switching, bundled seeding behavior, error handling, and selected-release preservation.

## Testing

* Added coverage proving an empty cache seeds from the bundled snapshot.
* Added coverage proving a newly active database is seeded after a previous database was already seeded.
* Added coverage proving newer bundled data replaces older cached release data.
* Added coverage proving older bundled data does not regress newer cached release data.
* Added coverage proving partial bundled snapshots leave missing release types untouched.
* Added coverage proving selected release state can populate when hardware lookup fails.
* Added coverage proving unknown hardware remains non-flashable.

## Migration Notes

* Unknown hardware remains blocked from flashing.
* This only changes firmware release discovery and update-screen state initialization.

## Summary by CodeRabbit

* **Bug Fixes**
  * Firmware releases now re-seed/refresh correctly when the active local database changes, rather than relying on a single one-time seed.
  * Bundled firmware data issues (missing/unreadable/invalid content) are treated as permanent, preventing repeated retry attempts.
  * Firmware update checks now clear stale device-related info when hardware lookup fails, ensuring the correct error state is shown.
  * Cache reads in stale-while-revalidate are more robust, avoiding unexpected failures when local data can’t be loaded.
* **Tests**
  * Added coverage for permanent bundled decode failure, reseeding on database switches, and stale-metadata clearing.

### DIFF
--- a/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
+++ b/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
@@ -172,22 +172,18 @@ class FirmwareUpdateViewModel(
                         enterRecoveryModeOrError()
                         return@launch
                     }
+                    val deviceHardware = getDeviceHardware(ourNode) ?: return@launch
+                    _deviceHardware.value = deviceHardware
+                    _currentFirmwareVersion.value = ourNode.firmwareVersion
+
                     val releaseFlow =
                         if (_selectedReleaseType.value == FirmwareReleaseType.LOCAL) {
                             flowOf(null)
                         } else {
                             firmwareReleaseRepository.getReleaseFlow(_selectedReleaseType.value)
                         }
-
-                    val deviceHardware = getDeviceHardware(ourNode)
-                    if (deviceHardware != null) {
-                        _deviceHardware.value = deviceHardware
-                        _currentFirmwareVersion.value = ourNode.firmwareVersion
-                    }
-
                     releaseFlow.collectLatest { release ->
                         _selectedRelease.value = release
-                        if (deviceHardware == null) return@collectLatest
 
                         val dismissed = bootloaderWarningDataSource.isDismissed(address)
                         val firmwareUpdateMethod =
@@ -247,12 +243,14 @@ class FirmwareUpdateViewModel(
     private suspend fun enterRecoveryModeOrError() {
         val recovery = firmwareRecoveryDataSource.pending.first()
         if (recovery == null) {
+            clearDeviceMetadata()
             _state.value = FirmwareUpdateState.Error(UiText.Resource(Res.string.firmware_update_no_device))
             return
         }
         pendingRecovery = recovery
         val hardware =
             deviceHardwareRepository.getDeviceHardwareByModel(recovery.hwModel, recovery.pioEnv).getOrElse {
+                clearDeviceMetadata()
                 _state.value =
                     FirmwareUpdateState.Error(
                         UiText.Resource(Res.string.firmware_update_unknown_hardware, recovery.hwModel),
@@ -587,6 +585,7 @@ class FirmwareUpdateViewModel(
     }
 
     private fun clearDeviceMetadata() {
+        _selectedRelease.value = null
         _deviceHardware.value = null
         _currentFirmwareVersion.value = null
     }
--- a/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
+++ b/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
@@ -179,11 +179,15 @@ class FirmwareUpdateViewModel(
                             firmwareReleaseRepository.getReleaseFlow(_selectedReleaseType.value)
                         }
 
-                    releaseFlow.collectLatest { release ->
-                        _selectedRelease.value = release
-                        val deviceHardware = getDeviceHardware(ourNode) ?: return@collectLatest
+                    val deviceHardware = getDeviceHardware(ourNode)
+                    if (deviceHardware != null) {
                         _deviceHardware.value = deviceHardware
                         _currentFirmwareVersion.value = ourNode.firmwareVersion
+                    }
+
+                    releaseFlow.collectLatest { release ->
+                        _selectedRelease.value = release
+                        if (deviceHardware == null) return@collectLatest
 
                         val dismissed = bootloaderWarningDataSource.isDismissed(address)
                         val firmwareUpdateMethod =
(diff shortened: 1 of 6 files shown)


feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -170,57 +170,60 @@ class FirmwareUpdateViewModel(
...
-                            val firmwareUpdateMethod =
-                                when {
-                                    radioPrefs.isSerial() -> {
-                                        // Serial OTA is not yet supported for ESP32 — only nRF52/RP2040 UF2.
-                                        if (deviceHardware.isEsp32Arc) {
-                                            FirmwareUpdateMethod.Unknown
-                                        } else {
-                                            FirmwareUpdateMethod.Usb
-                                        }
+                    releaseFlow.collectLatest { release ->
```
jamesarich, line 185: The description says release collection now starts before hardware resolution, but `releaseFlow` is a cold flow still collected here — after `getDeviceHardware()` returns at line 180, in the same coroutine. During a slow api.meshtastic.org hardware lookup the release state is still not populated any earlier. The real fix is the null-guard restructure below; the "collect before resolution" framing (and the construction reorder) is inert — worth correcting the PR body so it doesn't imply a latency win that isn't there.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -569,15 +572,22 @@ class FirmwareUpdateViewModel(
 
         return if (hwModelInt != null) {
             deviceHardwareRepository.getDeviceHardwareByModel(hwModelInt, target).getOrElse {
+                clearDeviceMetadata()
```
jamesarich, line 575: `clearDeviceMetadata()` was added only to the `getDeviceHardware` failure paths. The sibling recovery path (`enterRecoveryModeOrError`) still sets `Error` without clearing `_deviceHardware`/`_currentFirmwareVersion`, so the stale-metadata-alongside-Error state this PR fixes on the connected path survives there — and `FirmwareUpdateScreen` still renders the stale `DeviceInfoCard`. Same clear is needed on that branch.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -170,57 +170,60 @@ class FirmwareUpdateViewModel(
...
-                                    radioPrefs.isSerial() -> {
-                                        // Serial OTA is not yet supported for ESP32 — only nRF52/RP2040 UF2.
-                                        if (deviceHardware.isEsp32Arc) {
-                                            FirmwareUpdateMethod.Unknown
-                                        } else {
-                                            FirmwareUpdateMethod.Usb
-                                        }
+                    releaseFlow.collectLatest { release ->
+                        _selectedRelease.value = release
+                        if (deviceHardware == null) return@collectLatest
```
jamesarich: With `deviceHardware == null` the collector still runs before this guard, so `_selectedRelease` is populated while `_state` is `Error`. More importantly, `loadFromCache` in `staleWhileRevalidateFlow` isn't wrapped in `safeCatching` (only `fetch` is), so a throwing Room read here propagates into the outer `.onFailure` and overwrites the specific `firmware_update_unknown_hardware` error with a generic exception message. Guarding the cache read (or moving the null-guard above the collect) would keep the actionable error.

jamesarich (commented):
Re-reviewed after the force-push. The re-seed fix is solid — bundled releases now re-seed when the active device DB switches (good test), and the `bundleDecodeFailed` latch still prevents per-collection retry. Three carryovers inline; none are hard blockers, but the first two are worth tightening before merge.

jamesarich (commented):
Re-reviewed at 4f9ccd62 — two of the three are fixed:

- ✅ **Recovery path clears stale metadata** (`clearDeviceMetadata()` in both branches of `enterRecoveryModeOrError`), with `recovery hardware lookup failure clears stale device metadata`.
- ✅ **Specific unknown-hardware Error no longer clobbered** — on hardware-lookup failure the connected path now `return@launch`es before collecting the release flow, so a throwing cache read can't overwrite it (the new test stubs the cache to throw and asserts the specific error survives).

Two follow-ups:

1. **PR body now contradicts the code.** The `harden release discovery errors` commit *stops* collecting the release flow when hardware resolution fails (`getDeviceHardware(ourNode) ?: return@launch`) — the opposite of the Key Changes line "Ensured release state is still collected when hardware lookup fails." And neither head ever started collection *before* hardware resolution (it's a cold flow collected in the same coroutine). Please update the description to match what the code does.
2. **Same clobber survives on the recovery Ready path.** The fix short-circuited only the connected hardware-failure path. After a *successful* recovery lookup, `getReleaseFlow(type).collectLatest` still runs inside the outer `safeCatching`, and `loadFromCache()` in `StaleWhileRevalidateFlow` is still unguarded (only `fetch` is), so a throwing cache read there overwrites the recovery Ready state with a generic `DynamicString(e.message)`. Guarding `loadFromCache` at the source would cover both paths.

jeremiah-k:
@jamesarich I pushed a refreshed branch with the release-state cleanup tightened.

The description no longer claims release collection starts before hardware resolution. The code now keeps bundled seeding DB-aware, clears stale metadata on hardware lookup failures, and guards cache reads in the stale-while-revalidate flow so a transient cache/DB read failure does not overwrite the more specific firmware update state.

jamesarich (approved):
Traced the core logic (no deadlock / correct temp-file lifecycle & validation respectively) — LGTM. Details in review pass.

PR #6083, opened 2026-07-05
feat(firmware): Confirm local firmware files before flashing

jeremiah-k:
## Overview

This PR makes local firmware file selection safer.

Previously, selecting a file from the Android picker immediately started the update. On devices where long filenames are truncated, it was easy to select the wrong release bundle or firmware artifact without seeing the full name first.

This change resolves and validates the selected file, shows the full filename, and requires explicit confirmation before flashing begins.

## Key Changes

* Resolved the selected file display name.
* Added pending local firmware selection state.
* Required explicit confirmation before starting a local-file update.
* Showed the full filename, device name, and platformio target before flashing.
* Made the filename selectable for easier inspection.
* Validated local firmware filenames against the current device and update method.
* Revalidated pending selections immediately before starting the update.
* Rejected stale selections when target, update method, or address changed before confirmation.
* Supported selecting official release bundles and extracting the matching update payload.
* Preferred ESP32 `-update.bin` payloads when extracting official release bundles.
* Revalidated extracted payloads before confirmation.
* Preserved exclusions for non-app artifacts.
* Cleaned up temporary extracted files on dismiss, error, state change, cancellation, and ViewModel clear.
* Preserved USB file-save state during local and downloaded update flows.
* Added focused validation and ViewModel coverage.

## Testing

* Added coverage for local firmware filename validation.
* Added coverage for pending-file confirmation.
* Added coverage for stale target, update method, and address rejection.
* Added coverage for release bundle extraction.
* Added coverage for temporary artifact cleanup.
* Added coverage for preserving USB file-save state.
* Added coverage for cancelled pending-selection behavior.

## Migration Notes

* Users can continue selecting official release bundles.
* This only adds validation and confirmation before local firmware flashing begins.

## Summary by CodeRabbit

* **New Features**
  * Added a two-step local firmware update flow with a confirmation dialog before flashing.
  * Improved file selection by showing a readable filename and additional device/update details.
  * Added support for clearer guidance on required firmware file types and update methods.

* **Bug Fixes**
  * Better handles missing or unreadable firmware filenames.
  * Improves validation so local firmware files are checked against the selected device and update method.
  * Clears pending selections when the update context changes or is canceled.


feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -444,10 +466,188 @@ class FirmwareUpdateViewModel(
...
+        fileName: String,
+        state: FirmwareUpdateState.Ready,
+        fallbackReason: LocalFirmwareFileValidationReason,
+    ): LocalFirmwareResolution {
+        val payloadExtension = localFirmwarePayloadExtension(state.deviceHardware, state.updateMethod)
+        return if (payloadExtension == null) {
+            LocalFirmwareResolution.Invalid(reason = fallbackReason, fileName = fileName)
+        } else {
+            val extractedArtifact =
+                safeCatching { fileHandler.extractFirmware(uri, state.deviceHardware, payloadExtension) }
```
jamesarich: This extracts with `payloadExtension` (".bin") and no `preferredFilename`, so `AndroidFirmwareFileHandler` falls back to `minByOrNull { it.first.name.length }` and picks the shorter plain `firmware-<target>-<ver>.bin`. On pre-2.7.17 bundles that's the **merged** bootloader+partition+app image `FirmwareRetriever`'s own comment says `esp_ota_end` rejects — it deliberately prefers the manifest then `-update.bin`. Reuse that preference chain (pass `preferredFilename`) so a side-loaded official bundle doesn't extract an un-flashable image that then passes validation and gets confirmed.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareFileHandler.kt
```diff
@@ -126,3 +129,143 @@ internal fun isValidFirmwareFile(filename: String, target: String, fileExtension
...
+    missingSuffixReason: LocalFirmwareFileValidationReason,
+): LocalFirmwareFileValidation {
+    if (!fileName.endsWith(requiredSuffix)) {
+        return LocalFirmwareFileValidation.Invalid(missingSuffixReason)
+    }
+    return validateTargetMatch(fileName, target, fileExtension)
+}
+
+private fun validateTargetMatch(fileName: String, target: String, fileExtension: String): LocalFirmwareFileValidation =
+    if (isValidFirmwareFile(fileName, target, fileExtension)) {
```
jamesarich, line 301: The target match here delegates to `isValidFirmwareFile`, whose regex `.*[-_]<target>[-_.].*` accepts delimiter-bounded **superstrings**: a `tbeam` device validates `firmware-tbeam-s3-core-*.bin` as Valid (real colliding pairs in `device_hardware.json`: tbeam/tbeam-s3-core, t-echo/t-echo-lite, t-deck/t-deck-pro, nano-g1/nano-g1-explorer). Since this feature exists to block wrong-board files, the match needs to require the target be immediately followed by a version/extension delimiter (i.e. anchor the tail), not just appear delimiter-bounded.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareUpdateViewModel.kt
```diff
@@ -444,10 +466,188 @@ class FirmwareUpdateViewModel(
...
+            is LocalFirmwareFileValidation.Invalid ->
+                if (shouldTryLocalFirmwareBundle(fileName, validation.reason)) {
+                    resolveLocalFirmwareBundle(uri, fileName, state, validation.reason)
+                } else {
+                    LocalFirmwareResolution.Invalid(reason = validation.reason, fileName = fileName)
+                }
+        }
+    }
+
+    private suspend fun resolveLocalFirmwareBundle(
```
jamesarich, line 564: Bundle resolution here runs `getDisplayName` + a full multi-MB zip extraction with no user-visible progress — the old path set `Processing(firmware_update_extracting)` but it was removed and the string isn't even imported. The screen sits at `Ready` during extraction; a user seeing nothing happen may re-tap and spawn a second prepare. Set an extracting/Processing state around the resolve.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/FirmwareFileHandler.kt
```diff
@@ -126,3 +129,143 @@ internal fun isValidFirmwareFile(filename: String, target: String, fileExtension
...
+): LocalFirmwareFileValidation {
+    val currentTarget = currentState.deviceHardware.effectiveTarget
+    return when {
+        currentTarget != pendingFile.platformioTarget ->
+            LocalFirmwareFileValidation.Invalid(LocalFirmwareFileValidationReason.TargetMismatch)
+
+        // The user selected the file under a different connection (method or address) than the
+        // one active now — the file may still match the target, but the confirmation context is stale.
+        currentState.updateMethod != pendingFile.updateMethod || currentState.address != pendingFile.address ->
+            LocalFirmwareFileValidation.Invalid(LocalFirmwareFileValidationReason.ConfirmationContextChanged)
```
jamesarich, line 158: The context-changed fix only covers same-target/different-transport. `TargetMismatch` is still checked before this branch, so switching to a *different device* mid-dialog (new address **and** target) shows the scary "file does not match <device>" error instead of the intended "connection changed, reselect". Check `ConfirmationContextChanged` (address/method delta) before the target match.

jamesarich (changes requested):
Re-reviewed after the force-push. The confirmation-context ordering fix landed (context change now surfaces "reselect" instead of an extension error), and most temp-file orphan paths are cleaned up — thanks. But two confirmed holes defeat the feature's core purpose (flashing the *right* firmware), so I'm requesting changes. Details inline. Minor: one new string (`firmware_update_invalid_local_file_detail`) has unescaped `"%1$s"` quotes while its sibling escapes them — `sort-strings.py`/lint may flag it.

jeremiah-k:
@jamesarich I pushed a refreshed branch with the local-file edge cases tightened.

Official ESP32 release bundles now prefer the `-update.bin` payload, target matching rejects prefix collisions like `tbeam` vs `tbeam-s3-core`, extraction shows a processing state, stale connection context is checked before target mismatch, and older local filename shapes like `target-firmware-version.bin` still validate.

jamesarich (approved):
Traced the core logic (no deadlock / correct temp-file lifecycle & validation respectively) — LGTM. Details in review pass.

PR #6093, opened 2026-07-05
fix: harden against adversarial mesh-fuzz findings (crash, GC-thrash, unbounded growth, OOM)

jamesarich:
Found and fixed 4 real bugs by adversarially testing the app against a live replay-fuzz harness (via `meshtastic-mcp`'s replay engine) on a physical Pixel 6a — corrupted/malformed packets, connection churn, and large-mesh scale, escalating through `light` → `parser` → `adversary` → `chaos` fuzz presets. All four are reproduced live and each has a fix + regression test (or a live before/after re-run where a unit test wasn't practical).

## 🐛 Bug Fixes

- **Crash on a malformed buffered early packet.** `flushEarlyReceivedPackets()` replayed buffered packets in a bare `scope.launch`, unlike the live receive path which is guarded by `safeCatching`. A single corrupted NodeInfo/User packet arriving before the node DB / local node number resolve threw an uncaught `ProtocolException`/`EOFException` and crashed the whole process. Reproduced twice with a seeded `chaos` fuzz run. Fixed by wrapping each buffered packet the same way the live path already is.
- **GC-thrash risk in the Debug Panel.** `DebugViewModel.meshLog` re-decoded up to 5000 log rows — including proto `toString()` and a full node-DB snapshot **per row** — on every single packet insert, via Room's auto-invalidating `Flow` + `mapLatest`. Under a packet flood this can restart faster than it completes, generating a firehose of garbage (matches a prior live incident where a 70-minute soak pinned the Dalvik heap at 98.7%). Fixed with a 200ms debounce and by hoisting the per-row node-list snapshot out of the loop.
- **Unbounded FileInfo accumulator.** `RadioConfigRepositoryImpl.addFileInfo()` did an unconditional `+= info` with no cap; `clearFileManifest()` only runs on the *next* handshake, so a rogue peer emitting FileInfo packets in a loop grows it without limit for the life of a session. Capped at 4096 entries, same defensive pattern as the existing early-packet buffer.
- **OutOfMemoryError under combined large-mesh + adversarial load.** Reproduced live: a 2000-node NodeDB (the app's own largest `push_fake_nodedb` fixture — large public-event meshes like MeshCon/DEF CON/Burning Man are a *supported* scenario) + sustained `chaos` fuzz + rapid reconnects in one process exhausted the default ~256MB heap. Added `android:largeHeap="true"`; re-ran the identical crash-inducing sequence post-fix and it survived cleanly at the new 512MB ceiling.

## 🧹 Chores

- Added LeakCanary (stable 2.14, debug-only) for continuous leak detection on future soak/fuzz runs.

## Testing Performed

- New/updated unit tests: `MeshMessageProcessorImplTest` (regression test verified to fail without the fix, pass with it), `RadioConfigRepositoryImplTest` (new file, cap + handshake-clear behavior).
- `./gradlew spotlessCheck detekt` clean across the whole repo.
- `./gradlew :core:data:jvmTest :feature:settings:jvmTest` pass.
- Live device re-verification on a Pixel 6a for all four fixes: re-ran the exact seeded fuzz campaigns that originally triggered each bug and confirmed stability (no crash, flat memory, clean reconnect) post-fix.

## Summary by CodeRabbit

* **New Features**
  * Added debug-only leak detection tooling for development and testing.
  * Enabled Android large-heap mode to reduce out-of-memory risk during heavy stress.

* **Bug Fixes**
  * Prevented errors during replay of buffered “early” mesh packets from stopping processing of the rest.
  * Capped session file manifest growth and safely drops excess entries.
  * Improved the mesh log debug screen responsiveness by debouncing and reducing repeated annotation work.

* **Tests**
  * Added regression and repository tests covering early-packet flush resilience and file manifest capping/concurrency/clearing.

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -289,6 +289,12 @@ dependencies {
 
     debugImplementation(libs.androidx.compose.ui.test.manifest)
     debugImplementation(libs.androidx.glance.preview)
+    // LeakCanary auto-installs via a manifest-declared ContentProvider -- no init code needed.
+    // Debug-only (0 method count in release, per LeakCanary's own design). Added after adversarial
+    // mesh-fuzz testing surfaced a GC-thrash concern in the debug log view; this gives continuous,
+    // automatic leak detection (dumped heap + leak trace + notification) for future soak/fuzz runs
+    // instead of manually polling `dumpsys meminfo`.
+    debugImplementation(libs.leakcanary.android)
 
     googleImplementation(projects.feature.car)
     googleImplementation(libs.location.services)
--- a/androidApp/src/main/AndroidManifest.xml
+++ b/androidApp/src/main/AndroidManifest.xml
@@ -131,8 +131,17 @@
           full. largeHeap raises the ceiling (device dependent, commonly 2 to 4x) rather than reducing
           footprint; it buys headroom for legitimately large meshes while the real fix (bounding/streaming
           the in-memory log + node caches; see DebugViewModel and MeshMessageProcessorImpl) remains the
-          long-term mitigation. Android's own docs caution against relying on this alone; don't remove
-          without re-validating headroom under a similar soak.
+          long-term mitigation.
+
+          TREAT THIS AS A TEMPORARY STOPGAP, not a permanent setting. It is a global,
+          whole-app flag with a global cost: a larger heap increases the working set the GC scans
+          each cycle, which can raise GC pause times / jank for ALL sessions -- including typical
+          small-mesh use on low-RAM devices, which never come near the OOM this guards against. That
+          tradeoff (a tail-scenario safety net paid for by every user) is why Android's own docs
+          caution against relying on largeHeap alone. Remove it once the cache-bounding/streaming work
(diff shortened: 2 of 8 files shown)


jamesarich:
Both addressed in 23f066b:

**Race in `addFileInfo` (actionable):** Good catch — `handleFileInfo` dispatches each packet on its own `scope.handledLaunch`, so the read-then-write genuinely races. Switched to `MutableStateFlow.update {}` (atomic CAS loop) so the read/cap/append is atomic. Added a concurrency regression test that fires 2000 concurrent `addFileInfo` calls on `Dispatchers.Default` and asserts none are lost — verified it **fails** against the old read-then-write and **passes** with `update {}`.

**`largeHeap` stopgap (nitpick):** Agreed on the global GC-jank tradeoff. Expanded the comment to spell out that it's a temporary, whole-app flag with a global cost (larger GC working set → more pause/jank for *all* sessions, including small-mesh use on low-RAM devices that never approach the OOM it guards), and that removal is a tracked follow-up once the cache-bounding work lands — to be validated on a low-RAM device, not just the flagship soak target.

PR #6104, opened 2026-07-05
Add on-device translation for chat messages (google flavor)

### LINKED ISSUE #6091: Translate chat messages in-place (ML Kit, Google flavor only)
## What

Add a "Translate" / "Show Original" action to chat messages, matching the iOS reference implementation described in [meshtastic/design#100](https://github.com/meshtastic/design/issues/100). Original request: [#83](https://github.com/meshtastic/design/issues/83).

**Explicitly do not ship this on F-Droid.** Use Android's on-device ML Kit translation model on the `google` flavor only, with F-Droid falling back to a no-op (row hidden or reports unavailable) — the same pattern already used for AI features in this codebase. This is not a new architectural decision; it directly reuses infrastructure that already exists for a different feature.

## Reuse the existing docs-translation architecture — don't build a new pipeline

`feature/docs/.../translation/` already has exactly this shape, built for translating documentation pages:

- `DocTranslationService` — interface: `translatePage()`, `isLanguageAvailable()`, `downloadLanguageModel()`, returning `TranslationResult` (`Success` / `ModelDownloadRequired(locale, estimatedSizeMb)` / `Unavailable`).
- `androidApp/src/google/.../translation/MlKitDocTranslator.kt` — real implementation using `com.google.mlkit:translate` (already a dependency, `gradle/libs.versions.toml:68,204`). Auto-downloads the on-device model (~30MB/language) via `RemoteModelManager`, caches results, translates via `Translation.getClient(...)`.
- `androidApp/src/fdroid/.../di/FdroidAiModule.kt` — binds the same interface to `NoOpDocTranslator`, which always returns `TranslationResult.Unavailable`. This is how ML Kit (a proprietary, Play-Services-backed library) is already kept out of the F-Droid build today.

For chat messages:

1. Add a `MessageTranslationService` interface (or generalize `DocTranslationService` if it's a clean fit — it's already page/content agnostic aside from the `pageId` cache key) with an analogous `translate(text, targetLocale): TranslationResult` + `isLanguageAvailable()` + `downloadLanguageModel()`.
2. Implement it on the `google` flavor using ML Kit directly on message text (no markdown segmentation needed — `MlKitDocTranslator`'s `MarkdownTranslationSegmenter` step doesn't apply to plain chat text, everything else does).
3. Bind a no-op equivalent on the `fdroid` flavor (mirror `FdroidAiModule.kt`), so the feature is simply unavailable there — do not add the ML Kit dependency to any F-Droid source set.
4. Target locale should be the device's current locale (Android's `Locale.getDefault()` — consistent with how [meshtastic/design#99](https://github.com/meshtastic/design/issues/99) resolves locale elsewhere in the app), not a manual language picker, unless a later spec revision decides otherwise.

## UI

Add "Translate" / "Show Original" rows to `feature/messaging/.../component/MessageActionsBottomSheet.kt` (`MessageActionsContent`), alongside the existing Reply/Copy/Select/Delete rows — this is Android's equivalent of iOS's long-press context menu entry point, so no new interaction pattern is needed.

- Show "Translate" when no cached translation exists for the message.
- Show "Show Original" / "Show Translation" toggle once a translation is cached (mirror iOS's persisted toggle — don't discard the translation after viewing it once).
- If translation is unavailable (F-Droid, or `TranslationResult.Unavailable`), hide the row entirely rather than showing a disabled/error state.
- If a model download is required (`TranslationResult.ModelDownloadRequired`), show a brief one-time download prompt (~30MB) before translating, same as the docs feature already does.

## Persistence

`core/model/.../Message.kt` has no translated-text field today. Add one (and the corresponding DB column/migration) so a translation persists and can be toggled back to the original — mirror iOS's `saveTranslatedText` on `MessageEntity`, and reuse the docs feature's content-hash cache-key pattern (`DocTranslationCache`) if a similar on-disk/DB cache makes sense here, rather than re-translating on every view.

## Reference

Cross-platform alignment tracking issue: meshtastic/design#100. Original feature request: meshtastic/design#83.

jamesarich:
## Why
Users chatting across languages currently have no way to read a mesh message in their own language. `:feature:docs` already solved this exact problem for documentation pages via ML Kit (on-device, no cloud round-trip); this reuses that architecture for chat text so foreign-language messages can be translated in place, per meshtastic/design#100 (originally requested in meshtastic/design#83).

## 🌟 What's new
- `MessageTranslationService` (mirrors `DocTranslationService`): `translate()`, `isLanguageAvailable()`, `downloadLanguageModel()` → `TranslationResult` (`Success` / `ModelDownloadRequired` / `Unavailable`).
- Google flavor: `MlKitMessageTranslator` translates message text directly with ML Kit (no markdown segmenter needed) into the device's current locale, gated behind an explicit ~30MB model-download prompt (unlike the docs feature, which auto-downloads silently — chat messages are more frequent/ambient, so we ask first).
- Fdroid/desktop: `NoOpMessageTranslator`, mirroring `NoOpDocTranslator` — no ML Kit dependency added outside the google flavor.
- "Translate" / "Show Original" / "Show Translation" row added to the message actions bottom sheet, hidden entirely when translation is unavailable.
- Translated text persists on the message (new nullable `translated_text` column, Room DB v44→v45 auto-migration) so re-opening a conversation doesn't require re-translating; "Show Original" is a local display toggle.

## Testing Performed
- `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` — all green.
- Not yet manually verified on-device (no emulator pass in this session) — worth a manual check of the download-prompt flow and toggle behavior before merge.

Closes #6091.

### DIFF
--- a/.skills/compose-ui/strings-index.txt
+++ b/.skills/compose-ui/strings-index.txt
@@ -21,6 +21,9 @@ action_select_message
 action_select_network
 action_send_reply
 action_show_message_status
+action_show_original_message
+action_show_translated_message
+action_translate_message
 actions
 adc_multiplier_override
 adc_multiplier_override_ratio
@@ -578,6 +581,7 @@ firmware
 firmware_edition
 firmware_old
 firmware_recovery_banner
+firmware_recovery_ble_failed
 firmware_recovery_button
 firmware_recovery_dismiss
 firmware_recovery_explanation
@@ -1390,8 +1394,10 @@ should_update_firmware
 show_all_key_title
 show_iaq_legend
 show_layer
+show_original
 show_password
 show_precision_circle
+show_translation
 show_waypoints
 shutdown
 shutdown_node_name
@@ -1518,6 +1524,9 @@ traffic_management_rate_limit_max_packets
 traffic_management_rate_limit_window
 traffic_management_router_preserve_hops
 traffic_management_unknown_packet_threshold
+translate
+translate_download_model_body
+translate_download_model_title
 transmit_over_lora
 ### TRANSPORT ###
 transport
--- a/androidApp/src/fdroid/kotlin/org/meshtastic/app/di/FdroidAiModule.kt
+++ b/androidApp/src/fdroid/kotlin/org/meshtastic/app/di/FdroidAiModule.kt
@@ -24,6 +24,8 @@ import org.meshtastic.feature.docs.ai.AIDocAssistant
 import org.meshtastic.feature.docs.ai.KeywordFallbackAssistant
 import org.meshtastic.feature.docs.translation.DocTranslationService
 import org.meshtastic.feature.docs.translation.NoOpDocTranslator
+import org.meshtastic.feature.messaging.translation.MessageTranslationService
+import org.meshtastic.feature.messaging.translation.NoOpMessageTranslator
 
 /** Provides keyword-only fallback AI assistant for the F-Droid flavor (no on-device model). */
 @Module
@@ -33,4 +35,6 @@ class FdroidAiModule {
     @Single fun discoverySummaryAiProvider(fallback: AlgorithmicSummaryProvider): DiscoverySummaryAiProvider = fallback
 
     @Single fun docTranslationService(): DocTranslationService = NoOpDocTranslator()
+
+    @Single fun messageTranslationService(): MessageTranslationService = NoOpMessageTranslator()
 }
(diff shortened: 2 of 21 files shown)


jamesarich:
Closing in favor of #6103, which implements the same feature (translate chat message via ML Kit, closes #6091, same DocTranslationService-derived architecture) as a superset: persisted original/translated toggle (`show_translated` column, iOS parity), model-download progress UI, "already in your language" snackbar, search/TalkBack/selection-copy wiring, legacy locale normalization, and a full test suite. Nothing here to port that #6103 doesn't already cover.

PR #6126, opened 2026-07-07
fix(runtime): harden BLE profile, OTA setup, and DB access

jeremiah-k:
## Overview

This PR hardens runtime behavior seen during ESP32 BLE OTA testing, shared BLE profile setup, and device/database churn.

The BLE OTA path could connect to GATT and begin OTA setup before the expected ESP32 OTA service was fully usable. If service discovery lagged, required OTA characteristics were missing, notification observation failed, or reconnect attempts reused stale collectors/channels, the update flow could wait too long or poison state used by later attempts.

The BLE profile setup change is intentionally shared at the `core/ble` layer because profile access is used beyond OTA, including normal BLE radio and firmware-update paths. The goal is to make service-discovery and disconnect handling consistent for callers that enter a BLE profile.

The logs also showed Room/SQLite connection-pool pressure during device/database switching. This PR adds a short-term DatabaseManager containment step that routes `withDb` work through a single-lane dispatcher and allows DB-critical one-shot work to finish before cancellation is rethrown.

## Key Changes

* Wait for BLE service discovery before running profile setup.
* Run BLE profile setup on the IO dispatcher.
* Cancel profile setup promptly if the BLE connection scope closes.
* Surface connection loss during BLE profile setup as `NotConnectedException`.
* Validate ESP32 OTA notify/write characteristics before starting notification observation.
* Report exactly which OTA characteristic is missing when validation fails.
* Reset the BLE OTA response channel per connection attempt so a failed observation does not poison retries.
* Cancel stale BLE OTA notification collectors before reconnecting.
* Track and cancel the BLE OTA connection-state collector across reconnects.
* Reuse the discovered OTA service for firmware writes instead of repeatedly entering shared BLE profile setup per chunk.
* Fail fast if OTA notification observation fails before subscription readiness.
* Avoid logging raw BLE addresses in the BLE OTA connection/setup scan path.
* Preserve cancellation when BLE scan/receive timeouts are converted to OTA errors.
* Keep a bounded OTA notification-subscription readiness fallback.
* Route `DatabaseManager.withDb` work through a single-lane containment path to narrow Room/SQLite connection-pool churn during high-churn flows.
* Preserve DB-critical one-shot sections through cancellation and re-check cancellation afterward.
* Keep cancellation out of generic DB retry handling.
* Document that `withDb` containment is intended for bounded one-shot DB work, not long-lived Flow/Paging reads.

## Testing

* Added/updated BLE OTA coverage for missing OTA characteristics.
* Added coverage for notification observation failing before subscription readiness.
* Added coverage that OTA firmware writes reuse the discovered OTA service instead of re-entering profile setup per chunk.
* Reviewed F-Droid debug logs from BLE connection/handshake and OTA setup runs:
  * no `FATAL EXCEPTION` observed,
  * no SQLite Error code 5 / database-lock markers observed,
  * no `NoSuchElementException` observed,
  * BLE radio startup reached `MyNodeInfo` and local metadata,
  * BLE OTA setup path exercised the redacted connection/setup logging.

## Scope Note

The database change here is intentionally narrow. It is a short-term runtime containment measure for connection-pool churn seen during firmware update and device/database switching flows, not a new long-term database concurrency model.

The `withDb` containment path narrows the high-churn entry window for bounded one-shot DB work. It does not claim to globally serialize every Room executor operation, and long-lived Flow/Paging reads should continue to stay out of this path.

This PR hardens the runtime setup and failure paths that were blocking reliable OTA work. It does not claim to complete the broader OTA firmware-transfer work; that remains scoped to the follow-up OTA PR stack.

Broader database cleanup is intentionally left out of this PR. Areas to revisit separately include direct `currentDb.value` access, high-churn one-shot DAO calls, long-lived Flow/Paging reads, and where safe DB parallelism can be restored.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved BLE connection reliability during setup and reconnection.
  * Reduced the chance of stalled or cancelled database work during active state changes.
  * Made firmware upload handling more robust, including safer retries, timeouts, and response handling.
  * Improved BLE scan timeout handling so cancellations are no longer mistaken for timeouts.
* **Documentation**
  * Clarified BLE profile setup behavior in the developer-facing method guidance.

### DIFF
--- a/core/ble/src/commonMain/kotlin/org/meshtastic/core/ble/KableBleConnection.kt
+++ b/core/ble/src/commonMain/kotlin/org/meshtastic/core/ble/KableBleConnection.kt
@@ -29,15 +29,20 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.TimeoutCancellationException
+import kotlinx.coroutines.async
+import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.delay
+import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
+import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.job
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
+import org.meshtastic.core.common.util.ioDispatcher
 import kotlin.concurrent.Volatile
 import kotlin.time.Duration
 import kotlin.time.Duration.Companion.seconds
@@ -245,6 +250,7 @@ class KableBleConnection(private val scope: CoroutineScope, private val loggingC
         _deviceFlow.emit(null)
     }
 
+    @Suppress("ThrowsCount")
     override suspend fun <T> profile(
         serviceUuid: Uuid,
         timeout: Duration,
@@ -253,7 +259,32 @@ class KableBleConnection(private val scope: CoroutineScope, private val loggingC
         val p = peripheral ?: error("Not connected")
         val cScope = connectionScope ?: error("No active connection scope")
         val service = KableBleService(p, serviceUuid)
-        return withTimeout(timeout) { cScope.setup(service) }
+        return withTimeout(timeout) {
+            withContext(ioDispatcher) {
+                val profileExecution = async {
+                    p.services.first { it != null }
+                    cScope.setup(service)
+                }
+
+                val disconnectHandle =
+                    cScope.coroutineContext.job.invokeOnCompletion {
+                        profileExecution.cancel(CancellationException("Connection lost during BLE profile execution"))
+                    }
+
+                try {
+                    profileExecution.await()
+                } catch (e: CancellationException) {
+                    currentCoroutineContext().ensureActive()
+                    if (!cScope.coroutineContext.job.isActive) {
+                        throw NotConnectedException("Connection lost during BLE profile execution")
+                    }
+                    throw e
+                } finally {
+                    disconnectHandle.dispose()
+                    profileExecution.cancel()
+                }
+            }
+        }
     }
 
     override fun maximumWriteValueLength(writeType: BleWriteType): Int? = peripheral?.negotiatedMaxWriteLength()
--- a/core/ble/src/commonMain/kotlin/org/meshtastic/core/ble/KableBleConnection.kt
+++ b/core/ble/src/commonMain/kotlin/org/meshtastic/core/ble/KableBleConnection.kt
@@ -260,6 +260,8 @@ class KableBleConnection(private val scope: CoroutineScope, private val loggingC
         val cScope = connectionScope ?: error("No active connection scope")
         val service = KableBleService(p, serviceUuid)
         return withTimeout(timeout) {
+            // Shared BLE profile guard: wait for Kable service discovery before handing out the service, and map a
+            // connection-scope shutdown during caller setup to NotConnectedException instead of waiting for timeout.
             withContext(ioDispatcher) {
                 val profileExecution = async {
                     p.services.first { it != null }
--- a/core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseManager.kt
+++ b/core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseManager.kt
@@ -281,10 +281,12 @@ open class DatabaseManager(
         Logger.d { "Closed inactive database ${anonymizeDbName(dbName)} to free connections" }
     }
 
-    // Short-term runtime containment: serialize withDb calls to reduce Room/SQLite connection-pool churn seen during
-    // device/firmware update flows. Preserve DB-critical blocks through cancellation, then re-check cancellation so
-    // stale callers do not continue after the DB releases. Revisit after direct currentDb.value callers are audited and
-    // safe DB concurrency can be restored.
+    // Short-term runtime containment: route withDb entry through a single-lane dispatcher to narrow the Room/SQLite
+    // connection-pool churn window seen during device/firmware update flows. Room suspend DAOs may continue on Room's
+    // own executor after suspension, so this is not a strict global DB-I/O serialization guarantee. Preserve bounded
+    // one-shot DB-critical blocks through cancellation, then re-check cancellation so stale callers do not continue
+    // after the DB releases. Long-lived Flow/Paging reads must stay out of withDb; revisit after direct currentDb.value
+    // callers are audited and safe DB concurrency can be restored.
     private val limitedIo = dispatchers.io.limitedParallelism(1)
 
     /** Execute [block] with the current DB instance. Retries once if the pool closes during a DB switch. */
@@ -294,7 +296,7 @@ open class DatabaseManager(
         return withContext(limitedIo) {
             val queuedMillis = nowMillis - queuedAt
             if (queuedMillis >= WITH_DB_SLOW_OPERATION_MS) {
-                Logger.w { "withDb waited ${queuedMillis}ms for serialized DB access" }
+                Logger.w { "withDb waited ${queuedMillis}ms for the temporary DB containment lane" }
             }
 
             val startedAt = nowMillis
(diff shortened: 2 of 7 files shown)


core/ble/src/commonMain/kotlin/org/meshtastic/core/ble/KableBleConnection.kt
```diff
@@ -253,7 +259,32 @@ class KableBleConnection(private val scope: CoroutineScope, private val loggingC
         val p = peripheral ?: error("Not connected")
         val cScope = connectionScope ?: error("No active connection scope")
         val service = KableBleService(p, serviceUuid)
-        return withTimeout(timeout) { cScope.setup(service) }
+        return withTimeout(timeout) {
+            withContext(ioDispatcher) {
+                val profileExecution = async {
```
jamesarich, line 266: **`profile()` change is app-wide, not OTA-scoped — please flag the blast radius and add coverage.**

`profile()` lives in `core/ble` and is called by the main mesh-radio path (`BleRadioTransport`), `NymeaWifiService`, and both DFU transports — not only OTA. This rewrite (wait-for-discovery + `withContext(ioDispatcher)` + `async` + the `invokeOnCompletion` disconnect guard) therefore changes behavior for *every* BLE connection in the app, yet the PR is framed as OTA-only and `FakeBleConnection.profile()` still just runs `CoroutineScope(Unconfined).setup(service)` — it models neither the discovery-wait nor the disconnect cancellation, so nothing added here exercises the new logic.

Two asks:
1. Add a `KableBleConnection`-level test for the disconnect-during-profile path (connection scope closes → `NotConnectedException`), and note the app-wide scope in the PR description so reviewers know the main radio path is affected.
2. `BleOtaTransport.writeData()` calls `profile()` **once per chunk** (thousands of times per transfer), so each write now allocates a `withContext` + `async` + `invokeOnCompletion`/`dispose`. Please confirm that overhead is acceptable on the streaming hot path, or split a lighter write path that skips the setup/disconnect-guard machinery.

core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseManager.kt
```diff
@@ -279,36 +281,71 @@ open class DatabaseManager(
         Logger.d { "Closed inactive database ${anonymizeDbName(dbName)} to free connections" }
     }
 
-    private val limitedIo = dispatchers.io.limitedParallelism(4)
+    // Short-term runtime containment: serialize withDb calls to reduce Room/SQLite connection-pool churn seen during
+    // device/firmware update flows. Preserve DB-critical blocks through cancellation, then re-check cancellation so
+    // stale callers do not continue after the DB releases. Revisit after direct currentDb.value callers are audited and
+    // safe DB concurrency can be restored.
+    private val limitedIo = dispatchers.io.limitedParallelism(1)
```
jamesarich, line 290: **Comment over-states what `limitedParallelism(1)` guarantees.**

This bounds how many coroutines occupy the dispatcher's thread, but Room's suspend DAOs re-dispatch onto Room's own executor at their first suspension point — which releases this slot and lets the next `withDb` proceed. So actual DB-I/O concurrency is **not** pinned to 1; this narrows the churn *window* rather than hard-serializing access. Worth softening the comment ("serialize withDb calls") so a later reader doesn't rely on strict serialization that isn't actually guaranteed here.

core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseManager.kt
```diff
@@ -279,36 +281,71 @@ open class DatabaseManager(
...
+            } catch (retryEx: Exception) {
+                retryEx.addSuppressed(e)
+                throw retryEx
             }
         }
     }
 
+    private suspend fun <T> runCancellableDbBlock(db: MeshtasticDatabase, block: suspend (MeshtasticDatabase) -> T): T {
+        currentCoroutineContext().ensureActive()
+        val result = withContext(NonCancellable) { block(db) }
```
jamesarich, line 347: **`NonCancellable` on a single-slot dispatcher assumes `withDb` blocks stay one-shot — please document that.**

A genuinely hung one-shot query can't be cancelled and holds the containment slot until it completes. This is currently safe — the long-lived reads (`myNodeInfoFlow` / `nodeDBbyNumFlow` / `getNodesFlow`) go through `currentDb.flatMapLatest`, not `withDb`, and there's no nested `withDb` — so only short DAO calls are wrapped. But that's the load-bearing assumption of this whole containment change, so add a one-line note that `withDb` blocks must remain one-shot; a future long-lived block routed through here would hang the slot for all DB access.

jamesarich (changes requested):
Thanks for the careful hardening here — the coroutine idioms are right. In particular the `ensureActive()`-before-convert pattern (in `BleScanSupport.receiveWithin`, the connect-timeout catches, and the profile catch) is the correct fix for masking parent cancellation as a timeout, the per-attempt response-channel reset is race-safe (the collector captures the local channel, not the field), and `CoroutineStart.UNDISPATCHED` on the notification launch correctly closes the subscribe-vs-await race. Redacting BLE addresses from the logs is a nice privacy win too.

Requesting changes on two points before merge — details in the inline threads:

1. **`profile()` blast radius (core/ble).** The rewrite affects the main mesh-radio path (`BleRadioTransport`), WiFi provisioning, and DFU — not just OTA — but it's framed as OTA-only and has no direct coverage (the fake doesn't model the discovery-wait or disconnect cancellation). Please call out the app-wide scope and add a `KableBleConnection`-level test for the disconnect-during-profile path.

2. **DB containment comment vs. reality.** `limitedParallelism(1)` doesn't hard-serialize DB I/O because Room re-dispatches internally — please soften the comment, and add a note that `withDb` blocks must stay one-shot (the `NonCancellable` + single-slot combo would hang all DB access on a long-lived block).

The DB change is honestly scoped as short-term containment and the one-shot-only assumption checks out today, so I'd keep the approach — just make the assumptions explicit in-code.

Minor (non-blocking):
- `withCurrentDb` retry path: if the caller is cancelled between the first (NonCancellable) attempt and the retry, the retry's leading `ensureActive()` throws and the original closed-pool exception is dropped (not `addSuppressed`). Cancellation-wins is fine, just noting the lost context.
- `p.services.first { it != null }` treats an empty (non-null) service list as "discovered." Fine for OTA (`requireOtaCharacteristics` catches it), but the radio/DFU consumers don't re-validate.
- The stated motivation "a failed observation does not poison retries" (channel reset per attempt) isn't directly tested — the two new tests cover missing-chars and observe-fail, but not the failed→retry channel-reset cycle.

Note: I reviewed the diff statically and didn't run the build; worth a `:core:database` / `:core:ble` / `:feature:firmware` test pass before merge.

jeremiah-k:
@jamesarich Thanks, these were good catches.

I updated the PR description to call out that the `profile()` change is shared `core/ble` behavior, not OTA-only, since it affects the normal radio path and other firmware-update flows too.

For the OTA streaming hot path, I removed the repeated `profile()` entry from per-chunk writes. `connect()` now validates and stores the discovered OTA service once, and `writeData()` reuses that service for command/firmware writes instead of re-entering `profile()` for every packet. I also added coverage that streaming writes do not re-enter profile setup after connect.

On the DB side, I softened the wording so it no longer claims strict global DB I/O serialization. The comments now describe this as a single-lane `withDb` entry/containment path that narrows the Room/SQLite churn window, while acknowledging Room suspend DAOs may continue on Room’s own executor after suspension.

I also documented the load-bearing `NonCancellable` assumption: `withDb` blocks are expected to remain bounded one-shot DB work, and long-lived Flow/Paging reads should stay outside this path.

jamesarich (approved):
Re-reviewed after the feedback commit — all three blocking points are resolved, and the hot-path one better than requested. Thanks for the thorough follow-up.

- **`profile()` blast radius** — now documented on the `BleConnection.profile()` interface KDoc (shared by radio/OTA/DFU). 👍
- **Streaming hot path** — nicely done: `connect()` caches the discovered service in `otaService` and `writeData()` writes to it directly, so `profile()` runs once per connect instead of once per chunk. The `streamFirmware reuses discovered OTA service` test pinning `profileCalls == 1` is a good regression guard. This removes the concern entirely rather than just bounding it.
- **DB containment** — the `limitedIo` comment now correctly says this narrows the churn window rather than guaranteeing serialized DB I/O (Room re-dispatch caveat noted), and `runCancellableDbBlock` documents the bounded/one-shot requirement behind `NonCancellable`. Both reads I wanted.

CI is green (shard-core incl. `:core:ble`/`:core:database` and shard-feature `:feature:firmware:allTests`, plus lint/android-check).

One residual, non-blocking: the new `KableBleConnection.profile()` disconnect guard (async + `invokeOnCompletion` → `NotConnectedException`) still has no direct unit test — `FakeBleConnection.profile()` models the call, not the discovery-wait/disconnect mapping. That path is impractical to exercise without a real Kable peripheral, so the interface doc + device-log verification is a reasonable resolution; worth a line in the PR body that it's device-validated only. The earlier minors (retry drops the original closed-pool exception when cancelled between attempts; empty-service-list not re-validated by non-OTA callers) also remain and are fine to leave.

Approving. Nice work.

jamesarich (approved):

PR #6135, opened 2026-07-07
fix(discovery): ship a fresh LoRaConfig when switching to a beacon's preset

jamesarich:
## Why

Joining a Mesh Beacon via **SWITCH** copied the connected radio's *current* `LoRaConfig` and overrode only preset/region/`channel_num`. On the firmware side `use_preset = true` **only** gates `bandwidth`/`spread_factor`/`coding_rate` — every other RF field stays live. So any stale `channel_num`, `override_frequency`, or manual `bandwidth`/`spread_factor`/`coding_rate` on the radio carried over into the join and could strand it on the *old* mesh's slot instead of the offered one.

## 🐛 Fix

`MeshBeacon.toJoinChannelSet(SWITCH)` now builds a **fresh** `LoRaConfig` instead of copying the current one:

- `use_preset = true` + the advertised preset+region applied
- `channel_num` / `override_frequency` left at their zero defaults, so firmware re-derives the frequency from the offered channel name (Apple `014-mesh-beacons` FR-006)
- stale manual `bandwidth`/`spread_factor`/`coding_rate` and `tx_power` dropped

The config is applied as a **full LoRaConfig replacement** (`setLocalConfig(Config(lora = ...))`), so a truly-blank config would disable transmit. To stay safe, the fresh config still:

- carries `region` from the current config when the beacon doesn't advertise one (a zero region disables TX)
- sets the app's standard `hop_limit = 3` / `tx_enabled = true` defaults rather than zeroing them

ADD and NONE paths are unchanged. The temporary scan-dwell retune in `DiscoveryScanEngine` is intentionally left preserving the base config (it's a short-lived tune-back, documented in-place) and is out of scope here.

## Testing Performed

`./gradlew :core:model:allTests spotlessCheck detekt` — all green (13 `MeshBeaconOfferTest` cases).

Rewrote two SWITCH tests to lock in the new behavior:
- `SWITCH ships a fresh lora config so stale RF pins never survive the retune` — asserts `use_preset`, offered preset/region applied, and `channel_num`/`override_frequency`/`tx_power` all dropped from a config that had them set.
- `SWITCH without an offered preset keeps the current preset and region and resets channel_num` — asserts the region-carry fallback and preset preservation for a channel-only beacon.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved channel switching so a join now uses a fresh radio configuration rather than reusing older settings.
  * Prevented stale RF fields (including frequency/channel overrides and related RF controls) from being carried over during a switch.
  * For switches without an offered preset, the channel is reset while the current region behavior remains consistent.
  * Updated validation to match the new “clean config” switch behavior.

### DIFF
--- a/core/model/src/commonMain/kotlin/org/meshtastic/core/model/util/ChannelSet.kt
+++ b/core/model/src/commonMain/kotlin/org/meshtastic/core/model/util/ChannelSet.kt
@@ -148,20 +148,18 @@ fun MeshBeacon.toJoinChannelSet(option: BeaconJoinOption, currentLora: LoRaConfi
         BeaconJoinOption.ADD -> ChannelSet(settings = listOf(offerChannel))
 
         BeaconJoinOption.SWITCH -> {
-            // Ship a fresh LoRaConfig rather than a copy of the current one: any stale RF pin on the connected radio
+            // Start from the shared device default ([Channel.default] — use_preset=true, hop_limit/tx_enabled set,
+            // every RF field blank) rather than a copy of the current config: any stale RF pin on the connected radio
             // — an explicit channel_num, override_frequency, or manual bandwidth/spread_factor/coding_rate — would
             // otherwise strand it on the old slot. use_preset + the default (zero) channel_num/override_frequency lets
             // firmware re-derive the frequency from the offered channel name (Apple FR-006). This is sent as a full
-            // LoRaConfig replacement, so carry region (a zero region disables transmit) and set the standard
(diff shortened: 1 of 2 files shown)


jamesarich:
@coderabbitai Addressed the nitpick — the SWITCH branch now builds from `Channel.default.loraConfig.copy(...)` instead of re-hardcoding `hop_limit = 3` / `tx_enabled = true`, so the standard LoRa defaults live in one place (`Channel.kt`) and can't drift. No behavior change; `:core:model:allTests` + spotless + detekt green (5f2a78e5a).

Issue #5873, opened 2026-06-20
[Feature Request]: Add CO2 Humidity and temperature to the Airquality page


b8b8:
### Checklist

- [x] I have used the search function for **OPEN ISSUES** to see if someone else has already submitted the same feature request.
- [x] I have **also** used the search function for **CLOSED ISSUES** to see if the feature was already implemented and is just waiting to be released, or if the feature was rejected.

- [x] I will describe the request with as much detail as possible.

- [x] This request contains only one single feature, **not** a list of multiple (related) features.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Feature or improvement you want

The CO2 sensor also provides a temperature and humidity reading independent of the standard BME/BMP sensor most people install. This would be an easy way to "cheat" and have external CO2, Temp, Humidity readings on nodes while using the BMP sensors for interior moisture leakage and battery temp/health.

<img width="1080" height="1278" alt="Image" src="https://github.com/user-attachments/assets/9ca03afb-dc9d-4954-98a1-2fcc9aa2441e" />

DEBUG | 18:51:22 583 [AirQualityTelemetry] Reading SCD4X
DEBUG | 18:51:22 583 [AirQualityTelemetry] Got SCD4X readings: co2=1452, co2_temp=21.47, co2_hum57.91

Deuteranomalous1:
Wow what a great idea! People often want a weather station and monitoring their enclosure and the only way until now was to use a soil sensor instead.

DaneEvans:
Also get rid of the PM2.5 legend that is ever present while you're in there (likely future Dane)

PR #6145, opened 2026-07-08
fix(metrics): air-quality chart legend follows plotted data, not selection

jamesarich:
## Why

On the air-quality metrics screen, `selectedMetrics` defaults to `{PM2.5, CO2}` and the chart legend was built from that selection. On a node that never reports PM2.5 (e.g. a CO₂-only sensor), the chart correctly drew no PM2.5 line — but the legend still rendered a PM2.5 entry. That's the *ever-present PM2.5 legend* @DaneEvans flagged on #5873.

## What changed

🐛 The legend now follows the series that are **actually plotted** in the current time frame, not the raw selection. Added a small `internal fun metricsWithData(candidates, telemetries)` (the metrics with ≥1 reading in the frame — the same set the chart already uses to draw its lines) and drive `legendData` off it. Nodes that *do* report PM2.5 are unaffected; the fix only removes legend entries for series with no data.

Scoped to the legend/selection only — the CO₂/temp-humidity card work on this issue is untouched.

## Testing performed

- Added 2 regression tests to `AirQualityMetricsTest`: a CO₂-only node drops PM2.5 from the legend set; a series survives once it has any reading in the frame.
- `./gradlew :feature:node:allTests` — pass (incl. iOS simulator compile)
- `./gradlew :feature:node:detekt spotlessCheck` — clean

Refs #5873

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved air-quality chart legends so they only show metrics that actually have data.
  * Prevented selected but unavailable readings from appearing in the chart legend, reducing confusion when some series aren’t reported.

### DIFF
--- a/feature/node/src/commonMain/kotlin/org/meshtastic/feature/node/metrics/AirQualityMetrics.kt
+++ b/feature/node/src/commonMain/kotlin/org/meshtastic/feature/node/metrics/AirQualityMetrics.kt
@@ -97,6 +97,14 @@ internal enum class AirQuality(val labelRes: StringResource, val unit: String, v
     }
 }
 
+/**
+ * The subset of [candidates] with at least one reading in [telemetries]. The chart only draws series that have data, so
+ * the legend must use this rather than the raw selection — otherwise a default-selected series (PM2.5) shows a legend
+ * entry on nodes that never report it (issue #5873). Internal so it can be unit-tested.
+ */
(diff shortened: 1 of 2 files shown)


DaneEvans:
Perfect.  I'm adding the CO2 temp and humidity now, so I won't chase that one too

DaneEvans:
Nope, you've done that too ... I'll be patient then.

PR #6166, opened 2026-07-09
fix(ble): refresh GATT cache around ESP32 OTA profile changes

jeremiah-k:
## Overview

This PR makes ESP32 BLE OTA reliable across the full profile-switching flow on Android.

During an ESP32 BLE firmware update, the same BLE address can expose three different GATT states in sequence:

1. the normal Meshtastic radio profile,
2. the ESP32 BLE OTA loader profile,
3. the normal Meshtastic radio profile again after flashing.

Android can keep a stale GATT service cache for that address across those reboots. When that happens, the app may connect successfully but still see the wrong service table. That blocks OTA startup when the OTA loader characteristics are missing, or blocks post-update verification when Android keeps stale OTA handles after the device has rebooted back into normal radio firmware.

This change adds an Android-only, best-effort GATT cache refresh path for Kable-backed BLE connections and uses it at the ESP32 OTA handoff points so service discovery reflects the profile the device is actually running.

## Key Changes

* Added a BLE connection hook for best-effort GATT service-cache invalidation.
* Implemented guarded Android `BluetoothGatt.refresh()` support for Kable connections.
* Kept JVM and iOS behavior as safe no-ops.
* Refreshed the Android GATT cache before ESP32 BLE OTA service lookup.
* Kept the BLE OTA loader connection usable after a successful cache refresh.
* Reconnected once during BLE OTA setup when cache refresh succeeds but OTA characteristics are still missing, forcing Kable to rediscover services before failing.
* Added discovered-characteristic diagnostics for OTA service mismatch failures.
* Added a one-shot post-update BLE cache refresh request before reconnecting to normal radio firmware.
* Forced the post-update BLE radio connection to reconnect after cache refresh so Kable rediscovers the normal radio profile.
* Preserved pending post-update cache invalidation across the OTA `null -> same BLE address` reconnect path.
* Cleared stale one-shot cache-invalidation requests on explicit disconnect and genuine device-to-device switches.
* Gated post-update cache invalidation to BLE reconnects only.
* Preserved existing BLE bonds; this does not unpair or re-pair devices.
* Added fake and desktop no-op implementations for the new cache-invalidation API.
* Added focused coverage for OTA cache refresh retry and post-update invalidation behavior.

## Testing

* Added coverage proving legacy BLE-address post-update reconnect requests GATT cache invalidation.
* Added coverage proving BLE-prefixed post-update reconnect requests GATT cache invalidation.
* Added coverage proving TCP post-update reconnect does not request GATT cache invalidation.
* Added coverage proving BLE OTA reconnects once when cache refresh succeeds but required OTA characteristics are still missing.
* Added coverage proving the BLE radio transport consumes a pending post-OTA cache-invalidation request and reconnects after refresh.
* Added coverage proving GATT cache invalidation requests are consumed once.
* Added coverage proving disconnect clears a pending GATT cache invalidation request.
* Added coverage proving same-address reconnect preserves a pending GATT cache invalidation request.
* Added coverage proving device switch clears a pending GATT cache invalidation request.
* Added regression coverage proving the OTA `setDeviceAddress("n") -> request invalidation -> setDeviceAddress(original BLE address)` path preserves the pending cache-invalidation request.
* Verified on T-Beam ESP32 hardware through the firmware-update flow:

  * BLE OTA entered the ESP32 OTA loader.
  * BLE OTA service discovery reached the expected OTA service.
  * Firmware streamed successfully and returned `OK`.
  * Post-update reconnect requested GATT cache invalidation.
  * `refreshGattCache()` returned `true`.
  * The app reconnected after cache refresh.
  * BLE radio session became ready.
  * Mesh handshake resumed and local metadata was received.

## Migration Notes

* Existing BLE bonds are preserved.
* This only changes Android BLE service-cache handling around ESP32 OTA profile transitions.
* Non-Android platforms keep the existing no-op behavior.

## Summary by CodeRabbit

* **New Features**
  * Added a BLE cache refresh flow to help reconnects find the latest device services after updates.
  * Added support for retrying BLE connections when required OTA services aren’t found on the first attempt.

* **Bug Fixes**
  * Improved post-update reconnect behavior for Bluetooth devices.
  * Added better handling for stale BLE service data, including a reconnect retry when service discovery is outdated.

* **Tests**
  * Expanded coverage for BLE reconnect, cache invalidation, and OTA recovery scenarios.

### DIFF
--- a/core/ble/src/androidMain/kotlin/org/meshtastic/core/ble/KableGattCacheRefresh.kt
+++ b/core/ble/src/androidMain/kotlin/org/meshtastic/core/ble/KableGattCacheRefresh.kt
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2026 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.core.ble
+
+import android.bluetooth.BluetoothGatt
+import co.touchlab.kermit.Logger
+import com.juul.kable.Peripheral
+import kotlinx.coroutines.flow.StateFlow
+
+@Suppress("ReturnCount")
+internal actual fun Peripheral.refreshGattCache(): Boolean {
+    // Direct 2-hop reflection on Kable 0.43.1 internals.
+    // Path: BluetoothDeviceAndroidPeripheral.connection (MutableStateFlow<Connection?>) → .value → Connection.gatt
+    // Re-verify field names on Kable version bumps.
+    val gatt =
+        try {
+            extractBluetoothGatt()
+        } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+            Logger.w(e) { "refreshGattCache: BluetoothGatt extraction failed (${this.javaClass.name})" }
+            null
+        }
+            ?: run {
+                Logger.w { "refreshGattCache: BluetoothGatt unreachable via Kable internals (${this.javaClass.name})" }
+                return false
+            }
+    return try {
+        val refreshMethod = gatt.javaClass.getDeclaredMethod("refresh").apply { isAccessible = true }
+        val result = refreshMethod.invoke(gatt) as? Boolean ?: false
+        Logger.i { "refreshGattCache: refresh() returned $result" }
+        result
+    } catch (@Suppress("TooGenericExceptionCaught") e: Exception) {
+        Logger.w(e) { "refreshGattCache: refresh() invocation failed" }
+        false
+    }
+}
+
+/**
+ * Extracts the [BluetoothGatt] from Kable's internal object graph via a 2-hop field lookup.
+ *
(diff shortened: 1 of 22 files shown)


jeremiah-k:
I pushed a follow-up for the post-OTA cache-invalidation flag lifecycle.

During ESP32 OTA, the app temporarily deselects the device to free the BLE connection, then reselects the same BLE address after flashing. The pending GATT cache-invalidation request now survives that `null -> same BLE address` restore path, while still clearing on explicit disconnects and genuine device-to-device switches.

I also added regression coverage for that restore path so the pending request is not cleared before `BleRadioTransport` can consume it.

jamesarich (approved):
Reviewed the GATT cache-refresh path and the pending-invalidation state machine closely. The reflection helper is well-guarded (every hop null-safe, all exceptions caught, best-effort `false` on failure). The state machine holds up: one-shot consume is an atomic `getAndSet(false)`; on refresh failure it proceeds on the existing connection and on forced-reconnect failure it throws into `BleReconnectPolicy` (flag already consumed) so the radio session can't wedge disconnected; the flag is armed only for BLE addresses and `BleRadioTransport` is the sole consumer, so TCP/USB/nRF can't be affected; the in-OTA "reconnect once when characteristics still missing" retry is bounded to exactly one; and the `null→same-address` OTA reselect (preserve) vs genuine device switch (clear) discriminator is sound since both addresses derive from the same `radioPrefs.devAddr`. LGTM.

One LOW, non-blocking note: the device-switch clear at `SharedRadioInterfaceService.kt:521` is gated on `previousAddress != null`, so a stale flag can survive a deselect-through-null — arm for device A (OTA success) → A never reconnects → user deselects (`setDeviceAddress("n")`, flag preserved by design) → user picks a *different* BLE device B (`previousAddress == null`, not cleared) → B consumes the flag and does one spurious cache-invalidate + ~500 ms reconnect. Benign (healthy device just re-discovers the same services once, no data loss/wedge), and a picker path routing through `disconnect()` already clears it. Could also clear on the `previousAddress == null` branch if you want to close it, but it's cosmetic.

PR #6174, opened 2026-07-09
ci: merge-queue contention fixes, desktop builds on main, and 22.04-runner exit for desktop releases

jamesarich:
## Why

Three CI workstreams landed together while auditing the pipeline ahead of the Flathub submission and 2.8.0:

1. **Merge-queue runs were taking 48–96 min at peak for a pipeline that completes in 6–17 min off-peak** — pure runner-pool contention, not build time (measured from `gh run view` timing across the last 40 merge-queue runs; e.g. run 29030419053 sat 21 minutes between `setup` finishing and `lint-check` starting).
2. **Desktop distributable builds were dead code** — the `build-desktop` job was gated on a `run_desktop_builds` input that was never declared, so no caller could enable it. With Flathub in flight and 2.8.0 approaching, desktop packaging regressions need to surface before release time.
3. **The desktop release matrix was pinned to ubuntu-22.04 runners that begin brownouts 2026-09-17** (fully retired 2027-04-17), so the pin had a deadline.

## 🌟 New

- **Desktop distributable builds on every main push.** `run_desktop_builds` is now a declared input, enabled by `main-check.yml`: every merge produces `createDistributable` app images for macOS, Windows, and Linux (x64 + arm) as run artifacts. Post-merge rather than in the merge queue — `:desktopApp:test` in shard-app already covers desktop compilation there, and a 4-OS matrix (macOS/Windows queue times) would slow every merge.
- **`verify-flatpak.yml` now also triggers on `desktopApp/**` and `gradle/wrapper/**`.** The offline manifest pins the Gradle distribution independently of the wrapper, so wrapper bumps previously broke the offline Flatpak build silently; desktop build-config changes shape the uber jar the flatpak wraps.

## 🛠️ Improvements (merge-queue contention)

- **`merge-queue.yml`**: new `cancel-superseded` job. GitHub does not cancel workflow runs when a merge group is destroyed, and the old `concurrency` group keyed on `github.ref` (unique per queue entry) never matched anything — observed 3 simultaneous queued runs for PR #6166, all but one for already-destroyed merge groups, each holding ~7 heavy job slots. The new job cancels older `merge_group` runs for the same PR (matched by the `gh-readonly-queue/main/pr-N-` branch prefix, compared by run id).
- **`merge-queue.yml`**: new `check-changes` gate diffs the merge group against its base and skips the heavy pipeline when only `docs/**` / `*.md` changed (mirrors `main-check.yml`'s `paths-ignore`). `check-workflow-status` still reports success so the required check is satisfied.
- **`main-check.yml`**: passes `run_lint: false`. Every main commit is the exact merge commit the queue just verified, so re-running lint, screenshot-check, and rb-check on it was pure duplication — main pushes now build only the snapshot APKs and desktop distributables. `lint-check` is gated on `run_lint` at the job level, and `rb-check` runs solely in the merge queue.
- **`main-check.yml`**: `publish-snapshot` is guarded with `!cancelled()` so a desktop-matrix failure can't block the Android snapshot release (if the APK build itself failed, the artifact download fails and the job still goes red).

Net: ~4 fewer heavy jobs per merged PR in the gating path, stale runs no longer hold queue slots, and no gating coverage is lost — everything removed was a duplicate of a check the merge queue already ran on the same commit.

## 🐛 Fixes (desktop release off ubuntu-22.04)

- **Desktop release matrix: `ubuntu-22.04` → `ubuntu-24.04`** (x64 + arm). The pin existed because jpackage computes the `.deb` `Depends:` line from the build host, and Ubuntu 24.04 records time_t64 transition names (`libasound2t64`, `libpng16-16t64`) that don't exist on Debian 12 Bookworm / Raspberry Pi OS, making the `.deb` uninstallable there (#5233). It was never about glibc — the bundled JBR, jpackage launcher, and skiko are all vendor-prebuilt.
- **New `.deb` post-processing step** (before upload/attestation): unpack with `dpkg-deb -R`, strip the `t64` suffix from `lib*` names on the `Depends:` line, repack with `--root-owner-group`. Bookworm ships the plain names and the t64 packages `Provide` them on Trixie / Ubuntu 24.04+, so one artifact installs everywhere — strictly better than the 22.04-built one. This is the only host-independent fix: jpackage's `--linux-package-deps` is additive-only (JDK 21 `LinuxPackageBundler` appends custom deps to the host-computed list), the Compose plugin exposes no dep override, and it clears its `--resource-dir` mid-task so a custom control template can't be injected. The step logs before/after `Depends:` lines for verification on the first real release run.
- **`libfuse2` → `libfuse2t64`** for the AppImage tooling install (the old name has no install candidate on 24.04), and **`create-flatpak-src` moves to 24.04** (pure-Java uber jar + URL manifest, no host coupling).

## 🧹 Chores

- Documented the merge-queue hygiene rules in `.skills/testing-ci/SKILL.md`.

## Testing Performed

- `actionlint` on all touched workflows: the pre-existing `run_desktop_builds is not defined` error is resolved; the remaining findings are identical before/after (pre-existing shellcheck style notes).
- Docs-only merge-queue filter verified against POSIX grep for docs-only / mixed / code-only / empty-diff / `.mdx` cases; PR-prefix extraction verified against a real queue ref.
- t64 `sed` rewrite verified against a realistic `Depends:` line (versioned deps, multi-dash names, line-end), including idempotency on an already-clean line. The full unpack→rewrite→repack path can only run on a Linux release build — the step echoes before/after `Depends:` for the first release to confirm.

## Summary by CodeRabbit

* **New Features**
  * CI checks now cancel superseded merge-queue runs and skip heavy processing for docs-only changes.
  * Desktop-related packaging can be enabled or deferred as part of CI workflows, depending on the run type.

* **Bug Fixes**
  * Improved Linux desktop release packaging by adjusting dependency handling and correcting Debian package dependency names.
  * Flatpak verification now triggers when desktop app or build-wrapper files change, reducing missed failures.

* **Documentation**
  * Updated CI guidance to reflect the new merge-queue hygiene, change-based skipping, and adjusted build-check behavior.

### DIFF
--- a/.github/workflows/main-check.yml
+++ b/.github/workflows/main-check.yml
@@ -15,11 +15,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # Every commit on main arrives via the merge queue, which already ran lint, tests,
+  # screenshot validation, and rb-check on this exact merge commit. Re-running them here
+  # would be pure duplication — this workflow only needs to build the debug APKs for the
+  # snapshot release below (run_lint: false also skips screenshot-check and rb-check).
   validate-and-build:
     if: github.repository == 'meshtastic/Meshtastic-Android'
     uses: ./.github/workflows/reusable-check.yml
     with:
-      run_lint: true
+      run_lint: false
       run_unit_tests: false
       upload_artifacts: true
     secrets: inherit
--- a/.github/workflows/main-check.yml
+++ b/.github/workflows/main-check.yml
@@ -17,14 +17,18 @@ concurrency:
 jobs:
   # Every commit on main arrives via the merge queue, which already ran lint, tests,
   # screenshot validation, and rb-check on this exact merge commit. Re-running them here
-  # would be pure duplication — this workflow only needs to build the debug APKs for the
-  # snapshot release below (run_lint: false also skips screenshot-check and rb-check).
+  # would be pure duplication — this workflow builds the debug APKs for the snapshot
+  # release below (run_lint: false also skips screenshot-check and rb-check) plus the
+  # desktop distributables. Desktop packaging runs here post-merge rather than in the
+  # merge queue: :desktopApp:test in the queue's shard-app already covers compilation,
+  # and the 4-OS matrix (macos/windows queue times) would slow every merge.
   validate-and-build:
     if: github.repository == 'meshtastic/Meshtastic-Android'
     uses: ./.github/workflows/reusable-check.yml
     with:
       run_lint: false
       run_unit_tests: false
+      run_desktop_builds: true
       upload_artifacts: true
     secrets: inherit
 
@@ -34,7 +38,10 @@ jobs:
   # expire after 7 days).
   publish-snapshot:
     needs: validate-and-build
-    if: github.repository == 'meshtastic/Meshtastic-Android'
+    # !cancelled(): a desktop-matrix failure fails validate-and-build as a whole, but the
+    # Android APKs may still have built fine — attempt the snapshot regardless. If the
+    # APK build itself failed, the artifact download below fails and this job goes red.
+    if: github.repository == 'meshtastic/Meshtastic-Android' && !cancelled()
     runs-on: ubuntu-24.04-arm
     permissions:
       contents: write
--- a/.github/workflows/merge-queue.yml
+++ b/.github/workflows/merge-queue.yml
@@ -7,13 +7,74 @@ on:
 permissions:
   contents: read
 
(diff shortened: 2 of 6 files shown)


jamesarich:
Addressed the CodeRabbit nitpick in 369c27cf9: the t64 rewrite now also covers `Pre-Depends`/`Recommends`, and a post-rewrite assertion greps the whole control file for any surviving `lib*t64` name and fails the build if one remains (verified both directions: clean file passes, a missed field like `Suggests: libbazt64` trips it).

PR #6177, opened 2026-07-09
fix(database): recover firmware metadata reads after DB pool churn

jeremiah-k:
## Overview

This PR hardens firmware metadata database access during device switches and new-node connection flows.

The app can switch active per-device databases while firmware/device capability lookups are still in flight. Room KMP can surface that churn as transient connection-pool failures such as closed-pool errors or reader/writer connection acquisition timeouts.

This change routes the affected bounded firmware/device metadata reads through `DatabaseManager.withDb` and adds one-shot recovery paths for transient Room connection-pool churn.

## Key Changes

* Routed firmware release lookups through `DatabaseManager.withDb`.
* Routed device hardware capability lookups through `DatabaseManager.withDb`.
* Routed device-link writes and point-in-time reads through `DatabaseManager.withDb`.
* Kept device-link observation as a switching `flatMapLatest` flow so it resubscribes when the active database changes.
* Added transient DB-pool error detection for Room/KMP connection-pool churn.
* Retried once against the current DB when a stale DB is closed during a device switch.
* Reopened the active DB once when Room reports a reader/writer connection acquisition timeout on the current active DB.
* Left the replaced active Room instance open for the process lifetime after active-DB reopen to avoid closed-pool races with app-wide `currentDb` collectors.
* Isolated Room pool-acquire timeout message matching behind a narrow compatibility helper so non-DB connection errors do not trigger DB reopen recovery.
* Preserved safe defaults when the active database is unavailable:
  * empty lists for list reads,
  * `false` for boolean capability checks,
  * `0` for counts,
  * nullable passthrough for single-entity lookups.
* Removed redundant datasource-level dispatcher hops now that `withDb` owns the bounded DB access lane.

## Testing

* Added coverage for Room reader connection acquisition timeout detection.
* Added coverage for Room writer connection acquisition timeout detection.
* Added coverage for wrapped Room connection-pool timeout exceptions.
* Added coverage proving generic non-DB connection errors are not treated as DB pool acquisition timeouts.
* Added coverage proving BLE/GATT connection timeout messages are not treated as DB pool acquisition timeouts.
* Added coverage proving bare `database is locked` messages are not treated as active-pool acquisition timeouts.
* Updated repository tests for the simplified datasource constructors.
* Verified on device through repeated device switch / merge / unify / retire cycles without recurring Room pool acquisition timeouts.

## Migration Notes

* Existing per-device databases are preserved.
* This only changes runtime DB access/retry behavior during active database churn.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of local database access by recovering from closed connections, pool acquire timeouts, and transient locked/busy states.
  * Safer local reads and counts now fall back to empty/default values when no database result is available.
* **Reliability**
  * Improved local data observation so it always follows the currently active database.
* **Tests**
  * Added coverage for correctly detecting pool-acquire timeout errors and avoiding false positives for similar messages.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceHardwareLocalDataSource.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceHardwareLocalDataSource.kt
@@ -16,38 +16,33 @@
  */
 package org.meshtastic.core.data.datasource
 
-import kotlinx.coroutines.withContext
 import org.koin.core.annotation.Single
 import org.meshtastic.core.database.DatabaseProvider
 import org.meshtastic.core.database.entity.DeviceHardwareEntity
 import org.meshtastic.core.database.entity.asEntity
-import org.meshtastic.core.di.CoroutineDispatchers
 import org.meshtastic.core.model.NetworkDeviceHardware
 
 @Single
-class DeviceHardwareLocalDataSource(
-    private val dbManager: DatabaseProvider,
-    private val dispatchers: CoroutineDispatchers,
-) {
-    private val deviceHardwareDao
-        get() = dbManager.currentDb.value.deviceHardwareDao()
+class DeviceHardwareLocalDataSource(private val dbManager: DatabaseProvider) {
+    suspend fun insertAllDeviceHardware(deviceHardware: List<NetworkDeviceHardware>) {
+        dbManager.withDb { it.deviceHardwareDao().insertAll(deviceHardware.map { hw -> hw.asEntity() }) }
+    }
 
-    suspend fun insertAllDeviceHardware(deviceHardware: List<NetworkDeviceHardware>) =
-        withContext(dispatchers.io) { deviceHardwareDao.insertAll(deviceHardware.map { it.asEntity() }) }
-
-    suspend fun deleteAllDeviceHardware() = withContext(dispatchers.io) { deviceHardwareDao.deleteAll() }
+    suspend fun deleteAllDeviceHardware() {
+        dbManager.withDb { it.deviceHardwareDao().deleteAll() }
+    }
 
     suspend fun getByHwModel(hwModel: Int): List<DeviceHardwareEntity> =
-        withContext(dispatchers.io) { deviceHardwareDao.getByHwModel(hwModel) }
+        dbManager.withDb { it.deviceHardwareDao().getByHwModel(hwModel) }.orEmpty()
 
     suspend fun getByTarget(target: String): DeviceHardwareEntity? =
-        withContext(dispatchers.io) { deviceHardwareDao.getByTarget(target) }
+        dbManager.withDb { it.deviceHardwareDao().getByTarget(target) }
 
     suspend fun getByModelAndTarget(hwModel: Int, target: String): DeviceHardwareEntity? =
-        withContext(dispatchers.io) { deviceHardwareDao.getByModelAndTarget(hwModel, target) }
+        dbManager.withDb { it.deviceHardwareDao().getByModelAndTarget(hwModel, target) }
 
-    suspend fun hasAnyEntries(): Boolean = withContext(dispatchers.io) { deviceHardwareDao.count() > 0 }
+    suspend fun hasAnyEntries(): Boolean = dbManager.withDb { it.deviceHardwareDao().count() > 0 } ?: false
 
     /** All known `platformioTarget` values — used to determine which msh.to links are vendor links. */
-    suspend fun getAllTargets(): List<String> = withContext(dispatchers.io) { deviceHardwareDao.getAllTargets() }
+    suspend fun getAllTargets(): List<String> = dbManager.withDb { it.deviceHardwareDao().getAllTargets() }.orEmpty()
 }
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceLinkLocalDataSource.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceLinkLocalDataSource.kt
@@ -16,31 +16,32 @@
  */
 package org.meshtastic.core.data.datasource
 
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.withContext
+import kotlinx.coroutines.flow.flatMapLatest
 import org.koin.core.annotation.Single
 import org.meshtastic.core.database.DatabaseProvider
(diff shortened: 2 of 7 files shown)


jeremiah-k:
I pushed a follow-up for the @coderabbitai items.

The active-DB recovery path now emits the reopened DB before closing the replaced Room instance, so `currentDb.flatMapLatest` collectors can move to the replacement before the old pool is closed. I also kept the Room pool-acquire timeout detection narrow but moved the message matching behind a named compatibility helper with context for why the fallback exists.

jamesarich (commented):
Traced the recovery machinery closely. A lot of it is genuinely well-built — verified: the reopen is strictly one-shot (both retry arms call `runCancellableDbBlock` directly and rethrow on failure, so a second pool timeout can't cascade into another reopen); concurrent reopens are guarded by `mutex.withLock` plus a cache-identity re-check, so no double-close/double-build; the safe defaults only fire on the null-DB path, so genuine DB errors still propagate instead of being swallowed into `false`/`emptyList()`; `withDb`'s `limitedParallelism(1)` lane means dropping the datasource-level dispatcher hops doesn't put Room on Main; and the `flatMapLatest` device-link observation is a real fix — the old `currentDb.value` capture never followed DB switches. In-flight `withDb` callers holding the replaced instance also recover cleanly via the closed-during-switch retry branch.

One issue I'd want addressed before merge, and two smaller notes:

**1. `reopenActiveDatabaseIfStillCurrent` closes the active DB under the whole app's Room surface, not just the `withDb` lane.**

The deferred `managerScope.launch(io) { expectedDb.close() }` races every collector of `currentDb` app-wide — `PacketRepositoryImpl` (messages/contacts/unread flows), `SwitchingNodeInfoReadDataSource` (node DB), `TracerouteSnapshotRepositoryImpl`, the paging source factories. Only the `withDb` lane has closed-pool retry; the `flatMapLatest` observers have no `.catch`/resubscribe, and the point-in-time `currentDb.value.dao()` call sites have no recovery at all. The "close AFTER emitting so flatMapLatest collectors have already moved" ordering is weaker than it looks: those observers collect the *derived* `currentDb` (`_currentDb.filterNotNull().stateIn(managerScope)`), so the emission has to propagate through the `stateIn` pipeline before `flatMapLatest` even schedules its inner-flow switch — while the close is already launched concurrently. If the close wins, the inner Room flow throws "Connection pool is closed" and that exception propagates through `flatMapLatest` and kills the downstream flow (e.g. `observeAllLinks` has no catch).

The trigger widens the exposure: a reader-acquire timeout can be transient contention rather than a wedged pool, so one contended firmware read can close the active DB out from under healthy message/node queries. `switchActiveDatabase` refuses to close the previous DB for exactly this reason (the comment ~12 lines up: *"Closing the connection pool while those queries are executing causes 'Connection pool is closed' crashes"*), and this path breaks that invariant on the *active* DB, where the collector population is largest.

To be fair about the constraint: since reopen removes the old instance from `dbCache` and re-keys the name to the fresh one, LRU eviction can't reach the orphan — so *some* explicit close is needed if the reopen stays. Options: delay the close by a grace period so in-flight queries and the flatMapLatest switch complete; and/or add a closed-pool `retryWhen`-style resubscribe to the switching flows so a lost race self-heals instead of killing the flow.

**2. Pre-init window: reads no-op and writes silently drop (LOW).** `withCurrentDb` does `_currentDb.value ?: return null`, but `_currentDb` starts null and `init()` is fire-and-forget (`applicationScope.launch` in `MeshUtilApplication`), with a datastore read inside `switchActiveDatabase` before the first emission. The datasources this PR migrates previously used the eager, never-null `currentDb.value`, so they now return safe defaults — and `replaceFirmwareReleases`-style writes silently no-op — in that window. Self-healing (`ensureSeeded` re-evaluates per call, no one-shot latch), but a silently dropped write is a sharp edge; falling back to `currentDb.value` when `_currentDb` is null would close it.

**3. Classifier nits.** In `isRoomPoolAcquireTimeoutMessage`, the trailing `DB_TERMS.any` condition is dead — "reader connection"/"writer connection" already contain "connection" ∈ `DB_TERMS`, so it's always true when the phrase check passes. And since the tests assert against hardcoded strings, a Room version bump that rewords the message silently disables reopen recovery with no CI signal — acknowledged in the code comment, just flagging it as accepted fragility.

jeremiah-k:
@jamesarich Thanks. I really appreciate you calling out the active-DB close race, that was a great catch.

I pushed a follow-up that removes the close from the active-DB reopen path entirely. The replaced Room instance is now intentionally left open for the rest of the process because `currentDb` is derived through `stateIn`, so there is no deterministic point where all app-wide `flatMapLatest` collectors and paging factories have stopped touching the old instance. This mirrors the existing `switchActiveDatabase` no-sync-close discipline and avoids surfacing closed-pool errors to observers that do not have closed-pool recovery.

I also removed the redundant `DB_TERMS.any` condition from the Room pool-acquire timeout helper.

I left the pre-init `withDb` behavior unchanged in this PR. Falling back to `currentDb.value` can write to the sentinel default DB before init and become invisible after switching to the real device DB, so I’d prefer to handle that separately with an explicit init/await strategy rather than introducing a drive-by behavior change here.

jamesarich (approved):
Re-reviewed after `f8fa83f92`. The reopen path is now close-free — verified the only remaining closes are the LRU-eviction helper (name-keyed, non-active DBs) and full shutdown — so the app-wide collector race is gone, and the in-code comment documenting the bounded orphan + revisit condition is exactly the right record of the tradeoff. Classifier simplification is behavior-identical. Your pre-init counterpoint is well taken: persisting writes into the sentinel default DB would be worse than dropping them, so deferring to an explicit init/await strategy is the right call.

One nuance to note for the future, not a blocker: during the handoff window two Room instances share the DB file, so a collector still on the replaced instance won't observe invalidations from writes through the reopened one — transient staleness until `stateIn` propagates. Since those collectors were getting acquire timeouts before the reopen anyway, stale-then-recover strictly beats the alternative. LGTM.

jeremiah-k:
@jamesarich Thanks for the re-review and approval.

I also folded in the last `DatabaseManager.kt`-only CodeRabbit follow-ups:

* build the replacement DB before replacing the cache entry, so a failed build leaves the existing cache/current DB consistent;
* if active-DB reopen loses a concurrent switch race, retry against the current active DB instead of immediately throwing;
* tighten the switched-DB retry classifier so Room pool-acquire recovery uses the narrow reader/writer acquire-timeout predicate, while generic BLE/GATT acquire-timeout wording cannot match the DB retry path.

The close-free active-DB reopen tradeoff is unchanged, and I kept the pre-init behavior unchanged for the same reason discussed above.

PR #6190, opened 2026-07-10
fix: keep node-mute authoritative over @mentions

jamesarich:
## Why

A node the user explicitly muted could still force a notification by spamming `@!<userId>` mention tokens. The `@mention` bypass added in #6108 applied to **both** conversation/channel-mute *and* node-mute.

Per meshtastic/design#21, mentions are meant to break through a **muted channel** — a busy channel shouldn't hide a targeted ping. But muting a specific *node* is a stronger, per-sender signal, and a muted node shouldn't be able to override it. This restores that: mentions bypass conversation/channel mutes only; node-mute stays authoritative.

## 🐛 What changed

- `MeshDataHandlerImpl` — `isSilent` is now `nodeMuted || (conversationMuted && !mentionsMe)` (was `(conversationMuted || nodeMuted) && !mentionsMe`). Comment updated to reference design#21 and note the deviation.

## Testing Performed

Added two `MeshDataHandlerTest` cases:
- muted node + mention → **no** notification (`updateMessageNotification` not called)
- muted channel + mention from an unmuted node → notification (`isSilent = false`)

`./gradlew :core:data:allTests` and `:core:data:detekt` pass; `:core:data:spotlessCheck` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated mention notification precedence: a message that mentions you can still trigger audible notifications even when the conversation/channel is muted, provided the sending node isn’t muted.
  * If the sending node is muted, notifications stay suppressed even if the message mentions you.

* **Tests**
  * Added coroutine test coverage for mention vs. mute behavior, including cases where sender-node mute blocks notifications and where conversation mute is overridden by a mention.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshDataHandlerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshDataHandlerImpl.kt
@@ -415,11 +415,13 @@ class MeshDataHandlerImpl(
     ) {
         val conversationMuted = packetRepository.value.getContactSettings(contactKey).isMuted
         val nodeMuted = nodeManager.getNodeById(dataPacket.from.orEmpty())?.isMuted == true
-        // A mention of our own id is a targeted ping, so it should still notify even in a muted channel/contact.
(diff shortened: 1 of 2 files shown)


jamesarich:
@coderabbitai review

PR #6196, opened 2026-07-10
fix(service): route notification quick-replies through the send pipeline; drop dead parcelize plugin

jamesarich:
Post-AIDL cleanup audit follow-up: a sweep of the codebase after the AIDL service API removal (#5586) found two remaining items — one behavioral gap and one piece of dead build config. Everything else in the post-migration architecture checked out clean.

### 🐛 Bug Fixes

- **Notification quick-replies now go through `SendMessageUseCase`.** `ReplyReceiver` was calling `radioController.sendMessage()` directly, so a reply typed into a notification silently skipped everything an in-app send gets: homoglyph encoding, the optimistic DB-first history save, PKI contact sharing, and the durable `MessageQueue`/`SendMessageWorker` retry. Both send paths now have identical semantics.
- `goAsync()` is null-guarded in `ReplyReceiver`, which also makes the receiver directly invokable under Robolectric.

### 🧹 Chores

- Removed the `kotlin-parcelize` plugin from `core:database`, `core:datastore`, `androidApp`, the root `apply false`, and the version catalog. There are zero `@Parcelize` usages left in the repo — models stopped crossing a process boundary with the AIDL removal, and `DataPacket` is `@Serializable`.

### Testing Performed

- New `ReplyReceiverTest` (Robolectric + mokkery + Koin test module, following the `SendMessageWorkerTest` pattern):
  - reply goes through `SendMessageUseCase` and marks the conversation read
  - notification is cancelled even when the send fails
  - missing `RemoteInput` results does not send
- Full local baseline green: `spotlessApply spotlessCheck detekt assembleDebug test allTests`

## Summary by CodeRabbit

* **Bug Fixes**
  * Inline notification replies now follow the standard message-sending flow.
  * Conversations are marked as read and notifications are dismissed even when sending the reply fails.
  * Reply attempts are prevented when required reply input is missing.
* **Tests**
  * Added Robolectric coverage for reply success, send-failure cleanup, and missing reply input handling.
* **Chores**
  * Updated build configuration to remove Parcelize plugin usage and added ProGuard/R8 guidance to suppress related warnings.

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -28,7 +28,6 @@ plugins {
     alias(libs.plugins.meshtastic.android.application.compose)
     alias(libs.plugins.meshtastic.kotlinx.serialization)
     id("meshtastic.koin")
-    alias(libs.plugins.kotlin.parcelize)
     alias(libs.plugins.secrets)
     alias(libs.plugins.androidx.baselineprofile)
     id("meshtastic.aboutlibraries")
--- a/androidApp/proguard-rules.pro
+++ b/androidApp/proguard-rules.pro
@@ -40,6 +40,14 @@
 -dontwarn org.bouncycastle.**
 -dontwarn org.openjsse.**
 
+# ---- BLE (Kable) -------------------------------------------------------------
+# Kable annotates ScanResultAndroidAdvertisement with @Parcelize but doesn't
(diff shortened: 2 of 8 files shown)


jamesarich:
@coderabbitai full review

PR #6201, opened 2026-07-10
fix(dfu): harden Legacy nRF52 recovery after BLE stalls

jeremiah-k:
## Overview

This PR hardens Nordic Legacy nRF52 BLE DFU recovery after intermittent mid-stream link stalls and interrupted upload sessions.

During initial field testing on a RAK4631, an upload progressed normally before the BLE link went silent long enough to hit the negotiated supervision timeout. A later retry passed the same point successfully, indicating an intermittent link or bootloader stall rather than a deterministic transfer boundary.

The implementation was developed with reference to both the upstream [Adafruit nRF52 Bootloader](https://github.com/adafruit/Adafruit_nRF52_Bootloader) and the enhanced [OTAFIX bootloader](https://github.com/oltaco/Adafruit_nRF52_Bootloader_OTAFIX). Hardware validation was performed only with the latest recommended OTAFIX release for these boards, [`0.9.2-OTAFIX2.2-BP1.3`](https://github.com/oltaco/Adafruit_nRF52_Bootloader_OTAFIX/releases/tag/0.9.2-OTAFIX2.2-BP1.3).

This also aims to address related Legacy DFU failures reported in #6052. The recovery paths and diagnostics were informed in part by logs shared in that issue, although the specific device from the report was not available for direct reproduction.

Android cannot directly extend the peripheral-negotiated BLE supervision timeout. This change instead improves recovery behavior, retry accounting, diagnostics, and flow control on subsequent Legacy attempts.

## Key Changes

* Added a typed `MidStreamDisconnect` containing the host in-flight write offset, total image size, connection state, and last PRN-confirmed offset.
* Kept one disconnect tripwire active across packet writes and PRN waits, avoiding competing watchers that could misclassify a stream drop.
* Added bootloader identity, DFU version, stream profile, PRN interval, latency, and confirmed-checkpoint diagnostics.
* Preserved PRN 10 for the normal first attempt and introduced PRN 5 only for attempts following a typed mid-stream disconnect.
* Kept Legacy RESET as a `WITH_RESPONSE` control-point write and bounded acknowledgement waiting to one second.
* Treated missing RESET acknowledgement as ambiguous rather than definite success or failure.
* Skipped potentially ineffective same-connection cleanup after `INVALID_STATE` or a confirmed dropped link, while retaining bounded best-effort abort for ambiguous failures.
* Performed stale-session reset-prime recovery over a fresh connection.
* Separated real upload attempts from stale-session cleanup cycles.
* Preserved three real Legacy attempts and allowed up to two stale reset-prime cycles without consuming those attempts.
* Promoted an inconclusive Legacy probe to the full Legacy budget after its DFU service was confirmed.
* Preserved coroutine cancellation and `Error` propagation and guaranteed transport close through every cleanup path.

## Retry Behavior

A normal Legacy upload begins with PRN 10.

After a typed mid-stream disconnect:

1. The failed upload consumes one real attempt.
2. Subsequent Legacy attempts use PRN 5.
3. An `INVALID_STATE` response triggers reset-prime cleanup over a fresh connection.
4. Stale cleanup does not consume an upload attempt.
5. Upload and cleanup budgets remain independently bounded.

This supports recovery sequences such as:

* drop → stale cleanup → success
* drop → stale cleanup → second drop → stale cleanup → success

without creating an unbounded retry loop.

## Diagnostics

The logs distinguish the current host-side in-flight write boundary from bytes explicitly confirmed by a valid PRN.

The in-flight offset may include the chunk being attempted when a disconnect occurs inside the BLE write. It does not prove that the host stack accepted the chunk or that the bootloader received it. The last validated PRN remains the authoritative confirmed checkpoint.

Slow PRN responses are diagnostic only. They may indicate flash activity, bootloader backpressure, BLE scheduling delay, or degraded link conditions.

## Testing

Added or updated coverage for:

* NORMAL PRN 10 and RECOVERY PRN 5, including request and response cadence.
* Disconnect classification during packet writes and PRN waits.
* In-flight versus PRN-confirmed progress.
* Packet-receipt validation.
* `WITH_RESPONSE` RESET, bounded acknowledgement waiting, and ambiguous outcomes.
* RESET and ACTIVATE exception, cancellation, and `Error` behavior.
* Upload-attempt and stale-cleanup accounting.
* Repeated drop and stale-recovery sequences.
* Recovery-profile selection and Legacy budget promotion.
* Secure fallback after a pre-engagement Unknown failure.
* Cleanup behavior after completion, stale sessions, dropped links, ambiguous connection failures, and abort failures.

## Validation

Hardware updates completed successfully on:

* Multiple RAK4631s running [`0.9.2-OTAFIX2.2-BP1.3`](https://github.com/oltaco/Adafruit_nRF52_Bootloader_OTAFIX/releases/tag/0.9.2-OTAFIX2.2-BP1.3).
* Multiple Seeed SenseCAP Card Tracker T1000-Es running [`0.9.2-OTAFIX2.2-BP1.3`](https://github.com/oltaco/Adafruit_nRF52_Bootloader_OTAFIX/releases/tag/0.9.2-OTAFIX2.2-BP1.3).

The upstream Adafruit bootloader was reviewed during development but has not yet been validated on hardware with this change.

Local validation was limited to:

* `./gradlew :feature:firmware:spotlessApply`
* `git diff --check`
* commit-history and stable patch-ID inspection
* source-versus-PR-branch tree comparison
* direct static review of the changed DFU source and tests

## Scope and Compatibility

* The BLE supervision timeout is not modified.
* No fixed packet-loop pacing was added.
* PRN 5 is recovery-only and remains a resilience heuristic rather than a guarantee against bootloader-side stalls.
* Secure DFU transfer, retry, and cleanup behavior is unchanged.
* BLE bonding and protocol-detection ordering are unchanged.
* No user migration is required.
* Existing BLE bonds are preserved.
* Successful first-attempt Legacy DFU behavior remains unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved firmware update reliability during connection drops, stale sessions, timeouts, and device resets.
  * Added clearer failure diagnostics for interrupted transfers.
  * Refined retry and protocol fallback behavior, including recovery streaming after mid-transfer disconnects.
  * Ensured cancellation and serious errors are handled correctly during cleanup and reset operations.

* **Tests**
  * Expanded coverage for retry policies, fallback behavior, disconnect handling, packet receipt timing, and reset acknowledgements.

### DIFF
--- a/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/DfuUploadTransport.kt
+++ b/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/DfuUploadTransport.kt
@@ -43,7 +43,10 @@ interface DfuUploadTransport {
      */
     suspend fun transferFirmware(firmware: ByteArray, onProgress: suspend (Float) -> Unit): Result<Unit>
 
-    /** Best-effort abort of any in-flight transfer (for cancellation / error recovery). Never throws. */
+    /**
+     * Best-effort abort. Operational transport exceptions are swallowed; structured-concurrency cancellation and Error
+     * subtypes propagate.
+     */
     suspend fun abort()
 
     /** Disconnect and release resources. */
--- a/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuProtocol.kt
+++ b/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuProtocol.kt
@@ -167,6 +167,43 @@ internal fun legacyPrnRequestPayload(packets: Int): ByteArray = byteArrayOf(
 )
 
 // ---------------------------------------------------------------------------
+// Stream profiles
+// ---------------------------------------------------------------------------
+
+/**
+ * Default packet-receipt-notification interval (packets between flow-control ACKs). Higher values mean fewer
+ * notification round-trips per byte and therefore faster throughput, at the cost of a slightly longer recovery window
+ * if a packet is dropped.
+ *
+ * Capped at 10 to match Nordic's own Legacy DFU implementation, which force-limits legacy PRN to ≤10 with the comment:
+ * "DFU bootloaders from SDK 6.0.0 or older were unable to save incoming data to flash as fast as they are being sent …
+ * PRN = 10 may be the highest supported value" and treats status 6 (OPERATION_FAILED) as "data sent too fast — reduce
+ * PRN to 10 or less." The stock Adafruit bootloader shares this SDK11 flash-write path, so a higher value risks
+ * OPERATION_FAILED mid-stream on stock bootloaders. 10 is the safe ceiling that still batches flow-control ACKs.
+ */
+internal const val PRN_INTERVAL_PACKETS = 10
+
+/**
+ * Tighter PRN interval used only on a recovery upload that follows a mid-stream disconnect. Halving the interval
+ * reduces the burst depth between flow-control checkpoints. Used exclusively by [LegacyDfuStreamProfile.RECOVERY] as a
+ * recovery heuristic; it does not by itself prevent a bootloader-side stall.
+ */
+internal const val RECOVERY_PRN_INTERVAL_PACKETS = 5
+
+/**
+ * Legacy upload profile. Selects the PRN interval used in BOTH the `PACKET_RECEIPT_NOTIF_REQ` request and the
+ * receipt-await threshold inside [org.meshtastic.feature.firmware.ota.dfu.LegacyDfuTransport.streamFirmware].
+ * - [NORMAL]: the healthy first-attempt transfer profile — fewer PRN round-trips, higher throughput.
+ * - [RECOVERY]: selected after a [LegacyDfuException.MidStreamDisconnect] so subsequent attempts on the same run use
+ *   the tighter interval. A [LegacyDfuException.StaleSessionReset] alone does NOT switch the profile, since the link
+ *   did not actually drop mid-upload.
+ */
+internal enum class LegacyDfuStreamProfile(val prnIntervalPackets: Int) {
+    NORMAL(PRN_INTERVAL_PACKETS),
+    RECOVERY(RECOVERY_PRN_INTERVAL_PACKETS),
+}
+
+// ---------------------------------------------------------------------------
 // Exceptions
 // ---------------------------------------------------------------------------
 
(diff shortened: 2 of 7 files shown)


jamesarich (approved):
Reviewed the full diff plus the surrounding contracts it relies on (`safeCatching`, `withDisconnectTripwire`, `receiveWithin` idiom). The retry/cleanup policy checks out: the loop provably terminates (every iteration nets one upload attempt or one stale reset, both bounded), budget promotion happens before the exhaustion check, cancellation/`Error` contracts match what `safeCatching` actually does, and the bounded 1s `RESET` write fixes a latent indefinite hang inside the `NonCancellable` cleanup path — arguably the most valuable fix here beyond the headline feature. Test coverage is excellent, and CI (firmware shard, lint, detekt) is green.

**Approving with comments** — nothing blocking, but two items worth a reply:

1. **Classification race — a drop during `service.write()` can escape typing.** If link death makes the Kable write throw before the connection-state flow emits `Disconnected`, the raw write exception wins the `coroutineScope` race and the failure surfaces untyped — no RECOVERY profile switch, and cleanup attempts an abort on a dead link (now bounded to 1s, so cheap). Since the PR is about typing exactly this scenario, consider also checking `bleConnection.connectionState.value` in the generic write `catch` and wrapping into `MidStreamDisconnect` when disconnected. Fine as a follow-up.

2. **`Error` handling change — confirming it's intended.** `runUploadSession` narrowed `catch (Throwable)` → `catch (Exception)`, so `Error` subtypes now escape immediately instead of consuming retries/fallback first. They escaped `performUpdate`'s `catch (Exception)` before too, so the terminal outcome is unchanged — but the fail-fast is new. Looks deliberate per the PR description; flagging so it's a decision, not an accident.

Nits (take or leave):
- `bytesAtLastPrn` (local `Long`) is now redundant with `streamLastPrnOffset`; the local only feeds one final debug log.
- `MidStreamDisconnect.connectionState: String` — a "typed, don't parse the message" exception carrying a stringified `BleConnectionState` is a bit stringly-typed.
- `"Write FAILED at offset $streamOffset"` now reports the post-chunk end offset (publish-before-write), where the old log reported chunk start — anyone diffing field logs across versions will see offsets jump by one chunk.
- Test `acknowledged RESET completes and logs acknowledgement` asserts payload size, not logging — the name overpromises.
- Worst case is now up to ~6 connect/handshake cycles (3 uploads + 3 stale responses) plus reset waits — bounded and reasonable for recovery, just noting the user-visible duration can grow.

Nice work — the policy extraction into `runDfuRetryLoop`/`cleanupDfuSessionTransport` with lambda seams is exactly how this should be made testable, and the responder honoring the transport's actual requested PRN interval keeps the NORMAL/RECOVERY tests honest.

jeremiah-k:
@jamesarich Thanks for the thorough review. The `Error` fail-fast behavior is intentional—retries and fallback are for operational failures, not fatal runtime errors. Good catch on the write/state-flow race; I agree a Kable write could fail before `Disconnected` is emitted, and I’ll track that as a focused follow-up rather than widen this PR further.

PR #6209, opened 2026-07-10
fix(dfu): classify disconnects from failed Legacy writes

jeremiah-k:
## Overview

This is a focused follow-up to #6201.

During review of that PR, @jamesarich identified a narrow race where a Legacy DFU firmware write could fail before the connection-state watcher processed the corresponding disconnect. In that ordering, the raw write exception could escape without `MidStreamDisconnect` classification, preventing the next attempt from switching to the recovery profile.

The transport now re-checks the typed BLE connection state when a firmware write fails. If the link is already disconnected, it reports `MidStreamDisconnect` and preserves the original write failure as its cause. If the connection remains active, the original exception is left unclassified.

Thanks to @jamesarich for spotting this edge case.

## Changes

* Reclassifies failed firmware writes only when the BLE state is already `Disconnected`.
* Preserves the underlying write exception as the typed disconnect’s cause.
* Stores `BleConnectionState` directly instead of its string representation.
* Retains the host in-flight and last PRN-confirmed progress diagnostics.
* Removes redundant PRN-checkpoint tracking and clarifies failed-write logging.
* Aligns the test connection’s profile scope with the production factory contract.

## Testing

Added deterministic coverage for:

* a failed write occurring before the disconnect watcher processes the state change;
* typed state, byte offsets, and cause preservation in that path;
* leaving a write failure unclassified while the connection remains active.

Existing coverage for disconnects during a suspended write and during a PRN wait remains unchanged.

## Scope

This does not change PRN values, retry budgets, stale-session recovery, RESET behavior, packet pacing, Secure DFU behavior, protocol detection, or bonding.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved firmware update reliability when a Bluetooth connection drops during streaming.
  * More accurately classifies mid-update disconnects, including cases where the disconnect is detected during a failed write.
  * Preserves the underlying error details for improved diagnostics.
  * Improved firmware transfer progress reporting after streaming completes.

* **Tests**
  * Added coverage for disconnect and write-failure timing scenarios.
  * Confirmed active connections do not produce incorrect disconnect errors.

### DIFF
--- a/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuProtocol.kt
+++ b/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuProtocol.kt
@@ -18,6 +18,7 @@
 
 package org.meshtastic.feature.firmware.ota.dfu
 
+import org.meshtastic.core.ble.BleConnectionState
 import kotlin.uuid.Uuid
 
 // ---------------------------------------------------------------------------
@@ -258,15 +259,24 @@ sealed class LegacyDfuException(message: String, cause: Throwable? = null) : Dfu
      * PRN-confirmed checkpoint (the last byte the bootloader explicitly acknowledged); `-1` means no PRN was received
      * before the drop. Do not treat [bytesSent] as confirmed device progress.
      *
+     * [connectionState] is the typed [BleConnectionState] observed at the moment of the drop (always
+     * [BleConnectionState.Disconnected]); the message renders it via `toString()` only for human-readable logging.
+     *
+     * The optional [cause] carries the underlying operational write exception when classification was triggered by a
+     * failed BLE write on an already-disconnected link. Watcher-originated disconnects (where the state-flow watcher
+     * detected the drop independently) may have no cause.
+     *
      * Typed — callers must NOT parse the message to detect a mid-stream drop.
      */
     class MidStreamDisconnect(
         val bytesSent: Int,
         val totalBytes: Int,
-        val connectionState: String,
+        val connectionState: BleConnectionState,
         val lastConfirmedBytes: Int,
+        cause: Throwable? = null,
     ) : LegacyDfuException(
         "BLE link dropped mid-upload at host in-flight offset $bytesSent/$totalBytes " +
             "(lastConfirmed=$lastConfirmedBytes, state=$connectionState)",
+        cause,
     )
 }
--- a/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuTransport.kt
+++ b/feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuTransport.kt
@@ -349,13 +349,12 @@ internal constructor(
                 LegacyDfuException.MidStreamDisconnect(
                     bytesSent = streamOffset,
                     totalBytes = firmware.size,
-                    connectionState = state.toString(),
+                    connectionState = state,
                     lastConfirmedBytes = streamLastPrnOffset,
                 )
             },
         ) {
             var packetsSincePrn = 0
-            var bytesAtLastPrn = 0L
             bleConnection.profile(LegacyDfuUuids.SERVICE, timeout = STREAM_TIMEOUT) { service ->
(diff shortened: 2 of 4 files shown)


feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuProtocol.kt
```diff
@@ -258,15 +259,24 @@ sealed class LegacyDfuException(message: String, cause: Throwable? = null) : Dfu
...
+     * failed BLE write on an already-disconnected link. Watcher-originated disconnects (where the state-flow watcher
+     * detected the drop independently) may have no cause.
+     *
      * Typed — callers must NOT parse the message to detect a mid-stream drop.
      */
     class MidStreamDisconnect(
         val bytesSent: Int,
         val totalBytes: Int,
-        val connectionState: String,
+        val connectionState: BleConnectionState,
```
jamesarich, line 274: Non-blocking: since the KDoc guarantees this is always `BleConnectionState.Disconnected`, the property could be typed as `BleConnectionState.Disconnected` and the compiler would enforce that contract instead of the comment. Both construction sites already have a proven-`Disconnected` value; the only friction is `withDisconnectTripwire`'s `onDrop: (BleConnectionState) -> Throwable` parameter, which could become `(BleConnectionState.Disconnected) -> Throwable` by swapping the watcher's `first { it is … }` for `filterIsInstance<BleConnectionState.Disconnected>().first()`. That would also give callers direct access to `DisconnectReason` without a cast. Fine as a follow-up.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuTransport.kt
```diff
@@ -374,9 +373,26 @@ internal constructor(
...
-                    } catch (@Suppress("TooGenericExceptionCaught") e: Throwable) {
+                    } catch (e: Exception) {
+                        // The outer stream tripwire watches `connectionState` for Disconnected, but `service.write()`
+                        // can throw BEFORE the state-flow watcher processes the Disconnected emission. In that ordering
+                        // the raw write exception would escape without typed MidStreamDisconnect classification, so
+                        // later retries would not switch to the RECOVERY profile. Re-check the typed state here: if the
+                        // link is already Disconnected, classify as MidStreamDisconnect so the retry coordinator can
+                        // react. Error subtypes are NOT caught here — they propagate unchanged.
+                        val state = bleConnection.connectionState.value
+                        if (state is BleConnectionState.Disconnected) {
```
jamesarich, line 384: Question (non-blocking): a write that fails while the state reads `Disconnecting` still escapes unclassified — and by the time `Disconnected` lands, the tripwire watcher has already been cancelled, so that ordering ends in the same unclassified failure this PR fixes for `Disconnected`. I think that's fine in practice, since `Disconnecting` implies a local-initiated teardown that the cancellation path owns — but if that's the reasoning, one line here pinning it down would help the next reader.

jamesarich (commented):
Two non-blocking notes from review — the fix itself looks correct and the deterministic race test is a nice touch.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuProtocol.kt
jeremiah-k: Good point. Typing the property as `BleConnectionState.Disconnected` would make the contract stronger and expose the disconnect reason without a cast. I’m leaving the broader type here to keep this PR focused rather than changing the shared tripwire API, but I agree that would be a reasonable follow-up.

feature/firmware/src/commonMain/kotlin/org/meshtastic/feature/firmware/ota/dfu/LegacyDfuTransport.kt
jeremiah-k: Yes, that’s the reasoning. `Disconnecting` represents a local teardown already in progress, so I don’t want to reclassify a write failure there as a remote mid-stream drop. Only the terminal `Disconnected` state produces `MidStreamDisconnect`; otherwise the original write exception is preserved.

jamesarich (approved):

PR #6218, opened 2026-07-11
feat(map): send a waypoint as a DM or to a specific channel

joeyleake:
Waypoints could previously only be broadcast on the primary channel. The Meshtastic API can target a specific node or channel when sending a waypoint, so this brings the Android app's waypoint editor up to parity: a "Send to" picker lets you choose Broadcast/the primary channel (shown by its real configured name), any secondary channels, or a specific node for a private DM — matching the same routing options the Meshtastic API already supports.

The DM/channel routing plumbing (`BaseMapViewModel.sendWaypoint`) already accepted a destination `contactKey`; this was purely a UI gap.

## 🌟 New Features

- Added a "Send to" row and recipient picker to the waypoint editor (`EditWaypointDialog`), listing: the primary channel (by its configured name, e.g. "LongFast"), any secondary channels, a divider, then every known node — letting a waypoint be broadcast, sent to a specific channel, or DM'd to one node.
- Node list in the picker now follows whatever sort mode you've set on the Nodes tab (previously hardcoded to last-heard-first), and re-sorts live if you change that preference.

## 🐛 Bug Fixes

- Drawing a geofence box for a waypoint no longer silently resets the chosen recipient back to Broadcast — the selection is now carried through the box-authoring round trip.
- "Delete for everyone" now resends the expiry to the waypoint's original recipient (DM or channel) instead of always broadcasting it, so deletions actually reach where the waypoint was sent.
- DM contact keys now only use the PKC (public-key) channel when both the local and remote node support it, matching the existing convention in `NodeListViewModel.getDirectMessageRoute`, instead of assuming PKC unconditionally (which could silently misroute a DM to a non-PKC node).
- The recipient label no longer falsely reports "Broadcast" when the selected DM node's name is blank or the node drops out of range — it now falls back to the node's short name or raw ID.
- The recipient picker now still offers a Broadcast option even when the channel set is non-null but empty (e.g. right after a fresh install, before channel sync completes).

## 🧹 Chores

- Added `waypoint_recipient_broadcast`, `waypoint_recipient_channel`, and `waypoint_send_to` string resources.

## Testing Performed

- `feature/map/src/androidHostTest/.../EditWaypointDialogTest.kt` (new, Robolectric + Compose UI test): covers default primary-channel labelling, the generic-Broadcast fallback when disconnected, an empty-but-non-null `ChannelSet`, DM routing for both PKC-capable and non-PKC nodes, secondary-channel routing, and `initialContactKey` seeding (the geofence-box round trip).
- `feature/map/src/commonTest/.../BaseMapViewModelTest.kt`: added coverage for `sendWaypoint`'s default-broadcast and explicit-`contactKey` routing, `sendWaypoint(id = 0)` not sending anything, and the node list re-sorting live when the Nodes-tab sort preference changes.
- Full baseline verified locally: `spotlessApply spotlessCheck detekt assembleDebug test allTests` — all green.
- Manually verified end-to-end against a real device/mesh: waypoint editor shows the real primary channel name, the picker lists channels then nodes in the expected order, and selecting a node/channel routes correctly.

screen40

screen41

## Summary by CodeRabbit

* **New Features**
  * Added recipient selection to the waypoint editor, including broadcast, channel recipients, and direct-message nodes.
  * Added “send to” labels/prompts and a recipient picker with node filtering.
  * Preserved the selected waypoint recipient across the geofence box authoring/editing flow.
* **Bug Fixes**
  * Waypoint send/delete now expiry updates target the originally intended recipient/channel (avoids unintended broadcast).
  * Node list sorting updates immediately when the sort preference changes.
* **Tests**
  * Expanded Android UI and view-model test coverage for recipient selection, routing, and filtering.


feature/map/src/commonMain/kotlin/org/meshtastic/feature/map/BaseMapViewModel.kt
```diff
@@ -40,6 +42,8 @@ import org.meshtastic.core.repository.NotificationPrefs
 import org.meshtastic.core.repository.PacketRepository
 import org.meshtastic.core.repository.RadioConfigRepository
 import org.meshtastic.core.repository.RadioController
+import org.meshtastic.core.repository.UiPrefs
```
jamesarich, line 46: Pairs with the `waypointContactKey` helper suggested below (`isFromLocal` is already imported). Anchored on the nearest changed line, so it lands out of sorted order — `./gradlew spotlessApply` will slot it into place.

```suggestion
import org.meshtastic.core.model.isBroadcast
import org.meshtastic.core.repository.UiPrefs
```

feature/map/src/commonMain/kotlin/org/meshtastic/feature/map/BaseMapViewModel.kt
```diff
@@ -176,16 +185,15 @@ open class BaseMapViewModel(
...
         safeLaunch(context = ioDispatcher, tag = "deleteWaypoint") { packetRepository.deleteWaypoint(id) }
 
-    fun sendWaypoint(wpt: Waypoint, contactKey: String = "0${NodeAddress.ID_BROADCAST}") {
+    fun sendWaypoint(wpt: Waypoint, contactKey: String = "0${NodeAddress.ID_BROADCAST}"): Job? {
         // contactKey: unique contact key filter (channel)+(nodeId)
         val parsedKey = ContactKey(contactKey)
         val p = DataPacket(parsedKey.addressString, parsedKey.channel, wpt)
-        if (wpt.id != 0) sendDataPacket(p)
+        return if (wpt.id != 0) sendDataPacket(p) else null
     }
```
jamesarich, line 204: The `"${it.channel}${it.to}"` idiom the MapViews use for delete-for-everyone reads `to` unconditionally, but the app's canonical contact derivation (`MeshDataHandlerImpl.rememberDataPacket`) is `if (fromLocal || toBroadcast) to else from`. For a waypoint someone **DM'd to us**, `to` is our own node — the expiry gets addressed back to ourselves and never reaches the sender. Centralizing the rule here lets all four call sites (delete + edit-seed, both flavors — see the paired suggestions) share it.

```suggestion
    }

    /**
     * Contact key a waypoint packet was exchanged on: its destination for packets we sent (or broadcasts), otherwise
     * its sender — the same derivation as MeshDataHandlerImpl.rememberDataPacket, so edits and expiries route back to
     * the conversation the waypoint actually lives in (a raw `to` would self-address a waypoint someone DM'd to us).
     */
    fun waypointContactKey(packet: DataPacket): String {
        val contact = if (packet.isFromLocal(myNodeNum) || packet.isBroadcast) packet.to else packet.from
        return "${packet.channel}$contact"
    }
```

androidApp/src/google/kotlin/org/meshtastic/app/map/MapView.kt
```diff
@@ -702,27 +711,38 @@ fun MapView(
                 EditWaypointDialog(
                     waypoint = waypointToEdit,
                     displayUnits = displayUnits,
-                    onSend = { updatedWp ->
+                    nodes = nodes,
+                    ourNode = ourNodeInfo,
+                    channelSet = channelSet,
+                    initialContactKey = boxAuthoringDraftContactKey ?: "0${NodeAddress.ID_BROADCAST}",
```
jamesarich: **Editing an existing waypoint silently resets its destination to Broadcast.** DM a waypoint privately to one node, later edit its name, hit Send — it re-transmits mesh-wide unless the user notices the picker reset. The PR already routes *deletion* back to the original recipient; the same seeding belongs on the edit path. (Uses the `waypointContactKey` helper suggested in `BaseMapViewModel`.)

```suggestion
                    // Seed from the waypoint's original destination so editing a DM'd/channel waypoint re-sends it
                    // where it originally went instead of silently re-broadcasting a privately shared waypoint.
                    initialContactKey =
                    boxAuthoringDraftContactKey
                        ?: waypoints[waypointToEdit.id]?.let { mapViewModel.waypointContactKey(it) }
                        ?: "0${NodeAddress.ID_BROADCAST}",
```

androidApp/src/google/kotlin/org/meshtastic/app/map/MapView.kt
```diff
@@ -702,27 +711,38 @@ fun MapView(
...
                     },
                     onDelete = { wpToDelete ->
                         if (wpToDelete.locked_to == 0 && isConnected && wpToDelete.id != 0) {
-                            mapViewModel.sendWaypoint(wpToDelete.copy(expire = 1))
+                            // Route the expiry to wherever the waypoint was originally sent (a DM or a secondary
+                            // channel), not the default broadcast — otherwise a DM'd waypoint's deletion never
+                            // reaches its actual recipient.
+                            val originalContactKey =
+                                waypoints[wpToDelete.id]?.let { "${it.channel}${it.to}" }
+                                    ?: "0${NodeAddress.ID_BROADCAST}"
```
jamesarich, line 739: For a waypoint someone DM'd *to us*, `it.to` is our own node id, so this expiry self-addresses instead of reaching the sender — the canonical rule (`MeshDataHandlerImpl.rememberDataPacket`) is `if (fromLocal || broadcast) to else from`. The `waypointContactKey` helper suggested in `BaseMapViewModel` applies it.

```suggestion
                            val originalContactKey =
                                waypoints[wpToDelete.id]?.let { mapViewModel.waypointContactKey(it) }
                                    ?: "0${NodeAddress.ID_BROADCAST}"
```

androidApp/src/fdroid/kotlin/org/meshtastic/app/map/MapView.kt
```diff
@@ -894,7 +900,11 @@ fun MapView(
         EditWaypointDialog(
             waypoint = showEditWaypointDialog ?: return, // Safe call
             displayUnits = displayUnits,
-            onSend = { waypoint ->
+            nodes = nodes,
+            ourNode = ourNodeInfo,
+            channelSet = channelSet,
+            initialContactKey = geofenceBoxDraftContactKey ?: "0${NodeAddress.ID_BROADCAST}",
```
jamesarich: Same as the google flavor: editing an existing DM'd/channel waypoint resets the recipient to Broadcast, so a privately shared waypoint gets re-broadcast on re-send. Seed from the stored packet like the delete path does. (Uses the `waypointContactKey` helper suggested in `BaseMapViewModel`.)

```suggestion
            // Seed from the waypoint's original destination so editing a DM'd/channel waypoint re-sends it
            // where it originally went instead of silently re-broadcasting a privately shared waypoint.
            initialContactKey =
            geofenceBoxDraftContactKey
                ?: showEditWaypointDialog
                    ?.let { wp -> waypoints[wp.id] }
                    ?.let { mapViewModel.waypointContactKey(it) }
                ?: "0${NodeAddress.ID_BROADCAST}",
```

androidApp/src/fdroid/kotlin/org/meshtastic/app/map/MapView.kt
```diff
@@ -979,7 +992,13 @@ fun MapView(
                         TextButton(
                             onClick = {
                                 Logger.d { "User deleted waypoint ${waypoint.id} for everyone" }
-                                mapViewModel.sendWaypoint(waypoint.copy(expire = 1))
+                                // Route the expiry to wherever the waypoint was originally sent (a DM or a secondary
+                                // channel), not the default broadcast — otherwise a DM'd waypoint's "delete for
+                                // everyone" never reaches its actual recipient.
+                                val originalContactKey =
+                                    waypoints[waypoint.id]?.let { "${it.channel}${it.to}" }
+                                        ?: "0${NodeAddress.ID_BROADCAST}"
```
jamesarich, line 1005: Same received-DM edge as the google flavor: `it.to` self-addresses the expiry when the waypoint was DM'd to us.

```suggestion
                                val originalContactKey =
                                    waypoints[waypoint.id]?.let { mapViewModel.waypointContactKey(it) }
                                        ?: "0${NodeAddress.ID_BROADCAST}"
```

feature/map/src/androidMain/kotlin/org/meshtastic/feature/map/component/EditWaypointDialog.kt
```diff
@@ -97,37 +106,103 @@ import org.meshtastic.core.resources.send
 import org.meshtastic.core.resources.time
 import org.meshtastic.core.resources.waypoint_edit
 import org.meshtastic.core.resources.waypoint_new
+import org.meshtastic.core.resources.waypoint_recipient_broadcast
```
jamesarich, line 105: For the filter field suggested below. Anchored on the nearest changed line — `./gradlew spotlessApply` will sort it into place (`node_filter_placeholder` already exists in core resources, no new string/Crowdin churn).

```suggestion
import org.meshtastic.core.resources.node_filter_placeholder
import org.meshtastic.core.resources.waypoint_recipient_broadcast
```

feature/map/src/androidMain/kotlin/org/meshtastic/feature/map/component/EditWaypointDialog.kt
```diff
@@ -97,37 +106,103 @@ import org.meshtastic.core.resources.send
...
+@Composable
+private fun recipientLabel(contactKey: String, nodes: List<Node>, ourNode: Node?, channelSet: ChannelSet?): String {
+    val parsed = ContactKey(contactKey)
+    return if (parsed.address is NodeAddress.Broadcast) {
+        channelLabel(channelSet, parsed.channel)
+    } else {
+        val node = nodes.find { dmContactKey(it, ourNode) == contactKey }
+        node?.user?.long_name?.ifBlank { node.user.short_name }
+            ?: (parsed.address as? NodeAddress.ById)?.id
+            ?: stringResource(Res.string.waypoint_recipient_broadcast)
```
jamesarich: `ifBlank` returns the (still blank, non-null) short name when both names are blank, so the raw-id fallback described in the KDoc/PR body never fires — the button renders an empty label for a node with unset names.

```suggestion
        node?.user?.long_name?.takeIf { it.isNotBlank() }
            ?: node?.user?.short_name?.takeIf { it.isNotBlank() }
            ?: (parsed.address as? NodeAddress.ById)?.id
            ?: stringResource(Res.string.waypoint_recipient_broadcast)
```

feature/map/src/androidMain/kotlin/org/meshtastic/feature/map/component/EditWaypointDialog.kt
```diff
@@ -403,6 +492,85 @@ fun EditWaypointDialog(
...
+                items(distinctNodes, key = { it.num }) { node ->
+                    val contactKey = dmContactKey(node, ourNode)
+                    RecipientRow(
+                        label = node.user.long_name.ifBlank { node.user.short_name },
+                        selected = selectedContactKey == contactKey,
+                        onClick = { onSelect(contactKey) },
+                    )
+                }
+            }
+        },
```
jamesarich: **This picker gets unusable on big meshes.** Rendering is already fine at any scale (LazyColumn capped at 400dp, keyed after `distinctBy`, sort pushed into SQL) — but an MQTT-bridged node DB is thousands of entries and there's no way to search; `NodeListScreen` grew `NodeFilterTextField` for exactly this. A local filter field (reusing the existing `node_filter_placeholder` string — see import suggestion above) narrows nodes by long/short name while keeping the channel rows pinned. Also folds in a blank-name fallback on the row label (`ifBlank { node.user.id }`) so nodes with unset names don't render empty rows.

```suggestion
    var filterText by rememberSaveable { mutableStateOf("") }
    val distinctNodes = nodes.distinctBy { it.num }
    val filteredNodes =
        if (filterText.isBlank()) {
            distinctNodes
        } else {
            distinctNodes.filter {
                it.user.long_name.contains(filterText, ignoreCase = true) ||
                    it.user.short_name.contains(filterText, ignoreCase = true)
            }
        }

    AlertDialog(
        onDismissRequest = onDismissRequest,
        title = { Text(stringResource(Res.string.waypoint_send_to)) },
        text = {
            Column {
                OutlinedTextField(
                    value = filterText,
                    onValueChange = { filterText = it },
                    placeholder = { Text(stringResource(Res.string.node_filter_placeholder)) },
                    singleLine = true,
                    modifier = Modifier.fillMaxWidth(),
                )
                Spacer(modifier = Modifier.size(8.dp))
                LazyColumn(modifier = Modifier.heightIn(max = 400.dp)) {
                    items(count = channelCount(channelSet), key = { "channel_$it" }) { channelIndex ->
                        val contactKey = ContactKey.broadcast(channelIndex).value
                        RecipientRow(
                            label = channelLabel(channelSet, channelIndex),
                            selected = selectedContactKey == contactKey,
                            onClick = { onSelect(contactKey) },
                        )
                    }
                    if (filteredNodes.isNotEmpty()) {
                        item {
                            Spacer(modifier = Modifier.size(8.dp))
                            HorizontalDivider()
                            Spacer(modifier = Modifier.size(8.dp))
                        }
                    }
                    items(filteredNodes, key = { it.num }) { node ->
                        val contactKey = dmContactKey(node, ourNode)
                        RecipientRow(
                            label = node.user.long_name.ifBlank { node.user.short_name }.ifBlank { node.user.id },
                            selected = selectedContactKey == contactKey,
                            onClick = { onSelect(contactKey) },
                        )
                    }
                }
            }
        },
```

feature/map/src/androidHostTest/kotlin/org/meshtastic/feature/map/component/EditWaypointDialogTest.kt
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.feature.map.component
+
+import androidx.compose.ui.test.ExperimentalTestApi
+import androidx.compose.ui.test.assertIsDisplayed
+import androidx.compose.ui.test.onNodeWithText
+import androidx.compose.ui.test.performClick
```
jamesarich, line 24: For the filter test suggested at the end of this file.

```suggestion
import androidx.compose.ui.test.performClick
import androidx.compose.ui.test.performTextInput
```

feature/map/src/androidHostTest/kotlin/org/meshtastic/feature/map/component/EditWaypointDialogTest.kt
```diff
@@ -0,0 +1,215 @@
...
+import androidx.compose.ui.test.v2.runComposeUiTest
+import okio.ByteString.Companion.encodeUtf8
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.meshtastic.core.model.NodeAddress
+import org.meshtastic.core.resources.Res
+import org.meshtastic.core.resources.getString
+import org.meshtastic.core.resources.send
```
jamesarich, line 34: For the filter test suggested at the end of this file.

```suggestion
import org.meshtastic.core.resources.node_filter_placeholder
import org.meshtastic.core.resources.send
```

feature/map/src/androidHostTest/kotlin/org/meshtastic/feature/map/component/EditWaypointDialogTest.kt
```diff
@@ -0,0 +1,215 @@
...
+                onDismissRequest = {},
+            )
+        }
+
+        onNodeWithText("Secondary").assertIsDisplayed()
+        onNodeWithText(getString(Res.string.send)).performClick()
+
+        assertEquals("1${NodeAddress.ID_BROADCAST}", sentContactKey)
+    }
+}
```
jamesarich, line 252: Covers the filter field suggested in `EditWaypointDialog.kt` (needs the two import suggestions above). Verified locally: all 8 tests in this class pass via `:feature:map:allTests`.

```suggestion

    @Test
    fun filteringTheRecipientPickerNarrowsNodesButKeepsChannels() = runComposeUiTest {
        val bob = TestDataFactory.createTestNode(num = 43, userId = "!b2c3d4e5", longName = "Bob")
        setContent {
            EditWaypointDialog(
                waypoint = waypoint,
                displayUnits = DisplayUnits.METRIC,
                nodes = listOf(node, bob),
                ourNode = null,
                channelSet = channelSet,
                onSend = { _, _ -> },
                onDelete = {},
                onDismissRequest = {},
            )
        }

        onNodeWithText("Primary").performClick()
        onNodeWithText(getString(Res.string.node_filter_placeholder)).performTextInput("ali")

        // Case-insensitive name filter drops Bob but never the channel rows.
        onNodeWithText("Alice").assertIsDisplayed()
        onNodeWithText("Bob").assertDoesNotExist()
        onNodeWithText("Secondary").assertIsDisplayed()
    }
}
```

jamesarich (changes requested):
Really nice feature — the routing plumbing, the box-authoring round trip, and the test suite are all solid, and the CodeRabbit rounds (SQL-side sorting via the shared `UiPrefs.nodeSortOption`) made it better. Requesting changes for two behavioral issues and one at-scale UX gap; every request is an inline suggestion you can accept from the UI.

**The three asks** (details on each suggestion):
1. **Edit re-broadcasts private waypoints** — `initialContactKey` isn't seeded from the waypoint's original destination, so editing a DM'd waypoint silently re-sends it as a mesh-wide broadcast. The delete path already looks up the original recipient; the edit path needs the same.
2. **Delete-for-everyone self-addresses received DMs** — `"${it.channel}${it.to}"` uses `to` unconditionally; for a waypoint DM'd *to us* that's our own node. One shared `waypointContactKey` helper in `BaseMapViewModel` fixes both flavors and the edit-seed at once.
3. **No filter in the recipient picker** — rendering is fine at any node count, but scrolling thousands of MQTT-bridged nodes in a 400dp dialog isn't; a small filter field (existing `node_filter_placeholder` string, no new deps) plus a test.

Also folded in: blank-name fallbacks so nodes with unset names don't render empty labels/rows.

**Verification:** all suggestions were applied together locally and pass `spotlessCheck`, `detekt`, `:feature:map:allTests` (incl. the new test), and both `androidApp` flavor compiles. The two import suggestions are anchored to the nearest changed line, so run `./gradlew spotlessApply` after accepting — it slots them into sorted position.

**Non-blocking notes** (no action needed for this PR):
- `mapViewModel.nodes` is now collected unconditionally in the google `MapView`; moving the `collectAsStateWithLifecycle` inside the `editingWaypoint != null` block would avoid an always-on subscription.
- `dmContactKey` is a third copy of the DM-route rule (after `NodeListViewModel.getDirectMessageRoute`) — worth extracting next to `ContactKey` in `core:model` in a follow-up.
- Unnamed secondary channels resolve to the modem-preset name via `Channel.name`, so the picker can show two identical "LongFast" rows. Cosmetic.

joeyleake:
Thanks for the thorough review — all three required fixes are in, plus the blank-name fallback.

**1. Edit re-broadcasts private waypoints** — fixed. Added `BaseMapViewModel.waypointContactKey()` and seeded `initialContactKey` from it on the edit path in both flavors, same as the delete path already did.

**2. Delete-for-everyone self-addresses received DMs** — fixed via the same `waypointContactKey()` helper, used in both `MapView.kt` flavors' delete-for-everyone handlers.

**3. No filter in the recipient picker** — added, reusing `node_filter_placeholder`. Filters nodes by long/short name; channel rows stay unfiltered. Covered by a new test (`filteringRecipientPickerNarrowsNodesButKeepsChannelsVisible`).

Also fixed the blank-name `ifBlank` fallback bug.

Two small deviations from the literal suggested diffs, both intentional:
- The recipient picker row's final label fallback uses `NodeAddress.numToDefaultId(node.num)` instead of `node.user.id` — `node.user.id` can be blank on a bare/default node, whereas `node.num` is always populated and `numToDefaultId` is the existing convention (`NodeAddress.ByNum.toIdString()`) for deriving a display id from a bare node number.
- Also keyed `EditWaypointDialog`'s `waypointInput`/`selectedContactKey` state to `waypoint.id` (`remember(waypoint.id)` / `rememberSaveable(waypoint.id)`). Not something you flagged — the dialog is always fully unmounted/remounted between different waypoints today, so this isn't reachable via the current UI, but it's a cheap defensive fix against that changing later.

One structural change beyond the suggestions: pulling in the filter field pushed `EditWaypointDialog.kt` over detekt's function-count limit, so I split `WaypointRecipientPickerDialog` and its helpers out into a new `WaypointRecipientPicker.kt` in the same package.

Verified locally: `spotlessCheck`, `detekt`, `assembleDebug` (both flavors), `test`, `allTests` all green, including the new filter test.

Left the three non-blocking notes (unconditional `nodes` collection, `dmContactKey` duplication, unnamed-secondary-channel cosmetic issue) for a follow-up, as suggested.

Re-requesting review — let me know if I've missed anything.

jamesarich (approved):
All three requested changes landed in 74bd2401 — and improved on the suggestions rather than just accepting them: the picker moved to its own file with the filter memoized and extracted as a pure `filterNodesByName`, the row-label fallback uses `NodeAddress.numToDefaultId` (sturdier than a raw user id), the filter test asserts match *counts* to dodge the field-value/row ambiguity, and keying the dialog state to `waypoint.id` is a nice defensive touch I didn't ask for.

Verified locally at this head: `spotlessCheck`, `detekt`, `:feature:map:allTests` (8/8 in `EditWaypointDialogTest`, including the new filter test), and both `androidApp` flavor compiles — all green.

The earlier non-blocking notes (scoping the google `MapView` nodes collection, extracting `dmContactKey` to `core:model` beside `ContactKey`, duplicate preset-name labels for unnamed secondary channels) remain fine as follow-ups. Nice work!

PR #6247, opened 2026-07-12
feat(desktop): attach installable .flatpak bundles to GitHub releases

jamesarich:
## Why

Releases carry `flatpak-sources.json` (the offline build manifest for Flathub) but no installable Flatpak — Linux users must wait for the Flathub manifest bump to propagate, or build it themselves. Meanwhile `verify-flatpak.yml` already builds real per-arch `.flatpak` bundles on every relevant PR and then discards them as short-lived workflow artifacts. This PR runs that same proven build at release time and attaches the bundles to the GitHub release, giving sideloadable Flatpaks (`flatpak install ./….flatpak`) the moment a release exists — independent of Flathub propagation, and superseded cleanly by a later Flathub install of the same app id.

## 🌟 New Features

- **`build-flatpak-bundle` job in `release.yml`** (x86_64 + aarch64, gated on `build_flatpak_src` like the sources jobs): clones the Flathub packaging repo, overlays the shared offline manifest (`scripts/verify-flatpak/desktop-offline.yaml`) plus the **combined** `flatpak-sources.json` from `release-flatpak-src` (the two-arch union — strictly safer than verify's single-arch capture), rsyncs the **tag checkout** in as the build source, and builds fully offline. Exported as `org.meshtastic.MeshtasticDesktop-<version>-<arch>.flatpak`, provenance-attested, and picked up by the existing `github-release` artifact download (its `flatpak-multisrc-*` exclusion doesn't match).
- `github-release` now also `needs` the bundle job, keeping the strict all-or-nothing release semantics.

## Trade-offs

- **Flathub becomes a release-cut dependency**: the job pulls the freedesktop runtime/SDK + openjdk extension from Flathub at build time. A Flathub outage fails the cut; the existing retry path (failed internal release deletes its tag → fresh dispatch) covers it. Noted in a job comment.
- **Runner time**: two more jobs per internal cut (verify-flatpak's timings suggest ~20–40 min each, free-tier public runners — contention cost only).
- The build steps intentionally mirror `verify-flatpak.yml` rather than being extracted to a composite action; the substantive pins (runtime version, JBR, openjdk extension) already live in the shared overlay manifest, so drift risk is confined to mechanical steps.

## Testing Performed

- `actionlint` clean (pre-existing info-level shellcheck notes in untouched steps only).
- The build recipe is byte-for-byte the `verify-flatpak.yml` `build-flatpak` job — green on both arches as recently as #6219's PR checks — with three release-specific deltas: tag checkout instead of PR head, combined two-arch sources instead of single-host capture, and a versioned bundle filename.
- First real-artifact validation on the next internal release cut.

## Summary by CodeRabbit

* **New Features**
  * Added support for generating installable, versioned Flatpak bundles during releases.
  * Flatpak bundles are now included in draft GitHub Releases.
  * Generated bundles are uploaded with provenance attestations for improved distribution trust.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -510,6 +510,92 @@ jobs:
         with:
           subject-path: flatpak-sources.json
 
+  # Builds installable .flatpak bundles the same way verify-flatpak.yml does:
+  # the shared offline overlay manifest (scripts/verify-flatpak/desktop-offline.yaml)
+  # over the Flathub packaging repo, compiling the tag checkout with the
+  # combined offline sources. Sideloadable via `flatpak install ./<file>` —
+  # the runtime resolves from Flathub via --runtime-repo, and a later Flathub
+  # install of the same app id supersedes the sideload. Note: this job pulls
+  # the runtime/SDK from Flathub at build time, so Flathub availability is a
+  # release-cut dependency; a failed cut is retried as a fresh dispatch.
+  build-flatpak-bundle:
+    if: ${{ inputs.build_flatpak_src }}
+    runs-on: ${{ matrix.arch == 'aarch64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
+    needs: [prepare-build-info, release-flatpak-src]
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [x86_64, aarch64]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7.0.0
+        with:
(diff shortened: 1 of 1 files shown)


jamesarich:
@coderabbitai review

jamesarich:
Closing in favor of automating the Flathub manifest bump on production promotions (the actual latency gap — see discussion). Sideloaded bundles don't auto-update and still pull the runtime from Flathub, while the AppImage (#6246) covers the portable niche; not worth two heavy Flathub-dependent jobs on every internal cut. The verify-flatpak artifacts remain available for one-off bundle needs.

PR #6255, opened 2026-07-14
fix(database): reduce write-lane pressure from one-shot reads

jeremiah-k:
## Overview

This reduces unnecessary database coordination for bounded, one-shot reads without weakening database switching, writer admission, eviction, or shutdown safety.

Routine point queries previously entered the same `DatabaseManager.withDb` path used for coordinated writes. That serialized them through the temporary write-containment lane and refreshed last-used metadata for every operation, even though these reads do not mutate state.

This introduces a dedicated bounded-read path while preserving the existing coordination rules for writes, Flow and Paging subscriptions, database switching, association, and lifecycle-sensitive operations.

Builds on: #6256
Related: #6259

## Changes

* Added `DatabaseProvider.withReadDb` for bounded, non-replaying reads.
* Kept bounded reads outside writer admission and the serialized write-containment lane.
* Registered each read before capturing the published database.
* Tracked readers against the exact captured database instance.
* Prevented shutdown from closing pools beneath admitted read callbacks.
* Prevented LRU eviction from closing a switched-away pool while a reader or writer still owns it.
* Retried deferred eviction after the final database owner releases the pool.
* Routed device-hardware, firmware-release, node-info, and device-link point queries through the bounded-read path.
* Kept Flow and Paging consumers attached to `currentDb` so they continue to re-subscribe after a database switch.
* Removed per-operation last-used writes from the common access path while preserving updates at meaningful lifecycle boundaries.
* Preserved read exceptions rather than silently substituting empty, false, null, or zero fallback values.

## Validation

Added coverage for:

* bounded reads without writer admission;
* database capture across active-database switches;
* no automatic replay after read failures;
* cancellation-safe reader release;
* shutdown waiting for an admitted bounded read;
* rejection of new reads once shutdown begins;
* eviction deferral for a switched-away database with an active reader;
* deferred eviction after the reader completes;
* unchanged Flow and Paging relatching behavior.

Hardware validation switched between firmware 2.8 over TCP and firmware 2.7 over BLE in one process. Both configuration downloads completed, and the active database followed the selected device.

## Summary by CodeRabbit

* **Improvements**
  * Improved database read handling during switching, shutdown, and cache eviction using bounded read tracking.
  * Reads now complete against a consistent database instance without automatic retry/replay on failures.
  * Local data sources now execute queries in read-only mode and return results directly.
* **Documentation**
  * Clarified read/write access policy and shutdown behavior in database provider docs.
* **Tests**
  * Added coverage for bounded reads, cleanup/eviction coordination, cancellation, cache-limit eviction selection, and read failures.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceHardwareLocalDataSource.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceHardwareLocalDataSource.kt
@@ -33,16 +33,16 @@ class DeviceHardwareLocalDataSource(private val dbManager: DatabaseProvider) {
     }
 
     suspend fun getByHwModel(hwModel: Int): List<DeviceHardwareEntity> =
-        dbManager.withDb { it.deviceHardwareDao().getByHwModel(hwModel) }.orEmpty()
+        dbManager.withReadDb { it.deviceHardwareDao().getByHwModel(hwModel) }
 
     suspend fun getByTarget(target: String): DeviceHardwareEntity? =
-        dbManager.withDb { it.deviceHardwareDao().getByTarget(target) }
+        dbManager.withReadDb { it.deviceHardwareDao().getByTarget(target) }
 
     suspend fun getByModelAndTarget(hwModel: Int, target: String): DeviceHardwareEntity? =
-        dbManager.withDb { it.deviceHardwareDao().getByModelAndTarget(hwModel, target) }
+        dbManager.withReadDb { it.deviceHardwareDao().getByModelAndTarget(hwModel, target) }
 
-    suspend fun hasAnyEntries(): Boolean = dbManager.withDb { it.deviceHardwareDao().count() > 0 } ?: false
+    suspend fun hasAnyEntries(): Boolean = dbManager.withReadDb { it.deviceHardwareDao().count() > 0 }
 
     /** All known `platformioTarget` values — used to determine which msh.to links are vendor links. */
-    suspend fun getAllTargets(): List<String> = dbManager.withDb { it.deviceHardwareDao().getAllTargets() }.orEmpty()
+    suspend fun getAllTargets(): List<String> = dbManager.withReadDb { it.deviceHardwareDao().getAllTargets() }
 }
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceLinkLocalDataSource.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceLinkLocalDataSource.kt
@@ -29,7 +29,7 @@ class DeviceLinkLocalDataSource(private val dbManager: DatabaseProvider) {
     fun observeAll(): Flow<List<DeviceLinkEntity>> =
         dbManager.currentDb.flatMapLatest { db -> db.deviceLinkDao().observeAll() }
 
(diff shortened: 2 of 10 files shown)


jamesarich:
Nice find on the per-operation `markLastUsed` — that was a DataStore disk write plus a coroutine launch on every DB op, and removing it from the hot path is a real win. The removal is also safe for eviction correctness: `enforceCacheLimit` excludes the active DB by name, and `switchActiveDatabase` re-stamps both the new and previous DB, so LRU ordering stays accurate without per-op stamps.

One sequencing concern before this merges, though: **this PR is only safe on top of #6256, and #6256 is still in draft.**

The reads moved from `withDb` to `currentDb.value.dao()` lose their registration in `activeWriters`. On current `main`, the `associateDevice` merge path calls `drainWriters(source)` and then `retireDatabase(sourceName)`, which **physically closes and deletes** the source DB. A raw `currentDb.value` read is invisible to that drain, so a read that captures the source instance around the swap can still be running when retirement closes it — the exact "Connection pool is closed" failure class this series is eliminating. Two things widen the window:

1. The moved reads (device-hardware and firmware-release lookups) fire at connect time — precisely when `associateDevice` merges fire.
2. `currentDb` is a derived flow (`_currentDb.filterNotNull().stateIn(...)`), so `currentDb.value` can briefly lag the publish and hand out the stale source instance even after the swap.

#6256's logical retirement (never physically closing during process lifetime) removes the hazard entirely. Could we either land #6256 first and rebase this on top, or merge them together? If there's a reason to take this one alone, it would at least be worth noting the dependency in the PR body.

jeremiah-k:
@jamesarich Thanks — agreed on the sequencing concern.

The direct `currentDb` reads intentionally stop participating in writer admission, so this should not land while `main` still physically closes a merged source database. I’ll treat #6255 as dependent on #6256 rather than adding a second reader-drain mechanism here, which would duplicate the lifecycle work and broaden this otherwise small PR.

I’ll return this to draft, update the description to state that dependency, and rebase it onto `main` after #6256 lands. The per-operation `markLastUsed` removal remains independently valid, but the complete PR will merge after the lifecycle hardening and rerun CI on that base.

jamesarich:
Reviewed the evolved read path in depth — `withReadDb` is a solid design and it resolves everything I raised on the original version of this PR:

- Readers register in `activeReaders` atomically under the same lock the eviction filter holds, so eviction defers a pool with an in-flight read instead of closing it (and `endRead` retries the eviction when the last accessor leaves) — the stale-instance-vs-eviction window is closed.
- Reads count as manager operations, so `close()` drains them before closing pools.
- With `currentDb` now the raw `MutableStateFlow`, the derived-flow lag I flagged is gone entirely.
- Reads skipping the writer gate is the right trade — documented, and torn reads during association are acceptable for these point-in-time queries.
- The `markLastUsed` removal stays safe: current-DB exemption plus switch-time stamps preserve LRU ordering.

No defects found. Four test gaps worth covering while the machinery is fresh — `DatabaseManagerReadTest` covers stale-capture, shutdown, eviction-defer, and failure propagation, but not:

1. a read concurrent with an **association/merge** (gate armed) — proving reads aren't blocked and survive source retirement;
2. reader **cancellation** releasing the refcount and any deferred eviction;
3. **concurrent readers** on one pool (refcount > 1);
4. an assert that a victim can't acquire a new reader after the eviction filter has passed it (the lock ordering argues it's impossible — worth pinning with a test).

None of these block merge for me — happy to see this land after them, or with them as an immediate follow-up.

jeremiah-k:
@jamesarich Agreed—this is a good time to add the coverage while the concurrency behavior is fresh. I’ve added all four cases you identified; they’re now included in the updated branch.

PR #6271, opened 2026-07-14
fix(connections): show device long names in the connection manager (#5808)

### LINKED ISSUE #5808: [Feature Request]: Long device names in Bluetooth connection manager
### Checklist

### Contact Details

kahlerd@gmail.com

Would it be possible to also display device long names in the Bluetooth connection manager?

I have multiple devices with similar short names but unique long names. Would it be possible to also display the long names in the connection manager to make picking the device easier?

### Why should this be added?

Would ease selection of devices with similar short names.

_No response_

jamesarich:
## Why

Closes #5808. A user with several devices whose **short** names look alike (`Meshtastic_ab12`, `Meshtastic_ab13`, …) can't tell them apart in the Bluetooth connection manager, even though each has a distinct **long** name.

Digging in, this was two problems, not one:

1. **The long name was hidden even when we already had it.** The BLE *advertisement* only carries the short name; the long name is a mesh-identity attribute (`node.user.long_name`) learned after the first connection. For a previously-connected device (`node != null`) we hold that unique long name — but `DeviceHeadline` replaced the whole row headline with a tiny short-name-only `NodeChip` and threw the long name away.
2. **Advertised names truncated.** The headline was `maxLines = 1` + ellipsis, with the trailing RSSI/radio eating horizontal width, so long names were cut off (the literal symptom in the issue title).

## 🐛 / 🌟 Changes

- **Show the long name** (`node.user.long_name`) next to the `NodeChip` for known devices, falling back to the advertised name when the long name is empty. This is the core of #5808 — for the common "my devices" case (already paired/connected), the distinguishing field is now on screen.
- **Wrap instead of truncate:** the device name is now allowed `maxLines = 2` for both the known-node and advertised-name paths, so long names stay legible rather than being ellipsized at one line. The name column takes weighted width so it no longer competes with the RSSI/radio.
- Extracted a small `DeviceName` helper so both headline paths render identically.
- Added a `DeviceListItemWithLongNamePreview` (`@PreviewLightDark`) covering the known-node + long-name branch.

Only the row headline changed; the leading transport/connection icon, address supporting line, RSSI bars, and selection radio are untouched.

### Scope / follow-ups (intentionally not in this PR)
While reviewing the picker I noted a few things worth separate issues rather than expanding this change: surfacing `bonded`/pair state on the row, a clearer "currently connected" cue on non-selected rows, and an adaptive two-pane layout of the connections screen for tablet/foldable.

## Testing Performed

- `./gradlew spotlessCheck detekt assembleDebug test allTests` — all green locally.
- New `DeviceListItemWithLongNamePreview` for visual verification. Live PNGs weren't rendered here (the in-repo Compose→PNG path needs a running Android Studio, which wasn't available in this environment); a reviewer can render the preview, or I can attach before/after screenshots.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device list rendering for known mesh devices by showing the identifying chip alongside the full long name.
  * Long device names now wrap onto up to two lines for clearer readability.
  * Improved fallback device name display when mesh identity details aren’t available.

* **Tests**
  * Added a new UI preview to verify device list layout for long device names and wrapping behavior.

### DIFF
--- a/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/component/ConnectionsPreviews.kt
+++ b/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/component/ConnectionsPreviews.kt
@@ -30,6 +30,7 @@ import org.meshtastic.core.ble.BleConnectionState
 import org.meshtastic.core.ble.BleDevice
 import org.meshtastic.core.model.ConnectionState
 import org.meshtastic.core.model.DeviceType
+import org.meshtastic.core.model.Node
 import org.meshtastic.core.ui.icon.MeshtasticIcons
 import org.meshtastic.core.ui.icon.Search
 import org.meshtastic.core.ui.theme.AppTheme
@@ -43,6 +44,7 @@ import org.meshtastic.feature.connections.ui.components.DeviceSectionHeader
 import org.meshtastic.feature.connections.ui.components.DisconnectButton
 import org.meshtastic.feature.connections.ui.components.EmptyStateContent
 import org.meshtastic.feature.connections.ui.components.TransportSelector
+import org.meshtastic.proto.User
 
 private const val PREVIEW_BLE_RSSI = -60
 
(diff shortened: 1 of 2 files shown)


jamesarich:
Addressed the `ifEmpty` → `ifBlank` nitpick in fb056da3a, so a whitespace-only long name falls back to the advertised device name instead of rendering an empty gap.

PR #6281, opened 2026-07-15
feat(analytics): restore RUM view tracking on Nav3 (Google flavor)

jamesarich:
## Why

Android RUM sessions currently contain **no screens**. `RumConfiguration.Builder` sets no view-tracking strategy and there are no `startView`/`stopView` calls anywhere in the repo, so a session is just errors, ANRs, and the single custom `connect` action. Meanwhile Meshtastic-Apple auto-tracks views (and actions) via `DefaultSwiftUIRUMViewsPredicate`, so the cross-platform RUM comparison the 100% sampling exists for is apples-to-oranges.

This is also partly a **regression**: shipped builds ≤ 2.7.13 tracked views (RUM Resource spans carried `view.name: org.meshtastic.core.navigation.<X>Route.<Screen>`). The wiring — Datadog's `NavigationViewTrackingEffect` in the old `GooglePlatformAnalytics` — was dropped during the Navigation 3 migration (view tracking last present before #4735 removed it, no longer applicable once the app moved off Jetpack Navigation).

`NavigationViewTrackingEffect` (from `dd-sdk-android-compose`) targets **Jetpack Navigation** and cannot observe our Nav3 backstack (`MeshtasticNavDisplay` / `org.jetbrains.androidx.navigation3`), so the drop-in helper can't come back. This PR hooks view tracking into the Nav3 backstack manually instead.

## What

Restores per-screen RUM views on the Google flavor, wired through the existing `PlatformAnalytics` seam:

- **`PlatformAnalytics`** (commonMain) gains `startScreenView(key, name)` / `stopScreenView(key)` with default no-op bodies — commonMain stays free of any Datadog import.
- **`MeshtasticNavDisplay`** (Nav3 host, core/ui) observes the top of the active backstack and drives start/stop via a `DisposableEffect` keyed on the current destination. It takes an optional `analytics: PlatformAnalytics?` that defaults to `null`, so the intro flow and any non-app host stay untracked.
- **`GooglePlatformAnalytics`** implements the hook via `GlobalRumMonitor.get().startView/stopView`, guarded exactly like the existing `trackConnect`.
- **F-Droid / desktop** inherit the no-op defaults and emit nothing.
- View names are the route's fully-qualified class name (`NavKey.rumViewName()`), matching the historical convention so new data lines up with existing dashboards. *(Note: the sealed route types are named `<X>Route` today, so names are e.g. `org.meshtastic.core.navigation.NodesRoute.Nodes`; pre-migration they were `<X>Routes` — same package/screen, singular vs plural is the only drift and renaming route types was out of scope.)*

### 🌟 Features
- Per-screen RUM view tracking on Android (Google flavor), restoring cross-platform parity with iOS.

### 🛠️ Improvements
- View names derived from the Nav3 route FQN, aligned with pre-migration Datadog data.

## Out of scope (deliberate)

- **Action tracking.** iOS auto-tracks actions; Android keeps `trackFrustrations(false)` and only the custom `connect` action. `dd-sdk-android-compose` has **no** automatic action instrumentation for Compose (only manual `trackClick` / `TrackInteractionsEffect`), so adding it means annotating every interactive control — a much larger, noisier change. Views are the primary gap; actions can be a follow-up.

## Testing Performed

- `spotlessCheck` + `detekt` green across all touched modules (`core:repository`, `core:navigation`, `core:ui`, `androidApp`).
- `assembleDebug` green — both `googleDebug` and `fdroidDebug` compile; F-Droid inherits the no-op defaults and emits nothing.
- `:core:navigation:allTests` green — new `RumViewNameTest` (commonTest) pins the FQN view-name format for a `data object` and an argument-bearing `data class` route.
- `kmpSmokeCompile` green — confirms the common/iOS/JVM compile paths (the nav-display change lives in `commonMain`).
- CI runs the full `test allTests` matrix across every module.
- Not yet confirmed end-to-end in the Datadog RUM UI (US5, service `com.geeksville.mesh`) — that needs the Datadog org dashboard. The wiring mirrors the pre-migration implementation and reuses the existing `connect`-action init guards; debug builds run Datadog in developer mode (`startView`/`stopView` log to logcat) for local confirmation on an emulator.

## Summary by CodeRabbit

* **New Features**
  * Added automatic screen-view tracking for navigation destinations when analytics is enabled.
  * Integrated RUM-style start/stop screen tracking via the platform analytics interface.
  * Screen view names are generated consistently from the navigation route.
* **Bug Fixes**
  * Ensured tracked screen views end properly when navigating away to avoid stale tracking.
* **Tests**
  * Added coverage validating stable analytics names across navigation routes and route parameters.

### DIFF
--- a/androidApp/src/google/kotlin/org/meshtastic/app/analytics/GooglePlatformAnalytics.kt
+++ b/androidApp/src/google/kotlin/org/meshtastic/app/analytics/GooglePlatformAnalytics.kt
@@ -265,6 +265,16 @@ class GooglePlatformAnalytics(private val context: Context, private val analytic
         GlobalRumMonitor.get().addAction(RumActionType.CUSTOM, "connect", attributes)
     }
 
+    override fun startScreenView(key: String, name: String) {
+        if (!Datadog.isInitialized() || !GlobalRumMonitor.isRegistered()) return
+        GlobalRumMonitor.get().startView(key = key, name = name)
+    }
+
+    override fun stopScreenView(key: String) {
+        if (!Datadog.isInitialized() || !GlobalRumMonitor.isRegistered()) return
+        GlobalRumMonitor.get().stopView(key = key)
+    }
+
     private val isGooglePlayAvailable: Boolean
         get() =
             GoogleApiAvailabilityLight.getInstance().isGooglePlayServicesAvailable(context).let {
--- a/androidApp/src/main/kotlin/org/meshtastic/app/ui/Main.kt
+++ b/androidApp/src/main/kotlin/org/meshtastic/app/ui/Main.kt
@@ -30,6 +30,7 @@ import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.navigation3.runtime.NavKey
 import androidx.navigation3.runtime.entryProvider
 import co.touchlab.kermit.Logger
+import org.koin.compose.koinInject
 import org.koin.compose.viewmodel.koinViewModel
 import org.meshtastic.app.BuildConfig
 import org.meshtastic.core.model.ConnectionState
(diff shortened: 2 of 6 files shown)


jamesarich:
@coderabbitai Addressed the `DisposableEffect` nitpick in `MeshtasticNavDisplay.kt` — `analytics` is now passed as an effect key alongside `currentKey` (commit 0526c60). It's a stable Koin singleton, so this won't cause spurious restarts.

PR #6285, opened 2026-07-15
fix(connections): use a segmented button for transport selection

jamesarich:
The Connection screen's BLE / TCP / USB transport control was three independent `FilterChip`s. Community feedback flagged it as confusing: the selected chip's **filled** state was read as *"this transport is enabled/available"* rather than *"this transport is selected"* — the opposite of what it means — because a lone filled chip looks like a state toggle, not a single-choice picker.

This swaps it for a Material 3 [`SingleChoiceSegmentedButtonRow`](https://developer.android.com/reference/kotlin/androidx/compose/material3/SingleChoiceSegmentedButtonRow.composable), the canonical M3 control for picking one of a small set of related options.

### 🐛 Bug Fixes
- The three transports now read as **one grouped "pick one" control** (shared outline), so filled-vs-unfilled is no longer ambiguous.
- The **selected** segment shows a checkmark — M3's standard single-choice indicator — while **unselected** segments keep their transport icon so they still communicate what they are.

- `TransportSelector` reworked from a `FlowRow` of `FilterChip`s to a `SingleChoiceSegmentedButtonRow` of `SegmentedButton`s. Full-width, equal-weight segments so the control spans the same width as the connection card above it.
- Public API (`activeTransport: DeviceType`, `onSelectTransport: (DeviceType) -> Unit`) and `ScannerViewModel` selection logic are **unchanged** — this is purely presentational. No string or icon resources were touched.
- `TransportSelectorPreview` now renders inside a bounded width so its screenshot reference reflects the real full-width layout (a `fillMaxWidth` control renders content-sized under the unbounded default preview constraint).

### Screenshots

| Theme | Before | After |
|-------|--------|-------|
| Light |  |  |
| Dark |  |  |

### Testing Performed
- Full local baseline — `spotlessApply spotlessCheck detekt`, `assembleFdroidDebug`, `test`, `allTests` — all green.
- `:screenshot-tests:validateDebugScreenshotTest` passes against the regenerated `ScreenshotTransportSelector` goldens (light + dark), which are committed in this PR.
- No unit tests required changes — the component's API and the `ScannerViewModel` behavior are unchanged.

## Summary by CodeRabbit

* **UI Improvements**
  * Replaced the transport filter chips with a cohesive segmented control.
  * The selected transport is now indicated with a checkmark, while other options retain their transport icons.
  * The control now spans the available width for a more consistent layout.
  * Updated the preview to better reflect the Connections screen’s spacing and sizing.


roberthadow:
@jamesarich  Thank you.  My search terms were not robust enough to identify this duplicate item. Apoogies.

PR #6318, opened 2026-07-19
fix(messaging): reduce composer recomposition and typing jank

jeremiah-k:
## Overview

This reduces message-composer recomposition and per-keystroke work, particularly on active meshes where telemetry changes frequently and inline Markdown is enabled.

The composer previously received full `Node` objects, so updates to telemetry, position, metrics, last-heard time, or other unrelated fields could invalidate the input subtree even when mention-visible identity data had not changed. The live editor also performed more parsing, matching, and draft-persistence work than necessary during rapid typing.

This keeps stored and transmitted message text unchanged while narrowing reactive inputs and bounding mention matching, live styling, and draft persistence.

### Stable composer inputs

- Project nodes into lightweight mention candidates containing only the ID and display names used by the composer.
- Expose those candidates to `MessageInput` as a persistent `ImmutableMap`, giving Compose an explicit stability contract.
- Derive the candidate map with structural equality so telemetry-only node updates can leave the input subtree skippable instead of recreating its output transformation or platform input connection.

### Bounded typing work

- Cache normalized mention fields once per candidate and normalize each query once.
- Preserve source iteration order while stopping matching at the suggestion limit.
- Fast-path empty queries and plain text.
- Replace full GFM parsing in the live text field with lightweight detection for asterisk bold/italic, strikethrough, and inline code.
- Intentionally leave underscore emphasis raw while editing; sent message bubbles continue to use the full Markdown renderer.
- Debounce draft persistence to `SavedStateHandle` by 300 ms, cancel superseded writes, and keep send and explicit clear operations immediate.

### Presentation-only mentions and Markdown

- Parse Markdown ranges from the immutable source buffer before applying friendly-name substitutions.
- Remap style boundaries through shorter or longer mention replacements.
- Prevent Markdown delimiters in display names from introducing formatting absent from the stored text.
- Ignore bold, italic, and strikethrough delimiters inside inline-code spans while preserving outer emphasis around nested code.
- Preserve raw mention tokens and Markdown delimiters in stored and transmitted messages.

## Review guide

The commits are intended to be reviewed independently and in order:

1. `fix(messaging): stabilize composer lifecycle and hot paths`
   - stable mention projection;
   - bounded matching and live styling;
   - debounced, cancellation-safe draft persistence.
2. `fix(messaging): preserve markdown styling through mentions`
   - source-first presentation planning;
   - style-range remapping through mention substitutions;
   - delimiter and inline-code isolation.

## Regression coverage

Coverage includes:

- case-insensitive mention matching across IDs and display names;
- deterministic ordering and bounded early termination;
- plain-text and incomplete-Markdown fast paths;
- bold, italic, strikethrough, and inline-code styling;
- bold delimiters not being double-counted as italic;
- Markdown ranges before and after multiple mention replacements;
- friendly names containing Markdown delimiters;
- Markdown delimiters inside code spans;
- emphasis surrounding nested code;
- exact trailing-edge draft-persistence timing;
- rapid edits replacing the pending debounce window;
- clearing a draft cancelling pending persistence.

## On-device validation

On-device testing was performed while connected to an active mesh with ongoing telemetry updates.

Normal typing is noticeably more responsive, and the earlier focus, keyboard, selection, mention-suggestion, and live-formatting instability was not reproduced during ordinary input.

Holding backspace can still produce some stutter. That appears to be a narrower sustained-edit performance issue and will be investigated separately rather than expanding this PR beyond its current scope.

## Scope

This is limited to the message-composer hot path. It does not change transport, BLE scanning or reconnect behavior, database coordination, node identity, or application lifecycle behavior.

## Summary by CodeRabbit

- **New Features**
  - Improved message mentions with faster, more reliable autocomplete and clearer display names.
  - Mention matching now works case-insensitively across IDs and names.
  - Inline Markdown formatting is preserved when mentions are inserted or replaced.

- **Bug Fixes**
  - Draft messages now save smoothly after brief editing pauses, reducing unnecessary updates.
  - Clearing a draft immediately prevents pending changes from being restored.

### DIFF
--- a/feature/messaging/build.gradle.kts
+++ b/feature/messaging/build.gradle.kts
@@ -34,6 +34,7 @@ kotlin {
 
             implementation(libs.androidx.paging.common)
             implementation(libs.androidx.paging.compose)
+            implementation(libs.kotlinx.collections.immutable)
 
             // JetBrains Material 3 Adaptive (multiplatform ListDetailPaneScaffold)
             implementation(libs.jetbrains.compose.material3.adaptive)
--- a/feature/messaging/src/commonMain/kotlin/org/meshtastic/feature/messaging/Message.kt
+++ b/feature/messaging/src/commonMain/kotlin/org/meshtastic/feature/messaging/Message.kt
@@ -585,15 +585,88 @@ internal data class LiveStyleSpan(val range: IntRange, val style: InlineStyle)
  */
 internal fun liveInlineMarkdownStyleRanges(source: String): List<LiveStyleSpan> {
     if (!source.any { it in LIVE_STYLE_DELIMITER_SET }) return emptyList()
-    return buildList {
-        LIVE_BOLD.findAll(source).forEach { add(LiveStyleSpan(it.groups[1]!!.range, InlineStyle.Bold)) }
-        LIVE_ITALIC.findAll(source).forEach { add(LiveStyleSpan(it.groups[1]!!.range, InlineStyle.Italic)) }
-        LIVE_STRIKE.findAll(source).forEach { add(LiveStyleSpan(it.groups[1]!!.range, InlineStyle.Strikethrough)) }
-        LIVE_CODE.findAll(source).forEach { add(LiveStyleSpan(it.groups[1]!!.range, InlineStyle.Code)) }
+    val codeMatches = LIVE_CODE.findAll(source).toList()
+    // Returns true when either delimiter boundary of this match falls inside a code span. Markdown entirely inside
+    // code, or malformed markdown crossing a code boundary, must not add styling. A code span nested inside outer
+    // emphasis is still allowed, so outer bold text can retain both its bold span and an inner code span.
+    val isBlockedByCodeSpan: (MatchResult) -> Boolean = { match ->
+        codeMatches.any { codeMatch -> match.range.first in codeMatch.range || match.range.last in codeMatch.range }
     }
+    return buildList {
+        LIVE_BOLD.findAll(source).filterNot(isBlockedByCodeSpan).forEach {
+            add(LiveStyleSpan(it.groups[1]!!.range, InlineStyle.Bold))
+        }
+        LIVE_ITALIC.findAll(source).filterNot(isBlockedByCodeSpan).forEach {
+            add(LiveStyleSpan(it.groups[1]!!.range, InlineStyle.Italic))
+        }
+        LIVE_STRIKE.findAll(source).filterNot(isBlockedByCodeSpan).forEach {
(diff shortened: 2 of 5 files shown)


jamesarich:
Went through this in detail — the correctness story is solid: raw text in the buffer is what's stored and transmitted (`OutputTransformation` is display-only), so the wire format and iOS parity are untouched; the live-styling regexes are index-safe on unterminated/nested/multibyte input; rename propagation works through the derived candidate map; the span remapping math checks out through short/long replacements; and parsing before replacement closes the delimiter-injection hole. The cached normalization and bounded lazy matching do what the description says, with deterministic ordering. Nice work.

Four smaller items, none blocking:

1. **The headline recomposition claim is only partially realized.** `MessageScreen` still reads `nodes` directly, so it recomposes on every telemetry tick — and the candidate map is a plain `Map`, which is unstable to the Compose compiler, so `MessageInput` is compared by reference and re-executes anyway (cheap re-execution thanks to the inner memoization, but not the skipping the description implies). Passing a `kotlinx.collections.immutable.ImmutableMap` (or an `@Immutable` wrapper) would make the claim fully true for one line of change.

2. **Device validation.** The Validation section lists unit coverage only — for a jank PR I'd like the usual on-device pass (busy mesh, telemetry churn, inline Markdown on) before merge, even though the risk is low with the wire format untouched.

3. **PR body nit:** "Removed the ineffective `onCleared()` draft write" — there's no `onCleared` at the merge base or on main, and nothing of the sort in this diff. Worth correcting so the changelog doesn't record a phantom change.

4. **Cosmetic, worth a KDoc note:** the live detectors cover `*`, `**`, `~~`, and backticks but not underscore emphasis, so `_x_`/`__x__` previews unstyled in the composer while styling correctly in the sent bubble. Fine as a trade — just document it at `liveInlineMarkdownStyleRanges` so it doesn't come back as a bug report. (Also: `MessageComposerPerformanceTest` asserts correctness, not performance — a rename would keep expectations honest.)

With #2 done (and ideally #1), this is ready.

jeremiah-k:
@jamesarich I’ve addressed the earlier review feedback and completed another on-device pass. Normal typing is noticeably improved. There is still some stutter when holding backspace, which I plan to investigate further as part of follow-up lifecycle work, but I don’t think that should block this PR in its current scope.

During final testing I also encountered BLE reconnect and scan-lifecycle failures. Those issues are unrelated to the composer changes and are addressed separately in #6384, so there is no code or merge-order dependency between the two PRs.

jamesarich (approved):
Appreciate it - thank you!

PR #6319, opened 2026-07-19
fix: normalize F-Droid QR camera frames

RCGV1:
## Summary
- Normalize CameraX luminance rows before passing them to the F-Droid ZXing decoder.
- Add QR protocol vectors for replace, add, and shared-contact payloads.
- Add scanner regression coverage for both Android flavors.

## Root cause
The F-Droid analyzer treated the CameraX Y-plane buffer as tightly packed. CameraX can provide a rowStride wider than the visible image width, so the old decoder read padding bytes as pixels and corrupted the QR image. This affects any QR content, including codes rendered by the Apple app. The Google flavor uses ML Kit and is separately verified unchanged.

## Evidence
- FdroidQrDecoderTest renders a physical QR containing the Apple-compatible channel URL https://meshtastic.org/e/#CgMSAQESBggBQANIAQ, places every row in an 11-byte-padded camera buffer, then decodes the original URL successfully.
- QrInteroperabilityTest verifies canonical replace preserves LoRa, canonical add removes LoRa, and shared-contact CAE= round-trips.
- Passed:
  ./gradlew :core:barcode:testFdroidDebugUnitTest :core:barcode:testGoogleDebugUnitTest :core:model:jvmTest :core:barcode:spotlessCheck :core:model:spotlessCheck

No UI layout changes: this fixes the F-Droid camera decode path and adds protocol regression coverage.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved QR code scanning reliability for camera frames with padded luminance rows.
  - Ensured Apple-compatible channel URLs can be decoded correctly from supported camera input.

- **Tests**
  - Added coverage for QR decoding with camera buffers that include row padding, helping prevent regressions in barcode scanning.

### DIFF
--- a/core/barcode/src/fdroid/kotlin/org/meshtastic/core/barcode/BarcodeAnalyzerFactory.kt
+++ b/core/barcode/src/fdroid/kotlin/org/meshtastic/core/barcode/BarcodeAnalyzerFactory.kt
@@ -33,14 +33,14 @@ internal fun createBarcodeAnalyzer(onResult: (String) -> Unit): ImageAnalysis.An
 
     return ImageAnalysis.Analyzer { imageProxy ->
         try {
-            val buffer: ByteBuffer = imageProxy.planes[0].buffer
+            val lumaPlane = imageProxy.planes[0]
+            val buffer: ByteBuffer = lumaPlane.buffer.duplicate()
             val data = ByteArray(buffer.remaining())
             buffer.get(data)
 
             val width = imageProxy.width
             val height = imageProxy.height
-
-            val source = PlanarYUVLuminanceSource(data, width, height, 0, 0, width, height, false)
+            val source = PlanarYUVLuminanceSource(data, lumaPlane.rowStride, height, 0, 0, width, height, false)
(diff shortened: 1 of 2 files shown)


jamesarich:
The luminance/rowStride fix is the actual subject of this PR and it's fine. But the `shouldAdd` change in `core/model` is a separate fix in a separate module — URL fragment parsing has nothing to do with camera luminance. Unrelated changes bundled together are harder to review and harder to revert cleanly.

Please split the `shouldAdd`/`addParameter` change and its interop tests into their own PR. I'll review that one on its own. This PR should be just the barcode fix.

jamesarich (changes requested):
Please split the `shouldAdd`/`addParameter` change (and its interop tests) into its own PR — it's an unrelated fix in a different module. This PR should be scoped to just the F-Droid barcode luminance fix, which is good as-is. See inline comment for detail.

jamesarich (approved):
Scope addressed — this is now just the F-Droid barcode luminance fix (rowStride passed as dataWidth, native ZXing crop). CI green, CLA signed. LGTM.

PR #6320, opened 2026-07-19
fix: preserve add mode in legacy channel URLs

RCGV1:
## Summary
- parse legacy fragment query parameters by key, so `?source=ios&add=true` remains add mode
- preserve channel settings while stripping LoRa configuration for add imports
- cover all one-to-eight channel import sizes and legacy fragment URLs

Split from #6319 at maintainer request; this contains only the channel URL parser fix.

## Testing
- `:core:model:jvmTest`
- `:core:model:spotlessCheck`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved channel-set URL parsing to support multiple fragment parameters.
  * Recognized `add` flags with omitted values and case-insensitive values.
  * Preserved add-mode behavior without changing LoRa configuration.

* **Tests**
  * Added coverage for legacy URLs with additional parameters.
  * Verified channel settings and configuration are preserved across supported channel counts.


jamesarich (approved):
Clean split-out of the `shouldAdd` fix. `addParameter` correctly handles the `add` flag among multiple fragment params (e.g. `?source=ios&add=true`) where the old exact-match check silently fell through to replace-mode. Well covered by the multi-param and 1..8-channel round-trip tests. CI green. LGTM.

PR #6352, opened 2026-07-22
fix(map): retain camera across tab navigation

CatSu-OSM:
Switching away from the map recreated provider camera state, causing the map to reopen at a default or unrelated viewport. This retains the last camera across tab navigation while preserving the first-use behavior of fitting discovered nodes.

- Persist Google and F-Droid camera values atomically instead of treating temporary startup coordinates as saved state.
- Wait for persisted camera loading before initializing the map, preventing the default ocean-centered camera from winning startup races.
- Fit positioned nodes only when no camera has previously been saved.
- Add regression coverage for absent camera state, atomic restoration, and valid zero-valued coordinates.

## Testing Performed

- `./gradlew spotlessCheck detekt assembleDebug :core:prefs:jvmTest --tests org.meshtastic.core.prefs.map.MapPrefsImplTest :androidApp:testFdroidDebugUnitTest :androidApp:testGoogleDebugUnitTest --tests org.meshtastic.app.map.prefs.map.GoogleMapsPrefsTest`
- Verified F-Droid behavior on a physical Android device: pan/zoom, change tabs, and return to the same viewport.
- The repository-wide `spotlessApply spotlessCheck detekt assembleDebug test allTests --continue` invocation exceeded the 10-minute local command limit without reporting a map-related failure; the bounded cross-flavor matrix above completed successfully.

## Summary by CodeRabbit

* **New Features**
  * Map camera position is now restored when reopening the map (including target, zoom, tilt, and bearing where available).
  * When no saved camera exists, the map can automatically fit nodes after map data is ready.
* **Bug Fixes**
  * Improved camera initialization timing to avoid applying node bounds before nodes/maps are ready.
  * Camera saving now occurs at the right lifecycle points, including on effect disposal.
* **Tests**
  * Added/updated tests to verify camera position persistence and restore behavior, including the “absent before set” case.

### DIFF
--- a/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/MapView.kt
+++ b/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/MapView.kt
@@ -188,6 +188,13 @@ import org.meshtastic.proto.BoundingBox as ProtoBoundingBox
  */
 private class GeofenceOverlayPolygon : Polygon()
 
+private const val INITIAL_MAP_ZOOM = 1.5
+
+private fun MapView.saveCameraPosition(viewModel: MapViewModel) {
+    val center = mapCenter
+    viewModel.saveCameraPosition(center.latitude, center.longitude, zoomLevelDouble)
+}
+
 private fun MapView.updateMarkers(
     nodeMarkers: List<MarkerWithLabel>,
     waypointMarkers: List<MarkerWithLabel>,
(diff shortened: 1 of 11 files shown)


jamesarich (approved):
Approving. Solid camera-persistence refactor across both flavors.

The `Loading → FitNodes/Restored` state machine with the `!= Restored → return` save guard cleanly prevents the initial node-fit from clobbering a restored camera, and gating node-fitting on map-ready (`onMapLoaded` / `map.post {}`) avoids the `newLatLngBounds` layout crash. Nice touch preserving existing installs via the legacy Float→Double `ClassCastException` fallback, and good coverage on both prefs impls.

Two non-blocking notes:
- The fdroid `MapView` early-returns while `InitialCameraState.Loading`, so there is a blank frame until the camera loads from disk — fine in practice given DataStore is fast.
- CI is still queued at review time; enabling auto-merge so it lands once checks pass.

PR #6357, opened 2026-07-22
Fix ita

Angelk90:
Italian string fix: removed visible escapes `\"` that were incorrectly displayed in the permissions text.

Error:

### DIFF
--- a/core/resources/src/commonMain/composeResources/values-it/strings.xml
+++ b/core/resources/src/commonMain/composeResources/values-it/strings.xml
@@ -712,7 +712,7 @@
     <string name="paxcounter_config">Configurazione Paxcounter</string>
     <string name="paxcounter_enabled">Paxcounter abilitato</string>
     <string name="periodic_position_broadcast">Trasmissione periodica della posizione</string>
-    <string name="permission_missing_31">Meshtastic ha bisogno dei permessi \"Dispositivi nelle vicinanze\" abilitati per trovare e connettersi ai dispositivi tramite Bluetooth. È possibile disabilitare quando non è in uso.</string>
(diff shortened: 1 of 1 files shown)


jamesarich:
We generally prefer not to edit translations directly in the repository, as our strings are managed via Crowdin. Occasionally, Crowdin can struggle with Compose resource strings and their associated escaping. I will investigate whether we can automate the stripping of backslash escaping directly within the Crowdin workflow to prevent this issue from recurring.

Angelk90:
@jamesarich : These are the lines where the problem appears:


https://github.com/meshtastic/Meshtastic-Android/blob/bd25b1e6185c70066d6f2d2e9ec977055b4dfb26/core/resources/src/commonMain/composeResources/values-it/strings.xml#L199-L202

https://github.com/meshtastic/Meshtastic-Android/blob/bd25b1e6185c70066d6f2d2e9ec977055b4dfb26/core/resources/src/commonMain/composeResources/values-it/strings.xml#L715

https://github.com/meshtastic/Meshtastic-Android/blob/bd25b1e6185c70066d6f2d2e9ec977055b4dfb26/core/resources/src/commonMain/composeResources/values-it/strings.xml#L356

Angelk90:
@jamesarich :The last image also has punctuation formatting issues, such as commas and dot that shouldn't wrap. 
But these problems do occur in the app.

jamesarich:
@Angelk90 
I've added a post export sanitizer for escaped `\"` -> `"` which resolves the original issue of backslashes in translation strings.

Other spacing and punctuation will need to be edited via the Crowdin interface. https://crowdin.meshtastic.org/android

PR #6368, opened 2026-07-22
feat(messaging): redesign message bubbles toward the M3 conversation pattern

jamesarich:
## Why

The message screen had grown cramped and noisy: 8dp corners with hard-squared group edges, 8×2dp bubble interiors, and a metadata strip (time, delivery status, SNR/RSSI, hops, transport) jammed against the text inside every bubble. Hop and transport glyphs were also hardcoded to `Color.White`, making them invisible on light-theme bubbles. This audits the messaging UI against M3 conversation patterns (with Google Chat as the reference design) while keeping every Meshtastic-specific status affordance.

### 🌟 New
- **Timestamps move to the group header**: received runs show `node chip · sender name · time`; sent runs show a right-aligned muted time above the first bubble. Bubbles carry only content plus what is truly per-message.
- **Full timestamp in the long-press sheet**: since bubbles now only show a short header time per run, the message actions sheet gains a Timestamp row with the full date-time.
- **Time-windowed grouping**: same-sender runs now also break after a 10-minute gap, so the single header timestamp is always representative of every message beneath it.

### 🛠️ Improvements
- 18dp bubble corners; consecutive same-sender bubbles flatten only their facing corners to 4dp — conversation blocks stay visually marked, without the old hard squares.
- Mesh diagnostics collapse into one compact scrim-free row, ordered XEdDSA shield → transport → SNR/RSSI or hop count, so "how it arrived" reads before signal strength (follows #6367).
- Roomier 12×8dp bubble interiors with breathing room above the metadata row; reply-quote snippet padding matched.
- Sender name and timestamps de-emphasized to `onSurfaceVariant`; group rhythm is 10dp between runs, 2dp within.
- Bubbles cap at 480dp width with opposite-side margin, keeping sender sides scannable on tablets/desktop.

### 🐛 Fixes
- Hop-count and transport icons were tinted bare `Color.White` outside any scrim — illegible on light-theme bubbles. They now use `onSurfaceVariant`.
- Sender name next to the node chip had no explicit color and rendered near-black in dark mode.

### 🧹 Tests
- `MessageGroupingTest` pins the block rules: the 10-minute window boundary, sender changes, local-vs-remote runs, and out-of-order mesh arrival.
- A grouped-run preview + screenshot golden pins the block treatment end to end (header per run, 4dp facing corners, in-run vs between-run spacing, and a same-sender message past the gap starting a new block), driven by the real `isSameGroup` rule.

## Testing Performed

- `./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests` — all green, re-run after every commit (branch rebased onto main twice, most recently past #6370).
- `:screenshot-tests:updateDebugScreenshotTest` regenerated the six messaging goldens and added two grouped-run goldens; `validateDebugScreenshotTest` passes. Light/dark pairs visually reviewed at each step.
- Existing `MessageItemTest` assertions (delivery-status text, MQTT content description, a11y message semantics, status-click) pass unchanged.


jamesarich:
@coderabbitai review

PR #6369, opened 2026-07-22
fix(ui): add horizontal padding for position flags list item

dzmpr:
Fixed horizontal padding of position flags list item on Position settings screen to be consistent with other items in the list.

| Before | After |
|------|-----|
|  |  |

## Summary by CodeRabbit

* **Style**
  * Improved spacing within the exposed dropdown menu by using uniform padding for a more consistent, polished layout.
  * Simplified internal resource usage for button labels in the preference footer without changing the outward behavior or appearance.


jamesarich (approved):
small but nice, thanks

Issue #6317, opened 2026-07-18
[Bug]: Not visible second channel for remote node


mesb1:
### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [x] I actually read this list, and should be taken seriously.

### Affected app version

2.7.14 29321034 google

### Affected Android version

Android 13

### Affected phone model

Gigaset gx6, zte blade 20

### Affected node model

tdeck plus, heltec v4

### Affected node firmware version

2.7.26

### Steps to reproduce the bug

My every node had channels: LongFast default name and one with PIng name, both with AQ== 

Connect to remote node
Open channels section
Only one default channel visible, second is missing.

When you will try to add same existing channel, app will not do this, save button just in grey.
Please fix.

### Relevant log output

```shell

```

b8b8:
https://github.com/user-attachments/assets/4041773a-d514-4a3d-983f-3d932e37f96b

b8b8:
Affects 2.8 app latest internal build as well.

This has been an issue for a few months at least. Im not digging back to previous issues that may have covered it as this has the main points covered already.

b8b8:
The current work around is to press the cancel button as soon as you open the remote channel list page.

jamesarich:
This looks fixed by #6391 (merged 2026-07-23), which addresses exactly this symptom: the remote-channel editor was seeded once at first composition and never re-synced while the remote fetch streamed in the remaining channels — the footer "Cancel" (`replaceWith()`) was the accidental workaround. The editor now adopts the authoritative list while the fetch is still in flight.

Could you re-test on a build ≥ 2.8 that includes #6391 and confirm whether the second channel now shows without needing the Cancel workaround?

b8b8:
This is confirmed working for me with the latest android googleplay internal release build and the fdroid latest snapshot 2hours ago [Snapshot 29321611 (480334a0ad4c06e8ea879131557040647375499e)

PR #6391, opened 2026-07-23
fix(settings): stop dropping admin config responses (0% stall, missing remote channels)

### LINKED ISSUE #5592: [Bug]: Settings configuration hangs at 0% on Google Pixel 10 Pro (Bluetooth 6.0 / Android 16)
### Contact Details

rohitashs@gmail.com

### Checklist

### Affected app version

2.7.13 (29320069)

### Affected Android version

16

### Affected phone model

Google Pixel 10 Pro

### Affected node model

T-Echo, Station G2, Heltec V4

2.7.15

- Connect Device to Phone via bluetooth
- Go to settings and try to get to any of the Radio, Device or Module settings
- A dialog box pops up displaying 0% and stays there indefinitely. The actual settings do not get retrieved.

### Actual behavior

When connecting via Bluetooth or WiFi, maps/messages/node list all work correctly. Attempting to access ANY settings screen (Radio Configuration, Module Configuration, Channels) shows a 0% progress dialog that hangs indefinitely and never completes.

### Expected behavior

Clicking on any of the device related settings should fetch and display the settings and allow them to be changed.

Image

### Relevant log output

```shell

```

### Additional information

What works:
- USB serial connection via web client (client.meshtastic.org) - full settings access works perfectly on all devices

What was tried:
- Full uninstall and reinstall of app
- Rollback to 2.7.5
- Unpair all devices from Android Bluetooth, re-pair fresh through app
- All permissions granted (Location, Nearby Devices, Camera, Notifications)
- Battery optimization set to Unrestricted

jamesarich:
Firmware 2.8 (meshtastic/firmware#10967) delivers a locally connected node's admin responses through a synchronous loopback, so the response now reaches the phone **before** the `QueueStatus` ack for the request that produced it. The app registered each request's id for response correlation only **after** the ack-awaiting send returned — the response sailed through `meshPacketFlow` while the id set was still empty, was silently dropped by the correlation guard, and **every radio-config sub-screen stalled at a 0% loading overlay for 30s before a timeout dialog** against 2.8 firmware. Root-caused with a dual-ended capture (app logcat + firmware serial): the device handled `Get config: LoRa` and enqueued its reply to the phone in milliseconds; the app dropped it.

The same audit closed two adjacent holes behind #6317 (remote channel list shows only the first channel; the community workaround was "press Cancel"):

Likely also resolves #5592 (settings screens hang at 0% indefinitely) — same symptom, though that report is on 2.7.x firmware while the root cause fixed here is the 2.8 loopback-ordering race; worth confirming the reporter's stall clears before closing.

## 🐛 Fixes

- **Register request ids before the send is issued.** `RadioConfigUseCase` / `AdminActionsUseCase` request methods take an `onRequestId` callback invoked with the packet id ahead of the send; `RadioConfigViewModel` registers there instead of after the suspending call returns. Response ordering can no longer race registration, local or remote.
- **`QueueStatus.res = 35` is not a failure.** Firmware 2.8 returns its internal `ERRNO_SHOULD_RELEASE` ("delivered locally, packet consumed") for self-addressed packets; the handler treated any non-zero `res` as a failed send. Note: 35 numerically collides with `Routing.Error.PKI_UNKNOWN_PUBKEY` — `QueueStatus.res` carries `ErrorCode` semantics, not `Routing.Error` (diagnosis trap). The `handleQueueStatus` early return is scoped to the plain `res = 0` "accepted, now full" echo, so a `res = 35` delivery still completes its response even when the TX queue is full (`free = 0`) — otherwise it would re-introduce the same 5s-timeout stall under queue pressure.
- **Channel editor now shows channels as they stream in (#6317).** The editable list was seeded once, at first composition — i.e. the instant channel 0 arrived — and never re-synced while a remote fetch streamed the rest in; the footer *Cancel*'s `replaceWith()` was the accidental workaround. The editor adopts the authoritative list while the fetch is in flight (the loading overlay blocks edits during that window, so nothing can be clobbered).
- **Closed the request-chain gap.** `processPacketResponse` removed the completed request id inline, and could observe a momentarily-empty id set in the gap between chained `getChannel` requests — tearing the flow down (`clearPacketResponse`) and stranding the rest of the chain with a partial list. The removal now runs behind any chain continuation launched by the same response (FIFO on the main dispatcher; registration is the continuation's first act).
- **`NODEDB_RESET` marks an expected local restart.** Firmware reboots after a nodedb reset, but only `FACTORY_RESET` opened the restart window — a local nodedb reset surfaced the ensuing transport drop as a surprise disconnect instead of "restarting". `NODEDB_RESET` now calls `expectRestart()` for local resets, mirroring `FACTORY_RESET`.
- **A stale restart window can no longer mark an interrupted manual channel batch as "success".** A manual channel batch shares the save shape (empty route + `Loading`) that the restart-success paths key on. A local `MAY_RESTART` save that opened the 90s window but never actually rebooted left that window open; a transient disconnect or request timeout during a subsequent batch would then flip the *incomplete* batch to `Success`, silently misreporting a partial channel write. Both `completeRestartingSaveIfPending` and the request-timeout backstop now gate on `manualChannelBatchInFlight()`, which tracks the batch from enqueue through its ack-wait (past `finishManualChannelBatch`) via the pending batch request ids — the enqueue flag alone left the ack-wait window exposed.

## 🧹 Follow-up (firmware, not this PR)

`QueueStatus.res` leaking internal `ERRNO_*` values to clients is worth an upstream firmware issue — mapping `ERRNO_SHOULD_RELEASE` → 0 would spare every client the same trap.

## Testing Performed

- New unit tests: `getConfig invokes onRequestId with the packet id before issuing the send` (ordering regression guard); `handleQueueStatus treats ERRNO_SHOULD_RELEASE as success` / `other nonzero res as failure`; `handleQueueStatus completes ERRNO_SHOULD_RELEASE even when queue is full` (the `res = 35` + `free = 0` queue-full path — covered by test, since on-device runs didn't exercise a full TX queue); `NODEDB_RESET marks an expected local restart`; `manual channel batch is not completed as success by a stale restart window` (this one caught an insufficient first fix that guarded only the enqueue phase, not the ack-wait — both restart-window edge cases are test-covered rather than device-exercised).
- Full baseline `spotlessApply spotlessCheck detekt assembleDebug test allTests` green.
- On-hardware before/after (Pixel 6a ↔ Heltec DUT + M5Stack Cardputer, both on 2.8.0 develop firmware): before — LoRa/Security/Channels all stall at 0% → 30s timeout; after — LoRa loads in ~3s fully enabled, Channels runs the complete sequential fetch (4 chained requests) and renders every channel with the editor kept in sync throughout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved queue-status handling so a “should release” condition is treated as success.
  * Improved reliability completing radio/admin packet workflows and reducing request correlation timing issues.
* **New Features**
  * Added expected-restart tracking to smooth local reboot/reconnect behavior.
  * Added reboot-behavior support for configuration saves (including save-and-restart messaging).
* **User Experience**
  * Updated UI to clearly show “restarting” across connection/navigation and to prevent stale loading edits in channel config.
* **Tests**
  * Added/updated coverage for restart tracking, request-id ordering, and queue-status success/failure cases.

### DIFF
--- a/.skills/compose-ui/strings-index.txt
+++ b/.skills/compose-ui/strings-index.txt
@@ -252,6 +252,7 @@ config_position_broadcast_smart_minimum_interval_secs_summary
 config_position_flags_summary
 config_position_gps_update_interval_summary
 config_power_is_power_saving_summary
+config_saved_restarting
 config_security_admin_key
 config_security_debug_log_api_enabled
 config_security_is_managed
@@ -1149,6 +1150,7 @@ node_list_help_node_details
 node_list_help_title
 node_list_long_click_label
 node_number
+node_restarting
 node_sort_alpha
 node_sort_button
 node_sort_channel
@@ -1420,7 +1422,9 @@ rssi_definition
 rsyslog_server
 sample_message
 sats
+### SAVE ###
 save
+save_and_restart
 save_changes
 save_csv_in_storage_esp32_only
 save_rangetest
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConnectionManagerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConnectionManagerImpl.kt
@@ -51,6 +51,7 @@ import org.meshtastic.core.repository.MeshWorkerManager
 import org.meshtastic.core.repository.MqttManager
 import org.meshtastic.core.repository.NodeManager
 import org.meshtastic.core.repository.NodeRepository
+import org.meshtastic.core.repository.NodeRestartTracker
 import org.meshtastic.core.repository.PacketHandler
 import org.meshtastic.core.repository.PacketRepository
 import org.meshtastic.core.repository.PlatformAnalytics
@@ -94,6 +95,7 @@ class MeshConnectionManagerImpl(
     private val heartbeatSender: DataLayerHeartbeatSender,
     private val lockdownCoordinator: LockdownCoordinator,
     @Named("ServiceScope") private val scope: CoroutineScope,
+    private val nodeRestartTracker: NodeRestartTracker,
 ) : MeshConnectionManager {
     /**
      * Serializes [onConnectionChanged] to prevent TOCTOU races when multiple coroutines emit state transitions
@@ -139,6 +141,11 @@ class MeshConnectionManagerImpl(
 
         // Ensure notification title and content stay in sync with state changes
         serviceRepository.connectionState.onEach { updateStatusNotification() }.launchIn(scope)
+
+        // An expected node restart ends when the post-reboot config handshake completes.
+        serviceRepository.connectionState
+            .onEach { if (it == ConnectionState.Connected) nodeRestartTracker.onConnected() }
+            .launchIn(scope)
 
         scope.launch {
             try {
@@ -623,10 +630,16 @@ class MeshConnectionManagerImpl(
     }
 
     override fun updateStatusNotification(telemetry: Telemetry?) {
-        serviceNotifications.updateServiceStateNotification(
-            serviceRepository.connectionState.value,
-            telemetry = telemetry,
-        )
+        val state = serviceRepository.connectionState.value
+        // During an expected node restart the disconnect is transient; keep the persistent notification on the
+        // connecting presentation instead of flashing "disconnected".
+        val presented =
+            if (state == ConnectionState.Disconnected && nodeRestartTracker.restartExpected.value) {
+                ConnectionState.Connecting
+            } else {
+                state
+            }
+        serviceNotifications.updateServiceStateNotification(presented, telemetry = telemetry)
     }
 
     companion object {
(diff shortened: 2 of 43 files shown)


jamesarich:
**Remote admin verification (the #6317 scenario, on-hardware):**

Setup: Pixel 6a ↔ M5Stack Cardputer (local, BLE, fw 2.8.0.d9150e8) remote-administering a Heltec V3 (`!483ba531`, fw 2.8.0.8abae90) over LoRa via PKC admin.

Result with this branch: opening the remote **Channels** screen runs the full sequential fetch — `getChannel(0)` → `getLoraConfig` → `getChannel(1)` → `getChannel(2)` → `getChannel(3)`(DISABLED, terminates) — each request chained ~4s after the previous reply (LoRa RTT). All three channels (primary + two secondaries) stream into the editor as they arrive and the loading overlay clears cleanly at the end. No premature flow teardown, no 30s timeout, no partial list, and no need for the "press Cancel" workaround (which this audit traced to the footer Cancel's `replaceWith()` acting as an accidental refresh of the stale editor snapshot).

PR #6398, opened 2026-07-24
fix(android): avoid resource stalls in connection status

jeremiah-k:
## Overview

This avoids Compose resource-loading stalls in two connection-status paths that can overlap during normal Android lifecycle transitions.

The observed ANR had the Connections screen waiting in Compose resource loading while foreground-service telemetry formatting was resolving Compose resources at the same time. The UI lookup lived inside frequently re-entered animated content, while the service notification path used blocking non-composable resource helpers during foreground-service startup and telemetry updates.

The Connections UI now resolves stable connected-device display text outside the animated subtree, avoiding repeated entry into the blocking CMP resource state during animation frames. Service notifications return a cached or immediate application-label fallback for `startForeground`, then render localized status text asynchronously on the existing `ServiceScope`.

## Review Guide

The commits are intended to be reviewed in order:

1. **Leaf labels and UI composition** — allows signal and battery components to receive caller-resolved text and hoists stable connected-device display values out of animated content.
2. **Service notification foundation** — replaces blocking resource helpers with suspend formatting, an immediate fallback, and deterministic notification coverage.
3. **Structured test and failure handling** — contains fallback failures and keeps render work observable through structured test scopes.
4. **Ordered status rendering** — replaces generation/job coordination with one replaying `SharedFlow` collector, closes the last-writer window, and documents the temporary CMP workaround.

### Connection UI

* Allow `Rssi` and `MaterialBatteryInfo` to accept caller-provided labels while retaining localized defaults for other callers.
* Introduce an immutable `CurrentlyConnectedText` value for localized connected-device text.
* Resolve RSSI, unknown-battery, disconnect, and firmware-version text outside `AnimatedContent`.
* Keep these stable resource lookups out of connection-state animation transitions and deduplicate them across recompositions.
* Document the caller-provided label surface as a temporary CMP-6615 workaround for Compose Multiplatform 1.11.1, with an explicit removal condition.

### Service Notifications

* Use the Android application label or last rendered notification as an immediate `startForeground` fallback.
* Replace blocking resource helpers with suspend-based formatting on the established `ServiceScope`.
* Feed captured connection and telemetry snapshots through one replaying, drop-oldest `SharedFlow` collector.
* Use `collectLatest` to cancel obsolete resource work and serialize the complete render-and-notify action, preventing an older Binder post from overwriting newer notification text.
* Preserve the last telemetry message when newer state has no replacement metrics.
* Update cached state under `serviceNotificationLock`, then release the lock before posting through `NotificationManager`.
* Preserve cancellation and provide a safe fallback when localized rendering or notification posting fails.
* Clarify that `MeshService.startForeground` uses the cached or immediate fallback while localized rendering completes asynchronously.

## Testing

Added or updated coverage for:

* caller-provided RSSI and unknown-battery labels;
* immediate foreground-notification fallback behavior;
* deferred localized service-state rendering;
* latest-state selection when updates arrive before rendering completes;
* reposting a structurally identical state after notifications are cleared;
* valid oneof telemetry fixtures;
* notification-channel cleanup and active-notification behavior;
* conversation notification integration and structured render-scope cleanup.

Runtime validation repeatedly backgrounded, foregrounded, stopped, and restarted `MainActivity` while service telemetry continued to update. The original input-timeout and Compose resource-loading stall did not recur, and the capture contains no notification-render exception, ANR, or process restart.

## Migration Notes

* Notification channels and IDs are unchanged.
* Connection-state and telemetry semantics are unchanged.
* The caller-provided label parameters are temporary and can be removed after adopting a Compose Multiplatform release containing the CMP-6615 Android resource-loading fix.

## Summary by CodeRabbit

* **Improvements**
  * Foreground service notifications now update asynchronously while promptly displaying a fallback status.
  * Notification updates prioritize the latest connection state and preserve useful status details if rendering is delayed.
  * Connection details now support localized labels, optional firmware version display, and clearer battery and signal descriptions.
  * Signal strength and battery components support customized labels for improved contextual display.

* **Tests**
  * Added coverage for notification timing, state updates, reposting, and customizable connection labels.

### DIFF
--- a/core/service/src/androidHostTest/kotlin/org/meshtastic/core/service/MeshNotificationManagerImplConversationTest.kt
+++ b/core/service/src/androidHostTest/kotlin/org/meshtastic/core/service/MeshNotificationManagerImplConversationTest.kt
@@ -27,10 +27,14 @@ import dev.mokkery.every
 import dev.mokkery.everySuspend
 import dev.mokkery.matcher.any
 import dev.mokkery.mock
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.test.runTest
+import org.junit.After
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -68,6 +72,7 @@ class MeshNotificationManagerImplConversationTest {
     private val packetRepository: PacketRepository = mock(MockMode.autofill)
     private val nodeRepository: NodeRepository = mock(MockMode.autofill)
     private val radioConfigRepository: RadioConfigRepository = mock(MockMode.autofill)
+    private val notificationScope = CoroutineScope(Dispatchers.Unconfined + SupervisorJob())
 
     private val manager =
         MeshNotificationManagerImpl(
@@ -89,6 +94,7 @@ class MeshNotificationManagerImplConversationTest {
                 )
             },
             radioConfigRepository = lazy { radioConfigRepository },
+            scope = notificationScope,
         )
 
     @Before
@@ -107,6 +113,11 @@ class MeshNotificationManagerImplConversationTest {
                 ),
             )
         manager.initChannels()
+    }
+
+    @After
+    fun tearDown() {
+        notificationScope.cancel()
     }
 
     /** Newest-first message history, mirroring the repository's ordering. */
--- a/core/service/src/androidHostTest/kotlin/org/meshtastic/core/service/MeshNotificationManagerImplConversationTest.kt
+++ b/core/service/src/androidHostTest/kotlin/org/meshtastic/core/service/MeshNotificationManagerImplConversationTest.kt
@@ -29,12 +29,9 @@ import dev.mokkery.matcher.any
 import dev.mokkery.mock
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.SupervisorJob
-import kotlinx.coroutines.cancel
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.flowOf
-import kotlinx.coroutines.test.runTest
-import org.junit.After
+import kotlinx.coroutines.test.advanceUntilIdle
 import org.junit.Before
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -46,6 +43,7 @@ import org.meshtastic.core.model.Node
 import org.meshtastic.core.repository.NodeRepository
 import org.meshtastic.core.repository.PacketRepository
 import org.meshtastic.core.repository.RadioConfigRepository
+import org.meshtastic.core.testing.runWithRenderScope
 import org.meshtastic.proto.ChannelSet
 import org.meshtastic.proto.User
 import org.robolectric.annotation.Config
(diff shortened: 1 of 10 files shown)


jamesarich:
The notification half of this is a genuine fix, and I want to be clear about that up front: `MeshService.onStartCommand` -> `startForeground` was running a chain of blocking CMP `getString` calls on the binder thread inside Android's ~5s window. Moving that to `getStringSuspend` on the ServiceScope is the right call, and `@Named("ServiceScope")` is an established qualifier here (`CoreServiceModule.kt:31`), so the DI addition fits the existing shape.

Three things, none of which I consider merge blockers.

### 1. The concurrency machinery is heavier than the problem, and has a narrow race

`commitServiceNotification` validates the generation under `serviceNotificationLock`, then returns so that `notificationManager.notify()` runs outside the lock — deliberately, per the KDoc. That opens a last-writer window: an older job can pass its generation check, get descheduled, and have a newer generation's `notify()` land first; the older job's `notify()` then overwrites it with stale text. It needs the newer job to complete a full async render in the gap between the old job's commit and its very next statement, so it is unlikely rather than impossible, and it self-corrects on the next telemetry tick. But it is precisely the failure the generation counter exists to prevent.

The generation counter + `CoroutineStart.LAZY` + install-then-`start()` + manual `previousJob.cancel()` + `synchronized` blocks come to roughly 60 lines. A conflated flow gives the same guarantees for free, and is inherently ordered, so the race cannot arise:

```kotlin
private val serviceState = MutableSharedFlow<ServiceNotificationSnapshot>(
    extraBufferCapacity = 1,
    onBufferOverflow = BufferOverflow.DROP_OLDEST,
)

init {
    scope.launch { serviceState.collectLatest { snapshot -> renderAndNotify(snapshot) } }
}
```

`collectLatest` conflates and cancels the in-flight render for you. If you would rather keep the current structure, tracking a `lastNotifiedGeneration` at the `notify()` site closes the window.

### 2. The Compose half may not do what the title says

Hoisting `stringResource` out of `AnimatedContent` into `CurrentlyConnectedText` does not make the resource load non-blocking — it relocates the same blocking main-thread load to the initial composition of `ConnectionsScreen`. So the notification change genuinely avoids blocking, while this change moves it earlier and dedupes it across recompositions.

That may well be the outcome you want, since blocking during an animation frame is worse than blocking at screen entry. Was there a measured jank or ANR behind it? The cost is real API friction — `Rssi(label =)`, `MaterialBatteryInfo(unknownLabel =)`, and a required `text:` param on `CurrentlyConnectedInfo`, all workaround surface for an upstream bug. Could you add a TODO naming the CMP version that lets these params be deleted, so they do not calcify into permanent API?

### 3. Minor: misleading sequencing left behind

`MeshService.kt:135-136` now reads as though `connectionManager.updateStatusNotification()` feeds the notification handed to `startForeground`, but the update is async, so it no longer does. Not a bug — the async job re-`notify`s the same `SERVICE_NOTIFY_ID` moments later — but the adjacency implies a data dependency that has gone away. A comment would save the next reader the trip.

### Checked and fine

`runWithRenderScope` parents its `Job` to the test job and cancels in `finally`, so no `UncompletedCoroutinesError`; `kotlinx.coroutines.test` is already `api` in `core/testing`; `DisconnectButton` already had a `label` param; and starting an already-cancelled LAZY job in the install path is a correct no-op.

jeremiah-k:
@jamesarich Thanks for the detailed review. This work was driven by a captured ANR rather than a theoretical cleanup: the main thread was blocked in CMP resource loading from `stringResource` inside the connected-device `AnimatedContent`, while service-notification formatting was also resolving CMP resources. So the issue was a UI/resource-loading stall, not a database race.

I agree the current generation/job machinery is heavier than necessary, and the notify-after-generation-check window means it is not completely generation-safe. I’m going to move the PR back to draft and rework that path around a single ordered `StateFlow`/`collectLatest` renderer.

I’ll retain the Connections hoisting as mitigation for the measured animation-path ANR, add a TODO identifying the temporary CMP workaround, and clarify that the initial foreground notification is only the immediate fallback.

jeremiah-k:
I replaced the generation/job machinery with one replaying, drop-oldest `SharedFlow` collector using `collectLatest`. That cancels obsolete resource rendering and serializes the complete render-and-notify action, so an older Binder post can no longer land after a newer one.

jamesarich (approved):

PR #6399, opened 2026-07-24
fix(hardware): throttle repeated catalog refreshes for cache misses

jeremiah-k:
## Overview

This prevents missing or incomplete hardware models from triggering repeated full hardware-catalog downloads and device-link reconciliation.

Hardware lookups can be driven by packet processing and frequently recomposed UI paths. When a model is absent from both the bundled and remote catalogs, each lookup could start another refresh even though the latest complete catalog had already established that the model was unavailable.

The repository now applies a process-wide refresh gate and only replaces cached hardware after receiving a non-empty remote catalog.

## Key Changes

* Track the most recent refresh attempt and successful full-catalog response.
* Treat a successful non-empty catalog as authoritative for the normal cache lifetime.
* Throttle failed or empty refresh attempts for 15 minutes.
* Allow `forceRefresh` to bypass the gate without deleting cached data first.
* Replace the local catalog transactionally only after a non-empty response.
* Preserve cached remote-only hardware when a forced response is empty.
* Record hardware freshness before device-link reconciliation so a link-refresh failure cannot trigger another immediate full fetch.
* Preserve the existing single-flight and bounded-wait behavior.

## Testing

Added repository and gate coverage for:

* repeated unknown-model lookups sharing one successful refresh;
* retry throttling after failed or empty responses;
* successful-catalog TTL behavior;
* `forceRefresh` bypassing the gate;
* cached remote-only hardware surviving an empty forced response;
* non-empty catalog replacement and device-link reconciliation.

Extended runtime validation showed sustained packet queue and telemetry processing without repeated repository failures, Room or SQLite errors, or evidence of a tight refresh loop. The forced-empty-response and remote-only cache cases are covered directly by the focused repository tests.

## Migration Notes

* No database schema or user data migration is required.
* Hardware models and device-link formats are unchanged.
* Existing cached rows are preserved unless a non-empty replacement catalog is available.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device hardware catalog refresh logic to prevent unnecessary repeated network fetches.
  * Remote refreshes no longer overwrite the local cache when the remote catalog is empty.
  * Added refresh throttling after failed attempts, with a retry delay.
  * Force refresh continues to bypass throttling rules.
* **Tests**
  * Added JVM tests covering throttling, forced refresh behavior, empty-remote preservation, and cache/refresh gating scenarios.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceHardwareLocalDataSource.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceHardwareLocalDataSource.kt
@@ -28,8 +28,8 @@ class DeviceHardwareLocalDataSource(private val dbManager: DatabaseProvider) {
         dbManager.withDb { it.deviceHardwareDao().insertAll(deviceHardware.map { hw -> hw.asEntity() }) }
     }
 
-    suspend fun deleteAllDeviceHardware() {
-        dbManager.withDb { it.deviceHardwareDao().deleteAll() }
+    suspend fun replaceAllDeviceHardware(deviceHardware: List<NetworkDeviceHardware>) {
+        dbManager.withDb { it.deviceHardwareDao().replaceAll(deviceHardware.map { hw -> hw.asEntity() }) }
     }
 
     suspend fun getByHwModel(hwModel: Int): List<DeviceHardwareEntity> =
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/repository/DeviceHardwareRepositoryImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/repository/DeviceHardwareRepositoryImpl.kt
@@ -17,6 +17,7 @@
 package org.meshtastic.core.data.repository
 
 import co.touchlab.kermit.Logger
+import kotlinx.atomicfu.atomic
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.withContext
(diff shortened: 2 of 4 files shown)


jamesarich:
**Review: approve** ✅ — correct, well-scoped fix for the unknown-`hwModel` refresh loop. Gate logic verified across the key paths (non-empty catalog missing model → single fetch then gated; empty/failed remote → throttled + cache preserved; `reconcile()` failure can't re-trigger a fetch; `forceRefresh` bypass; atomic `replaceAll`; clock-skew fails open). Tests are solid and KMP-clean.

Follow-up nits — **all low priority**, fine as-is or a quick follow-up:

- [ ] Add a one-line comment noting `successTtlMs` is deliberately tied to `CACHE_EXPIRATION_TIME_MS` — a genuinely new remote model won't surface for ~24h without a force-refresh (intended trade-off, but non-obvious to a future reader).
- [ ] `DeviceHardwareDao.replaceAll`'s `require(deviceHardware.isNotEmpty())` is unreachable — the only caller already guards on `remoteHardware.isNotEmpty()`. Harmless belt-and-suspenders; the invariant just lives in two places.
- [ ] Optional test: a repository-level case for the *non-forced* empty-response throttle (the mechanism is already covered by `DeviceHardwareRefreshGateTest`).

None block merge. Just confirm the local baseline (`./gradlew spotlessCheck detekt test allTests`) is green first per verify-before-push.

jeremiah-k:
I'll get these while I have it handy.

jeremiah-k:
Err, if it's not cancelled. I'll let it go.

PR #6400, opened 2026-07-24
fix(firmware): decode manifests independent of content type

jeremiah-k:
## Overview

This makes firmware release manifest decoding independent of the HTTP response content type.

GitHub release assets can return valid JSON as `application/octet-stream`. Passing those responses through Ktor content negotiation can produce `NoTransformationFoundException` even though the request succeeded and the response body contains a valid firmware manifest.

The repository now reads the response body as text and decodes it explicitly with the firmware JSON configuration.

## Key Changes

* Read release manifests as text rather than relying on Ktor response transformation.
* Decode release and nightly metadata with one JSON configuration.
* Ignore unknown fields for forward compatibility.
* Preserve serialization failures for malformed payloads.
* Leave request URLs, transport behavior, and manifest models unchanged.

## Testing

* Added coverage for a release-style manifest containing unknown fields.
* Added coverage proving malformed JSON still fails decoding.
* Verified manifest decoding independently of the response content type.

## Migration Notes

* Firmware manifest models and repository interfaces are unchanged.
* This only changes how a downloaded manifest body is decoded.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved firmware release manifest fetching when delivered as plain text, with more resilient JSON decoding.
  * Enhanced decoding robustness by ignoring unknown JSON fields and coercing explicit `null` values to safe defaults.
  * Kept existing behavior for missing nightly firmware (HTTP 404) while ensuring unexpected statuses still surface errors.
* **Tests**
  * Added coverage for manifest retrieval/decoding (including octet-stream responses) and for malformed or non-JSON inputs.

### DIFF
--- a/core/network/build.gradle.kts
+++ b/core/network/build.gradle.kts
@@ -68,6 +68,7 @@ kotlin {
         commonTest.dependencies {
             implementation(projects.core.testing)
             implementation(libs.kable.core) // Kable exception types for BLE failure-injection tests
+            implementation(libs.ktor.client.mock)
         }
     }
 }
--- a/core/network/src/commonMain/kotlin/org/meshtastic/core/network/service/ApiService.kt
+++ b/core/network/src/commonMain/kotlin/org/meshtastic/core/network/service/ApiService.kt
@@ -38,7 +38,17 @@ import org.meshtastic.core.model.NetworkFirmwareReleases
 private const val NIGHTLY_INDEX_URL =
     "https://raw.githubusercontent.com/meshtastic/meshtastic.github.io/master/firmware-nightly/index.json"
 
-private val nightlyIndexJson = Json { ignoreUnknownKeys = true }
+private val firmwareJson = Json {
+    isLenient = true
+    ignoreUnknownKeys = true
+    coerceInputValues = true
+}
+
+/**
+ * Decodes a GitHub release manifest independently of Ktor content negotiation. GitHub release assets commonly use
+ * `application/octet-stream` even when their payload is JSON.
+ */
+internal fun decodeFirmwareReleaseManifest(body: String): FirmwareReleaseManifest = firmwareJson.decodeFromString(body)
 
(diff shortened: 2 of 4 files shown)


jamesarich (approved):

jamesarich:
Nice fix — reading `bodyAsText()` and decoding explicitly is the right way to dodge the `NoTransformationFoundException` from `application/octet-stream` release assets, and the two tests cover the important cases.

One non-blocking note on the JSON config. The new `firmwareJson = Json { ignoreUnknownKeys = true }` is **stricter** than the JSON the old `.body()` path used. That path went through `ContentNegotiation`, which is wired to the DI `provideJson()` (`core/network/.../di/CoreNetworkModule.kt`), configured with:

```kotlin
isLenient = true
ignoreUnknownKeys = true
coerceInputValues = true
```

Since every field in `FirmwareReleaseManifest`/`FirmwareTarget` is a non-nullable type with a default, `coerceInputValues = true` was previously coercing an explicit `null` (e.g. `"platform": null`) to the default `""`. Under the new config that same payload now throws `SerializationException` — a small behavioral narrowing. So `decodeFirmwareReleaseManifest` isn't quite "the same JSON configuration" the rest of the app uses; it's a new, narrower one.

Suggestion (low real-world likelihood, but a free robustness win): either inject the DI `Json` into `ApiServiceImpl` and reuse it, or just add `isLenient = true` and `coerceInputValues = true` to `firmwareJson` to preserve the prior tolerance and stay consistent with the surrounding code.

jeremiah-k:
@jamesarich Good catch. You’re right that the explicit decoder currently narrows the tolerance provided by the shared `Json` configuration.

I’ll align `firmwareJson` with the existing settings by adding `isLenient` and `coerceInputValues`, and add coverage for an explicit `null` falling back to the model default.

jeremiah-k:
I just added that last coderabbit nit test too.

jamesarich (approved):

PR #6403, opened 2026-07-24
fix(android): avoid retaining activity in Coil singleton factory

jeremiah-k:
## Overview

This moves Coil's process-wide singleton `ImageLoader` factory from `MainActivity` to `MeshUtilApplication`.

Registering the singleton factory from activity composition allowed the process-wide factory lambda to capture an activity instance. Recreating the activity could therefore leave the destroyed activity and its associated UI graph reachable through Coil's singleton configuration.

The application now owns the singleton factory and continues to return the existing Koin-configured `ImageLoader`.

## Key Changes

* Implement `SingletonImageLoader.Factory` on `MeshUtilApplication`.
* Resolve the existing Koin-provided `ImageLoader` from application scope.
* Remove activity-level singleton factory registration from `MainActivity`.
* Preserve the existing HTTP client, SVG decoder, cache, logger, and flavor-specific image-loader configuration.
* Leave Desktop image loading unchanged.

## Testing

* Added a Robolectric regression test that starts Koin, invokes the application factory, and verifies it returns the exact configured singleton instance.
* Ensured Koin and image-loader resources are cleaned up after the test.

Runtime validation repeatedly paused, stopped, restarted, and resumed `MainActivity` in the same process without image-loader initialization errors or process replacement. This exercises the lifecycle pattern that previously allowed activity ownership to leak into the process-wide singleton.

## Migration Notes

* Image URLs, cache formats, and rendering behavior are unchanged.
* This only changes ownership of the process-wide Coil singleton factory.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved image-loading lifecycle management for more consistent, reliable rendering across the app.
  * Streamlined startup image-loader configuration to prevent conflicting initialization paths.

* **Tests**
  * Added Robolectric test coverage to confirm the app provides the configured singleton image loader and that it is reused consistently during runtime.

### DIFF
--- a/androidApp/src/main/kotlin/org/meshtastic/app/MainActivity.kt
+++ b/androidApp/src/main/kotlin/org/meshtastic/app/MainActivity.kt
@@ -44,11 +44,8 @@ import androidx.core.splashscreen.SplashScreen.Companion.installSplashScreen
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.lifecycle.lifecycleScope
 import co.touchlab.kermit.Logger
-import coil3.ImageLoader
-import coil3.compose.setSingletonImageLoaderFactory
 import com.eygraber.uri.toKmpUri
 import kotlinx.coroutines.launch
-import org.koin.android.ext.android.get
 import org.koin.android.ext.android.inject
 import org.koin.androidx.viewmodel.ext.android.viewModel
 import org.koin.compose.koinInject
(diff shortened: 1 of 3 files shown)


jamesarich (approved):
Verified the pieces that could break this:

- `org.koin.android.ext.android.get` is already imported and used in `MeshUtilApplication.kt`, so `get<ImageLoader>()` resolves.
- `GoogleMeshUtilApplication` extends `MeshUtilApplication`, so both flavors inherit `newImageLoader`.
- No Koin leak across Robolectric classes: `onTerminate()` already calls `stopKoin()` (`MeshUtilApplication.kt:147`), and Robolectric drives `onTerminate` at teardown.
- `newImageLoader` is invoked lazily on first image load, always after `onCreate` starts Koin.

One thing worth adding to the description, since it is a real improvement that is easy to miss: `setSingletonImageLoaderFactory` only bridged Compose call sites, whereas `SingletonImageLoader.Factory` on the Application covers all process-wide Coil use — including Glance/widget and any non-Compose loads, which previously fell back to a default `ImageLoader` with no custom `HttpClient` or SVG decoder.

Approving; queued to merge on green.

PR #6424, opened 2026-07-24
fix(strings): correct mistranslated subnet string

Jonher937:
The subnet string used the Spanish "Subred" instead of localized text in the French, Romanian, and base (English) resources.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the displayed subnet label from “Subred” to “Subnet” for clearer terminology.


jamesarich:
translations are sourced from Crowdin - https://meshtastic.crowdin.com/android

jamesarich:
Reopening just the base-resources half of this in #6425 — you were right that the English string is wrong. The fr/ro edits would get clobbered by the Crowdin sync, but `values/strings.xml` is the Crowdin *source*, so fixing it there invalidates those translations and they'll be retranslated properly. Credited you on the PR. Thanks for catching it!

Jonher937:
I was too slow, I changed my PR to only edit the English, but your Claude was faster :)

jamesarich:
crowdin sync here shows the update http://github.com/meshtastic/Meshtastic-Android/pull/6426/changes

PR #6433, opened 2026-07-25
fix(node): show node counts without truncation

### LINKED ISSUE #6268: [Bug]: Nodes tab, description of the total number of nodes is partially obscure
### Contact Details

_No response_

### Checklist

### Affected app version

2.8.0 close 9

### Affected Android version

Android 16

### Affected phone model

All

### Affected node model

All

2.7.26

1. Go to node tab
2. At top of the label "Nodes", below there are a description of nodes totalizer 76 online / 398 shown / xxx total) [xxx total or something like that]
The shortname of the node, partially obscures the description of the total number of nodes.

### Actual behavior

The shortname of the node, partially obscures the description of the total number of nodes.

### Expected behavior

NOT obscure the description of the total number of nodes.

Image

### Relevant log output

```shell

```

### Additional information

_No response_

jamesarich:
Fixes #6268

## Root cause

The count string was rendered in the app bar's **subtitle** slot, which is hard-clipped to one ellipsized line:

- `NodeListScreen.kt:171` (pre-fix) passed `subtitle = stringResource(Res.string.node_count_template, ...)`
- `MainAppBar.kt:96-106` renders that subtitle with `maxLines = 1, overflow = TextOverflow.Ellipsis` at `titleSmall`

The width left for the title column is whatever survives the branding icon (`MainAppBar.kt:119`), the our-node shortname chip (`:152-155`), and the two `IconButton`s this screen supplies (`NodeListScreen.kt:177-188`). In an M3 `TopAppBar` the actions win that space and the title column absorbs the shortfall — which is exactly the reporter's "the shortname obscures the description".

The deficit is **absolute, not a distribution problem**: `(76 online / 398 shown / 1247 total)` at `titleSmall` needs ~230dp, while chip + 2 icons + branding consume ~200dp of a 360dp phone. And because the single-row app bar has a fixed container height, simply dropping `maxLines` would clip *vertically* instead of wrapping. **The slot cannot be fixed in place.**

## Approaches considered

1. **Rebalance the Row weights / shrink the chip** — rejected; the deficit is absolute, no weighting makes it fit, and longer locales or large font scales widen the gap.
2. **Drop `maxLines`, or switch to a two-row / flexible top app bar** — rejected; fixed height clips wrapped lines, and a taller bar still competes with the actions row, so it only postpones truncation while spending vertical space on a list screen.
3. **Abbreviate (`76/398/1247`) or autosize** — rejected outright; that hides the problem and drops the labels that make the numbers meaningful.
4. **Chosen: move the counts into the list's sticky header as a wrap-capable `FlowRow` of three independent labels.**

Why 4: it gets the full content width with nothing competing, it can grow vertically, and because each count is its own child it reflows at meaningful boundaries instead of mid-phrase. `FlowRow` is layout-direction aware, so RTL mirrors for free. It sits directly above the filter field — arguably its right home, since the filter is what changes "shown".

## Changes

- **`NodeCountSummary.kt`** (new) — `FlowRow` of three labels, 12dp/2dp gaps, `labelLarge` on `onSurfaceVariant`, **no `maxLines`/overflow anywhere**. `Modifier.clearAndSetSemantics` collapses the three into one screen-reader announcement built from `node_count_template`.
- **`NodeCountSummaryPreviews.kt`** (new) — 6 previews driving an `internal` layout half with injected labels, so long-locale cases don't need a translated bundle.
- **`NodeListScreen.kt`** — drops the `subtitle` argument (with a comment on why counts must not go back there); sticky header now hosts `NodeCountSummary` above `NodeFilterTextField`.
- **`strings.xml`** — adds `node_count_online`, `node_count_shown`, `node_count_total`. `node_count_template` is **retained** (now used for the a11y label), so no existing Crowdin translation is orphaned. `sort-strings.py` run; index regenerated.

## Verification

- `spotlessApply spotlessCheck :feature:node:detekt :core:resources:detekt` → clean
- `:feature:node:allTests`, `:core:konsist:allTests` → pass
- `assembleDebug` → `BUILD SUCCESSFUL`
- No new tests: pure layout change with no logic to assert.

## Reviewer notes

- ⚠️ **This relocates visible UI.** Users who read the counts next to the "Nodes" title now look ~40dp lower. That's a deliberate UX change rather than a pure bug fix, and the call I'd most like confirmed. Worth a release-note line.
- ⚠️ **Never visually verified.** The Compose preview renderer is broken on the dev machine used here (`android` CLI has no `studio render-compose-preview`), so all 6 previews are unrendered. Please render `NodeCountSummaryLongLocaleLargeFontPreview` (long locale @ 2× font) before merging — that's the worst case. Others: `...Default`, `...Narrow` (320dp), `...LongLocale`, `...LargeFont`, `...Wide` (840dp).
- **Residual risk:** at 2× font on a 320dp phone the sticky header grows to ~130dp and eats list rows. Deliberate legibility-over-density trade; if too heavy, move `NodeCountSummary` from `stickyHeader` into a plain `item {}` so it scrolls away — one line, keeps the no-truncation guarantee.
- Labels have no separator glyph (spacing only) and sit on `surfaceDim` while the filter field paints its own background — a subtle two-tone band worth a design eyeball.

## Summary by CodeRabbit

* **New Features**
  * Added a node count summary showing online, shown, and total nodes.
  * Moved the summary to a sticky header area above the node filters.
  * Supports responsive wrapping for narrow screens, long translations, and larger font sizes.
  * Added accessibility text covering the full node count summary.
* **Documentation**
  * Added new translation strings for the node count labels and details in the Compose UI strings index.

### DIFF
--- a/.skills/compose-ui/strings-index.txt
+++ b/.skills/compose-ui/strings-index.txt
@@ -1112,7 +1112,10 @@ no_usb_devices_found
 no_usb_devices_hint
 no_usb_devices_seen
 ### NODE ###
+node_count_online
+node_count_shown
 node_count_template
+node_count_total
 node_filter_exclude_infrastructure
 node_filter_exclude_mqtt
 node_filter_ignored
--- a/core/resources/src/commonMain/composeResources/values/strings.xml
+++ b/core/resources/src/commonMain/composeResources/values/strings.xml
@@ -1145,7 +1145,10 @@
     <string name="no_usb_devices_hint">Connect a device with a USB data cable to use serial.</string>
     <string name="no_usb_devices_seen">No USB devices detected</string>
     <!-- NODE -->
+    <string name="node_count_online">%1$d online</string>
+    <string name="node_count_shown">%1$d shown</string>
     <string name="node_count_template">(%1$d online / %2$d shown / %3$d total)</string>
+    <string name="node_count_total">%1$d total</string>
     <string name="node_filter_exclude_infrastructure">Exclude infrastructure</string>
     <string name="node_filter_exclude_mqtt">Exclude MQTT</string>
     <string name="node_filter_ignored">You are viewing ignored nodes,\nPress to return to the node list.</string>
--- a/feature/node/src/commonMain/kotlin/org/meshtastic/feature/node/component/NodeCountSummary.kt
+++ b/feature/node/src/commonMain/kotlin/org/meshtastic/feature/node/component/NodeCountSummary.kt
@@ -1,0 +1,88 @@
+/*
+ * Copyright (c) 2026 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.feature.node.component
+
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.ExperimentalLayoutApi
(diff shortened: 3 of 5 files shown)


jamesarich:
Addressed in 16f4e5425 — renamed to `ENGLISH_LABELS` / `LONG_LOCALE_LABELS` rather than the suggested lowerCamelCase, since these are top-level constant lists and the same file already uses `ENGLISH_DESCRIPTION`; constant-case keeps them consistent with their neighbour while fixing the same PascalCase issue. `spotlessCheck` and `:feature:node:detekt` pass.

PR #6449, opened 2026-07-26
fix(settings): retain config session across navigation

jeremiah-k:
## Overview

Navigation 3 gives each navigation entry its own `ViewModelStoreOwner`. Resolving `RadioConfigViewModel` independently from each entry therefore recreated the radio-configuration session while moving between configuration pages, restarting its collectors and losing transient state.

This change introduces retained local and per-node remote configuration sessions shared across their navigation entries on Android and Desktop. Session stores survive Android configuration changes, remain available while outgoing transition entries are still composed, and are cleared once the corresponding configuration flow is no longer active and its final entry lease has been released.

Direct local configuration routes, including Connections → LoRa, participate in the same retained-session lifecycle even when the back stack does not contain a Settings root.

## Session lifecycle

* Model inactive, local, and per-node remote configuration sessions explicitly.
* Keep child `ViewModelStore`s in a shell-scoped holder `ViewModel` so the active session survives Android configuration recreation.
* Reuse one `RadioConfigViewModel` while navigating within the same local or remote configuration session.
* Preserve each Settings root’s exact destination while outgoing and incoming entries overlap.
* Lease a session store for each composed navigation entry, delaying eviction until its exit transition is complete.
* Clear abandoned local or remote stores and cancel their collectors after their final lease is released.
* End the session when the user leaves the active radio-configuration flow.
* Start the appropriate new local or remote session when the user later returns.

## Loading behavior

* Continue requesting route-specific configuration when a configuration page opens.
* Keep the existing connect-time configuration visible while local refreshes complete.
* Avoid briefly replacing local content with a full-screen `0%` progress overlay.
* Continue showing blocking progress for genuinely remote configuration reads.
* Continue showing progress for saves and other explicit configuration-changing operations.
* Treat an explicitly addressed local node as local rather than inferring that every non-null destination is remote.

## Navigation integration

* Pass the retained provider through the Android and Desktop navigation hosts.
* Observe back-stack changes with `snapshotFlow` and emit only logical session changes.
* Support direct radio-configuration routes opened from outside the Settings root.
* Remove the unused entry-scoped production fallback.
* Prevent duplicate pushes of the current Settings route.
* Pass the exact Settings-root arguments to the provider during overlapping transitions.
* Add and apply a real `Modifier` parameter to `DesktopMainScreen`.
* Keep the Desktop suppression limited to `ViewModelForwarding`.

## Testing

Added focused coverage for:

* inactive, local, and per-node remote session classification;
* direct local configuration routes outside the Settings tab;
* unrelated settings routes remaining inactive;
* reuse of the same store within one session;
* configuration recreation retaining the active store;
* direct-route handoff without a zero-lease store clear;
* destination changes and configuration-flow exits;
* delayed clearing while outgoing entries remain composed;
* explicit Settings-root destination precedence;
* safe fallback during inactive exit transitions;
* duplicate-route suppression;
* non-blocking local route refreshes;
* blocking remote route refreshes;
* explicitly addressed local-node refreshes;
* progress behavior for direct configuration operations.

## Scope

* Radio configuration payloads and firmware behavior are unchanged.
* Local and remote refresh requests are still performed.
* Session retention is limited to the active radio-configuration flow; collectors are not retained throughout unrelated app navigation.
* This removes redundant ViewModel creation and transient local loading overlays but does not claim a measured performance improvement.

## Summary by CodeRabbit

* **Improvements**
  * Settings navigation now preserves the correct configuration state when switching between local and remote devices.
  * Configuration screens provide clearer loading behavior, including progress overlays only when appropriate.
  * Navigation returns to the appropriate starting screen based on whether a device is connected.

* **Bug Fixes**
  * Improved cleanup and retention of settings state when changing destinations, leaving settings, or recreating the app.
  * Prevented duplicate settings routes from being added.

### DIFF
--- a/androidApp/src/main/kotlin/org/meshtastic/app/ui/Main.kt
+++ b/androidApp/src/main/kotlin/org/meshtastic/app/ui/Main.kt
@@ -27,7 +27,6 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
-import androidx.lifecycle.viewmodel.compose.LocalViewModelStoreOwner
 import androidx.navigation3.runtime.NavKey
 import androidx.navigation3.runtime.entryProvider
 import co.touchlab.kermit.Logger
@@ -55,6 +54,7 @@ import org.meshtastic.feature.map.navigation.mapGraph
 import org.meshtastic.feature.messaging.navigation.contactsGraph
 import org.meshtastic.feature.node.navigation.nodesGraph
 import org.meshtastic.feature.settings.lockdown.LockdownDialog
+import org.meshtastic.feature.settings.navigation.rememberSettingsRadioConfigViewModelStoreOwner
 import org.meshtastic.feature.settings.navigation.settingsGraph
 import org.meshtastic.feature.settings.navigation.settingsRadioConfigViewModel
 import org.meshtastic.feature.settings.radio.channel.channelsGraph
@@ -68,7 +68,7 @@ fun MainScreen() {
     val multiBackstack = rememberMultiBackstack(initialTab)
     val backStack = multiBackstack.activeBackStack
     val scrollToTopEvents = viewModel.scrollToTopEventFlow
-    val appViewModelStoreOwner = checkNotNull(LocalViewModelStoreOwner.current)
+    val settingsViewModelStoreOwner = rememberSettingsRadioConfigViewModelStoreOwner(backStack)
 
     AndroidAppVersionCheck(viewModel)
 
@@ -110,7 +110,9 @@ fun MainScreen() {
                     channelsGraph(backStack)
                     connectionsGraph(backStack)
                     discoveryGraph(backStack)
-                    settingsGraph(backStack) { settingsRadioConfigViewModel(backStack, appViewModelStoreOwner) }
+                    settingsGraph(backStack) {
+                        settingsRadioConfigViewModel(backStack, settingsViewModelStoreOwner)
+                    }
                     docsEntries(backStack)
                     firmwareGraph(backStack)
                     wifiProvisionGraph(backStack)
--- a/androidApp/src/main/kotlin/org/meshtastic/app/ui/Main.kt
+++ b/androidApp/src/main/kotlin/org/meshtastic/app/ui/Main.kt
@@ -27,6 +27,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.remember
 import androidx.compose.ui.Modifier
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import androidx.lifecycle.viewmodel.compose.LocalViewModelStoreOwner
 import androidx.navigation3.runtime.NavKey
 import androidx.navigation3.runtime.entryProvider
 import co.touchlab.kermit.Logger
@@ -55,6 +56,7 @@ import org.meshtastic.feature.messaging.navigation.contactsGraph
 import org.meshtastic.feature.node.navigation.nodesGraph
 import org.meshtastic.feature.settings.lockdown.LockdownDialog
 import org.meshtastic.feature.settings.navigation.settingsGraph
+import org.meshtastic.feature.settings.navigation.settingsRadioConfigViewModel
 import org.meshtastic.feature.settings.radio.channel.channelsGraph
 import org.meshtastic.feature.wifiprovision.navigation.wifiProvisionGraph
 
@@ -62,16 +64,11 @@ import org.meshtastic.feature.wifiprovision.navigation.wifiProvisionGraph
 fun MainScreen() {
     val viewModel: UIViewModel = koinViewModel()
     // Land on Connections for first-run / no-device-selected; otherwise on Nodes (seeded from prefs).
-    val initialTab = remember {
-        if (viewModel.currentDeviceAddressFlow.value.isNullOrSelectedNone()) {
-            TopLevelDestination.Connect.route
-        } else {
-            NodesRoute.Nodes
-        }
-    }
+    val initialTab = remember { initialRoute(viewModel.currentDeviceAddressFlow.value) }
     val multiBackstack = rememberMultiBackstack(initialTab)
     val backStack = multiBackstack.activeBackStack
     val scrollToTopEvents = viewModel.scrollToTopEventFlow
+    val appViewModelStoreOwner = checkNotNull(LocalViewModelStoreOwner.current)
 
     AndroidAppVersionCheck(viewModel)
 
@@ -113,7 +110,7 @@ fun MainScreen() {
                     channelsGraph(backStack)
                     connectionsGraph(backStack)
                     discoveryGraph(backStack)
-                    settingsGraph(backStack)
+                    settingsGraph(backStack) { settingsRadioConfigViewModel(backStack, appViewModelStoreOwner) }
                     docsEntries(backStack)
                     firmwareGraph(backStack)
                     wifiProvisionGraph(backStack)
@@ -124,6 +121,9 @@ fun MainScreen() {
         }
     }
 }
+
+private fun initialRoute(deviceAddress: String?): NavKey =
+    if (deviceAddress.isNullOrSelectedNone()) TopLevelDestination.Connect.route else NodesRoute.Nodes
 
 /** True when no device address is persisted, or the address is the "none" sentinel (`"n"`). */
 private fun String?.isNullOrSelectedNone(): Boolean = isNullOrBlank() || this == "n"
--- a/androidApp/src/main/kotlin/org/meshtastic/app/ui/Main.kt
+++ b/androidApp/src/main/kotlin/org/meshtastic/app/ui/Main.kt
@@ -54,9 +54,8 @@ import org.meshtastic.feature.map.navigation.mapGraph
 import org.meshtastic.feature.messaging.navigation.contactsGraph
 import org.meshtastic.feature.node.navigation.nodesGraph
 import org.meshtastic.feature.settings.lockdown.LockdownDialog
-import org.meshtastic.feature.settings.navigation.rememberSettingsRadioConfigViewModelStoreOwner
+import org.meshtastic.feature.settings.navigation.rememberSettingsRadioConfigViewModelProvider
 import org.meshtastic.feature.settings.navigation.settingsGraph
-import org.meshtastic.feature.settings.navigation.settingsRadioConfigViewModel
 import org.meshtastic.feature.settings.radio.channel.channelsGraph
 import org.meshtastic.feature.wifiprovision.navigation.wifiProvisionGraph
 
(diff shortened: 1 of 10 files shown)


jamesarich (changes requested):
The diagnosis here is right and worth fixing: Navigation 3 gives every entry its own `ViewModelStoreOwner`, so `getRadioConfigViewModel` was minting a **separate** `RadioConfigViewModel` per settings entry — each one re-running eight `launchIn(viewModelScope)` collectors and its own remote fetch, with the "session" living in whichever instance the visible screen happened to resolve. Hoisting to one store keyed by destination is the right shape, and `:feature:settings:allTests` is green locally.

Two things I'd want addressed before merge, both downstream of holding the store in a plain `remember`.

---

### 1. Rotation now destroys the config session

`SettingsNavigation.kt:107-111` keeps the owner in a plain `remember` and clears it from a `DisposableEffect`. `MainActivity` declares no `configChanges` (there is no `configChanges`/`screenOrientation` attribute anywhere in `androidApp/src/main/AndroidManifest.xml`), so rotation recreates the activity, disposes the composition, fires `onDispose { owner.clear() }`, and kills the `RadioConfigViewModel` mid-session — losing `_radioConfigState`, any in-flight `responseState`/pending save, and forcing `ensureLoadingForRemote` to refetch.

The entry-scoped path this replaces did **not** have that problem. `MeshtasticNavDisplay.kt:132` installs `rememberViewModelStoreNavEntryDecorator`, whose stores are held by a `ViewModelStoreProvider` parented to the activity's store; upstream is explicit about it:

> `removeViewModelStoreOnPop`: This parameter was a workaround for detecting configuration changes … **Configuration changes are now handled internally.**

Entry stores are cleared on `onPop`, not on composition disposal, so a config session survived rotation before this change and won't after. That's the same class of bug the PR is fixing, just triggered by a different gesture — and rotating mid-config is a pretty ordinary thing to do.

Suggested fix: park the per-session `ViewModelStore`s in a retained holder `ViewModel` obtained at the shell scope (`koinViewModel()` in `MainScreen`/`DesktopMainScreen`), keyed by `session.viewModelKey`, clearing stores for keys that are no longer the active session. That keeps the cross-entry sharing this PR wants *and* the config-change retention nav3 was already giving you.

### 2. The local session is retained for the whole app-shell lifetime

`settingsRadioConfigSession` (`SettingsNavigation.kt:137`) returns `destination = null` whenever the active backstack contains no `SettingsRoute.Settings` — which is true on **every non-settings tab**, not just in local settings. `SettingsRadioConfigSession` is a data class over a single nullable `Int`, so `remember(session)` at `:108` sees the same value on the Nodes tab, the Map tab, everywhere, and hands back the same owner. Net effect: once a user opens local settings, that `RadioConfigViewModel` is never cleared for the life of the composition, and its collectors keep running everywhere in the app — including `serviceRepository.meshPacketFlow.onEach(::processPacketResponse)` (`RadioConfigViewModel.kt:344`), which processes every mesh packet.

That's precisely the invariant the new KDoc claims to establish:

> …and clears it when the destination changes, so an abandoned remote session cannot retain collectors for the rest of the activity or window lifetime.

True for remote sessions, inverted for the local one. `SettingsRadioConfigSession` needs to distinguish "no settings destination in the stack" (→ drop the session) from "local settings destination" (→ keep it), rather than collapsing both to `null`.

---

### Non-blocking

- **Tab switching still tears the session down.** `activeBackStack` swaps on tab change (`MultiBackstack.kt:39-40`), so leaving the Settings tab mid-remote-session flips the session to local, clears the remote owner, and returning rebuilds the ViewModel from scratch. Fine as a design choice, but the description's "retain the same `RadioConfigViewModel` while navigating within one local or remote settings session" reads stronger than what ships — worth stating the boundary.

- **The shell now recomposes on every navigation.** `rememberSettingsRadioConfigSession` calls `backStack.toList()` outside `remember` (`:134`), so `MainScreen`/`DesktopMainScreen` subscribe to backstack mutations; previously the shell only read `currentTabRoute`. Every push/pop now invalidates the shell and rebuilds the `entryProvider`. Probably minor next to the ViewModel churn you're removing, but it does cut against the "improves perceived UI responsiveness" claim, so I'd either measure it or confine the read below the shell.

- **`entryScopedRadioConfigViewModel` is dead in production.** Both Android (`Main.kt:113`) and Desktop pass a provider; the only caller of the default is `androidApp/src/test/.../NavigationAssemblyTest.kt:57`. So the default keeps a second code path compiling that nothing ships, and lets the assembly test assert a path no user hits. Either drop the default and update the test, or add a comment on why it stays.

- **`@Suppress("ViewModelForwarding", "ModifierMissing")`** on `DesktopMainScreen` — `ModifierMissing` is unrelated to this change and the suppression is function-wide. Worth narrowing to just what the new lambda triggers.

- **Nit:** `entry<SettingsRoute.Settings>` (`:159`) no longer honors `args.destNum`, deriving destNum from the last `Settings` key instead. Settings is single-pane so these agree at rest; while pushing `Settings(1234)` onto `Settings()` both entries are briefly composed and the outgoing local screen resolves the remote ViewModel for the crossfade. Cosmetic, and awkward to avoid given the shared owner — flagging only because the override was deliberate.

- **Tests.** The new coverage is over the pure key-derivation helpers, plus one case (`clearing a settings session store clears its view models`) that mostly asserts `ViewModelStore.clear()` calls `onCleared()` — androidx's behavior, not this PR's. Neither of the two claims that actually matter is covered: that local → `DeviceConfiguration` → `LoRa` yields the *same* ViewModel instance, and that changing destination clears the previous store. A Compose test over `settingsRadioConfigViewModel` with a fake owner would pin both, and would have caught (2).

jeremiah-k:
@jamesarich Thanks for the detailed review. Both lifecycle blockers and the related implementation notes have now been addressed.

The session stores are owned by a shell-scoped retained ViewModel, so Android configuration recreation no longer clears an active `RadioConfigViewModel`. The session model explicitly distinguishes inactive, local, and per-node remote states, preventing the local session and its collectors from remaining active throughout unrelated app navigation.

Each composed navigation entry leases its captured session store. Destination changes and configuration-flow exits therefore retain the outgoing ViewModel only through its transition, then clear the abandoned store after the final lease is released. Settings roots pass their exact `args`, so overlapping local and remote entries cannot resolve one another’s ViewModel.

Back-stack observation now uses `snapshotFlow` with logical-session deduplication, the unused entry-scoped production fallback was removed, and `DesktopMainScreen` exposes and applies a real `Modifier` with only the ViewModel-forwarding suppression retained.

I also covered direct configuration routes such as Connections → LoRa, which do not have a Settings root but still require a retained local session. Local route refreshes now keep the existing connect-time configuration visible instead of briefly showing a `0%` overlay; genuinely remote reads and explicit configuration operations continue to show progress.

The expanded tests cover same-session reuse, configuration recreation, direct-route handoff, destination changes, configuration-flow exit, transition-safe eviction, exact-root precedence, inactive fallback, duplicate-route suppression, and local-versus-remote loading behavior.

jamesarich (approved):
Re-reviewed at `82c533a`. Both blockers are properly addressed, and the fixes go further than what I asked for — thanks for that.

**Rotation (was blocking).** `SettingsRadioConfigSessionHolder` is a retained `@KoinViewModel` resolved from the shell's `LocalViewModelStoreOwner` (the activity), so the per-session stores now outlive activity recreation. The lease design is the part I'd have gotten wrong: `activate()` marking a session current, plus refcounted `retain`/`release` from each entry's `DisposableEffect`, means the store survives the window where the old composition has released but the new one hasn't acquired. I traced both interleavings — release-then-activate and activate-then-release — and they converge on the same eviction either way, which is what `configuration recreation keeps the active settings store` and `direct local route transition does not clear its active store between entries` pin down. `onCleared` sweeping the map closes the last exit.

**Local session retention (was blocking).** The sealed `Inactive`/`Local`/`Remote` split is the right shape, and `usesRadioConfigSettingsSession` catching the `Connections -> LoRa` direct-entry case is a path I'd missed entirely — good catch. `meshPacketFlow` collectors no longer survive on unrelated tabs.

Also resolved: the dead default provider is gone (`NavigationAssemblyTest` now fails loudly instead of exercising a shipped-nowhere path), `ModifierMissing` is fixed properly with a real `modifier` param rather than suppressed, `entry<SettingsRoute.Settings>` takes `args` again so the crossfade nit is moot, and the new holder tests cover same-instance retention and cross-destination eviction directly.

I also went looking for a stuck overlay in `avoid local refresh loading flash` — `ensureLoadingForRemote` now fires on `destNum != null` (synchronous) instead of `!state.isLocal` (async, defaults `false`), so a remote route sets `Loading` on the first frame; if `loadConfigRoute` then early-returned for a non-refreshable module the overlay would never resolve. It doesn't: `setResponseStateLoading` writes `ResponseState.Empty` on that path (`RadioConfigViewModel.kt:757-760`). Not a bug. Worth noting the `destNum != null` gate does change behavior for remote-administering *your own* node (`destNum == myNodeNum` was `isLocal`, so no overlay; now it shows one) — that reads as a fix to me, since admin round-trips really are in flight there, but it's a deliberate change rather than a no-op refactor.

`:feature:settings:allTests`, `:androidApp:testFdroidDebugUnitTest`, and `:feature:settings:detekt` are green locally.

---

One non-blocking item, and it's specifically about the comment rather than the behavior:

> Observes back-stack mutations without invalidating the app shell for every settings submenu push or pop.

I don't think that holds. `produceState`'s `initialValue` is an ordinary argument expression, re-evaluated on every recomposition, and it calls `settingsRadioConfigSession(backStack.toList())` — so the snapshot read of the backstack still happens on the way in. Because `rememberSettingsRadioConfigSession` and `rememberSettingsRadioConfigViewModelProvider` both return values, neither is a restartable composable, so that read is attributed to the nearest restartable ancestor: `MainScreen` / `DesktopMainScreen`. Every push and pop should still invalidate the shell.

What `produceState` + `distinctUntilChanged` *does* buy is that the returned `State` only changes on real session transitions, so `holder.activate` isn't churning — worth keeping. But the shell invalidation is unchanged from before, just now with an extra coroutine hop that delays `activate` by a frame (harmless as far as I can tell — I walked the eviction orderings above with the lag in mind).

`derivedStateOf` would actually deliver what the comment claims, and is simpler and synchronous:

```kotlin
@Composable
internal fun rememberSettingsRadioConfigSession(backStack: NavBackStack<NavKey>) =
    remember(backStack) { derivedStateOf { settingsRadioConfigSession(backStack.toList()) } }
```

Reads inside a `derivedStateOf` are tracked by the derived-state object, not the enclosing scope, so the shell only invalidates when the session value actually changes. Worth confirming with a recomposition counter before changing anything — this is a subtle enough corner of the compiler that I'd rather you verify than take my word for it. Either way, please reword the comment to match whichever behavior ships.

Two smaller nits, neither worth another round:

- `remember(entrySession, holder) { holder.ownerFor(entrySession) }` mutates `stores` via `getOrPut` from inside a `remember` calculation. An abandoned composition would leave a zero-lease entry behind — self-healing on the next `clearUnusedStores()`, so it's benign, but it is a side effect in a slot-table calculation.
- `usesRadioConfigSettingsSession` derives the `ConfigRoute`/`ModuleRoute` cases automatically but hardcodes the four hand-written entries (`Settings`, `DeviceConfiguration`, `ModuleConfiguration`, `Administration`). A future entry added to `settingsGraph` has to be mirrored here. The lease keeps a live screen safe if someone forgets, so this is a maintenance note, not a hazard.

Approving.

PR #6469, opened 2026-07-27
fix(network): resync stream framing and stop reporting expected disconnects as errors

jamesarich:
Three error groups that appeared together on 2.8.0 and fire within minutes of each other on the same sessions turned out to have **two unrelated causes**, not one bad frame. This hardens the stream framing reader so a corrupt frame costs one frame instead of desyncing the parser, and stops reporting ordinary unreachable-host/peer-disconnect events as application errors.

### What the telemetry actually showed

The reported cluster was:

- `java.net.ProtocolException: Unexpected field encoding` at `wire.ProtoReader.nextTag` (2 events)
- `java.lang.IllegalStateException: Unexpected call to beginMessage()` in okio (3 events)
- `java.io.EOFException: Not enough data available` at `ktor …ByteReadChannelOperationsKt.readByte` (6 events in RUM; 352 events / 113 users/day in Crashlytics `0691eaf8…`)

Pulling a sample event for the `EOFException` shows it is **not** the radio framing path — it carries `tag: MqttConnection` / `message: Fatal error — tearing down connection`, and the session log reads `Error/MqttConnection: Read loop error: Not enough data available` at the same instant Wi-Fi drops (`TCP connection error` → `Liveness check failed`). It also first appeared in 2.7.14, not the new build. The three errors co-occur because one flaky network kills the radio socket *and* the MQTT socket together — not because one frame corrupts three parsers.

The two wire/okio errors are the genuine framing bug, and `StreamFrameCodec` had a concrete defect that produces exactly them.

### 🐛 Bug Fixes

- **`StreamFrameCodec` lost a whole frame when a start byte repeated, then desynced into the next payload.** On a byte that broke framing the state machine reset to the hunting state but *discarded that byte*. For `94 94 c3 …` the second `0x94` failed the `START2` check and was thrown away, taking the real header with it — the frame vanished silently and the machine resumed hunting **inside the payload of the following frame**, where it would re-sync on an incidental `0x94`/`0xc3` pair and hand a misaligned byte range to Wire. That is precisely `ProtocolException: Unexpected field encoding` and `Unexpected call to beginMessage()`. The behaviour was parity-dependent: odd-length runs of `START1` happened to work, even-length runs ate the frame.
  - A repeated `START1` is now treated as padding (the wake sequence is four of them) rather than a desync.
  - A byte that genuinely breaks framing is re-examined as a possible `START1` instead of dropped — load-bearing when a rejected length prefix ends in `0x94`.
  - Desyncs are counted (`framingDesyncCount`) and logged with structural detail only — reason and declared length, never payload bytes. The log is warn-once-per-disruption, then debug, so a peer emitting non-protocol bytes cannot flood.
  - The desync log was `Logger.e`, which itself fed the error trackers; it is now `w`/`d`.
- **Expected connection failures are no longer reported as errors.** Added `Throwable.isExpectedConnectionFailure()` (`expect`/`actual`, so `commonMain` stays free of `java.*`) covering the `IOException` family on every target plus `UnresolvedAddressException` on JVM/Android — that one extends `IllegalArgumentException`, so it slipped past every `catch (IOException)` in the transport stack and landed in the generic error handler. Applied in two places:
  - `TcpTransport` — an unreachable radio logs at warn and feeds the normal retry/backoff loop instead of `Logger.e`. This is the `UnresolvedAddressException` (195/day) and `ConnectException: Connection refused` (76/day) noise.
  - `KermitMqttLogger` — the MQTT client reports a broker-closed socket at `ERROR` and then reconnects on its own; when the throwable is an expected connection failure it is logged at warn. This is the 352/day `EOFException`.

- New `jvmAndroidMain` source set in `:core:network` (the module already applies the `kmp.jvm.android` convention) for the shared JVM/Android actual.
- `KermitMqttLogger` takes an injectable base `Logger` (defaulting to the global) so the severity mapping is testable.
- `desyncCount` is `@Volatile` — it is published through a public getter for diagnostics, so an external reader needs safe publication. Matches the metrics counters in `TcpTransport`; `atomicfu` is reserved there for `heartbeatNonce`, which needs `getAndIncrement()` atomicity across concurrent senders.

### Known limitation

The framing protocol has no checksum, so a frame whose *length prefix* is corrupt but in-range cannot be detected as corrupt — the reader will consume that many bytes and may swallow the frame that follows. That is inherent to the wire format; what the reader now guarantees is that it always recovers and parses the next well-formed frame. The truncated-frame test asserts exactly this, including the frame that is unavoidably lost.

## Testing Performed

21 new tests in `:core:network` (34 total across the four classes), all passing.

`StreamFrameCodecTest` (commonTest) — resynchronization after malformed input:
- `parses a frame whose header is preceded by a repeated START1` — the regression; asserts the frame survives *and* that padding is not counted as a desync
- `parses a frame after a START1 run of any length` — runs of 1–6, covering the parity-dependent failure
- `parses a frame that arrives immediately after wake bytes`
- `re-examines the rejected byte when a corrupt length ends in START1`
- `recovers from a corrupt length prefix and parses the next frame`
- `recovers from a truncated frame and parses a later frame`
- `recovers from garbage between two frames`
- `recovers from a frame header split by a desync in the middle of a payload` — a payload containing an incidental `94 c3` is still delivered verbatim
- `always parses a well-formed frame following non-protocol noise` — property test over arbitrary noise containing no `START1`
- `desync count tracks framing errors and resets with the codec`

`ConnectionFailuresTest` (commonTest) and `ConnectionFailuresJvmTest` (jvmTest) — the `IOException` family, `UnresolvedAddressException`, and negative cases proving the classification does not widen to every `IllegalArgumentException`.

`KermitMqttLoggerTest` (commonTest) — an expected connection failure at `ERROR` is downgraded to warn; a genuine error and an error with no throwable both stay at error; other levels map straight through and keep the library tag.

Baseline green locally: `./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` — BUILD SUCCESSFUL, 1876 tasks. `kmpSmokeCompile` included because this adds an `iosMain` actual.

### DIFF
--- a/core/network/src/commonMain/kotlin/org/meshtastic/core/network/repository/KermitMqttLogger.kt
+++ b/core/network/src/commonMain/kotlin/org/meshtastic/core/network/repository/KermitMqttLogger.kt
@@ -17,6 +17,7 @@
 package org.meshtastic.core.network.repository
 
 import co.touchlab.kermit.Logger
+import org.meshtastic.core.network.transport.isExpectedConnectionFailure
 import org.meshtastic.mqtt.MqttLogLevel
 import org.meshtastic.mqtt.MqttLogger
 
@@ -33,15 +34,32 @@ import org.meshtastic.mqtt.MqttLogger
  * Note: The production log level should be set to [MqttLogLevel.WARN] (not INFO) to prevent the library's own
  * INFO-level messages (which include endpoint addresses and topic strings) from reaching remote analytics sinks.
  */
-class KermitMqttLogger : MqttLogger {
+class KermitMqttLogger(
+    /** Base logger to tag and delegate to. Injectable so tests can capture severities. */
+    private val baseLogger: Logger = Logger,
+) : MqttLogger {
     override fun log(level: MqttLogLevel, tag: String, message: String, throwable: Throwable?) {
-        val logger = Logger.withTag(tag)
+        val logger = baseLogger.withTag(tag)
         when (level) {
             MqttLogLevel.TRACE -> logger.v(throwable) { message }
+
             MqttLogLevel.DEBUG -> logger.d(throwable) { message }
+
             MqttLogLevel.INFO -> logger.i(throwable) { message }
+
             MqttLogLevel.WARN -> logger.w(throwable) { message }
-            MqttLogLevel.ERROR -> logger.e(throwable) { message }
+
+            // The library reports a broker that closed the socket, an unreachable host, or Wi-Fi dropping mid-read at
+            // ERROR ("Read loop error: Not enough data available"), and the client then reconnects on its own. Those
+            // are expected for a mobile client and dominated our error tracking, so log them at WARN — diagnosable in
+            // a support session, but not counted as application errors.
+            MqttLogLevel.ERROR ->
+                if (throwable?.isExpectedConnectionFailure() == true) {
+                    logger.w(throwable) { message }
+                } else {
+                    logger.e(throwable) { message }
+                }
+
             MqttLogLevel.NONE -> return
         }
     }
--- a/core/network/src/commonMain/kotlin/org/meshtastic/core/network/transport/ConnectionFailures.kt
+++ b/core/network/src/commonMain/kotlin/org/meshtastic/core/network/transport/ConnectionFailures.kt
@@ -1,0 +1,37 @@
+/*
+ * Copyright (c) 2026 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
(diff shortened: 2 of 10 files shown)


jamesarich:
> **Guard the newly-exposed `framingDesyncCount` against cross-thread visibility issues.**

Agreed on the concern and fixed in 67c7bdd3b — `desyncCount` is now `@Volatile`.

One correction on the cited precedent: `heartbeatNonce` uses `atomic()` because it needs `getAndIncrement()` atomicity across potentially concurrent senders. This counter has a single writer (the transport read thread) and only needs safe publication to external readers, so the closer analogue is the metrics counters in the same file — `packetsReceived`, `bytesReceived`, `timeoutEvents` — which are all `@Volatile`. Went with that for consistency.

Left `consecutiveDesyncs` as a plain var: it is private and only throttles logging on the read thread, so it has no cross-thread reader to guarantee anything for.

PR #6473, opened 2026-07-27
feat(widget): Glance error UI, documented resource bridge, and widget tests

jamesarich:
## Why

Split out of #6467, which is the actual 2.8.0 crash fix and should stay a one-line manifest removal. This PR is the quality work found while investigating that crash — none of it is release-critical, and none of it should hold up 2.8.0.

Three pre-existing Glance best-practice gaps in `feature/widget`, checked against the [official Glance docs](https://developer.android.com/develop/ui/compose/glance).

### 1. No error UI at all

`LocalStatsWidget` had no `errorUiLayout` and no `onCompositionError` override, so any composition throw showed users a bare *"Can't show content"* with no diagnostics and no signal to us — a silently broken widget.

This matters more than usual because of gap 2: the widget's string rendering depends on an undocumented workaround, and its failure mode is exactly a composition throw.

**Worth knowing for anyone else hitting this:** the [error-handling guide's](https://developer.android.com/develop/ui/compose/glance/error-handling) `GlanceAppWidget(errorUiLayout = R.layout.x)` **does not compile against 1.2.0-rc01** — that constructor parameter is `internal`. Overriding `onCompositionError` is the only route from outside the library, and doing so means `errorUiLayout` is never read at all.

The message is resolved from the Crowdin-managed Compose-resources catalog through the project's existing blocking `getString()`, so a `RemoteViews` layout that cannot reference Compose resources still shows translated text rather than shipping an English-only string.

### 2. The Compose-resources bridge was undocumented

Glance deliberately does not provide the `androidx.compose.ui` composition locals — a Glance tree becomes `RemoteViews`, it is never laid out by Compose UI. But Compose Multiplatform's `stringResource` reads `LocalContext`, `LocalConfiguration`, and `LocalDensity` to find the asset loader, locale, and density qualifier. So the widget hand-provides them.

That bridge was an inline `CompositionLocalProvider` with a one-line comment, despite essentially every visible string in the widget depending on it. Now `WidgetResourceEnvironment`, with the rationale, the failure mode, and a warning that `LocalDensity` here selects resource qualifiers and must not be used to compute layout.

### 3. No tests, and `glance-testing` was not in the catalog

`feature/widget` had zero tests. Both `androidx.glance:glance-testing` and `glance-appwidget-testing` are now catalogued.

## 🐛 Bug Fixes

- Add an error UI for `LocalStatsWidget` via `onCompositionError`, and log the throwable so a failed composition stops being silent.
- `widget_local_stats_info.xml` referenced `@layout/widget_local_stats_preview`, which lived in `androidApp` — the widget module's resources could not link on their own. Moved into this module alongside the provider info that references it.
- `widget_ic_refresh.xml` tinted with `?attr/colorControlNormal`, an AppCompat theme attribute this module does not depend on and which a launcher's theme need not define. Now a literal — inert at render time, because Glance's `CircleIconButton` tints the icon via `contentColor` regardless.

## 🧹 Housekeeping

- Extract the composition-local bridge into `WidgetResourceEnvironment` with full rationale.
- Drop a hardcoded user-facing string from the preview layout in favour of `@string/widget_local_stats_label`.

## Testing Performed

Four tests on Glance's own harness ([`runGlanceAppWidgetUnitTest`](https://developer.android.com/develop/ui/compose/glance/testing)) — Robolectric-backed, because resolving Compose Multiplatform strings needs a real `Context`:

- connected content renders node identity, battery, and footer counts
- disconnected content renders the translated status string
- the refresh affordance exists and carries its content description
- the narrowest responsive size still renders rather than collapsing

Two notes on these:

- **The bridge test genuinely guards.** Verified by making it fail: bypassing `WidgetResourceEnvironment` fails all four with `IllegalStateException: CompositionLocal LocalConfiguration not present`. A guard that cannot fail is worthless, so this was checked rather than assumed.
- **They pass an explicit 30s timeout** rather than Glance's 2s `DEFAULT_TIMEOUT`, which was observed failing one test with `UncompletedCoroutinesError` during a full baseline while passing in isolation. `configureTestOptions()` sets `maxParallelForks` to every available processor in CI, so the default is not a safe budget. Re-verified 3/3 green under `-Pci=true`.

Also included: `:feature:widget` is wired into the `shard-app` test shard. This is not optional — `check-changes` has a drift guard that fails when a module listed as test-less gains test sources, and it correctly caught that these tests would otherwise have been compiled but **never run in CI**. Task names were verified against the module's real task list: it is unflavored, so `testDebugUnitTest`, not the `test{Flavor}DebugUnitTest` pair `:feature:car` needs.

Full baseline green on this branch's base: `./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests` (543 tests). The widget tests were additionally forced to execute with `--rerun-tasks --no-build-cache -Pci=true` — 4/4 — because a cached green tick proves nothing about whether a test ran.

**Not device-verified.** The error UI has only been exercised by construction, never by forcing a real composition failure on hardware. Placing the widget to confirm it still refreshes and opens the app, and ideally forcing a throw to see the error layout, would be worth doing before merge.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -143,7 +143,6 @@ jobs:
               ':core:di',
               ':core:nfc',
               ':core:resources',
-              ':feature:widget',
           }
 
           shards = check.split('# ── Sharded Unit Tests')[1].split('# ── Android Build')[0]
--- a/.github/workflows/reusable-check.yml
+++ b/.github/workflows/reusable-check.yml
@@ -332,7 +332,8 @@ jobs:
   #   shard-core:    remaining core:* KMP module tests (allTests)
   #   shard-feature: feature:* KMP module tests + :core:service
   #   shard-app:     Pure-Android/JVM tests (androidApp, desktopApp,
-  #                  core:barcode) + :core:database + :core:network
+  #                  core:barcode, feature:car, feature:widget)
+  #                  + :core:database + :core:network
   test-shards:
     runs-on: ubuntu-24.04
     permissions:
@@ -410,6 +411,7 @@ jobs:
               :core:network:allTests
               :feature:car:testFdroidDebugUnitTest
               :feature:car:testGoogleDebugUnitTest
+              :feature:widget:testDebugUnitTest
             kover: >-
               :androidApp:koverXmlReportFdroidDebug
               :androidApp:koverXmlReportGoogleDebug
@@ -420,6 +422,7 @@ jobs:
               :core:network:koverXmlReport
               :feature:car:koverXmlReportFdroidDebug
               :feature:car:koverXmlReportGoogleDebug
+              :feature:widget:koverXmlReportDebug
 
     steps:
       - name: Checkout code
--- a/feature/widget/build.gradle.kts
+++ b/feature/widget/build.gradle.kts
@@ -26,6 +26,10 @@ android {
     resourcePrefix = "widget_"
 
     defaultConfig { minSdk = 26 }
+
+    // Glance unit tests resolve Compose-resources strings through a real Context, so the merged
+    // Android resources and assets must be on the Robolectric test classpath.
+    testOptions { unitTests { isIncludeAndroidResources = true } }
 }
 
 dependencies {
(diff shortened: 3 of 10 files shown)


jamesarich:
Split out of #6467 so that PR can stay a one-line manifest removal for the 2.8.0 line. Independent of #6467 — the two touch disjoint files and either can merge first.

PR #6483, opened 2026-07-28
feat(event): DEF CON 34 logo and full brand palette in event theming

jamesarich:
## 🌟 Why

Event branding only ever read **one** color per edition: the accent wash behind the app bar. For DEF CON 34 that color is `#0D294A` — a navy dark enough that washed at 12% over a dark surface it is invisible. Connected to DEF CON firmware, the app showed a generic Meshtastic logo and no visible theming at all.

The metadata was never the problem. `event_firmware.json` already carries DEF CON's six-color palette, its named brand colors, the theme name ("Agency"), and the tagline. None of it reached the UI.

## 🛠️ What

**Logo** — bundles the DEF CON 34 mark as `img_event_defcon.png` (384×384, 18.6 KB) and registers it in `eventIconFor()`, and points the edition's `iconUrl` at the hosted copy shipping in meshtastic/api#109. Hosted first, bundled as the offline fallback — the existing `EventBrandingIcon` contract, so the icon appears whether or not the API change has landed.

**Palette** — `EventTheme` now carries the full `palette` and a high-contrast `highlight` alongside the wash `accent`. The palette is drawn as a gradient rule under the app bar and as a strip closing the info-sheet header. A single wash color can never show more than one color; the rule is where the rest of the palette lives.

**Unused metadata** — the info sheet now renders the theme name (as a subtitle to the event name) and the tagline.

**Contrast** — new `pickLegible()` in `ColorContrast.kt`. Brand palettes are authored for the event's own print and web materials, not for our light *and* dark surfaces: DEF CON's navy fails on dark, FAB26's `#EAB14B` fails on light. Info-sheet tints now walk highlight → palette → `onSurfaceVariant`, taking the first that clears WCAG AA for graphical objects.

Everything is data-driven off the manifest, so every edition gets whatever colors it publishes. DEF CON is just the first with a full palette to spend; Hamvention and Open Sauce publish one color and render exactly as before.

## 🎨 Design notes

- The header band stays a single flat color so the title keeps a predictable contrast ratio. The palette gets its own strip rather than running under text.
- The brand rule is 3 dp under the app bar, 6 dp in the sheet — brand presence without spending vertical space.
- An edition with one color draws a flat rule instead of a gradient; an edition with none draws nothing.
- All of it stays behind the existing "Use event theme" opt-out.

## 🧪 Testing performed

- `EventBrandingTest` — new coverage for `brandPalette()` (authored palette preferred, malformed hex dropped, falls back to named colors then accent, de-duplicated, empty when nothing is published) and `brandHighlightOrNull()` (accent → secondary; a bare primary is the wash color and is never promoted to a highlight).
- `PickLegibleTest` — new. Same palette resolving to different colors on light vs dark backgrounds, threshold honoured, fallback when nothing reads.
- `./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests` — green.

Not yet verified against a physically connected DEF CON device; the branding path itself (hosted icon → bundled fallback → Meshtastic logo) is unchanged, and the palette work is exercised by unit tests.

## 🔗 Related

- meshtastic/api#109 — hosts `defcon34.png` and sets the matching `iconUrl` upstream. Independent: this PR's bundled drawable covers the gap until it deploys.

## Summary by CodeRabbit

* **New Features**
  * Added DEF CON 34 event icon and richer event theming, including a color palette strip and theme subtitle in the event header.
  * Event details screens now use event-specific palette/highlight colors for improved visual branding, including a consistent theme toggle row.
  * Event themes now support highlights, palettes, and optional custom fonts.
* **Bug Fixes**
  * Improved text/icon/link color legibility by selecting colors that meet contrast requirements across light and dark backgrounds.
* **Tests**
  * Added coverage for legible color selection and event palette/highlight derivation.


jamesarich:
Both CodeRabbit nitpicks addressed in a679d2b:

- **Shared palette strip** — extracted `EventPaletteStrip(palette, height)` into `core:ui`; `MainAppBar` and `EventInfoSheet` both call it, and both copies of the private `MIN_GRADIENT_STOPS` are gone.
- **`distinct()` on the authored palette** — a repeated hex was becoming a repeated gradient stop, flattening the gradient over that span. Both paths now de-dupe, preserving authored order. Covered by a new `brandPaletteDeDupesAuthoredEntriesPreservingOrder` test.

No inline review threads were opened on this PR, so there was nothing to reply to or resolve individually.

PR #6503, opened 2026-07-28
fix(connections): send set_time_only at MyNodeInfo instead of onNodeDbReady

jamesarich:
### 🛠️ Why

The app has sent `set_time_only` on every config handshake since #4103 — but at the *end* of the handshake (`onNodeDbReady`). By then the firmware has already begun flushing its queued packet backlog to the phone, so the backlog is stamped with the device's old clock. Firmware meshtastic/firmware#11274 introduces backlog re-timing, which only helps if the phone's time arrives **before** the packets are requested.

### What changed

- `set_time_only` now goes out the moment the `MyNodeInfo` frame arrives — the first Stage 1 frame, and the earliest point the local node number is authoritative (new `MeshConnectionManager.onMyNodeInfoReceived` hook).
- The send bypasses the outbound packet queue via a new `CommandSender.sendAdminImmediate`: the queue only drains while `Connected`, which would otherwise hold the packet until after the backlog flush (verified on hardware — the queued path delayed it ~4s, landing after the flush began).
- The `get_owner_request` session-passkey seeding stays at `onNodeDbReady`, unchanged.

### 🧪 Testing Performed

- New unit test pins the `sendAdminImmediate` contract: direct `ToRadio` dispatch (not the Connected-gated queue), destination, `ADMIN_APP` portnum, `want_response=false`, and `session_passkey` injection.
- `spotlessCheck detekt :core:data:allTests :core:repository:allTests :core:service:allTests :core:takserver:allTests` all green.
- E2E on Pixel 9 Pro (fdroid debug) + RAK4631 (WisMesh Pocket) over BLE, watching the firmware serial log:
  - **Before:** `Client received set_time_only command` arrived ~100 ms *after* the first backlog `phone downloaded packet`.
  - **After:** it arrives during the Stage 1 config stream (`state=3`), ~5 s *before* `Config Send Complete` and the backlog/replay drain. Handshake completes normally — no GATT contention from the single early write.
  - (Test node has GPS, so firmware correctly ignored the NTP-quality time — the ordering is what was under test.)

## Summary by CodeRabbit

- **Improvements**
  - Device time is now set earlier during connection setup, improving message timing and synchronization.
  - Local node information is processed immediately after handshake progress updates.
  - Administrative commands can be sent immediately when needed, bypassing the normal outbound packet queue.
- **Documentation**
  - Expanded guidance on when local node number information is first authoritative during the connection stream.
- **Tests**
  - Updated and added coverage for immediate admin message dispatch, including passkey injection and expected packet attributes.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/CommandSenderImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/CommandSenderImpl.kt
@@ -182,6 +182,12 @@ class CommandSenderImpl(
         packetHandler.sendToRadio(packet)
     }
 
+    override fun sendAdminImmediate(destNum: Int, initFn: () -> AdminMessage) {
+        val adminMsg = initFn().copy(session_passkey = sessionManager.getPasskey(destNum))
+        val packet = buildAdminPacket(to = destNum, adminMessage = adminMsg)
+        packetHandler.sendToRadio(ToRadio(packet = packet))
+    }
+
     override suspend fun sendAdminAwait(
         destNum: Int,
         requestId: Int,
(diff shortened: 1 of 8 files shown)


jamesarich:
Added regression coverage in `CommandSenderImplTest.sendAdminImmediate_dispatchesDirectToRadioWithPasskeyAndNoResponse` — asserts direct ToRadio dispatch, destination, ADMIN_APP portnum, no want_response, and session_passkey injection (72edaa4d7).

PR #6529, opened 2026-07-31
fix(lora): validate 2.4 GHz bandwidth options

RCGV1:
## Summary

- constrain manual `LORA_24` bandwidth choices to the firmware's canonical wire codes `200`, `400`, and `800`
- expose `1600` only for directly connected PlatformIO targets that are positively identified as SX128x-only
- keep mixed `TLORA_T3_S3`, LR1121, remote, and unknown targets on the conservative three-option set
- require an explicit bandwidth when custom `LORA_24` mode contains the preset-only `0` sentinel, with "Choose a bandwidth" guidance instead of calling it unsupported
- retain genuinely unsupported persisted values in the UI with an explanation, and block Save until a valid value is selected
- refresh bandwidth capability when a same-identity device reports a different `pioEnv` after a reflash
- preserve the existing numeric input for sub-GHz regions and preserve protobuf defaults in preset mode

This follows the behavior established in [Meshtastic-Apple#2225](https://github.com/meshtastic/Meshtastic-Apple/pull/2225) and the firmware validation in [firmware#11317](https://github.com/meshtastic/firmware/pull/11317).

## Capability limitation

Current protobuf metadata does not expose the radio chip selected at runtime. `MyNodeInfo.pio_env` is available only for the directly connected node, so remote and mixed targets cannot safely prove SX128x-only capability. The implementation therefore fails closed: `1600` is enabled only for exact firmware targets verified to compile exclusively for SX128x, while known mixed/LR1121 hardware models explicitly veto it.

Sub-GHz remains a numeric input in this PR. A complete client-side sub-GHz allowlist has the same missing runtime-radio-capability problem and should follow the protocol decision rather than add another partial board matrix here.

## Validation

- merged current upstream `main` without rewriting branch history
- `./gradlew spotlessCheck detekt`: passed
- `./gradlew :feature:settings:allTests`: passed after isolating two pre-existing `Dispatchers.Main` cleanup races; 196 settings tests, including the bandwidth policy, same-node `pioEnv` refresh, and Compose save-gating flow
- isolated dispatcher-race tests passed before the clean module rerun
- full repository gate was attempted; unrelated current-main `ConnectionsViewModelTest` and `NodeDetailCompassLifecycleTest` failures prevent claiming a green full-suite run

No physical devices were accessed. Firmware-side hardware validation is tracked in firmware#11317.


jamesarich:
what discussion, issue, spec, or related upstream work in firmware or protos to support this change? where did it come from and why?

RCGV1:
There was some discussion the discord in the Apple channel. Firmware and apps did not gate the bad with allowed for 2.4ghz radios leading to issues configuring custom bandwidths.

RCGV1:
I also made this pr https://github.com/meshtastic/firmware/pull/11317 to validate the bandwidth options in the firmware.

jamesarich:
Reviewed this together with meshtastic/Meshtastic-Apple#2225 and meshtastic/firmware#11317.

### `bandwidth = 0` is a valid firmware state, and this flags it as unsupported

`loRaBandwidthSelection(0, LORA_24, …)` finds 0 absent from `{200, 400, 800}`, so `invalidPersistedValue = 0` — the picker shows "Unsupported (0 kHz)" and Save is disabled for the whole LoRa screen.

But 0 is firmware's documented "use the default" sentinel, not an unsupported bandwidth. Two places in firmware say so explicitly:

- `Default.h`: *"When use_preset is true, config.lora.spread_factor and bandwidth may be 0 because applyModemConfig() sets them on RadioInterface, not on config.lora"*
- `AdminModule.cpp`: *"In preset mode bandwidth 0 is expected (the preset supplies it), so leave it untouched"*

And `checkOrClampConfigLora` only ever assigns `loraConfig.bandwidth` inside its failure branches, so a healthy preset-mode LORA_24 node genuinely reports 0. Toggling Use Preset off lands on it immediately.

This only bites in custom mode (`allowsSave` short-circuits on `usePreset`) and the user can escape by picking a value, so it isn't a deadlock. Two things are still wrong: the label is factually incorrect, and it gates the entire screen — a node persisted with `use_preset = false` + `bandwidth = 0` can't have its hop limit or TX power changed without also picking a bandwidth. If forcing an explicit choice is the intent, the copy should say that rather than claim the radio can't do it.

`unsetRegion_preservesProtobufDefault` covers `0` for `RegionCode.UNSET` only; a `LORA_24` + `0` case would have caught this.

### Side effect inside `.map`, upstream of `distinctUntilChanged()`

In `RadioConfigViewModel`, the new `_radioConfigState.update` runs on every `myNodeInfo` emission and duplicates the `isLocal` write that the `flatMapLatest` branches already perform. It also writes the new node's `pioEnv` *before* `flatMapLatest` cancels the previous inner flow — which is the window the comment three lines above says doesn't exist.

`.onEach { }` after `distinctUntilChanged()`, or folding `pioEnv` into the `isLocal` branch's `combine`, avoids both.

### Allowlist completeness

Checked `SX128X_ONLY_TARGETS` against firmware `variants/` (uncommented `#define USE_SX1280`, no other radio defined). Beyond the four listed, `my-esp32s3-diy-oled` and `my-esp32s3-diy-eink` also qualify. Fail-closed so not a bug, and I couldn't confirm those ship in release builds — but "all proven SX128x-only targets" in the description isn't quite accurate.

### Sub-GHz coverage gap

The firmware PR also tightens sub-GHz — its own tests assert `{31, 42, 83, 101, 200}` are invalid on LR11x0. This PR leaves sub-GHz as a free numeric `EditTextPreference`, so a Muzi/LR1121 user can still enter 31 and have the device clamp it away. If #11317 lands, the app-side policy should cover sub-GHz too.

### On provenance

Following up my earlier question. The deeper issue is that this allowlist statically mirrors firmware's build matrix in two other repos, because the protocol doesn't expose the runtime radio chip — which the description acknowledges. Three copies of the stopgap means three places to update per new board, with nothing catching drift.

Worth deciding whether a `DeviceMetadata` radio-family or supported-bandwidth field is the right home before all three of these merge.

jamesarich (commented):
The fail-closed design and the test coverage here are solid — the remote-node `pioEnv` scoping (gateway's SX128x can't unlock 1600 kHz for a remote LR1121) is exactly the right call. One question to resolve before merge, plus a housekeeping item:

1. **Stored bandwidth `0` in LORA_24 manual mode blocks Save — is that intended?** `LoRaBandwidthPolicy` treats `bandwidth = 0` as an unsupported persisted value ("Unsupported (0 kHz)") that disables Save. But 0 is the protobuf default meaning "firmware derives bandwidth from the modem preset" — a node switched to `use_preset = false` in LORA_24 without an explicit bandwidth would land in this state, and the user is then unable to save *any* change on the LoRa screen until they pick a bandwidth. If forcing an explicit choice is the intent (consistent with the fail-closed philosophy), could you state that in the PR description and add a bandwidth=0 test? If not, 0 probably deserves a selectable "Default (preset)" entry rather than the blocking "Unsupported" treatment. There's a test for region=UNSET preserving the proto default, but none for bandwidth=0 within LORA_24.

2. **Rebase** — the branch is ~6 weeks behind main, so the green CI is stale; a rebase will re-run it against current code.

RCGV1:
Addressed the review feedback in `4b19b7365` and merged current `main` into the branch.

- Custom `LORA_24` `bandwidth=0` now remains blocked until an explicit choice is made, but the UI says **Choose a bandwidth** and explains the custom-mode requirement instead of labeling zero unsupported. Policy and Compose tests cover preset-vs-custom behavior and Save gating.
- The `myNodeInfo` flow no longer mutates state inside `map`. `(isLocal, pioEnv)` is the distinct key, and the local combine applies both together. A regression test reflashes the same node identity from mixed `tlora-t3s3-v1` to SX128x-only `my-esp32s3-diy-oled` and proves the available 1600 kHz capability changes.
- Added `my-esp32s3-diy-oled` and `my-esp32s3-diy-eink` to the proven SX128x-only targets.
- The PR description now explicitly scopes sub-GHz numeric validation as a protocol-capability follow-up instead of claiming the local board allowlist solves it.

Validation: `spotlessCheck` and `detekt` pass; `:feature:settings:allTests` passes after isolating two existing `Dispatchers.Main` cleanup races. The full repository gate was also attempted and remains red in unrelated current-main firmware-notice and compass lifecycle tests.

RCGV1:
Updated the protobuf-default bandwidth handling in edc65d0f6. Following existing client patterns for proto scalar defaults, wire value `0` is preserved and shown as `Default (812.5 kHz)` for LORA_24 instead of being treated as missing. Unsupported nonzero values remain visible and block saving.

Validation passed:
- `spotlessCheck`
- `detekt`
- focused `LoRaBandwidthPolicyTest` and `LoRaBandwidthUiTest` JVM tests

jamesarich (approved):
Re-reviewed — the bandwidth `0` question is resolved the way I'd hoped: wire value 0 is preserved and shown as a first-class **Default (812.5 kHz)** entry, so a node that landed on the protobuf default can still save, while unsupported nonzero values stay visible-but-disabled with an explanation and block Save until replaced. The policy tests (`lora24ProtobufDefault_…`, `unsetRegion_…`) and the end-to-end UI test cover exactly the cases I asked about, and the conflicting hwModel/pioEnv fail-closed tests are a nice touch. Rebase is done and CI is green against current main.

One housekeeping item before merge: the four new strings aren't alphabetized — `python3 scripts/sort-strings.py` wasn't run (`bandwidth_option_khz` sorts after `bandwidth_default`, and the index has the same order). Please run it and commit the result so the next contributor's run doesn't produce a noisy diff.

Approving with that one nit — everything substantive is addressed.

jamesarich:
Pushed the `sort-strings.py` run for you as `fa8c35426` so this isn't blocked on a round trip — nothing but generated housekeeping, no behaviour change.

It did three things:

- Swapped `bandwidth_default` / `bandwidth_option_khz` into alphabetical order, in both `strings.xml` and `strings-index.txt`.
- Added the `<!-- BANDWIDTH -->` / `### BANDWIDTH ###` section marker, which the script emits automatically now that the `bandwidth_*` group has reached five members.
- Restored the trailing newline at the end of `strings.xml`. Your diff had removed it; the file is now byte-identical to `main` at the tail, so this actually shrinks the diff.

Verified the result is the sorter's canonical form by re-running it and confirming zero further changes, so a future contributor's run won't produce surprise churn.

That was the only outstanding item from my review — the bandwidth `0` handling and its test coverage were already right. Enqueueing now. Thanks for working through the `bandwidth = 0` question; landing on "preserve the wire value and label it `Default (812.5 kHz)`" is the correct call and matches how the rest of the client treats proto scalar defaults.

PR #6538, opened 2026-08-01
fix(nodes): suppress alerts during initial database sync

jeremiah-k:
## Overview

During the initial configuration handshake, `clear()` resets node state before `setNodeDbReady(true)` marks the initial database synchronization complete. Live or firmware-queued `NODEINFO_APP` packets can arrive during that pre-ready window and are processed through `handleReceivedUser()`.

If one of those packets lost a node-state compare-and-set and retried after readiness changed, it could become notification-eligible even though it originally arrived as synchronization traffic. A packet from a previous radio session could similarly retry into freshly-cleared state if the session changed during reduction.

This change pins notification eligibility and radio-session ownership to packet arrival so pre-ready packets remain synchronization traffic across retries and stale-session packets cannot populate a new session.

## Changes

* Capture node-database readiness when a user packet enters `handleReceivedUser()` and preserve that notification-admission decision across CAS retries.
* Capture the current radio-session generation and stop processing if the session is cleared or replaced before the state update commits.
* Continue installing pre-ready user records in memory while leaving persistence controlled by the existing readiness and write gates.
* Keep genuine post-sync discoveries notification-eligible once the node database is ready.
* Preserve existing identity reconciliation, retired-node handling, distinct-key reuse, and notification revalidation behavior.

## Testing

* Added coverage proving a user packet received before node-database readiness installs in memory without persistence or a new-node notification while database writes are otherwise enabled.
* Added a deterministic CAS-race test that changes readiness and node state between reduction and commit, verifies the reducer executes twice, and confirms the packet remains notification-suppressed.
* Added coverage proving a packet from the previous radio session is discarded when `clear()` starts a new session before CAS.
* Updated live-notification tests to explicitly run with the node database marked ready.

## Scope

* This applies to live or firmware-queued `NODEINFO_APP` packets that arrive during the pre-ready handshake window.
* The handshake NodeDB replay path through `installNodeInfo()` is unchanged.
* Node identity matching, persistence formats, radio protocol behavior, and database schema are unchanged.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/NodeManagerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/NodeManagerImpl.kt
@@ -25,6 +25,7 @@ import kotlinx.collections.immutable.persistentMapOf
 import kotlinx.collections.immutable.persistentSetOf
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.update as updateStateFlow
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import okio.ByteString
@@ -50,13 +51,12 @@ import org.meshtastic.core.resources.new_node_seen
 import org.meshtastic.proto.DeviceMetadata
 import org.meshtastic.proto.FirmwareEdition
 import org.meshtastic.proto.HardwareModel
+import org.meshtastic.proto.NodeInfo as ProtoNodeInfo
 import org.meshtastic.proto.Paxcount
+import org.meshtastic.proto.Position as ProtoPosition
 import org.meshtastic.proto.StatusMessage
 import org.meshtastic.proto.Telemetry
 import org.meshtastic.proto.User
-import kotlinx.coroutines.flow.update as updateStateFlow
-import org.meshtastic.proto.NodeInfo as ProtoNodeInfo
-import org.meshtastic.proto.Position as ProtoPosition
 
 /**
  * Resolves a validated public-key correlation hint from a raw [ByteString]. Returns null unless the key is exactly
@@ -172,7 +172,7 @@ class NodeManagerImpl(
                     .sortedWith(
                         compareByDescending<Node> { isDefaultIdentityPlaceholder(it) }
                             .thenBy { it.lastHeard }
-                            .thenBy { it.num },
+                            .thenBy { it.num }
                     )
             val toDrop = evictable.take(byNum.size - maxNodes)
             return toDrop.fold(this) { index, node -> index.remove(node.num) }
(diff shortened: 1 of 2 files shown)


jamesarich (commented):
The session-generation guard here is correct and worth having on its own. Before this, a `clear()` landing mid-reduction would fail the CAS, retry against the freshly-cleared state, and install the previous session's node into the new one. Pinning the generation before the loop, re-checking after the reduction hook, and exiting on mismatch is the right shape, and `User packet is discarded when clear starts a new session before CAS` covers it properly.

One substantive request, and one on the tests.

## The description names a path that can't reach this code

> Initial node-database replay uses the same user-packet reduction path as live discoveries.

It doesn't, as far as I can trace it. The handshake replay goes `HandshakeState.ReceivingNodeInfo.nodes` → `installNodeInfo()` (`NodeManagerImpl.kt:742`) → `updateNodeState()`, which never constructs a `ReceivedUserTransition` and has no notification path at all. `handleReceivedUser` has four production call sites: `MeshDataHandlerImpl.kt:366` (live NODEINFO_APP mesh packets), `MessagingControllerImpl.kt:123` (contact import), and `AdminControllerImpl.kt:64`/`:82` (setOwner / ham mode). The device's NodeDB replay isn't among them.

What the change actually covers is the window between `clear()` and `setNodeDbReady(true)` (`MeshConfigFlowManagerImpl.kt:303`) — during which live and firmware-queued NODEINFO_APP mesh packets arrive and, before this PR, each produced a new-node alert. That is a real bug and this is a reasonable fix for it; the title "suppress alerts during initial database sync" describes it accurately. It's just that the mechanism in the body is the wrong one, and anyone approving on the replay story is approving something else.

Could you update the Overview to describe the pre-ready window and the queued-mesh-packet path? The code can stay as-is.

(The other two call sites are effectively unreachable while `isNodeDbReady` is false, since `clear()` nulls `myNodeNum` and both early-return on that — so the blast radius really is just the handshake window. Worth stating explicitly in Scope.)

## `initial node database replay installs nodes without new-node notifications`

Two things:

1. It calls `handleReceivedUser`, not `installNodeInfo`, so it doesn't exercise the replay path its name claims. Renaming it to something like `user packet before node-db ready does not notify` would make it say what it tests.

2. The `verifySuspend(mode = not) { nodeRepository.upsert(any()) }` assertion passes because `shouldPersist()` (`NodeManagerImpl.kt:584`) already requires `isNodeDbReady`, which this PR didn't change. Per the repo's own review guidance (`.skills/code-review/SKILL.md`, "Tests that pass for the wrong reason" — *assert the side effect only the intended path produces*), that half of the test would still pass with this change reverted. The notification assertion is the one carrying weight; the upsert assertion is testing a pre-existing gate.

`initial-sync packet stays notification-suppressed across a CAS retry` is the better of the two — driving the retry through `setMyNodeNum` in the hook is a clean deterministic race. Consider counting reductions rather than a boolean `reductionHookRan`, so the test proves a retry actually occurred rather than just that the hook fired once.

jeremiah-k:
@jamesarich Thanks — you were right about the path distinction. I retraced the call sites and updated the description to cover the actual pre-ready window: live or firmware-queued `NODEINFO_APP` packets reaching `handleReceivedUser()` after `clear()` and before `setNodeDbReady(true)`. The handshake NodeDB replay itself remains on `installNodeInfo()`.

I also renamed the test so it no longer claims to exercise replay, enabled DB writes while leaving readiness false so the persistence assertion isolates that gate, and changed the CAS-race test to count reductions and require exactly two passes.

The session-generation guard and production behavior otherwise remain unchanged.

jamesarich (approved):
Re-reviewed at `a360b603`. All three points addressed, and I checked the code rather than the summary.

- The Overview now describes the actual pre-ready window, and the replay claim is gone. Accurate to what `handleReceivedUser` can be reached by.
- `user packet before node-db ready installs without persistence or notification` is honestly named now, and `setAllowNodeDbWrites(true)` with readiness left false makes `isNodeDbReady` the only false term in `shouldPersist()` — the assertion is unconfounded even though it documents a pre-existing gate.
- The CAS test counting reductions and asserting exactly 2 is the one carrying the weight: `allowNotificationAtArrival` is pinned at arrival, so the retry after `setNodeDbReady(true)` stays suppressed. That fails if the change is reverted, which is what I wanted from it.

Adding `setNodeDbReady(true)` to the six pre-existing tests is the correct fallout of pinning admission at arrival rather than a workaround — those cases intend live-discovery semantics.

LGTM.

PR #6539, opened 2026-08-01
fix(ble): pause background RSSI polling

jeremiah-k:
## Overview

The connections screen's RSSI coroutine could continue issuing BLE GATT reads while the app was backgrounded because the polling delay itself does not depend on Compose frame production.

This change makes RSSI reads follow the owning lifecycle, so polling exists only while the app is at least `STARTED`. Connection-state changes are observed directly so disconnecting cancels the active read loop immediately and reconnecting resumes it without a separate timer poll.

## Changes

* Run RSSI polling only while the owning lifecycle is at least `STARTED`.
* Use a 3-second polling interval while the BLE device is connected and the screen is active.
* Observe `BleDevice.state` directly with `collectLatest`.
* Cancel the active RSSI loop and clear stale RSSI whenever the device leaves the connected state.
* Resume polling immediately when the same device reconnects, without periodic disconnected-state polling.
* Avoid Compose state writes when the RSSI value has not changed.
* Preserve the existing bounded RSSI-read timeout and propagate coroutine cancellation normally.
* Preserve the throwable and stack trace for unexpected read failures while avoiding explicit interpolation of arbitrary exception messages.

## Testing

* Added JVM Compose coverage with an explicit lifecycle owner.
* Verified RSSI reads occur while the lifecycle is `STARTED`.
* Verified reads stop when the lifecycle drops below `STARTED` and resume when it returns.
* Verified BLE disconnect stops reads and clears the active polling loop.
* Verified reconnect resumes RSSI reads directly from connection-state observation rather than a timer poll.

## Scope

* RSSI is still read only for the currently displayed BLE device.
* Connection establishment, transport ownership, reconnect policy, packet traffic, and device configuration are unchanged.
* No persistence or protocol changes are introduced.

### DIFF
--- a/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/components/CurrentlyConnectedInfo.kt
+++ b/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/components/CurrentlyConnectedInfo.kt
@@ -37,6 +37,9 @@ import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.text.style.TextOverflow
 import androidx.compose.ui.unit.dp
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.compose.LocalLifecycleOwner
+import androidx.lifecycle.repeatOnLifecycle
 import co.touchlab.kermit.Logger
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.TimeoutCancellationException
@@ -58,7 +61,7 @@ import org.meshtastic.proto.Paxcount
 import org.meshtastic.proto.User
 import kotlin.time.Duration.Companion.seconds
 
-private const val RSSI_DELAY = 2
+private const val RSSI_DELAY = 3
 private const val RSSI_TIMEOUT = 1
 
 /**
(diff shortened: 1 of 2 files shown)


jamesarich (commented):
The lifecycle scoping is the right fix and it closes the leak completely. Worth spelling out why, since it also bears on the interval change:

`MeshtasticNavDisplay` registers only the saveable-state and ViewModel-store decorators (`MeshtasticNavDisplay.kt:131-132`) — no per-entry lifecycle decorator — so `LocalLifecycleOwner` here resolves to the Activity/window rather than the nav entry. In-app navigation was already covered, because Navigation 3 disposes the composition of entries that aren't on top. The case that actually leaked is "screen still composed, app backgrounded": `delay()` doesn't need the frame clock, so the loop kept polling GATT indefinitely. `repeatOnLifecycle(STARTED)` closes exactly that. Good fix.

## Can we keep the 2s interval?

Given the above, the lifecycle gate already eliminates all the background traffic on its own. The remaining polling only happens while the screen is visible and in front of the user — which is precisely when RSSI is being watched, e.g. aiming an antenna or walking a device around a site to find dead spots. At 10s that workflow gets noticeably worse, and it's a separate UX tradeoff from the bug being fixed.

I'd suggest keeping `RSSI_DELAY = 2`, or 3-5s as a compromise, and letting the lifecycle scoping carry the win. Happy to be overruled if there's a specific stack-pressure symptom driving the 10s number — if so, could you mention it in the description?

## Keep the throwable in the log

```kotlin
Logger.d { "Failed to read RSSI; cause=${e::class.simpleName ?: "Exception"}" }
```

This drops the stack trace that `Logger.d(e)` carried. The privacy rationale doesn't quite apply here — the original logged `e.message`, not PII — and for a wedged GATT connection the stack is the useful part. `Logger.d(e) { "Failed to read RSSI" }` keeps it without interpolating arbitrary exception text, which I think is what you were actually after.

## Minor

- The `while (true)` + `DISCONNECTED_POLL_DELAY` branch replaces the old `while (isConnected)` exit, so it now ticks at 1Hz forever while the screen is visible and the device is disconnected. Cheap, but it's a poll where an observable would do. If `DeviceListEntry.Ble` exposes connection state as a flow, keying the effect off it removes the loop entirely.

- No test, while the other two PRs in this batch add coverage. `feature:connections` already has `jvmTest` wired for Compose UI tests (`build.gradle.kts:47`); a test that drives a `TestLifecycleOwner` through STARTED → STOPPED and asserts `readRssi` stops being called would lock in the fix and prevent it regressing the next time this composable is touched.

jeremiah-k:
@jamesarich Thanks, these were good catches. I tightened this back around the lifecycle issue.

The visible polling cadence is now 3 seconds rather than 10. Ten seconds was an early adjustment made while initially chasing the issue, and 3 seconds is a better middle ground. The disconnected 1 Hz loop is gone; `BleDevice.state` is observed directly with `collectLatest`, so disconnect cancels the active read loop and clears stale RSSI, and reconnect starts reads again immediately.

I also kept the throwable with `Logger.d(e) { "Failed to read RSSI" }` so the stack remains available without interpolating the exception message.

Finally, I added a JVM Compose lifecycle test covering STARTED → below STARTED → STARTED plus disconnect/reconnect behavior. The test verifies reads actually stop and resume on those transitions.

jamesarich (approved):
Re-reviewed at `be6e77ba`. This came back better than what I asked for.

Observing `BleDevice.state` with `collectLatest` is the right shape — disconnect cancels the active read loop and clears stale RSSI, reconnect resumes immediately, and the 1 Hz disconnected poll is gone entirely rather than just made cheaper. 3s is a sensible cadence now that the lifecycle gate carries the background win. `Logger.d(e) { "Failed to read RSSI" }` keeps the stack without interpolating exception text.

`CurrentlyConnectedInfoRssiLifecycleTest` drives STARTED → CREATED → STARTED plus disconnect/reconnect and asserts read counts actually stop and resume, so the fix is locked in against the next edit to this composable.

LGTM.

PR #6540, opened 2026-08-01
fix(database): recover wedged observer pools

jeremiah-k:
## Overview

Long-lived Room `Flow` observers can remain attached to a database connection pool that has stopped granting reader or writer connections. Existing bounded database operations already recover the recognized pool-acquisition timeout, but a failing DAO `Flow` could otherwise remain latched to the failed pool indefinitely.

This change adds a shared observation path that can replace the still-current database after that specific timeout and automatically relatch long-lived DAO flows to the replacement, while bounding both repeated failures and lifetime pool retention.

## Changes

* Add `DatabaseProvider.observeCurrentDb()` as the explicit contract for long-lived DAO `Flow` observation.
* Require each `DatabaseProvider` implementation to define its relatching policy rather than silently inheriting a non-recovering default.
* On the existing recognized Room connection-pool acquisition timeout, replace only the still-current database and let `flatMapLatest` attach observers to the fresh pool.
* Never replay the failed query on the same database instance.
* Preserve coroutine cancellation and propagate unrelated database or query failures unchanged.
* Converge concurrent recovery attempts on the same replacement instead of reopening once per failing collector.
* Limit each collector to three consecutive automatic Flow pool recoveries without a successful emission.
* Independently limit successful Flow-triggered replacements to six for the lifetime of a `DatabaseManager`, preventing intermittent recover/emit/wedge cycles from accumulating detached Room instances without bound.
* Keep the manager ownership-transfer portion of an admitted reopen non-cancellable so a replacement cannot be abandoned between construction, cache replacement, and publication.
* Route existing long-lived read flows through the shared helper while leaving one-shot write/replay semantics unchanged.
* Keep non-production providers on explicit ordinary `flatMapLatest` relatching behavior.

## Testing

* Added coverage proving a Flow pool timeout reopens the database and relatches without repeating the failed pool.
* Added coverage proving three consecutive failing replacement pools exhaust the per-collector recovery budget.
* Added coverage proving successful emissions do not reset the manager-lifetime recovery budget and intermittent failures stop after six Flow-triggered replacements.
* Added coverage proving shutdown suppresses recovery and does not build another database.
* Added coverage proving unrelated database failures propagate unchanged and do not trigger a replacement.
* Existing shutdown and detached-pool ownership coverage continues to verify retained databases are reclaimed during orderly close.

## Scope

* Recovery remains limited to the existing recognized Room connection-pool timeout condition.
* Ordinary query failures are not retried.
* One-shot writes are not replayed.
* Recovery used by bounded one-shot operations remains separate from the Flow-specific lifetime budget.
* Existing PagingSource creation from `currentDb.value` is unchanged; paging relatching is a pre-existing follow-up concern rather than part of this change.
* Database schema, persisted data, DAO contracts, and migration behavior are unchanged.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceLinkLocalDataSource.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/DeviceLinkLocalDataSource.kt
@@ -18,7 +18,6 @@ package org.meshtastic.core.data.datasource
 
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.flatMapLatest
 import org.koin.core.annotation.Single
 import org.meshtastic.core.database.DatabaseProvider
 import org.meshtastic.core.database.entity.DeviceLinkEntity
@@ -27,7 +26,7 @@ import org.meshtastic.core.database.entity.DeviceLinkEntity
 class DeviceLinkLocalDataSource(private val dbManager: DatabaseProvider) {
     @OptIn(ExperimentalCoroutinesApi::class)
     fun observeAll(): Flow<List<DeviceLinkEntity>> =
-        dbManager.currentDb.flatMapLatest { db -> db.deviceLinkDao().observeAll() }
+        dbManager.observeCurrentDb { db -> db.deviceLinkDao().observeAll() }
 
     suspend fun getAll(): List<DeviceLinkEntity> = dbManager.withReadDb { it.deviceLinkDao().getAll() }
 
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/EventFirmwareEditionLocalDataSource.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/EventFirmwareEditionLocalDataSource.kt
@@ -18,7 +18,6 @@ package org.meshtastic.core.data.datasource
 
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.withContext
 import org.koin.core.annotation.Single
 import org.meshtastic.core.database.DatabaseProvider
@@ -40,7 +39,7 @@ class EventFirmwareEditionLocalDataSource(
 
     @OptIn(ExperimentalCoroutinesApi::class)
     fun observeByEdition(edition: String): Flow<EventFirmwareEditionEntity?> =
-        dbManager.currentDb.flatMapLatest { db -> db.eventFirmwareEditionDao().observeByEdition(edition) }
+        dbManager.observeCurrentDb { db -> db.eventFirmwareEditionDao().observeByEdition(edition) }
 
     suspend fun upsertAll(editions: List<EventFirmwareEditionEntity>) {
         withContext(dispatchers.io) { dbManager.withDb { it.eventFirmwareEditionDao().upsertAll(editions) } }
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/SwitchingChannelSetDataSource.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/datasource/SwitchingChannelSetDataSource.kt
@@ -18,7 +18,6 @@ package org.meshtastic.core.data.datasource
 
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.distinctUntilChanged
-import kotlinx.coroutines.flow.flatMapLatest
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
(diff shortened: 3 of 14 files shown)


jamesarich (commented):
Traced this one carefully against the existing recovery path — the design is right and the tests are unusually precise. A few things worth confirming before merge.

## What I verified and liked

- It extends the existing policy rather than inventing a second one: same narrow `isDbPoolAcquireTimeoutException` matcher, same never-replay-on-the-same-pool rule, same detached-pool discipline as `withCurrentDb` (`DatabaseManager.kt:1171`).
- **Convergence genuinely works.** `reopenFlowDatabaseIfStillCurrent` releases the mutex before `reopenActiveDatabaseIfStillCurrent` re-acquires it, so two collectors can both pass the first check — but the re-check at `:873` runs under the lock and rejects the loser. One wedge event produces one replacement pool, not one per collector.
- `shouldRethrowFlowFailure` reading `currentDbState` directly, to avoid `_currentDb`'s lazy `getOrCreateCurrentDbState()` rebuilding a pool while shutdown is clearing the publication, is a subtle and correct detail.
- The conversion is complete — no `currentDb.flatMapLatest` sites remain anywhere outside the implementation and the interface default.
- `flowPoolRecoveryStopsAfterConsecutiveReplacementPoolsFail` asserting the *detached-pool count* rather than just the throw is the right assertion to have written.

## Main question: the recovery bound is per-streak, not per-session

`MAX_CONSECUTIVE_FLOW_POOL_RECOVERIES` bounds a run of failures with **no successful emission in between** — `onEach { consecutivePoolRecoveries = 0 }` resets the counter on every DAO emission. Meanwhile `detachedDatabases` is append-only (`:903`), cleared only in `close()` (`:1602`), and `:1311` explicitly protects those names from cache eviction.

So an intermittent wedge pattern — wedge, recover, emit, wedge again — reopens without limit, and each reopen retains an open Room instance and its file descriptors for the rest of the process.

That retention is pre-existing in the `withDb` path, so this isn't a regression you introduced. But this PR widens the trigger set from `withDb` callers to every long-lived flow collector in the app, which multiplies how often it happens. Is a session-lifetime cap (or at least a counter/metric on total reopens) worth adding here, so a persistently sick pool can't accumulate detached instances indefinitely?

## Minor

1. **`consecutivePoolRecoveries` is an unsynchronized `var`** crossing coroutine boundaries — reset in `onEach`, incremented in `catch`, across `flatMapLatest` inner flows. `flatMapLatest`'s cancel-before-collect ordering makes this benign in practice, but `atomic(0)` costs nothing and removes the question. Note also that the reset runs on every DAO emission, i.e. the hot path.

2. **The `NonCancellable` reopen can outlive its collector.** It takes the manager `mutex`, which `associateDevice` holds across `drainWriters` (bounded at `WRITER_DRAIN_TIMEOUT_MS` = 5s) plus datastore and merge I/O. A ViewModel scope cancelled in that window won't finish tearing down until the reopen returns. Consistent with the `withDb` discipline, so probably intended — flagging only so it's a decision rather than an accident.

3. **`DatabaseProvider.observeCurrentDb` has a default impl** that silently falls back to plain `flatMapLatest` with no recovery. Only `DatabaseManager` implements it in production today, so this is fine — but a future implementor loses recovery with no compile error. Consider making it abstract and letting the fakes opt in explicitly.

## Follow-up, not a blocker

`PacketRepositoryImpl` still builds paging sources from `dbManager.currentDb.value` (lines 69, 181, 210) — the message and contact lists. Those don't re-latch on a wedge, but they don't re-latch on an ordinary device switch either, so it's a pre-existing gap this PR neither causes nor worsens. Worth a follow-up issue rather than scope creep here.

jeremiah-k:
@jamesarich Thanks for tracing this one so carefully. The lifetime-bound concern was valid.

There are now two independent limits: three consecutive recoveries per collector, plus six successful Flow-triggered replacements for the lifetime of the `DatabaseManager`. A successful DAO emission still resets the local failure streak, but it does not reset the manager-wide budget, so an intermittent wedge → recover → emit cycle can no longer accumulate detached pools indefinitely.

The manager-wide check and increment happen under the same mutex as the reopen/publication path, so concurrent collectors cannot race past the limit.

I also changed the local streak counter to `atomic`, made `DatabaseProvider.observeCurrentDb` abstract and gave the test providers explicit plain `flatMapLatest` behavior, and documented why the admitted ownership transfer remains `NonCancellable`. The shutdown and unrelated-failure tests now also assert that no replacement database was built.

I left the PagingSource/currentDb.value behavior alone as the pre-existing follow-up you called out rather than expanding this PR further.

jamesarich (approved):
Re-reviewed at `7a6fabd1`. The lifetime bound is correct — I traced the concurrency rather than taking the summary for it.

`flowPoolRecoveriesThisLifetime`'s check (`DatabaseManager.kt:895`) and increment (`:933`) sit in the same `mutex.withLock` scope as the build and the `_currentDb` publication, so concurrent collectors can't race past the budget. Exhausting it returns null, `shouldRethrowFlowFailure` is true because the database is still current, and the original timeout surfaces — so it degrades to pre-PR behavior instead of wedging silently. That's the right fallback.

`flowPoolRecoveryStopsAfterIntermittentFailuresReachTheManagerLifetimeLimit` is the assertion that actually proves it: 7 successful emissions against 6 builds tests the wedge → recover → **emit** → wedge cycle specifically, which is the pattern the per-collector streak alone could not bound.

The rest is all in: `atomic` for the streak counter, `observeCurrentDb` made abstract with both fakes opting into plain `flatMapLatest` explicitly, the `NonCancellable` ownership transfer documented as a decision, and the shutdown/unrelated-failure tests now asserting no replacement was built.

One observation, explicitly not blocking: the budget is never reset — not on `switchActiveDatabase`, and `lifecycleState` never returns to `OPEN` after `close()`, so "manager lifetime" is accurate as documented. A device switch builds a fresh pool but still spends from the old budget. Arguable either way, and the hard bound is the safer default; noting it only so a future reader knows it was considered.

Paging left as the pre-existing follow-up — agreed, that shouldn't grow this PR.

LGTM.

PR #6598, opened 2026-08-08
fix(lifecycle): harden packet admission and transport ownership

jeremiah-k:
Superseded by #6716 #6717 #6718

## Overview

This makes outbound packet admission explicit across the queue, service, and transport layers and keeps admitted work bound to the connection or transport generation that owns it.

Rejected or stale sends no longer create dispatch evidence, response timers, optimistic state, or follow-on work. Accepted sends retain their ownership through acknowledgements, cancellation, reconnects, and teardown. The same lifecycle boundaries are applied to settings transactions, post-handshake requests, remote-admin session refresh, and discovery home restoration.

## Review Guide

The five commits are intended to be reviewed in order:

1. **Packet ownership and connection lifecycle** — reserves packet ownership through terminal completion, correlates connection epochs, and applies those boundaries to settings commits.
2. **Transport admission** — makes transport handoff explicit, binds queued writes to active sessions, and rejects work that cannot be admitted.
3. **Remote-admin latency** — allows a realistic response window for multi-hop metadata/session refresh without changing late-response persistence.
4. **Lifecycle hardening** — closes teardown/callback races, binds post-handshake work to its connection generation, and serializes discovery cleanup and restoration.
5. **Review-identified state races** — closes remaining request, transport, and discovery edge cases found during review and aligns supporting test contracts with production lifecycle semantics.

### Packet admission and command outcomes

* Reserve packet IDs from admission through terminal completion so duplicate sends cannot take ownership of acknowledgements.
* Distinguish queue rejection, radio rejection, timeout, transport departure, and successful response outcomes.
* Publish dispatch evidence only after transport admission and ignore acknowledgements that arrive before dispatch ownership exists.
* Avoid optimistic status, timer, position, or request side effects when the underlying packet was not admitted.
* Treat terminal routing/admin failures as the end of the current request flow so stale request timers cannot replace the specific failure later.
* Reject requests that require a local node identity when that identity is unavailable instead of substituting node number `0`.

### Transport lifecycle

* Require transport handoff to report synchronous admission across BLE, serial, TCP, replay, mock, and inert transports.
* Bind queued writes and callbacks to the transport generation that admitted them so replacement connections cannot inherit stale work.
* Bound pending stream sends and drain or cancel admitted work before physical teardown.
* Keep connect/disconnect callbacks inside lifecycle admission and reject stale serial-device selection before transport creation.
* Treat bounded teardown timeouts as operational failures to report rather than exceptions that can abort the remaining shutdown sequence.
* Preserve legacy single-device USB serial address recovery while rejecting ambiguous stale selections.
* Release Android serial operation ownership before wake-failure cleanup begins so teardown cannot depend on draining the operation that initiated it.

### Admin and connection lifecycle

* Stage local settings projections behind accepted commit boundaries and preserve the original edit failure when commit cleanup also fails.
* Capture local-destination ownership consistently for the lifetime of a settings transaction.
* Bind post-handshake passkey and telemetry requests to the connected lifecycle generation that scheduled them while keeping independent requests isolated from sibling failures.
* Give remote-admin metadata/session refreshes a 30-second user-facing deadline so multi-hop responses beyond the previous 10-second window can complete normally.
* Keep late remote-admin responses eligible to refresh durable session state.

### Discovery recovery

* Serialize terminal session persistence and home-radio restoration.
* Keep interrupted-session recovery bound to the selected device and prevent recovery from reconfiguring the radio underneath a newly admitted scan.
* Wake pending restoration when either connection state or the selected device changes.
* Prevent overlapping starts or resets from displacing active scan/cleanup ownership.
* Bound initial radio-configuration snapshot acquisition so scan startup cannot suspend indefinitely before publishing a result.
* Keep optional NeighborInfo requests best-effort so a request failure cannot terminate the scan loop before home-radio restoration.
* Preserve the intended terminal status for completed sessions that are still awaiting restoration, including legacy rows without a usable restore plan.
* Preserve terminal status when a late background restore completes after another terminal path has already finalized the session.

### User-facing rejection handling

* Surface rejected node and configuration requests without starting success timers or applying local success state.
* Keep queue/session rejection typed through command APIs so callers can distinguish admission failure from a later response timeout.
* Resolve asynchronous node-request feedback without blocking a coroutine while loading string resources.

## Testing

Added or extended coverage for:

* packet-ID reservation, duplicate admission, dispatch ordering, acknowledgement handling, timeout, cancellation, and reconnect generations;
* unavailable local-node identity, NeighborInfo request-ID normalization, and terminal request cleanup;
* unavailable/opening transports, bounded send queues, completion ownership, close/send races, stale callbacks, replacement sessions, legacy USB selection, and wake-failure cleanup;
* edit-settings begin/commit boundaries, failure preservation, cancellation cleanup, local departure evidence, and transaction-scoped destination ownership;
* remote-admin timeout, cancellation, in-flight request sharing, routing errors, and responses arriving after the former 10-second deadline;
* discovery restoration ordering, interrupted and legacy pending sessions, selected-device changes, overlapping cleanup, bounded startup, rejected restore writes, and active-scan ownership;
* rejected node requests and optimistic side-effect suppression;
* lifecycle-aware fakes and test fixtures so repeated connection states and configured command failures match production behavior.

On-device F-Droid validation exercised BLE, Wi-Fi/TCP, and USB serial transport paths, including transport switching, handshake, packet admission, node and configuration exchange, telemetry, admin-session refresh, and remote-admin traffic. All three transports completed normal connection setup and handshake, carried admitted outbound traffic, and continued exchanging expected queue, configuration, telemetry, and session data. Transitions between transports also cleared the outgoing session and established a new connection generation without stale work being carried forward.

## Scope

* No changes to Meshtastic protocol behavior, message formats, or firmware-level communication are introduced.
* Internal transport and command-admission contracts are updated consistently across in-repo implementations and callers.
* Existing reconnect policy and transport selection behavior are unchanged except where required to reject stale or unowned work or preserve legacy single-device serial recovery.


jeremiah-k:
@coderabbitai I consolidated the follow-up here rather than posting a series of small inline replies.

The issues raised in this CodeRabbit review have been addressed across the current five-commit series:

* `CommandSenderImpl.requestNeighborInfo` now rejects an unavailable local identity instead of constructing a request with node `0`. The discovery caller treats the optional NeighborInfo request as best effort, contains request failures, and still propagates cancellation.
* Legacy Android USB serial selections retain the single-device recovery path in both address validation and connection resolution, while ambiguous stale selections remain rejected.
* BLE and TCP lifecycle timeouts are reported without throwing from teardown. Replay teardown follows the same rule, and TCP now derives its operation-drain bound from its operation timeout. The JVM serial `stat` lookup is bounded and always releases its subprocess.
* Restore waits observe both connection state and selected-device address, so a device change can wake an otherwise stationary connection wait. Retryable interrupted-session restore failures are logged as warnings rather than reporting expected reconnect behavior as defects.
* Recoverable discovery rows distinguish unfinished scans from terminal sessions awaiting restoration. `restore_pending_*` states are mapped back to their intended terminal status, including the no-restore-plan path, and the DAO contract documents that recoverable rows are not necessarily unfinished scans.
* The bounded home-restore wait runs outside the scan mutex. The final race check under the mutex is non-suspending and verifies pending-restore ownership before scan admission, so stop/reset/packet collection are not held behind the restore timeout.
* The scan-start refusal paths were consolidated so the failure state is applied consistently without duplicating the mutex-protected logic.
* `UiText.resolve()` uses the suspend resource path, including nested `StringResource` and `UiText` arguments, rather than blocking through the synchronous resolver.
* The routing-error timeout regression test now verifies that the underlying request was actually initiated before virtual time is advanced.
* `ReplayFuzzTest` starts the replay transport before sending arbitrary outbound data, so the test continues to exercise the decode/tolerance path rather than succeeding on pre-start rejection.
* The discovery test support was consolidated where the fakes had the same contract: packet-collection coverage now uses `SharedInMemoryDiscoveryDao`, and restore timing uses the named production settle-delay constant.
* `DiscoverySessionStatus.RECOVERABLE` is declared from the status constants rather than parsed from the Room SQL literal, with a common test asserting that the in-memory set and SQL filter stay synchronized.
* Settings transactions capture whether the destination is local once and use that same value for staging and commit-boundary handling, so a changing local-node identity cannot make the two halves of one transaction disagree.

While applying those fixes, I also tightened a few adjacent lifecycle cases for consistency:

* Discovery startup now bounds acquisition of its initial radio-configuration snapshot rather than allowing the start call to suspend indefinitely before publishing a result.
* Scan preparation carries the selected-device identity and connection-lifecycle version across the suspending persistence step and revalidates them before publishing the session; stale provisional state is removed if ownership changed meanwhile.
* Home restoration uses capped exponential retry backoff while it still owns the selected device, avoiding a hard retry limit without retrying at a fixed high rate during a prolonged admission outage.
* Android serial wake-failure cleanup begins only after the admitted operation lease is released, avoiding teardown depending on the operation that initiated it.
* Supporting lifecycle fakes and tests were aligned with production version/admission semantics, including completion ownership and zero-request-ID coverage.

jamesarich (commented):
First: this is impressive work — the admission-through-terminal-completion model is the right shape, the five review-ordered commits and the on-device validation across all three transports show real care, and roughly half the added lines being tests is exactly what this code needs. Three structural asks before line-level review:

1. **Please split this into its three actual PRs.** (a) Packet admission + transport ownership — the true subject; (b) the discovery-restoration rewrite (~900 production lines, three new coordinator classes); (c) the remote-admin 10s→30s deadline + dedup fix. They're separable with small interfaces, and (a) touches the most safety-critical path in the app — if a regression ships we need to be able to bisect and revert it without dragging the discovery rewrite along. (b) and (c) will merge faster on their own, too.

2. **The session-lease invariant softening needs a different fix.** `releaseSessionOperation` downgrades the "session changed before release" and underflow `check()`s to `Logger.e`, while `admittedSessionOperations` stays a single counter shared across session generations. A late release from a dead generation can decrement the new session's count, complete a drain waiter early, and let teardown proceed under a live operation — the exact class of bug #6587 closed. The reentrancy case behind the softening (`ReentrantRestartTransport`) is real, but the answer is generation-scoped accounting — which `TransportLifecycleGate` already demonstrates per-transport — rather than log-only checks.

3. **Consider a sealed result over throwing from `CommandSender`.** `PacketQueueRejectedException`/`LocalNodeUnavailableException` make every future call site a latent crash if a catch is forgotten; a sealed `SendOutcome` makes "radio busy" unrepresentable as a crash. If you'd rather keep exceptions, the current-caller sweep needs verification at `InstallProfileUseCase` → `setFixedPosition` in particular.

Heads-up on sequencing: #6603 touches the same admission seams (`PacketHandlerImpl`/`MeshDataHandlerImpl`) and is small and green, so it will likely land first — part (a) will need a rebase over it, and your `PendingResponse`/typed-outcome machinery may subsume its queue/routing split entirely, which would be a welcome simplification in (a).

jeremiah-k:
Superseded by #6716 #6717 #6718

Issue #2659, opened 2025-08-07
[Bug]: 2.6.33 - 2.7 +  Phone battery drain


mjvolante:
### What happened?

Noticing excessive battery drain since 2.6.33. can only measure it by saying phone (pixel 7) has hit 20% warning each night since update. See screenshot. Did not occur on previous versions.

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/bb98bde4-fe08-4835-be9c-983310503c25" />

### App Version

2.6.33

### Phone

Pixel 7, Android 15 latest update.

### Device

T1000e

### Firmware

2.6.11

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

DaneEvans:
Can you roll back and get a comparison? 

I'd say the BT issues are likely to blame, but it would be good to confirm that this is real, not just a perception.

mjvolante:
Please advise how to roll back to 2.6.32.

mjvolante:
Does not happen on 2.6.30. Half the battery usage with similar same screen and background use compared to 2.6.33.

Pitel:
I've had it since 2.6.32. And now with 34. With both BT and WiFi connection.

As a workaround, I've disabled background run in Android settings for the app.

DaneEvans:
Evidence that it's still present on a current build. Include details of phone hardware.
And if it's Samsung its almost certainly their fault. 

I've run it basically constantly on my Moto and not seen any obvious impact.

ramseymarka:
Updated to 2.7.1 directly from 2.6.3. Battery drain went very high. Had to sideload back to 2.6.3 and can verify it fixed battery drain issue.

dzaragoza:
I have the issue in 2.7.0 and 2.7.1, it is consuming excessive battery when I leave the app connected to a Bluetooth node. Workaround is to disconnect from the node and only reconnect when I'm checking messages.

weebl2000:
Same issue with Pixel 6 and 2.7.1, will try to figure out why.

Timendus:
Same here on a Fairphone 5. Has been great with battery life until Meshtastic app update. Also the node list seems really sluggish from time to time, could be related?

![Screenshot_20251010-224307~2.png](https://github.com/user-attachments/assets/28cc5b6e-38c0-4a6e-ade1-5b4176e2a1c0)

Edit: version 2.7.2 by the way

[...]

PR #6602, opened 2026-08-11
fix(map): scope location tracking to the visible lifecycle

simulationstation:
## Summary

The Google map starts a high-accuracy fused-location request from a Compose effect. Backgrounding the activity does not remove the map from composition, so callbacks can continue every 2–5 seconds after `ON_STOP`.

### Bug fixes

- Run Google map location updates only while the lifecycle is at least `STARTED` and tracking plus permission remain enabled.
- Remove location callbacks on stop, disable, permission loss, or disposal, and resume them on start only when tracking is still enabled.
- Replace direct window flag mutation in both map flavors with the project's disposal-safe `KeepScreenOn` utility.
- Add regression coverage for stop, resume, disable, and composition disposal.

This addresses only map-owned tracking; it does not change the separate "provide phone location" flow or BLE connection behavior.

Related to #2659, specifically https://github.com/meshtastic/Meshtastic-Android/issues/2659#issuecomment-3728447817

## Root cause

The previous `LaunchedEffect` was keyed only to the tracking toggle and permission. Activity lifecycle changes did not cancel it, and the `DisposableEffect` cleanup ran only when the composable left composition. `LifecycleStartEffect` gives the callback registration an explicit `STARTED`-lifecycle owner while preserving the user's tracking toggle across background/foreground transitions.

## Testing Performed

- `:androidApp:spotlessKotlinCheck`
- `:androidApp:detekt`
- `:androidApp:compileFdroidDebugKotlin`
- `:androidApp:compileGoogleDebugKotlin`
- `:androidApp:testFdroidDebugUnitTest --tests org.meshtastic.app.map.MapLifecycleEffectsTest`
- `:androidApp:testGoogleDebugUnitTest --tests org.meshtastic.app.map.MapLifecycleEffectsTest`

Both flavor lifecycle tests passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Location tracking now starts and stops reliably as the app moves between foreground and background.
  * The screen remains awake only while active location tracking is enabled and permission is granted.
  * Improved cleanup when tracking is disabled or the map screen is closed.

* **Tests**
  * Added coverage verifying lifecycle-aware tracking behavior, including backgrounding, disabling, and resuming tracking.

### DIFF
--- a/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/MapView.kt
+++ b/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/MapView.kt
@@ -147,6 +147,7 @@ import org.meshtastic.core.ui.icon.Lens
 import org.meshtastic.core.ui.icon.Map
 import org.meshtastic.core.ui.icon.MeshtasticIcons
 import org.meshtastic.core.ui.icon.PinDrop
+import org.meshtastic.core.ui.util.KeepScreenOn
 import org.meshtastic.core.ui.util.PermissionStatus
 import org.meshtastic.core.ui.util.formatAgo
 import org.meshtastic.core.ui.util.rememberLocationPermissionState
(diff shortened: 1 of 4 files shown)


simulationstation:
@coderabbitai review

Issue #4988, opened 2026-04-04
[Bug]: Message delivery Status: Unrecognized (was "Encrypted send failed"?)


domusonline:
### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [x] I actually read this list, and should be taken seriously.

### Affected app version

2.7.14 (29320391)

### Affected Android version

Android 16

### Affected phone model

Samsung A34

### Affected node model

Heltec WSL V3

### Affected node firmware version

2.7.20

### Steps to reproduce the bug

1. Choose a node that is both on the application and radio NodDB
2. Send a private message to the node. It works
3. Remote that node from the radio (via CLI)
4. Confirm (via CLI) the node is not in the radio
5. Reconnect the APP to the radio
6. Repeat step 2. The cloud will show the message was not sent

### Actual behavior

Press the message to see the options. It will show "unrecognized" (see screenshot)
![Image](https://github.com/user-attachments/assets/bf71a192-84ff-4128-a29d-762c687b37d8)

Sequence of messages:

![Image](https://github.com/user-attachments/assets/fb2b403e-928e-4097-a8ef-6d39cdb7f547)

### Expected behavior

The application should send the node's contact info and the message should be sent (as it happens on iOS application)

### Relevant log output

```shell
Log from the APP (Mesh logs)

The successful message:
Apr 4, 00:14 [Packet]
MeshPacket{from=1442612248, to=-617952656, channel=0, id=1410262548, rx_time=1775258063, rx_snr=5.0, hop_limit=2, want_ack=true, priority=UNSET, rx_rssi=-92, via_mqtt=false, hop_start=2, public_key=[size=0], pki_encrypted=false, next_hop=0, relay_node=24, tx_after=0, transport_mechanism=TRANSPORT_LORA}
decoded {
  Data{portnum=ROUTING_APP, payload=[hex=1800], want_response=false, dest=0, source=0, request_id=1543458770, reply_id=0, emoji=0, bitfield=0}
[...]

PR #6603, opened 2026-08-11
fix(messaging): wait for shared-contact routing ACK

simulationstation:
### Bug fixes

- Keep transport queue admission separate from the routing ACK/NAK awaited by `sendAdminAwait`.
- Prevent an ordinary `QueueStatus(res = 0)` from releasing the first PKI DM before the local `add_contact` packet is processed.
- Correlate zero-ID queue-status fallbacks by packet ID instead of completing an unrelated pending routing response.
- Propagate routing NAKs, queue rejection, disconnected state, and transport-send failure as failed awaited sends.
- Fail a strict await immediately when disconnected instead of retaining its packet for transmission after reconnect.
- When a routing NAK arrives before `QueueStatus`, fail both deferred phases so the packet-queue worker is released.
- Cover queue admission, routing ACK/NAK, and immediate transport failure with regression tests.

## Root cause

This is a residual sequencing path from #4988 and a follow-up to #4992 and #5005.

#4992 made shared-contact installation suspending, but `sendToRadioAndAwait` used the same deferred that the packet-queue worker uses for `QueueStatus`. `QueueStatus(res = 0)` reports that firmware accepted or queued the packet; it is not the later routing acknowledgement. On firmware 2.7.20, a self-addressed packet is queued for local processing and QueueStatus is returned before `AdminModule` runs. That allowed `SendMessageUseCase` to enqueue the dependent PKI DM before `add_contact` reached the firmware NodeDB.

The fix tracks those two phases separately. QueueStatus continues to release the transport queue worker, while the strict caller remains suspended until the correlated routing ACK/NAK arrives. A queue or local transport failure still fails both phases, and the existing synchronous `ERRNO_SHOULD_RELEASE` success remains supported.

This intentionally changes two shared-path behaviors: `sendToRadioAndAwait` now returns failure immediately while disconnected rather than leaving the packet queued for post-reconnect transmission, and a routing NAK received before its `QueueStatus` also fails the queue deferred so queue processing cannot remain blocked.

Firmware registers `RoutingModule` last, after `AdminModule`; for a valid shared contact, `NodeDB::addFromContact` completes before the routing ACK is generated. This makes that ACK a processing barrier for this path without changing the wire protocol.

Relevant protocol and firmware paths:

- QueueStatus schema: https://github.com/meshtastic/protobufs/blob/master/meshtastic/mesh.proto#L2309-L2320
- Local packet enqueue: https://github.com/meshtastic/firmware/blob/v2.7.20.6658ec2/src/mesh/Router.cpp#L239-L248
- `add_contact` handling: https://github.com/meshtastic/firmware/blob/v2.7.20.6658ec2/src/modules/AdminModule.cpp#L355-L363
- RoutingModule registration order: https://github.com/meshtastic/firmware/blob/v2.7.20.6658ec2/src/modules/Modules.cpp#L252-L254
- NodeDB mutation and save: https://github.com/meshtastic/firmware/blob/v2.7.20.6658ec2/src/mesh/NodeDB.cpp#L1756-L1814

## Related work

- Addresses the residual ordering race reported in #4988.
- Follow-up to #4992, #4996, and #5005.
- #6598 overlaps the packet-admission files, but its current implementation still completes awaited sends on ordinary QueueStatus; this PR separates the routing phase.

## Testing Performed

- `:core:data:testAndroidHostTest` — 464/464 after the review follow-up.
- `:core:repository:testAndroidHostTest` — 21/21.
- Final focused `PacketHandlerImplTest` — 15/15, including zero-ID response correlation.
- `:core:data:spotlessCheck :core:repository:spotlessCheck`
- `:core:data:detekt :core:repository:detekt`
- `kmpSmokeCompile`

## Limitations

A missing routing ACK is not proof that firmware failed to install the contact, so the existing caller remains best-effort after a NAK or timeout and still enqueues the DM. This PR fixes the deterministic queue-admission ordering race; it does not change message failure UX or retry policy.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshDataHandlerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshDataHandlerImpl.kt
@@ -436,7 +436,7 @@ class MeshDataHandlerImpl(
                     packetRepository.value.updateReaction(updated)
                 }
             }
-            packetHandler.removeResponse(requestId, complete = true)
+            packetHandler.removeResponse(requestId, complete = isAck)
         }
     }
 
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/PacketHandlerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/PacketHandlerImpl.kt
@@ -190,10 +190,14 @@ class PacketHandlerImpl(
                         routingResponse.remove(requestId)?.complete(success)
                     }
                 } else {
-                    queueResponse.values.firstOrNull { !it.isCompleted }?.complete(success)
-                    if (!success || queueStatus.res == ERRNO_SHOULD_RELEASE) {
-                        routingResponse.values.firstOrNull { !it.isCompleted }?.complete(success)
-                    }
+                    queueResponse.entries
+                        .firstOrNull { !it.value.isCompleted }
+                        ?.let { (packetId, response) ->
+                            response.complete(success)
+                            if (!success || queueStatus.res == ERRNO_SHOULD_RELEASE) {
+                                routingResponse.remove(packetId)?.complete(success)
+                            }
+                        }
                 }
             }
         }
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/PacketHandlerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/PacketHandlerImpl.kt
@@ -87,6 +87,7 @@ class PacketHandlerImpl(
 
     private val responseMutex = Mutex()
     private val queueResponse = mutableMapOf<Int, CompletableDeferred<Boolean>>()
+    private val routingResponse = mutableMapOf<Int, CompletableDeferred<Boolean>>()
 
     override fun sendToRadio(p: ToRadio) {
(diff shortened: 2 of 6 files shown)


simulationstation:
@coderabbitai review

jamesarich (commented):
Nice fix — the queue/routing phase split is the right mechanism, and the test matrix (queue-acceptance-doesn't-complete, ACK, NAK, disconnected, transport-throw, zero-ID) is exactly what this path needs. Merging as-is; two asks for the record and one follow-up:

1. **Please add the two semantic changes to the description**: (a) `sendToRadioAndAwait` now fails fast when disconnected instead of leaving the packet queued for post-reconnect transmission; (b) a routing NAK arriving before its `QueueStatus` now releases the queue deferred as failure. Both look right, but they're behavior changes on a shared path and should be discoverable from the PR record.

2. **Resolving the CodeRabbit threads as non-blocking**, with rationale: the zero-ID fallback correlation issue (queue and routing responses picked independently, potentially from different packets) is real but confined to the rare `requestId == 0` fallback, and the strict routing path currently has a single production caller — the pre-existing ambiguity there predates this PR. Worth a follow-up fix (correlate both maps through the same selected entry), as is the suggested `assertFalse(result.isCompleted)` mid-test assertion and the KDoc on strict-await outcomes.

3. Heads-up: #6598 rewrites these same admission seams and will rebase over this — its typed-outcome machinery may eventually subsume the two-deferred split.

PR #6604, opened 2026-08-11
fix(konsist): normalize source paths on Windows

simulationstation:
## Summary

Konsist returns backslash-delimited source paths on Windows, while the architecture tests match slash-delimited project fragments. This makes the BLE privacy guard fail its scope assertion and lets the commonMain framework guard pass without inspecting any source files.

### Bug fixes

- Normalize project paths before filtering, excluding, and allowlisting files.
- Apply the same path handling to the BLE and commonMain architecture guards.
- Assert that the commonMain scan reaches a known source so an empty scope cannot pass silently.
- Normalize file paths in failure diagnostics.

## Testing Performed

- `:core:konsist:allTests` — 6/6 architecture tests passed on Windows.
- `:core:konsist:spotlessCheck`
- `:core:konsist:detekt`

## Summary by CodeRabbit

* **Tests**
  * Improved project file scanning and path handling across platform environments.
  * Added validation that shared source files are discovered correctly.
  * Enhanced BLE scan checks with more consistent filtering, allowlist validation, and diagnostic reporting.
  * Added exclusions for internal configuration files to reduce false positives in automated checks.


simulationstation:
@coderabbitai review

Issue #4544, opened 2026-02-12
[Bug]: Map performance issues


dpedu:
### Contact Details

Reply here :)

### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [x] I actually read this list, and should be taken seriously.

### Affected app version

2.7.12-google, 2.7.13-google

### Affected Android version

Android 16

### Affected phone model

Google Pixel 9a

### Affected node model

N/A

### Affected node firmware version

N/A

### Steps to reproduce the bug

1. Acquire a large number of nodes
2. Zoom into a portion of the map where there are few or no nodes
3. See that it performs poorly - input lag, single-digit fps, high CPU usage

### Actual behavior

Maps performs poorly - input lag, single-digit fps, high CPU usage.

### Expected behavior

Map performs decently regardless of zoom level.

### Screenshots/Screen recordings

https://www.youtube.com/watch?v=o3Uu2-eyW2c

### Relevant log output

```shell
N/A
```

### Additional information

I have ~2000 nodes in my app. 

When I zoom out on the map - for example, such that I can see the entire United States - performance is decent.

When I zoom into a narrow area on the map, where there are few (~3) or even zero nodes, performance dips. The map is rendered at single-digit frames per second, input lag increases, etc. 

I would have expected the full zoom out where more content is shown to perform worse than a narrow zoom where little content is shown. This is strange, so I suspect something is wrong.

rjDipcord:
Can confirm I am experiencing the same behavior on a Google Pixel 10 Pro XL. 

When zoomed out, the nodes condense into the larger groups to take up less screen space and make it easier to see where the dense concentrations of nodes are without the high number of entities on screen.

When zoomed in, these condenesed groups expand, increasing the entity count. I suspect something has changed with this mechanism, and it has caused performance issues related the quantity of entities rendered.

I still notice jarring movement and low FPS, even when zoomed all the way out.

Also on 2.7.12

dpedu:
[...]

PR #6605, opened 2026-08-11
fix(map): reuse cluster items across camera recompositions

simulationstation:
## Summary

- Cache Google Maps `NodeClusterItem` materialization while the filtered node values are unchanged.
- Rebuild when node data, local-node identity, localized labels, or the minute bucket changes.
- Advance the minute bucket only while the map lifecycle is started, so relative labels refresh without background work.
- Keep the complete filtered node set; this does not add viewport filtering or change clustering semantics.

`MapView` observes camera state, so map movement recomposes it. Previously, every such recomposition rebuilt every node's `LatLng`, `NodeClusterItem`, and relative-time title even when the filtered nodes were value-equal. With 2,000 nodes, 30 camera-only recompositions performed 62,000 unnecessary item materializations before the fix.

This removes that deterministic O(node-count) allocation pass from camera-only frames. It addresses one current hotspot reported in #4544 without claiming that all native marker or clustering costs are eliminated.

Related to #4544.

## Testing

- Added a Google-flavor Compose/Robolectric regression with 2,000 deterministic positioned nodes and 30 camera-only recompositions. The extracted pre-fix path failed in three runs; after the fix all 31 observations reuse one 2,000-item list.
- Added coverage that node changes and minute rollover still rebuild the items, and that the ticker advances exactly at the next minute boundary.
- Clean upstream-based branch: `./gradlew :androidApp:compileGoogleDebugKotlin :androidApp:testGoogleDebugUnitTest --tests org.meshtastic.app.map.MapNodeClusterItemsTest`
- Source validation matrix: `./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests lintFdroidDebug lintGoogleDebug testFdroidDebugUnitTest testGoogleDebugUnitTest`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map cluster-marker timestamp refreshes, including updates when relative times cross minute boundaries.
  * Added localized labels for “now” and “unknown” marker states.
  * Improved map performance and stability during camera-only updates, including large node lists.

* **Tests**
  * Added coverage for marker reuse, node and timestamp updates, and minute-boundary refresh behavior.

### DIFF
--- a/androidApp/src/google/kotlin/org/meshtastic/app/map/MapView.kt
+++ b/androidApp/src/google/kotlin/org/meshtastic/app/map/MapView.kt
@@ -108,6 +108,9 @@ import com.google.maps.android.data.renderer.model.PolygonStyle
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.withContext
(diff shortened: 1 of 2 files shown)


simulationstation:
@coderabbitai review

PR #6619, opened 2026-08-11
fix(android): preserve shared text in deep links

simulationstation:
## Summary

- encode shared message text as a URI query parameter instead of interpolating it into the deep-link string
- preserve ampersands, fragments, query delimiters, plus signs, line breaks, emoji, and non-Latin text
- add a Robolectric regression covering exact Android URI construction and the full `CommonUri` through `DeepLinkRouter` decode path

## Root cause

`MainActivity` previously appended the raw message after `?message=`. URI delimiter characters in user text were therefore interpreted as query or fragment syntax, truncating or changing the text delivered to the share screen.

`Uri.Builder.appendQueryParameter` applies the platform's query encoding while preserving the existing deep-link contract.

## Impact

Plain-text sharing now delivers the exact selected message even when it contains URI-significant or Unicode characters. Messages without those characters are unchanged.

## Testing

- `:androidApp:testFdroidDebugUnitTest --tests org.meshtastic.app.ShareMessageDeepLinkTest`
- `:androidApp:testGoogleDebugUnitTest --tests org.meshtastic.app.ShareMessageDeepLinkTest`
- `:androidApp:spotlessApply :androidApp:spotlessCheck :androidApp:detekt :androidApp:assembleDebug`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sharing of messages containing special characters, spaces, line breaks, emoji, and non-Latin text.
  * Ensured shared links preserve message content accurately and avoid unintended URL fragments or extra parameters.

* **Tests**
  * Added coverage for sharing and restoring a wide range of message formats.

### DIFF
--- a/androidApp/src/main/kotlin/org/meshtastic/app/MainActivity.kt
+++ b/androidApp/src/main/kotlin/org/meshtastic/app/MainActivity.kt
@@ -368,9 +368,8 @@ class MainActivity : AppCompatActivity() {
     }
 
     private fun createShareIntent(message: String): PendingIntent {
-        val deepLink = "$DEEP_LINK_BASE_URI/share?message=$message"
         val startActivityIntent =
-            Intent(Intent.ACTION_VIEW, deepLink.toUri(), this, MainActivity::class.java).apply {
+            Intent(Intent.ACTION_VIEW, createShareMessageDeepLink(message), this, MainActivity::class.java).apply {
                 flags = Intent.FLAG_ACTIVITY_SINGLE_TOP
             }
 
@@ -393,3 +392,6 @@ class MainActivity : AppCompatActivity() {
         const val EXTRA_SKIP_ONBOARDING = "skip_onboarding"
     }
 }
+
+internal fun createShareMessageDeepLink(message: String): Uri =
+    "$DEEP_LINK_BASE_URI/share".toUri().buildUpon().appendQueryParameter("message", message).build()
--- a/androidApp/src/test/kotlin/org/meshtastic/app/ShareMessageDeepLinkTest.kt
+++ b/androidApp/src/test/kotlin/org/meshtastic/app/ShareMessageDeepLinkTest.kt
@@ -1,0 +1,85 @@
+/*
+ * Copyright (c) 2026 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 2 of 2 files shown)


jamesarich (commented):
Looks good — swapping raw interpolation for `Uri.Builder.appendQueryParameter` is the right fix, and it correctly handles `&`, `#`, `=`, `+`, newlines and non-ASCII. Empty messages still degrade to `Share("")` as before, no commonMain framework bleed, and `"message"` is a wire key so no string resource is needed.

One gap worth closing, since it's the seam the fix actually depends on: the new test round-trips through Android's `Uri` (the producer), but the **consumer** parses through uri-kmp — `uri.toKmpUri()` → `DeepLinkRouter.routeContacts`'s `getQueryParameter("message")` at `core/navigation/.../DeepLinkRouter.kt:124` — and the existing `DeepLinkRouterTest` only covers `%20`. So nothing currently asserts that what `android.net.Uri` encodes is what uri-kmp decodes. Low risk, since uri-kmp mirrors `android.net.Uri` semantics, but one case feeding the produced URI string through `CommonUri.parse` + `DeepLinkRouter.route` would make that guarantee explicit.

Minor placement thought, take or leave: the helper could live next to `DEEP_LINK_BASE_URI` and `DeepLinkRouter` in `:core:navigation`, keeping build-and-parse of the share link in one module.

simulationstation:
Thanks — added the suggested end-to-end regression in `d5d3b502b`. It now feeds Android's `Uri.Builder` result through `CommonUri.parse()` and `DeepLinkRouter.route()` and verifies the exact original text for URI delimiters, line breaks, emoji, and non-Latin text in both app flavors.

simulationstation:
Addressed the remaining test-strength feedback in 69f0b0b59. The regression now invokes MainActivity's actual share callback PendingIntent factory, inspects the resulting ACTION_VIEW intent, and verifies its component, flags, encoded message query, fragment handling, and DeepLinkRouter destination. The focused Android app test plus Spotless/Detekt passed; both debug flavors assembled in the full gate.

simulationstation:
Rebased this branch onto current `main` to refresh GitHub's merge ref and current-head CI. The focused regression now passes in both F-Droid and Google flavors, along with `:androidApp:spotlessCheck` and `:androidApp:detekt` (Gradle capped at four workers). Current head: `ec93912d3`.

PR #6620, opened 2026-08-11
fix(node): stop compass updates while backgrounded

simulationstation:
## Summary

- share one `ActiveWhileStarted` lifecycle primitive from `:core:ui` across Maps and the node compass
- run compass location and heading updates only while the lifecycle is `STARTED`, with key-based restart and optional enablement
- stop on `ON_STOP` or overlay disposal and resume only while the compass remains open
- retain the selected identity and units while using the matching live node state for refreshed positions
- consolidate lifecycle coverage in `:core:ui` for enable/disable, key changes, background/resume, and disposal

## Root cause

Opening the node compass called `CompassViewModel.start()` immediately. Its only cleanup hooks were bottom-sheet dismissal and composition disposal, but pressing Home leaves that composition alive. The ViewModel therefore kept its one-second GPS/network location flow and motion-sensor listener registered while the app was backgrounded.

The shared lifecycle primitive now owns both the Google map callback and compass updates. It invokes `stop()` from `onStopOrDispose`; that cancels the provider job immediately, while the provider `awaitClose` unregisters the platform listeners when cancellation resumes.

## Impact

Backgrounding the app from an open compass releases its location and heading flows. Returning restarts the same selection and picks up a refreshed matching node position; dismissing the overlay prevents later lifecycle changes from restarting it.

## Testing

- `:core:ui:jvmTest`
- `:feature:node:jvmTest`
- affected `:core:ui`, `:feature:node`, and `:androidApp` Spotless/Detekt checks
- `kmpSmokeCompile`

The focused 303-task gate passed. The full project gate passed Spotless, Detekt, assembly, and ordinary tests; only the known Windows Android-host DataStore atomic-rename fixture failed in untouched `:core:database:testAndroidHostTest` and reproduced when rerun alone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compass lifecycle handling so location tracking starts and stops reliably when the compass overlay opens, closes, or is paused.
  * Ensured compass updates apply to the selected node and display units.
  * Prevented duplicate cleanup and improved cancellation of location listeners when tracking stops.

* **Refactor**
  * Improved lifecycle utility behavior, including support for restarting active effects when relevant settings change.

### DIFF
--- a/androidApp/src/google/kotlin/org/meshtastic/app/map/MapView.kt
+++ b/androidApp/src/google/kotlin/org/meshtastic/app/map/MapView.kt
@@ -162,6 +162,7 @@ import org.meshtastic.core.ui.icon.Map
 import org.meshtastic.core.ui.icon.MeshtasticIcons
 import org.meshtastic.core.ui.icon.TripOrigin
 import org.meshtastic.core.ui.theme.TracerouteColors
+import org.meshtastic.core.ui.util.ActiveWhileStarted
 import org.meshtastic.core.ui.util.KeepScreenOn
 import org.meshtastic.core.ui.util.PermissionStatus
 import org.meshtastic.core.ui.util.formatAgo
@@ -351,7 +352,7 @@ fun MapView(
         }
     }
 
-    ActiveWhileStarted(isLocationTrackingEnabled && locationPermission.isGranted) {
+    ActiveWhileStarted(enabled = isLocationTrackingEnabled && locationPermission.isGranted) {
         val locationRequest =
             LocationRequest.Builder(Priority.PRIORITY_HIGH_ACCURACY, 5000L).setMinUpdateIntervalMillis(2000L).build()
         try {
--- a/core/ui/src/commonMain/kotlin/org/meshtastic/core/ui/util/ActiveWhileStarted.kt
+++ b/core/ui/src/commonMain/kotlin/org/meshtastic/core/ui/util/ActiveWhileStarted.kt
@@ -14,7 +14,7 @@
  * You should have received a copy of the GNU General Public License
  * along with this program.  If not, see <https://www.gnu.org/licenses/>.
  */
-package org.meshtastic.app.map
+package org.meshtastic.core.ui.util
 
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.getValue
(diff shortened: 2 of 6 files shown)


jamesarich (commented):
The fix is correct — moving start/stop onto `LifecycleStartEffect` with `onStopOrDispose` gives synchronous teardown on ON_STOP, which is exactly the hazard a coroutine-based cleanup misses, and carrying node/displayUnits on the `NodeDetailOverlay.Compass` state means a background/foreground cycle restarts the same selection. One structural request before merge:

**Please hoist the helper instead of duplicating it.** `CompassUpdatesWhileStarted` is the same `rememberUpdatedState` + `LifecycleStartEffect` + `onStopOrDispose` shape as `ActiveWhileStarted`, which landed a few hours ago in your own #6602 (`androidApp/src/main/kotlin/org/meshtastic/app/map/MapLifecycleEffects.kt:33`). You can't reuse it literally — it's `internal` to `:androidApp` and the compass lives in `:feature:node/commonMain` — but `:core:ui` is already a `feature/node` dependency and would host one shared copy for both call sites. As it stands main gains two copies of the same primitive plus two near-identical lifecycle tests, which is exactly the kind of thing that drifts apart later.

Two smaller notes:

- **The test proves the framework, not the wiring.** `CompassUpdatesLifecycleTest` drives the helper with recording lambdas; nothing covers `NodeDetailOverlays` actually reaching `viewModel::stop` on sheet dismissal. Non-blocking, but the regression this guards against lived in the wiring rather than in `LifecycleStartEffect`.
- **Teardown is one hop later than ON_STOP.** `CompassViewModel.stop()` only cancels the job, so the actual sensor/GPS unregistration happens in the providers' `awaitClose` when the cancelled coroutine resumes. Prompt in practice, just not strictly synchronous with ON_STOP.

Non-blocking observation: `NodeDetailOverlay.Compass(node, …)` freezes the node at click time, so the sheet's "request position" refresh never reaches the compass target. That's pre-existing, but now that `node` is a `LifecycleStartEffect` key, passing live `uiState.node` would make it reactive essentially for free.

simulationstation:
Thanks — addressed the structural request in `75c150fd6`. `ActiveWhileStarted` now lives in `:core:ui/commonMain` and is shared by Maps and the compass, with enablement and restart-key coverage consolidated into one lifecycle test. The compass retains the selected identity/units but consumes the matching live node state after refresh. I also corrected the teardown wording: `stop()` cancels immediately, while provider `awaitClose` unregisters when cancellation resumes. Focused UI/node/app validation and KMP smoke compilation pass.

simulationstation:
Added the requested feature-level wiring regression in bf131cf62. NodeDetailCompassLifecycleTest drives the real NodeDetailScreen path: it scrolls to and selects Open Compass, confirms the real CompassViewModel subscribes while STARTED, verifies cleanup on ON_STOP, restart on ON_START, then dismisses the actual bottom sheet and verifies final cleanup with no later restart.

The test uses Compose UI test v2 in jvmTest because this module documents a known Compose idling crash in its Android host-test stubs. Local validation passed: feature:node JVM tests, core:ui ActiveWhileStartedTest, Spotless and Detekt for both touched modules, and kmpSmokeCompile. Fresh CI is running for the amended head.

PR #6622, opened 2026-08-11
fix(usb): validate permission callbacks

simulationstation:
## Summary

- register the USB permission callback receiver as not exported
- accept a callback only when its `UsbDevice` matches the request
- treat a callback with no `EXTRA_DEVICE` as a denial that completes and unregisters the flow
- verify a reported grant against `UsbManager.hasPermission()` before returning success
- cover private registration, absent/spoofed/wrong-device callbacks, authoritative grants, and concurrent requests

## Root cause

The callback receiver used the project's exported default and trusted only `EXTRA_PERMISSION_GRANTED`. Any matching broadcast could therefore complete an in-flight permission flow, and concurrent requests all consumed the first result regardless of which USB device it described. The first revision also treated an absent device like a different device, leaving the permission flow waiting forever.

The receiver is now private to the app, correlates a non-null `EXTRA_DEVICE` to the requested device, completes an absent-device result as denial, and treats the platform permission state as authoritative.

## Impact

A foreign or unrelated permission result can no longer grant or terminate the wrong USB connection attempt. A malformed/incomplete platform result no longer hangs the request. Legitimate permission callbacks keep the mutable `PendingIntent` required for the USB service to add its result extras.

## Testing

- `:core:network:testAndroidHostTest --tests org.meshtastic.core.network.repository.UsbManagerTest`
- full `:core:network:testAndroidHostTest`
- `:core:network:spotlessApply :core:network:spotlessCheck :core:network:detekt`
- `kmpSmokeCompile`

The Robolectric regression verifies the actual dynamic-receiver flag and drives the registered receiver with missing-device, wrong-device, ungranted, granted, and overlapping-device results. The focused 353-task gate passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved USB permission handling by validating the requested device and permission state.
  * Prevented unrelated or incomplete permission broadcasts from being accepted.
  * Ensured permission receivers are safely cleaned up and not externally accessible.
  * Improved handling of simultaneous USB permission requests.

### DIFF
--- a/core/network/src/androidHostTest/kotlin/org/meshtastic/core/network/repository/UsbManagerTest.kt
+++ b/core/network/src/androidHostTest/kotlin/org/meshtastic/core/network/repository/UsbManagerTest.kt
@@ -1,0 +1,226 @@
+/*
+ * Copyright (c) 2026 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.core.network.repository
+
+import android.app.Application
+import android.content.Context
+import android.content.Intent
+import android.hardware.usb.UsbDevice
+import android.hardware.usb.UsbManager
+import androidx.core.content.ContextCompat
+import androidx.test.core.app.ApplicationProvider
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.async
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.test.runCurrent
+import kotlinx.coroutines.test.runTest
+import org.junit.After
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.Shadows.shadowOf
+import org.robolectric.annotation.Config
+import org.robolectric.shadow.api.Shadow
+import org.robolectric.util.ReflectionHelpers
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+@RunWith(RobolectricTestRunner::class)
+@Config(sdk = [34])
+class UsbManagerTest {
+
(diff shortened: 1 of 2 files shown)


jamesarich (commented):
Good hardening, and the reasoning is sound — `RECEIVER_NOT_EXPORTED`, device correlation, and treating `UsbManager.hasPermission()` as authoritative over the spoofable `EXTRA_PERMISSION_GRANTED` extra are all the right calls. I also verified the firmware-update path is unaffected: the nRF DFU / USB-serial flow goes through `SerialConnectionImpl.connect()` → `openDevice()` → `TransportDisconnectReason.UsbPermissionDenied`, and never touches `usbPermissionResultFlow`. One fix before merge:

**Silent hang when the result carries no device** — `core/network/src/androidMain/.../repository/UsbManager.kt:41`. The receiver now returns early whenever `EXTRA_DEVICE` is absent or unequal, and `null != device` is true for the absent case. So a result broadcast without the extra makes the flow never emit and never `close()`. The sole caller (`AndroidScannerViewModel.requestPermission`) has no timeout, so the user gets neither a connection nor the `usb_permission_denied` message, and the receiver stays registered for the ViewModel's lifetime. Before this change, any result terminated the flow.

AOSP always populates `EXTRA_DEVICE` so this shouldn't fire in practice, but distinguishing *null device* (treat as denial, close the flow) from *wrong device* (ignore, keep waiting) keeps the whole anti-spoofing win with no dead end.

Two cosmetic notes, take or leave: the `checkNotNull(...)` around `PendingIntentCompat.getBroadcast` can't trip with `flags = 0` (only `FLAG_NO_CREATE` returns null), so it would surface as a bare `IllegalStateException` inside the flow; and inside the wrapper lambda `requestPermission(device, permissionIntent)` resolves to the two-arg `UsbManager` member while a same-named one-arg parameter is in scope — correct, but easy to misread.

The `hasPermission()` gate is worth keeping as-is. AOSP persists the grant before firing the `PendingIntent`, so ordering is safe on stock Android; just flagging that it's the one change only device testing can fully clear.

simulationstation:
Thanks — fixed the absent-device path in `78c6c8f4a`. A callback without `EXTRA_DEVICE` now completes as denial and unregisters immediately; a different non-null device remains ignored, and a matching device still uses the authoritative platform grant. The new Robolectric regression verifies completion, no permission query, and cleanup; the full network Android-host suite passes.

simulationstation:
Addressed the remaining permission-path test feedback in 7380236e9. The tests now prove requestPermission is invoked before callback dispatch and separately prove that a broadcast grant is rejected when UsbManager.hasPermission remains false. All five UsbManager Android-host tests, Spotless, and Detekt pass; KMP smoke compilation and both debug flavor assemblies completed in the full gate.

simulationstation:
Rebased this branch onto current `main` to refresh GitHub's merge ref and current-head CI. The complete `UsbManagerTest` Android-host suite now passes (5/5), including the false-platform-permission grant and request-callback invocation coverage, along with affected Spotless, Detekt, and KMP smoke compilation. Gradle was capped at four workers. Current head: `566a39694`.

PR #6640, opened 2026-08-12
fix(map): open Site Planner for the selected node

### LINKED ISSUE #6591: [Bug]: Site Planner not going to the location of the selected node
### Contact Details

_No response_

### Checklist

### Affected app version

2.8.0 (29321705)

### Affected Android version

Android 16

### Affected phone model

Samsung

### Affected node model

A25 5G

2.7.26

1. Go to Nodes
2. Select a Node (not your nodes)
3. Verify it has position (map)
4. Click in "Estimate coverage"

### Actual behavior

The Mash Map remains in the last position it was previously viewed/positioned.
"Use current node location" do not got to the Node's location you had selected.

Eg node SHT1
Android:

Image

Image

in IOS it goes to the nodes location you had selected.

Image

Image

### Expected behavior

The Site Planner goes to the location of the selected node

_No response_

### Relevant log output

```shell

```

### Additional information

_No response_

simulationstation:
## Summary

The Node Details route already carries `sitePlannerNodeNum`, but Navigation 3 did not forward it through the map host, while the flavor ViewModels incorrectly expected it in their own `SavedStateHandle`. As a result, the selected-node request never reached Site Planner.

- forward the selected node through the Navigation 3 map entry and platform map providers
- resolve the one-shot request against the live NodeDB in both map flavors
- prefill and center the planner on the selected node
- keep the node-location shortcut on that selected node for route launches, while preserving connected-node behavior for manual map launches
- retain consumed route identity in the entry-scoped ViewModel so composition exit/re-entry cannot re-arm the same request
- consume only the exact request that was rendered so a stale effect cannot clear a newer selection
- cover route forwarding, both flavor ViewModels, live NodeDB replacement, selected-versus-connected shortcut behavior, and Navigation 3 composition exit/re-entry

## Validation

- focused F-Droid and Google request-state tests and the deterministic Navigation 3 composition lifecycle regression passed
- Spotless, Detekt, `kmpSmokeCompile`, and both debug APK assemblies passed
- the broad `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile --max-workers=4` gate passed the changed suites and stopped only on the known unrelated Windows `FileStorage.kt:114` atomic-rename test-fixture failures in `:core:database` and `:core:datastore`

Physical-device UI validation was not performed.

Fixes #6591

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -333,6 +333,7 @@ dependencies {
 
     testImplementation(kotlin("test-junit"))
     testImplementation(libs.androidx.work.testing)
+    testImplementation(projects.core.testing)
     testImplementation(libs.koin.test)
     testRuntimeOnly(libs.junit.vintage.engine)
     testImplementation(libs.kotlinx.coroutines.test)
--- a/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/FdroidMapViewProvider.kt
+++ b/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/FdroidMapViewProvider.kt
@@ -27,9 +27,15 @@ import org.meshtastic.core.ui.util.MapViewProvider
 @Single
 class FdroidMapViewProvider : MapViewProvider {
     @Composable
-    override fun MapView(modifier: Modifier, navigateToNodeDetails: (Int) -> Unit, waypointId: Int?) {
+    override fun MapView(
+        modifier: Modifier,
+        navigateToNodeDetails: (Int) -> Unit,
+        waypointId: Int?,
+        sitePlannerNodeNum: Int?,
+    ) {
         val mapViewModel: MapViewModel = koinViewModel()
         LaunchedEffect(waypointId) { mapViewModel.setWaypointId(waypointId) }
+        LaunchedEffect(sitePlannerNodeNum) { mapViewModel.setSitePlannerNodeNum(sitePlannerNodeNum) }
         org.meshtastic.app.map.MapView(
             modifier = modifier,
             mapViewModel = mapViewModel,
--- a/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/MapView.kt
+++ b/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/MapView.kt
@@ -158,7 +158,6 @@ import org.meshtastic.feature.map.component.DeleteWaypointDialog
 import org.meshtastic.feature.map.component.EditWaypointDialog
 import org.meshtastic.feature.map.component.MapButton
 import org.meshtastic.feature.map.component.MapControlsOverlay
-import org.meshtastic.feature.map.component.SitePlannerParams
 import org.meshtastic.feature.map.component.WaypointInfoDialog
 import org.meshtastic.proto.Waypoint
 import org.osmdroid.bonuspack.utils.BonusPackHelper.getBitmapFromVectorDrawable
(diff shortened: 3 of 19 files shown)


jamesarich (commented):
Good fix — the `savedStateHandle.get<Int>("sitePlannerNodeNum")` reads in the flavor ViewModels genuinely never resolved, and threading the value through the Nav3 entry is the right shape. The `compareAndSet` in `consume()` is well done: a stale effect can't clear a newer selection, and the tests cover that. No conflict with the #6581 sheet-title revert (that was strings-only; this touches no strings).

**One question I'd like answered before merge — the one-shot may re-arm on back-navigation.**

Both providers do `LaunchedEffect(sitePlannerNodeNum) { mapViewModel.setSitePlannerNodeNum(sitePlannerNodeNum) }` (`FdroidMapViewProvider.kt:30`, `GoogleMapViewProvider.kt:155`). The route arg is immutable for the entry's lifetime, so the effect only re-runs if the entry leaves and re-enters composition.

It looks like it does: `MeshtasticNavDisplay` installs `rememberSaveableStateHolderNavEntryDecorator` and `rememberViewModelStoreNavEntryDecorator` (`core/ui/.../MeshtasticNavDisplay.kt:131-132`), so a non-top entry leaves composition while its ViewModel survives in the entry-scoped store. If that's right, then Map → tap a node → NodeDetail → back re-runs the effect with the same key and re-arms a value the user already consumed, reopening Site Planner unexpectedly. Previously the pending value lived only in the ViewModel and stayed consumed.

I haven't run this on device, so I may be wrong about whether the map entry actually leaves composition in the adaptive two-pane scene — could you confirm? If it does re-arm, keying the effect on something navigation-scoped (or having `setSitePlannerNodeNum` ignore a value already consumed for that nodeNum) would close it.

**Nit:** `core/ui/detekt-baseline.xml:23` still carries a `CompositionLocalAllowlist` entry keyed to the old 3-parameter `LocalMapMainScreenProvider` signature. This PR makes it 4 parameters, so that entry can never match again and is dead; you added an inline `@Suppress` at `LocalMapMainScreenProvider.kt:796` instead, which works but diverges from how the other eight CompositionLocals in that baseline are handled. Regenerating the baseline entry and dropping the inline suppression would be tidier — at minimum the dead line should go.

**Heads-up on ordering:** this and #6641 both modify five identical files (`androidApp/build.gradle.kts` plus `MapView.kt`/`MapViewModel.kt` in both flavor source sets). Whichever lands second needs a rebase; I'd suggest this one goes first since it's the smaller bug fix.

simulationstation:
Confirmed: the Map entry leaves composition when Node Detail is pushed, while its entry-scoped ViewModel survives. A deterministic Navigation 3 test now demonstrates that the route effect starts again after popping back to Map.

I changed `SitePlannerRequestState` to retain pending/consumed identity in that ViewModel. Reinjecting the same consumed route argument is now ignored; a different request can still replace a pending request, and stale consumption cannot clear it. The regressions pass in both map flavors.

I also removed the inline `CompositionLocalAllowlist` suppression and regenerated the baseline for the four-parameter provider signature. The focused tests, formatting, Detekt, KMP smoke compile, and both debug assemblies pass. The broad Windows gate only reproduced the unrelated `FileStorage.kt:114` atomic-rename fixture failures in the database/datastore modules.

@jamesarich, ready for re-review.

jamesarich (approved):
Re-reviewed — both items resolved.

**Re-arm on back-navigation**: confirmed real by your new `mapRouteEffectRestartsWhenEntryReturnsFromBackStack` test (the entry's effect does restart with the same route arg), and the fix is the right shape: `SitePlannerRequestState`'s `Pending → Consumed(nodeNum)` transition lives in the entry-scoped ViewModel, so the re-injected arg is a no-op after consumption while a *different* nodeNum still arms normally, and a stale `consume()` can't clear a newer request. The flavor-duplicated ViewModel tests cover all three transitions. Retaining the selected node as the "use node location" source in `SitePlannerLaunch` for route launches is a genuine improvement over silently switching to the connected node.

**Detekt baseline**: the dead 3-parameter `CompositionLocalAllowlist` entry is regenerated to the new signature and the inline suppress is gone, consistent with the other CompositionLocals.

On ordering with #6641: this one is now rebased and green, so as suggested it should land first; #6641 takes the rebase.

LGTM.

PR #6654, opened 2026-08-12
feat(analytics): report key user interactions as named RUM actions

jamesarich:
## Why

Datadog RUM tap actions from release builds are unreadable. R8 renames the Compose/View classes that the SDK's auto-tracker uses to label a tap, so the dashboard shows `tap on fhy(0x-1)` and `tap on ImageView(0x-1)`. The only legible action today is the custom `connect` one, because it is the only action the app names explicitly.

That made feature-level usage analysis impossible for the DEF CON 34 telemetry report — we could see that people tapped, but never what they tapped.

This names the interactions that actually carry analytical value, so RUM can answer "how many people sent a message / ran a traceroute / started a firmware update" instead of counting anonymous taps.

## 🌟 What

Adds `PlatformAnalytics.trackAction(name, attributes)`, following the same shape as the existing `trackConnect`:

- **Default no-op** on the interface, so fdroid and desktop are untouched (neither has RUM).
- **google flavor only** overrides it, emitting a Datadog `RumActionType.CUSTOM` action behind the same `isInitialized` / `isRegistered` guard the other RUM calls use.

Datadog's `interactionPredicate` was evaluated and rejected: it hooks the Android View system and cannot name Compose Multiplatform targets, which is what this app's UI is built from. Explicit calls at the handler layer are also the pattern already established here by `trackConnect` and `track("data_send")`.

Instrumented at the ViewModel/controller layer rather than in composable click lambdas, so one call site covers every entry point into the action:

| Action | Attributes | Where |
| --- | --- | --- |
| `message_send` / `waypoint_send` / `data_send` | `port_num`, `num_bytes` | `MessagingControllerImpl.sendMessage` |
| `reaction_send` | — | `MessagingControllerImpl.sendReaction` |
| `traceroute_request` | — | `CommonNodeRequestActions` |
| `position_request` | — | `CommonNodeRequestActions` |
| `user_info_request` | — | `CommonNodeRequestActions` |
| `telemetry_request` | `telemetry_type` | `CommonNodeRequestActions` |
| `node_favorite` | `favorite` | `NodeManagementActions.setFavorite` |
| `channel_update` | `num_channels` | `ChannelViewModel.setChannels` |
| `channel_share` | — | `ChannelViewModel.trackShare` |
| `firmware_update_start` | `update_method`, `is_recovery`, `release_version` | `FirmwareUpdateViewModel.startUpdate` |
| `admin_action` | `route`, `is_remote` | `RadioConfigViewModel.sendAdminRequest` |

Two notes on placement:

- `sendMessage` is a single funnel for both messaging and map waypoint sends, so one call covers both without touching `MessageViewModel` or `BaseMapViewModel`.
- `admin_action` sits at the `when (route)` dispatch, so reboot, shutdown, factory reset, nodedb reset and set-time are all covered by one call, keyed by route.

Deliberately **not** instrumented: contact open, map open, and settings entries. Those are navigations and already arrive in RUM as named views via `startScreenView`; adding actions would double-count them.

## 🛠️ Housekeeping

- `ChannelViewModel.trackShare()` was dead code — defined and unit-tested, but never called from production. Wired it to the QR/share button in `ChannelScreen`, so the existing `share` event starts firing too.
- `feature/firmware` gains an explicit `implementation(projects.core.repository)`; it was relying on a transitive `api` from `:core:service`.
- `FirmwareUpdateMethod` is a sealed class, so its update-method label is mapped explicitly rather than read from a class name that R8 would rename — the same failure mode this PR exists to fix.

## Privacy

No PII, location, or key material. Every attribute is a count, boolean, enum name, or firmware version string. In particular there are no node numbers or long names, no coordinates, no message content, and `channel_share` carries no channel URL (that URL embeds the PSK).

## Testing Performed

`spotlessApply spotlessCheck detekt assembleDebug test allTests` — spotless/detekt/assembleDebug green, 250 tests passed.

Two failures surfaced in that run, `:core:navigation:jvmTest` and `:core:network:jvmTest`, both `java.nio.file.NoSuchFileException: in-progress-results-generic.bin`. Both are unrelated to this change and were confirmed as such:

- `:core:navigation` does not depend on `:core:repository` at all (only `core:common` + `core:resources`), so nothing in this diff can reach it.
- Re-running both modules in isolation: **BUILD SUCCESSFUL in 51s.**

The signature is a test worker killed mid-run under host memory pressure, not a failing assertion.

Targeted re-verification after the final refinement — `spotlessApply spotlessCheck detekt :feature:settings:allTests :feature:node:allTests :feature:firmware:allTests :core:service:allTests` — **BUILD SUCCESSFUL in 2m 9s.**

Existing tests were updated for the new constructor parameters; analytics call sites are not separately asserted, matching how the pre-existing `track("data_send")` call is treated.

## Summary by CodeRabbit

* **New Features**
  * Added custom analytics tracking for messaging, reactions, firmware updates, node requests, favorites, channel changes and sharing, and radio administration actions.
  * Added support for sending tracked actions to Datadog on Android.
  * Recorded relevant context such as action type, update method, recovery status, channel count and remote-target status.
* **Tests**
  * Updated test fixtures to support the new analytics tracking capabilities.

### DIFF
--- a/androidApp/src/google/kotlin/org/meshtastic/app/analytics/GooglePlatformAnalytics.kt
+++ b/androidApp/src/google/kotlin/org/meshtastic/app/analytics/GooglePlatformAnalytics.kt
@@ -275,6 +275,11 @@ class GooglePlatformAnalytics(private val context: Context, private val analytic
         GlobalRumMonitor.get().addAction(RumActionType.CUSTOM, "connect", attributes)
     }
 
+    override fun trackAction(name: String, attributes: Map<String, Any>) {
+        if (!Datadog.isInitialized() || !GlobalRumMonitor.isRegistered()) return
+        GlobalRumMonitor.get().addAction(RumActionType.CUSTOM, name, attributes)
+    }
+
     override fun startScreenView(key: String, name: String) {
         if (!Datadog.isInitialized() || !GlobalRumMonitor.isRegistered()) return
         GlobalRumMonitor.get().startView(key = key, name = name)
--- a/core/repository/src/commonMain/kotlin/org/meshtastic/core/repository/PlatformAnalytics.kt
+++ b/core/repository/src/commonMain/kotlin/org/meshtastic/core/repository/PlatformAnalytics.kt
@@ -53,6 +53,17 @@ interface PlatformAnalytics {
     }
 
     /**
+     * Tracks a key user interaction as a named custom RUM action. Auto-tracked tap targets are R8-obfuscated in
+     * release builds, so analytically important interactions must be reported explicitly with a stable name.
+     *
+     * @param name Stable snake_case action name (e.g. "message_send", "traceroute_request").
+     * @param attributes Optional structured attributes attached to the action.
+     */
+    fun trackAction(name: String, attributes: Map<String, Any> = emptyMap()) {
+        // Default no-op for platforms that don't support RUM (fdroid, desktop)
+    }
+
+    /**
      * Starts tracking a screen as a RUM view, aligned with the Meshtastic-Apple Datadog integration (which auto-tracks
      * SwiftUI views) so per-screen RUM data lines up across platforms.
      *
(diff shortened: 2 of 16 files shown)


jamesarich:
@coderabbitai full review

jamesarich:
@coderabbitai full review

PR #6665, opened 2026-08-12
fix(konsist): anchor path rules to the scanned checkout root so the commonMain guard isn't vacuously green in agent worktrees

jamesarich:
## Why

`CommonMainFrameworkBoundaryTest` enforces the **No Framework Bleed** rule from `AGENTS.md` — no `java.*` or `android.*` imports in any `commonMain` source set. It was silently verifying **nothing** for any agent session working from a worktree.

Both guards excluded agent scratch sources with `"/.claude/" in it.path` — matched against the **absolute** path. Agent worktrees are rooted at `<main checkout>/.claude/worktrees/<name>`, so that substring is present in *every file of the entire checkout*. The scan set became empty.

Today that surfaces as a confusing failure, because the "scan actually reaches sources" self-check trips. But the guard was one relaxed assertion away from a vacuous green — and a rule that passes without scanning anything is worse than no rule, because it reads as verification. CI checks out at a normal path, so CI has been the only real gate for these two rules.

## 🛠️ What changed

Konsist resolves its project root by walking up for `gradlew` (`GradleProjectRootDirResolver`), so `KoFileDeclaration.projectPath` is relative to **whichever checkout is being scanned**. That is the correct anchor, and every path rule now keys off it:

- **`scanPath`** (`ProjectPath.kt`) — normalized `projectPath`; replaces all uses of the absolute `path`, in filters and in offender messages alike.
- **`isNestedAgentWorktree()`** — matches `.claude/` only as the *first segment* relative to the scanned root. Excludes nested worktrees when scanning the main checkout; excludes nothing when the scan runs from a worktree.
- **`emptyScanMessage()`** — the empty-scan self-check stays a hard failure (it did its job here), but now names the cause rather than the symptom and points at `ProjectPath.kt`.

The exclusion is **not** vestigial and was deliberately kept: `scopeFromProject` genuinely does sweep nested `.claude/worktrees/` checkouts, whose stale copies resurface long-fixed lines as phantom offenders. Correct scoping — not deletion — is the fix. See the control run below.

## Testing Performed

`:core:konsist:allTests`, both path shapes, plus a negative test and a control for each direction:

| Scenario | Result |
|---|---|
| From `.claude/worktrees/<name>` (this branch's worktree) | ✅ 6/6, both scan-reaches self-checks green |
| ⬆ same, with `import java.util.UUID` injected into `core/model` commonMain | ❌ **fails as intended** — `commonMain declares no java imports` names `Message.kt`; injection reverted |
| From a normal path (fresh worktree under `/private/tmp`) | ✅ 6/6 |
| ⬆ same, with a violating file planted at `.claude/worktrees/fake-agent-session/core/ble/.../StaleOffender.kt` | ✅ passes — exclusion still holds |
| ⬆ same, control run with `isNestedAgentWorktree()` forced to `false` | ❌ **fails**, naming `StaleOffender.kt` — proving the nested sweep is real and the exclusion is load-bearing |

That last row is the one worth a reviewer's attention: it rules out the tempting simpler "fix" of just deleting the filter.

`spotlessApply spotlessCheck detekt` — clean, no findings.

## Reviewer notes

- Test-only change; scoped to `core/konsist`, no production code touched.
- Any **new** path rule added to this module must use `scanPath`, never the absolute `path`.
- Konsist failures seen from an agent worktree are now *real findings* — the previous advice to dismiss them as an environment artifact (and to pass `--continue` so they don't abort a baseline run) no longer applies.

## Summary by CodeRabbit

* **Tests**
  * Improved architectural validation scans for BLE-related and shared application code.
  * Added clearer diagnostics when scans return no results.
  * Standardized path handling and exclusions to make checks more reliable across development environments.

* **Chores**
  * Centralized project-path utilities used by validation checks.
  * Improved handling of nested development worktrees without affecting application behavior.

### DIFF
--- a/core/konsist/src/jvmTest/kotlin/org/meshtastic/core/konsist/BleAddressLoggingTest.kt
+++ b/core/konsist/src/jvmTest/kotlin/org/meshtastic/core/konsist/BleAddressLoggingTest.kt
@@ -53,17 +53,15 @@ class BleAddressLoggingTest {
      */
     private fun scannedFiles() = Konsist.scopeFromProject()
         .files
-        // scopeFromProject sweeps .claude/worktrees/ checkouts too; stale copies there
-        // resurface long-fixed lines as phantom offenders (paths match "/core/ble/").
-        .filterNot { "/.claude/" in it.path.normalizedProjectPath() }
-        .filter { file -> scannedPathFragments.any { it in file.path.normalizedProjectPath() } }
-        .filterNot { file -> identityUseAllowlist.any { file.path.normalizedProjectPath().endsWith(it) } }
+        .filterNot { it.isNestedAgentWorktree() }
+        .filter { file -> scannedPathFragments.any { it in file.scanPath } }
+        .filterNot { file -> identityUseAllowlist.any { file.scanPath.endsWith(it) } }
 
     @Test
     fun `the scan actually reaches the BLE sources`() {
-        val paths = scannedFiles().map { it.path.normalizedProjectPath() }
+        val paths = scannedFiles().map { it.scanPath }
 
-        assertTrue(paths.isNotEmpty(), "scoped scan matched no files at all — the path filter is wrong")
+        assertTrue(paths.isNotEmpty(), emptyScanMessage("BLE-scoped scan"))
         assertTrue(
             paths.any { it.endsWith("KableBleConnection.kt") },
             "expected core/ble sources in scope; got ${paths.size} files, e.g. ${paths.take(3)}",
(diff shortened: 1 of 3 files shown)


jamesarich:
### CI triage: the 11 failures were inherited from the base commit, not this change

Codecov flagged 11 failing tests. None are konsist — the full set is:

- `:core:service` — `MeshNotificationManagerImplTest` (3), `MeshNotificationManagerImplConversationTest` (1)
- `:core:ui` — `ConnectionsViewModelTest` (2)
- `:feature:connections` — `ScannerViewModelTest` (3), `AndroidScannerViewModelBondingTest` (1)
- `:feature:node` — `NodeDetailCompassLifecycleTest` (1)

That is the CMP 1.12.0-rc01 regression set. This branch was cut from `1b89b6392`, which is exactly one commit *before* the revert in #6664 landed, and `git merge-base --is-ancestor bc4e9da3a HEAD` confirmed the bad bump was in its ancestry (the branch pinned `compose-multiplatform = "1.12.0-rc01"`).

**Fix:** rebased onto `origin/main` (`d8361ccd1`, the revert). The pin is now `1.11.1` / `1.11.0-alpha07` and the diff against main is unchanged — still the same three files under `core/konsist`, 45 insertions / 17 deletions, no production code.

Re-verified on the rebased tree: `:core:konsist:allTests --rerun-tasks` 6/6 green including both "the scan actually reaches …" self-checks, plus `spotlessCheck` and `detekt` clean.

No code changes were made in response to review feedback — CodeRabbit reported no actionable comments, so there are no threads to reply to.

Issue #6263, opened 2026-07-14
[Bug]: App-provided nodes appear in list with spurious 0SNR 0RSSI 0HOPS


NomDeTom:
### Contact Details

I'm not giving my details to a machine

### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [x] I actually read this list, and should be taken seriously.

### Affected app version

2.8.0 CLOSED9

### Affected Android version

?

### Affected phone model

Galaxy S10

### Affected node model

promicro diy

### Affected node firmware version

2.8.0

### Steps to reproduce the bug

reconnect to previously used node with wiped nodeDB and some nodeDB entries
check nodelist
Note that some nodes are shown with spurious heard/hops/etc.

### Actual behavior

several nodes are listed in the app that are not in the node's own node db, nor are they transmitted over the bluetooth connection. Therefore they exist in the app only.

### Expected behavior

nodes don't keep coming back as last heard now, 0 hops, 0snr, 0rssi, etc.

### Screenshots/Screen recordings

<img width="1080" height="2044" alt="Image" src="https://github.com/user-attachments/assets/11c4e8a4-ac9c-4fec-ab65-4a296d91ceb9" />

### Relevant log output

```shell
Nothing relevant in the log - the nodes are coming from inside the building.

```

jamesarich:
Logs? Expected behavior/logic?
It seems like this might have been an underscoped feature.

NomDeTom:
> Logs? Expected behavior/logic? It seems like this might have been an underscoped feature.

Updated. I ruled out firmware-side effects and have a potential improvement there, but this is almost certainly app-side.

jamesarich:
Re-diagnosed this via the two closed auto-generated PRs against it (#6682, #6693) — removing the Stale label, this is still valid.

[...]

PR #6682, opened 2026-08-13
fix(database): prune nodes absent from radio snapshot

simulationstation:
## Summary

A completed Stage 2 NodeDB download is an authoritative snapshot of the radio's current nodes, but Android retained database and in-memory entries that were absent from that snapshot. Reconnecting to a wiped or different radio could therefore leave stale app-only nodes visible.

- delete absent node rows and metadata transactionally after a complete Stage 2 snapshot
- remove absent in-memory nodes before publishing readiness
- preserve app-local fields for nodes that remain in the snapshot
- cover DAO pruning and the Stage 2 cache transition with deterministic regressions

Addresses #6263.

## Testing

- `spotlessCheck detekt assembleDebug kmpSmokeCompile :core:data:allTests :core:database:jvmTest --max-workers=4 --console=plain`
- branch-relevant DAO and flow-manager regressions passed
- the canonical all-target gate stops in unchanged Windows Android-host DataStore/database failures previously reproduced on clean official main

Packaged emulator or physical-device UI acceptance was not performed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved radio configuration synchronization by removing outdated nodes that are no longer included in the latest snapshot.
  * Ensured refreshed node information is applied while retaining local notes.
  * Prevented stale node metadata from remaining after configuration updates.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConfigFlowManagerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConfigFlowManagerImpl.kt
@@ -298,6 +298,13 @@ class MeshConfigFlowManagerImpl(
             nodeManager.applyTrustedIdentityMigrations(removedNums)
         }
 
+        val installedNums = entities.mapTo(mutableSetOf()) { it.num }
+        val staleNums = nodeManager.nodeDBbyNodeNum.keys.filterNot { it in installedNums }
+        staleNums.forEach(nodeManager::removeByNodenum)
+        if (staleNums.isNotEmpty()) {
+            Logger.i { "Removed ${staleNums.size} node(s) absent from the completed radio snapshot" }
+        }
+
         val published =
             runForSession(session) {
                 nodeManager.setNodeDbReady(true)
--- a/core/data/src/commonTest/kotlin/org/meshtastic/core/data/manager/MeshConfigFlowManagerImplTest.kt
+++ b/core/data/src/commonTest/kotlin/org/meshtastic/core/data/manager/MeshConfigFlowManagerImplTest.kt
@@ -659,6 +659,38 @@ class MeshConfigFlowManagerImplTest {
     }
 
     @Test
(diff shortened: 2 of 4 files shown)


jamesarich:
Closing this — it inverts a deliberate design invariant: **the phone's node database is meant to grow independently of the radio's NodeDB.** The only sanctioned way for it to shrink is an explicit, user-initiated reset while connected.

## Why

**That sanctioned path already exists.** `AdminActionsUseCase.nodedbReset(..., isLocal = true)` sends the admin reset and then calls `nodeRepository.clearNodeDB(preserveFavorites)`. This PR performs a stronger version of that on every reconnect, with no user intent and no opt-out.

**It's not an edge case, it's the steady state.** The radio's NodeDB is capped and evicts entries. On any mesh larger than the radio's capacity, `radio snapshot ⊂ phone DB` is normal — so every reconnect would delete real history. Multi-radio users fare worse: connect to the base station and the handheld's nodes are pruned; connect back and the base's are pruned.

**It destroys favorites and app-local state.** The PR's own DAO test asserts this — `stale = NodeEntity(num = 2, user = User(id = "!2"), isFavorite = true)` is expected to be deleted. Even the explicit reset path honours `preserveFavorites`; this path has no such notion. `deleteNodesAndMetadata` also takes `notes`, ignore state, and the per-node `metadata` row.

**The completeness signal isn't strong enough to justify deletion.** "Complete Stage 2" is inferred from a nonce-matched `handleConfigComplete`. Any truncated or partially-dropped download that still lands the completion becomes a destructive, unrecoverable event.

## It also doesn't fix #6263

That report is about *display*: app-only nodes rendering 0 SNR, 0 RSSI, 0 hops, and last-heard-now. That's a presence-vs-sentinel-zero defect, not a retention problem. The fix belongs in rendering — don't paint sentinel zeros where the proto carries no presence, don't fabricate `lastHeard`, and optionally badge nodes that aren't in the connected radio's NodeDB. That resolves the reported symptom without changing what the phone remembers.

A PR along those lines would be welcome.

PR #6700, opened 2026-08-14
fix(notifications): stop repeated protected position alerts

ayysasha:
## Summary

- recognize the firmware's protected-position advisory on Android using an exact, fail-closed predicate
- replace repeated advisory alerts in one tagged notification slot with `setOnlyAlertOnce(true)`
- suppress the advisory's blocking global modal on Android while preserving normal behavior on other platforms
- use one shared identity for generic ClientNotification posting and cancellation

## Problem

Event firmware can reject phone-provided coordinates and return a `ClientNotification` with `Location sharing is disabled on this channel`. When phone location updates continue, Android receives a fresh reply roughly every 30 seconds. Each reply previously produced both a global modal and a high-importance system notification, repeatedly interrupting the user.

## Approach

The Android-specific path recognizes only notifications that match all current firmware characteristics: the exact message, `WARNING` level, a nonzero `reply_id`, no structured security payload, and non-OTA status.

Matching advisories:

- do not populate the global modal state on Android
- remain visible as a system notification
- reuse the tagged slot `protected_position_advisory` / ID `1`
- set `onlyAlertOnce`, so subsequent firmware replies update rather than repeatedly alert

All generic, structured security, OTA, Desktop, and other platform behavior keeps the existing path. This is stateless and does not infer event channels, inspect keys, or retain packet/session correlation state.

## Verification

- `./gradlew --rerun-tasks :core:data:jvmTest :core:service:testAndroidHostTest`
- `./gradlew :core:data:allTests :core:service:allTests :core:data:spotlessCheck :core:repository:spotlessCheck :core:service:spotlessCheck :core:ui:spotlessCheck :core:data:detekt :core:repository:detekt :core:service:detekt :core:ui:detekt :androidApp:assembleGoogleDebug`
- `git diff --check`

The Android host tests cover the exact advisory, all current predicate near-misses and structured payload variants, modal suppression/default-platform behavior, tagged replacement, `FLAG_ONLY_ALERT_ONCE`, generic alert behavior, stale sessions, and matched generic post/cancel identity.

No live emulator UI result is claimed because KVM acceleration was unavailable in the test environment; Robolectric/Android host tests and the Google debug assembly passed.

## Summary by CodeRabbit

* **New Features**
  * Added improved handling for client notifications, including platform-specific delivery and modal suppression.
  * Protected-position advisories now use stable notification IDs, tagged updates, and single-alert behavior.
  * Notification clearing now reliably cancels the corresponding alert.

* **Bug Fixes**
  * Prevented stale or retired sessions from updating notification state.
  * Improved replacement and cancellation behavior for recurring alerts.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/FromRadioPacketHandlerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/FromRadioPacketHandlerImpl.kt
@@ -35,6 +35,7 @@ import org.meshtastic.core.repository.RadioInterfaceService
 import org.meshtastic.core.repository.RadioSessionContext
 import org.meshtastic.core.repository.ServiceStateWriter
 import org.meshtastic.core.repository.XModemManager
+import org.meshtastic.core.repository.notificationId
 import org.meshtastic.core.resources.Res
 import org.meshtastic.core.resources.client_notification
 import org.meshtastic.core.resources.duplicated_public_key_title
@@ -154,7 +155,11 @@ class FromRadioPacketHandlerImpl(
 
     private fun handleClientNotification(cn: ClientNotification, session: RadioSessionContext) {
         val admitted =
-            radioInterfaceService.runIfSessionActive(session) { serviceStateWriter.setClientNotification(cn) }
+            radioInterfaceService.runIfSessionActive(session) {
+                if (!notificationManager.suppressClientNotificationModal(cn)) {
+                    serviceStateWriter.setClientNotification(cn)
+                }
+            }
         if (!admitted) {
             Logger.d { "Discarding client notification from stale transport session" }
             return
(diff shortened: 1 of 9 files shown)


jamesarich:
Thanks for this — the shape of the fix is right (suppress the modal, keep the tray notification, `onlyAlertOnce` for the re-alert noise), and the fail-closed predicate with its near-miss test coverage is nicely done.

While reviewing I found one issue in the identity half, so I've pushed a commit on top (19a59b3) rather than round-tripping review comments:

`notificationId()` was `toString().hashCode()`, which folds in `reply_id` and `time` — both change on every firmware reply. Anything that misses the protected-position predicate (firmware wording drift, a different `level`, `reply_id == 0`) fell through to the generic path with a fresh id every ~30s, stacking a new `IMPORTANCE_HIGH` tray entry per reply instead of coalescing — noisier than the pre-PR behavior the fix targets. It also meant the "one shared identity for posting and cancellation" goal didn't hold for the advisory itself, since posting overrode to a fixed id+tag that cancellation never used.

The commit derives the id from `message` + `level.value` + payload-variant key:

- stable across repeats, so every client notification coalesces into one tray slot regardless of whether the advisory predicate matches;
- hashes the enum's wire value rather than the enum object (enum hashCodes are identity-based, and the tray outlives the process — a repost after a service restart must land on the same id);
- lets the fixed-id/tag scheme in `AndroidNotificationManager` go away entirely, so posting and cancellation genuinely share one identity.

Added `ClientNotificationIdentityTest` plus a regression test that repeated generic client notifications coalesce (failed `expected:<1> but was:<2>` before the change). Full local run is green: `:core:repository`/`:core:service`/`:core:data` tests, spotless, detekt, and `:androidApp:assembleGoogleDebug`.

PR #6703, opened 2026-08-14
feat(node): label SNR quality on the Node Details signal row

beecho01:
## Why

The node list already labels signal quality (e.g. "-17.0 dB · Good") via `SignalInfo.kt`'s pill, using `determineSignalQuality`/`Quality` from `LoraSignalIndicator.kt` (#5903). The Node Details screen's `SignalRow` was missed - it only showed the raw SNR/RSSI numbers, with no interpretation.

## What changed

- `SignalRow` (`feature/node/.../NodeDetailsSection.kt`) now labels SNR with its quality word, reusing the existing, already-tested `determineSignalQuality` - no new thresholds introduced.
- Format is `"<raw value> · <quality>"` (e.g. `"-17.0 dB · Good"`), matching the value-then-quality order already used by the node-list pill in `SignalInfo.kt`, rather than inventing a new convention.
- `InfoItem` (`NodeDetailComponents.kt`) gained one optional `valueColor` parameter (default unchanged) so the SNR value can be tinted by quality, same as elsewhere in the app.
- RSSI is deliberately left unlabeled. The app's existing signal-quality logic is SNR-only (`LoraSignalIndicatorTest`'s `RSSI does not influence the rating`), because RSSI alone can't indicate quality without the noise floor - adding a quality word to RSSI here would contradict that existing, tested design decision.
- The raw value is never removed, only the quality word is added alongside it.

## Test plan

- New test: `feature/node/src/jvmTest/.../SignalRowQualityLabelTest.kt` - 6 desktop Compose UI tests (see the comment in `feature/node/build.gradle.kts` for why Compose UI tests live in `jvmTest`) covering all 4 quality bands (Good/Fair/Bad/None) at the `LONG_FAST` preset boundaries, confirms RSSI shows only the raw value, and confirms the SNR row is absent (not mislabeled) when a node has no SNR reading.
- `LoraSignalIndicatorTest` (the underlying quality logic) is unchanged and still passing - this PR only adds a new call site.
- Screenshot goldens regenerated and validated for the 4 affected previews (light/dark): `ScreenshotNodeDetailsSection`, `ScreenshotNodeDetailsSectionSigned`, `ScreenshotNodeDetailContentLocal`, `ScreenshotNodeDetailContentRemote`.
- Local baseline green: `spotlessApply spotlessCheck detekt assembleDebug test allTests` and `kmpSmokeCompile`.
- **Hardware smoke test:** installed the fdroid debug build on a physical device (Samsung Galaxy S24+) connected to a real Meshtastic radio, with a directly-heard node in range. Confirmed the Node Details SNR label matches the existing node-list pill for the same node, same quality word and color.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


jamesarich:
Thanks for this, it's a really nice first PR. Reusing `determineSignalQuality` instead of introducing new thresholds is exactly the right call, and I appreciate that you left RSSI unlabeled and wrote down why. The hardware smoke test is above and beyond.

Three small things and then I'm happy to get this in:

1. **`VerificationRow`'s KDoc** picked up a "tappable" to "tap-able" change that looks incidental to the SNR work. Mind reverting it? "tappable" is what we use elsewhere (`AutoLinkText.kt`, `NodeSignedStatusIcon.kt`, `FdroidMapOverlayRenderer.kt`), so it keeps us consistent.

2. **Description vs. code on the format string:** the PR body says `"-17.0 dB - Good"` but the code emits `" · "`. The middot is fine, could you just update the description to match? It ends up in the commit history, so I like keeping those honest.

3. **Minor labelling one:** the test plan calls these "Robolectric Compose tests", but they're actually desktop Compose UI tests in `jvmTest`. There's a comment in `feature/node/build.gradle.kts` explaining that Compose UI tests live in `jvmTest` specifically because the Robolectric idling strategy NPEs on the null `Build.FINGERPRINT` the androidHostTest stubs leave behind. The tests themselves are good, just the label.

No rush on any of it. Thanks again for the test coverage, especially the "SNR row is absent rather than mislabeled" case.

beecho01:
Hi @jamesarich,

These should all be complete now:
1. This was an error on my part. VSCode was complaining about the spelling, and I didn't check the correct terminology before making the change.
2. Updated the description to use `·` so it matches what the code actually uses.
3. Updated the test plan to just describe them as desktop Compose UI tests living in `jvmTest`, rather than calling them `Robolectric`.

Thanks for the thorough review, appreciate you catching all three. Let me know if you need or recommend anything more.

jamesarich (approved):
Nice add, thanks!

Issue #6704, opened 2026-08-14
Fresh-setup preset default overwrites a preset that is still legal in the chosen region


NomDeTom:
Tom says "Clod, remember to be nice when you make the request"

### What happens

Community/vendor firmware builds can pin a modem preset (`USERPREFS_LORACONFIG_MODEM_PRESET`) while leaving the region unset, so the user picks their own region on first connect. A fresh flash of such a build reaches the app as **region `UNSET` + `SHORT_TURBO`** (or whatever the mesh runs on).

In `LoRaConfigItemList.kt`, selecting a region from `UNSET` discards that preset unconditionally:

```kotlin
val freshSetup = formState.value.region == RegionCode.UNSET
val repaired = regionPresetMap.repairPresetFor(region, formState.value.modem_preset)
val preset = if (freshSetup) {
        regionPresetMap.constraintFor(region)?.defaultPreset ?: defaultPresetFor(region) ?: repaired
    } else { repaired }
```

Picking **US** with `SHORT_TURBO` selected yields `LONG_FAST` (the firmware's advertised US default) on 2.8 firmware, or `LONG_TURBO` (`REGION_DEFAULT_PRESETS`) on older firmware. Note `repaired` would have **kept** `SHORT_TURBO` — it's in `PRESETS_STD` and perfectly legal in US. It's the `freshSetup` branch that drops it.

The user gets no indication this happened. The node ends up on different airtime parameters and a different frequency slot from the mesh it was flashed for, and simply sees nobody.

Unlike Apple's equivalent, the branch isn't region-scoped, so it fires for every region rather than just US.

### Why it's reasonable today

For a stock install the rule is right: region `UNSET` normally carries a meaningless `LONG_FAST` placeholder, and adopting the region's default is the friendly thing to do. The app has no way to tell a placeholder from a deliberate preset, because the firmware doesn't currently say.

### Suggestion

Two parts, and the first stands alone:

1. **Only override when the current preset is actually illegal in the new region.** That's already what `repairPresetFor()` computes, so the `freshSetup` branch could be narrowed to fire only when `repaired != current` — preserving the intent of the rule (a placeholder `LONG_FAST` still becomes the region default where the firmware says so) without discarding a legal, deliberate choice.

[...]

PR #6711, opened 2026-08-14
feat(lora): honor a pinned-preset intent advertised for UNSET

jamesarich:
## Why

Second half of #6704 (stacked on the placeholder-gate PR).

The placeholder heuristic in the previous PR can't tell a vendor build that deliberately pins `LONG_FAST` from a stock install: both report region `UNSET` + `LONG_FAST`. meshtastic/firmware#11507 closes that gap by advertising an `UNSET` entry in `LoRaRegionPresetMap` when, and only when, the build pins a preset, stating the pin as the group's sole entry and default.

## What changed

🌟 `presetForRegionChange()` now treats a non-null `constraintFor(RegionCode.UNSET)` as an explicit statement that the current preset is deliberate: fresh setup keeps it (legality-repaired only), even when it is `LONG_FAST`. Firmware that sends no `UNSET` entry (stock builds, and everything shipped today) falls through to the placeholder heuristic unchanged.

## Notes for reviewers

- The `UNSET` lookup is scoped to the fresh-setup decision only: the picker's constraint lookup now explicitly skips `UNSET`, so a pinned build's single-entry group never collapses the dropdown while the region is still unset — matching #11507's "intent, not enforcement", and confirmed as the intended reading by the map's author on #6704. Behavior-neutral on firmware without the entry (`constraintFor(UNSET)` was null there anyway).
- Inert on today's firmware: no shipped release emits an `UNSET` entry, so behavior is identical to the previous PR until firmware#11507 lands.

## Testing Performed

- `:core:model:allTests` — new tests: a map with an `UNSET` entry keeps a pinned `LONG_FAST` at fresh setup; the same pin is still legality-repaired when illegal in the chosen region; maps without an `UNSET` entry keep the placeholder heuristic.
- `:feature:settings:allTests`, `spotlessCheck`, `detekt`, `kmpSmokeCompile` — green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserves deliberately selected radio presets during fresh setup, including `LONG_FAST`.
  * Repairs presets when they are not valid for the selected region.
  * Keeps preset selection unrestricted while the region remains unset.

### DIFF
--- a/core/model/src/commonMain/kotlin/org/meshtastic/core/model/RegionPresetConstraint.kt
+++ b/core/model/src/commonMain/kotlin/org/meshtastic/core/model/RegionPresetConstraint.kt
@@ -82,6 +82,9 @@ fun LoRaRegionPresetMap?.repairPresetFor(region: RegionCode, current: ModemPrese
  * [current] is the [ModemPreset.LONG_FAST] placeholder a factory-flashed node reports (proto default), the region's
  * advertised default is adopted instead (the firmware map's when present, else the app's built-in default). Any other
  * preset at UNSET was set deliberately (e.g. a vendor build pinning USERPREFS_LORACONFIG_MODEM_PRESET) and is kept.
+ *
+ * A build that pins a preset can also advertise it as an UNSET map entry (firmware #11507), stating outright that the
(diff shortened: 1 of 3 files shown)


jamesarich:
@coderabbitai full review

PR #6716, opened 2026-08-15
fix(lifecycle): harden packet admission and transport ownership

jeremiah-k:
## Overview

This makes outbound packet admission explicit across the queue, service, and transport layers and keeps admitted work bound to the connection or transport generation that owns it.

Rejected or stale sends no longer create dispatch evidence, response timers, optimistic state, or follow-on work. Accepted sends retain their ownership through transport admission, queue status, routing completion, cancellation, reconnects, and teardown.

The same admission boundary is carried through settings transactions and long-lived command callers so transient lifecycle rejection remains a specific, recoverable outcome rather than turning into stale timers, misleading success state, or unrelated secondary failures.

### Transport lifecycle

* Require transport handoff to report synchronous admission across BLE, serial, TCP, replay, mock, and inert transports.
* Bind queued writes, delayed callbacks, heartbeat work, and shared-service operations to the transport generation that admitted them.
* Bound framed-send queues and allow already-admitted stream frames to drain before cancellation; a forced drain timeout makes the underlying stream terminal rather than permitting partial-frame reuse.
* Serialize BLE session retirement so profile-failure, reconnect, and explicit-close paths cannot independently clean up the same retired GATT session.
* Keep connect/disconnect callbacks inside lifecycle ownership and prevent stale callbacks from acting on replacement sessions.
* Reject stale serial-device selection before transport creation. A stale saved address can recover when exactly one serial device is attached; with multiple devices, the ambiguous selection is rejected instead of silently choosing the first driver.
* Keep detached serial cleanup scoped to the transport lifetime even when its caller is cancelled during close.
* Treat bounded close/drain failures as reported operational outcomes rather than exceptions that abandon the remaining teardown work.
* Document and account for nested lifecycle-close budgets where one transport owns another bounded lifecycle.

### Packet admission and ownership

* Reserve packet IDs from admission through every required terminal stage so duplicate sends cannot take ownership of acknowledgements.
* Keep the expected connection version attached to connection-owned queued packets and revalidate it immediately before dispatch.
* Distinguish transport queueing from routing completion while keeping both stages under one pending packet owner.
* Start response timers and record dispatch evidence only after the active transport accepts the frame.
* Own strict-routing expiry in service scope so caller cancellation cannot strand a packet-ID reservation; a late expiry is identity-checked so it cannot affect a reused ID.
* Correlate packet-specific queue status with dispatched work while keeping unscoped full-queue status advisory.
* Preserve specific routing ACK/NAK, timeout, transport-stop, and send-failure outcomes instead of allowing later queue state to overwrite them.
* Resolve persisted outgoing packets through one shared mesh-identity predicate and reject ambiguous packet or reaction matches instead of mutating an arbitrary row.
* Keep packet and reaction persistence aligned with terminal ownership, including timeout handling for persisted ENROUTE work.
* Keep `TRANSPORT_STOPPED` distinct from a known pre-dispatch rejection so callers can use dispatch evidence before deciding whether a retry is safe.

### Command and settings boundaries

* Keep queue and local-node rejection typed through command APIs instead of allowing it to degrade into a later response timeout.
* Capture the local destination once for the lifetime of a settings transaction.
* Require begin, staged writes, and commit handling to respect the active connection lifecycle.
* Preserve the original edit failure when commit cleanup also fails, while retaining the cleanup failure for diagnostics.
* Keep fixed-position removal distinct from an unavailable local node identity.
* Treat an omitted routing `error_reason` as absence of a routing error instead of constructing an error result with a null reason.
* Preserve specific routing failures through settings response windows and queued channel operations.
* Keep position and serial-open diagnostics useful without writing raw location data or local OS-user information to logs.

### Caller containment

* Retry connection bootstrap work only while the same connected lifecycle still owns the request.
* Bind history replay and other connection-owned follow-up work to the generation that scheduled it.
* Keep optional telemetry, discovery, history, and widget requests from escalating expected transport transitions.
* Prevent rejected node and configuration requests from starting optimistic state, success callbacks, or response timers.
* Run node-management success callbacks after the guarded radio mutation so callback failures cannot be misreported as send rejection.
* Surface localized feedback for actionable node-request rejection while leaving unexpected failures visible.
* Keep lifecycle-aware fakes and fixtures aligned with the production admission, settings, and ownership contracts.

## Testing

Added or extended coverage for:

* unavailable/opening transports, bounded send queues, active-frame drain, close/send races, stale callbacks, replacement sessions, serialized BLE cleanup, heartbeat rejection, unambiguous single-device USB recovery, ambiguous multi-device rejection, and serial lifecycle cleanup;
* Android serial selection/factory behavior and TCP lifecycle races in addition to the common transport suites;
* packet-ID reservation, duplicate admission, stale connection generations, dispatch ordering, zero-ID and packet-specific queue status, ambiguous persisted identities, routing ACK/NAK, owner-independent timeout after caller cancellation, ID reuse after expiry, reconnect, and terminal persistence;
* edit-settings begin/commit boundaries, cancellation cleanup, failure preservation, destination ownership, fixed-position rejection, and full profile installation;
* absent versus explicit routing errors, connection-epoch invariants, and concurrent lifecycle publication using matching real-time dispatch;
* connection bootstrap generation changes, history replay ownership, rejected node/configuration requests, optimistic side-effect suppression, telemetry/history/discovery best-effort callers, and widget refresh behavior;
* lifecycle-aware production fakes and caller fixtures used throughout the affected service, command, persistence, and transport tests.

## Merge Order

Merge this PR before #6717. Discovery restoration in #6717 uses this PR's admission-aware local settings transaction so a cleared or undispatched commit remains retryable instead of being recorded as a successful restore. #6718 is independent of this ordering.

The combined stack has also been validated together, so this ordering is about the runtime contract between #6716 and #6717 rather than resolving source conflicts.

## Validation

Validated as part of the combined integration stack with an overnight soak and repeated real-device handoffs.

* Switched between several physical radios across BLE, USB/serial, and TCP, including repeated disconnect, reconnect, and device-selection changes while the app remained active.
* Exercised transport replacement under normal traffic rather than only idle connections. The captured run retired the previous connection and packet queue, switched per-device state, advanced transport ownership, and completed bootstrap on the replacement radio.
* Verified normal packet and service traffic continued after device replacement rather than inheriting stale work from the previous connection.
* Exercised packet-specific queue admission and routing completion throughout the soak, including successful queue acknowledgements and routing ACKs after transport handoffs.
* Sent normal messages and background/service-generated traffic while changing radios and interfaces.
* Repeated app background/foreground and connection lifecycle transitions without observing a stuck transport, stale packet appearing on a replacement radio, or lifecycle-related crash.
* The rebuilt PR branch is green in CI after the final standalone-branch cleanup and review fixes.

The overnight soak exercised these changes together with the discovery and remote-admin stacks, including repeated remote-admin traffic across multiple nodes and interface switches, providing additional coverage of the shared connection-generation and teardown boundaries under realistic concurrent load.

## Scope

* No Meshtastic protocol, message-format, or firmware behavior changes are introduced.
* Internal transport, packet-admission, persistence, and command contracts are updated consistently across their in-repo implementations and callers.
* Existing reconnect and transport-selection policy is unchanged except where stale or unowned work must be rejected or admitted work must be drained safely.
* Serial fallback is deliberately limited to an unambiguous single attached device. A stale saved address with multiple attached serial devices is rejected rather than selecting an arbitrary driver.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of radio connections, reconnections, shutdowns, and queued sends.
  * Prevented stale connection sessions from processing packets or affecting current sessions.
  * Added clearer handling for rejected requests, unavailable local nodes, timeouts, and routing failures.
  * Fixed persistence and status tracking for outgoing messages and emoji reactions.
  * Node actions now show localized feedback when requests cannot be sent.

* **Improvements**
  * Added lifecycle-aware connection status and more accurate send acceptance reporting.
  * Improved settings transactions, history replay, telemetry, discovery scans, and fixed-position removal handling.
  * Added safeguards for fan-out configuration reads and cancellation of manual writes.

### DIFF
--- a/.skills/compose-ui/strings-index.txt
+++ b/.skills/compose-ui/strings-index.txt
@@ -1237,6 +1237,7 @@ node_list_help_node_details
 node_list_help_title
 node_list_long_click_label
 node_number
+node_request_send_failed
 node_restarting
 node_sort_alpha
 node_sort_button
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/CommandSenderImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/CommandSenderImpl.kt
@@ -32,13 +32,18 @@ import org.meshtastic.core.model.NodeAddress
 import org.meshtastic.core.model.Position
 import org.meshtastic.core.model.TelemetryType
 import org.meshtastic.core.model.util.isWithinSizeLimit
+import org.meshtastic.core.repository.AwaitedSendResult
 import org.meshtastic.core.repository.CommandSender
+import org.meshtastic.core.repository.LocalNodeUnavailableException
 import org.meshtastic.core.repository.NeighborInfoHandler
 import org.meshtastic.core.repository.NodeManager
 import org.meshtastic.core.repository.PacketHandler
+import org.meshtastic.core.repository.PacketQueueRejectedException
 import org.meshtastic.core.repository.RadioConfigRepository
 import org.meshtastic.core.repository.SessionManager
 import org.meshtastic.core.repository.TracerouteHandler
+import org.meshtastic.core.repository.toFixedPositionAdminMessage
+import org.meshtastic.core.repository.toFixedPositionProto
 import org.meshtastic.proto.AdminMessage
 import org.meshtastic.proto.AirQualityMetrics
 import org.meshtastic.proto.ChannelSet
@@ -89,6 +94,8 @@ class CommandSenderImpl(
     override fun getCachedChannelSet(): ChannelSet = channelSet.value
 
     override fun getCurrentPacketId(): Long = currentPacketId.value
+
+    private fun Int.nonZeroRequestId(): Int = takeUnless { it == 0 } ?: generatePacketId()
 
     override fun generatePacketId(): Int {
         val numPacketIds = ((1L shl PACKET_ID_SHIFT_BITS) - 1)
@@ -151,10 +158,15 @@ class CommandSenderImpl(
             p.status = MessageStatus.QUEUED
         }
 
-        sendNow(p)
+        if (!sendNow(p)) {
+            p.status = MessageStatus.ERROR
+            // Persistence owners treat a normal return as successful admission; throw so they can requeue or fail it.
+            throw PacketQueueRejectedException("Data packet")
+        }
+        p.time = nowMillis
     }
 
-    private suspend fun sendNow(p: DataPacket) {
+    private suspend fun sendNow(p: DataPacket): Boolean {
         val meshPacket =
             buildMeshPacket(
                 to = resolveNodeNum(NodeAddress.fromString(p.to)),
@@ -170,15 +182,47 @@ class CommandSenderImpl(
                     emoji = p.emoji,
                 ),
             )
-        p.time = nowMillis
-        packetHandler.sendToRadio(meshPacket)
+        return packetHandler.sendToRadio(meshPacket)
     }
 
+    private suspend fun enqueueOrThrow(packet: MeshPacket, operation: String, expectedConnectionVersion: Long? = null) {
+        val accepted =
+            if (expectedConnectionVersion == null) {
+                packetHandler.sendToRadio(packet)
+            } else {
+                packetHandler.sendToRadioForConnection(packet, expectedConnectionVersion)
+            }
+        if (!accepted) throw PacketQueueRejectedException(operation)
+    }
+
+    private fun buildAdminMessagePacket(
+        destNum: Int,
+        requestId: Int,
+        wantResponse: Boolean,
+        initFn: () -> AdminMessage,
+    ): MeshPacket = buildAdminPacket(
+        to = destNum,
+        id = requestId.nonZeroRequestId(),
+        wantResponse = wantResponse,
+        adminMessage = initFn().copy(session_passkey = sessionManager.getPasskey(destNum)),
+    )
+
     override suspend fun sendAdmin(destNum: Int, requestId: Int, wantResponse: Boolean, initFn: () -> AdminMessage) {
-        val adminMsg = initFn().copy(session_passkey = sessionManager.getPasskey(destNum))
-        val packet =
-            buildAdminPacket(to = destNum, id = requestId, wantResponse = wantResponse, adminMessage = adminMsg)
-        packetHandler.sendToRadio(packet)
+        enqueueOrThrow(buildAdminMessagePacket(destNum, requestId, wantResponse, initFn), "Admin command")
+    }
+
+    override suspend fun sendAdminForConnection(
+        destNum: Int,
+        expectedConnectionVersion: Long,
+        requestId: Int,
+        wantResponse: Boolean,
+        initFn: () -> AdminMessage,
+    ) {
+        enqueueOrThrow(
+            buildAdminMessagePacket(destNum, requestId, wantResponse, initFn),
+            "Admin command",
+            expectedConnectionVersion,
+        )
     }
 
     override fun sendAdminImmediate(destNum: Int, initFn: () -> AdminMessage) {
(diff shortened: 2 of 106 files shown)


jamesarich (approved):
Ran this through a deep multi-reviewer pass (transports, repository contracts, data managers, service/UI) plus a combined-tree build with #6717 and #6718. The architecture holds up: synchronous admission is genuinely synchronous across all seven transports, terminal-outcome ownership and the DELIVERED-vs-timeout races are sound in both orders, cancellation is rethrown everywhere I traced, and the fixed-position change quietly fixes a latent presence-vs-sentinel bug (the old `pos != Position(0.0, 0.0, 0)` compared `time` too, so removals with a non-matching timestamp were sent as set_fixed_position). Moving `sendToRadio` to `RadioTransportWriter` to stay under the detekt function cap without baseline edits was the right call. I also verified the octopus merge of all three PRs compiles and passes the overlapping modules' tests, so no merge-queue surprises in any order.

Two comments worth a response before or after merge, neither blocking:

1. The single-device USB fallback was narrowed, not preserved. `SerialRadioTransport.resolveSerialDevice` is now `devices[address] ?: devices.values.singleOrNull()`, where main used `firstOrNull()`. With a stale saved address and two or more serial devices attached, the old code self-healed onto the first driver; the new code fails with "Serial device not found" and `isAddressValid` goes false. Reasonable hardening, but the PR text says the historical fallback is kept. Either revert to `firstOrNull` or own it as a deliberate change in the description.

2. Packet-ID reservation leaks when a strict-await caller is cancelled mid-flight. `awaitAdmittedPacket` deliberately keeps ownership on caller cancellation, but the 30s routing timeout lives only in the cancelled caller's coroutine. If the queue stage completed ACCEPTED and the routing verdict is lost, the `queueResponse` entry stays non-terminal until disconnect, so the ID is rejected as DuplicateId for the rest of the connection. That contradicts the PacketHandler KDoc's promise that a completed ID may be reused by a retry. The reservation wants an owner-independent expiry.

Non-blocking observations from the sweep: `stopPacketQueue` drains at coroutine-run time rather than call time, so a packet admitted on a fast reconnect in that window can get stamped TRANSPORT_STOPPED (parity with main's async structure, but a generation check inside the drain block would close it); the settings transaction is not pinned to a connection version (the begin/write/commit callers are back-to-back today so the FIFO saves you, but a future suspending editSettings block could span connections); the commit-departure heuristic treats dispatched-then-departed as durable, which is the right call for the LoRa reboot case but worth remembering it applies staged projections on inference; and TCP is the one transport without a backlog cap on `launchConnectionOperation`.

jeremiah-k:
@jamesarich Thanks for the deep pass. Both points make sense.

The serial fallback narrowing is intentional. With one attached serial device, a stale saved address can still recover unambiguously. With multiple devices, I don't want a stale address to silently select whichever driver happens to be first, so I'll update the description to state that explicitly rather than claiming the previous fallback is fully preserved.

The cancelled strict-await reservation is a real gap. Caller cancellation should not release ownership immediately because the admitted packet may still complete, but the reservation also shouldn't depend on that caller remaining alive to reach its terminal timeout. I'll move that expiry under owner-independent lifecycle ownership and keep the existing terminal/routing race semantics intact.

The other observations are useful, but I agree they don't need to expand this PR's scope.

jeremiah-k:
@jamesarich Follow-up is complete.

The serial behavior is intentional and the PR description now says so directly: a stale saved address can recover only when there is exactly one attached serial device. With multiple devices, the selection is ambiguous and is rejected rather than silently choosing the first driver.

Strict-routing expiry is now owned by service scope after dispatch instead of by the awaiting caller. Cancelling that caller still does not release ownership prematurely, but the 30-second terminal expiry now completes the pending response and releases the packet ID independently. The late-expiry path retains the pending-response identity check so it cannot affect a newer owner that reused the same ID. Coverage pins both an active strict waiter timing out and a cancelled waiter releasing the ID for retry after expiry.

I also made the merge order explicit: #6716 should land before #6717 because discovery restoration relies on this PR's admission-aware local settings commit boundary. I left the other non-blocking observations out of this already-large change.

PR #6718, opened 2026-08-15
fix(admin): retain session refresh across multi-hop latency

jeremiah-k:
## Overview

Remote-admin metadata/session refresh can be shared by multiple callers, but the in-flight refresh previously remained vulnerable to whichever caller returned or cancelled first. On multi-hop meshes, the existing 10-second user-facing deadline could also report failure before a valid response arrived.

This keeps one refresh owned by service scope until the shared deferred actually completes and extends the user-facing metadata deadline to 30 seconds. Caller cancellation no longer removes work still awaited by another participant, and valid responses arriving after the former deadline remain eligible to refresh durable session state.

## Key Changes

* Keep one in-flight metadata/session refresh per remote node and own it from service scope until completion.
* Identity-check completion cleanup so an older or cancelled caller cannot remove a newer refresh.
* Allow one participant to cancel without cancelling or orphaning the shared work needed by another caller.
* Extend the remote-admin metadata deadline from 10 to 30 seconds for realistic multi-hop response latency.
* Keep late valid responses eligible to update durable remote-admin session state.
* Derive timeout behavior from one production constant and avoid reusing already-completed shared refresh work.
* Keep disconnect and refresh completion outcomes type-safe at the terminal select boundary.

## Testing

Added coverage for:

* two callers sharing one in-flight refresh;
* cancellation of one participant while the remaining caller receives the shared result;
* completion cleanup not removing a newer refresh for the same node;
* a valid response arriving after the former 10-second deadline;
* the 30-second terminal timeout;
* completed refresh work not being reused by a later request.

## Validation

Remote admin was exercised repeatedly during the combined integration soak, including across device changes and reconnects.

* Tested remote-admin operations against multiple nodes while switching among several physical radios and all three supported connection interfaces.
* Verified connection departure clears the existing remote-admin session entries before the replacement transport is established, so a later connection does not inherit the previous radio's cached session state.
* Observed multiple successful metadata/session refreshes from different remote nodes during the captured run, including fresh session passkeys being recorded after the device handoff.
* Captured a particularly relevant multi-hop latency case: an ADMIN request was dispatched at `12:22:03.033` and admitted by the packet queue at `12:22:03.301`; the corresponding remote metadata/session response was received and recorded at `12:22:13.231`, approximately 10.2 seconds after dispatch. That response arrived beyond the former 10-second user-facing deadline and was still accepted as a valid session refresh.
* Continued issuing admin traffic after the delayed refresh; subsequent ADMIN packets were admitted normally and routing ACKs continued to complete successfully.
* Repeated remote-admin use, reconnects, and device changes during the overnight soak did not expose stale-session reuse or a caller cancellation/lifecycle transition that stranded subsequent admin access.

This provides real-device coverage of the behavior targeted by the change in addition to the focused shared-refresh, cancellation, reconnect, completed-work reuse, and 30-second timeout regression tests.

## Scope

* This is limited to remote-admin metadata/session refresh ownership and its user-facing response deadline.
* Routing, packet transport, reconnect policy, admin message formats, and firmware behavior are unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved remote admin session handling during concurrent requests.
  - Session checks now respond correctly when connectivity is lost and recover cleanly after reconnection.
  - Prevented canceled requests from interrupting other active session checks.
  - Increased the session refresh timeout from 10 to 30 seconds, allowing more time for delayed responses.

- **Documentation**
  - Updated session timeout documentation to reflect the new 30-second limit.

### DIFF
--- a/core/domain/README.md
+++ b/core/domain/README.md
@@ -50,7 +50,7 @@ Ensures a per-node remote-admin passkey session exists before entering the remot
 sealed interface EnsureSessionResult {
     data object AlreadyActive   : EnsureSessionResult  // passkey already fresh
     data object Refreshed       : EnsureSessionResult  // metadata response arrived
-    data object Timeout         : EnsureSessionResult  // no response within 10 s
+    data object Timeout         : EnsureSessionResult  // no response within 30 s
     data object Disconnected    : EnsureSessionResult  // radio not connected
 }
 ```
--- a/core/domain/src/commonMain/kotlin/org/meshtastic/core/domain/usecase/session/EnsureRemoteAdminSessionUseCase.kt
+++ b/core/domain/src/commonMain/kotlin/org/meshtastic/core/domain/usecase/session/EnsureRemoteAdminSessionUseCase.kt
@@ -125,7 +125,7 @@ open class EnsureRemoteAdminSessionUseCase(
                 try {
                     if (disconnected.isCompleted) return@coroutineScope EnsureSessionResult.Disconnected
                     radioController.refreshMetadata(destNum)
-                    select {
+                    select<EnsureSessionResult> {
                         disconnected.onAwait { EnsureSessionResult.Disconnected }
                         refreshed.onAwait { EnsureSessionResult.Refreshed }
                     }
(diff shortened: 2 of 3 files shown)


jamesarich (approved):
The core fix is right. The old caller-side finally could never run its map cleanup when the awaiting caller was cancelled (a cancelled coroutine can't acquire the suspending mutex), which stranded a completed entry in the dedupe map and let later callers pick up a stale result with no fresh dispatch. Moving cleanup into the shared deferred's own finally under NonCancellable with the identity check, plus the isCompleted guard on lookup, closes that properly. The select over refreshed/disconnected sits inside a try whose finally cancels both children, so the inner coroutineScope can't hang on the losing subscription, and the timeout path completes the deferred normally so cleanup still runs. Test coverage hits the right cases and the late-response test deriving its boundary from the production constant is a nice touch. The 10.2s real-device trace justifies the 30s deadline.

Two non-blocking nits, no action needed:

1. select is biased to its first clause, so a simultaneous refresh + disconnect returns Disconnected even though the session was durably refreshed. Harmless since the UI observes the durable SessionStatus flow and self-heals, which the KDoc already covers.
2. A caller awaiting a serviceScope-cancelled deferred receives a CancellationException it didn't cause (the classic external-deferred await leak). Pre-existing behavior, not introduced here.

PR #6742, opened 2026-08-16
fix(database): raise default cache limit and warn before eviction

### LINKED ISSUE #6186: [Feature Request]: Increase the default Device DB cache limit
### Checklist

### Contact Details

_No response_

The default value for `node_db_cache_limit` is too low; this can cause problems for users with a large number of devices or those involved in building nodes.

### Why should this be added?

Increasing the default value of `node_db_cache_limit` from 3 to a higher value will prevent the sudden loss of chat histories for users working with a large number of nodes.

_No response_

jamesarich:
## Summary
- Raises `DatabaseConstants.DEFAULT_CACHE_LIMIT` from 3 to 5, so users who never touched Settings > Persistence get a more forgiving default.
- Adds a confirmation dialog before a lowered device DB cache limit would actually evict (and permanently delete the local history of) currently-cached devices. Shared between Android and Desktop settings screens via a new `CacheLimitPreference` composable. Raising or leaving the limit unchanged applies immediately with no prompt.

Fixes #6186.

## Background
The eviction warning was originally deferred with the reasoning that all `enforceCacheLimit()` call sites were background fire-and-forget coroutines with no UI context. An adversarial review found that claim was wrong: `SettingsViewModel.setDbCacheLimit()` — wired directly to the Settings > Persistence dropdown on both Android and Desktop — calls `DatabaseManager.setCacheLimit()` directly, which does have full UI/ViewModel context available. This PR implements the warning there instead of deferring it.

## Testing Performed
- Full baseline (`spotlessApply spotlessCheck detekt assembleDebug test allTests`) green.
- `:screenshot-tests:validateDebugScreenshotTest` green — no goldens needed regeneration (the new dialog only renders when an eviction is pending, so default states are unaffected).
- New unit tests: `SettingsViewModelTest` covers `cachedDeviceCountExceeding` for both the no-eviction and eviction-pending cases.

### DIFF
--- a/.skills/compose-ui/strings-index.txt
+++ b/.skills/compose-ui/strings-index.txt
@@ -151,6 +151,7 @@ broadcast_interval
 busy_noise_floor
 button_gpio
 buzzer_gpio
+cache_limit_eviction_warning
 calculating
 call_sign
 call_sign_summary
--- a/core/common/src/commonMain/kotlin/org/meshtastic/core/common/database/DatabaseManager.kt
+++ b/core/common/src/commonMain/kotlin/org/meshtastic/core/common/database/DatabaseManager.kt
@@ -29,6 +29,9 @@ interface DatabaseManager {
     /** Sets the database cache limit. */
     fun setCacheLimit(limit: Int)
 
+    /** Returns how many device-specific databases are currently cached on disk, subject to eviction. */
+    suspend fun cachedDeviceDbCount(): Int
+
     /** Switches the active database to the one associated with the given [address]. */
     suspend fun switchActiveDatabase(address: String?)
 
--- a/core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseConstants.kt
+++ b/core/database/src/commonMain/kotlin/org/meshtastic/core/database/DatabaseConstants.kt
@@ -26,7 +26,12 @@ object DatabaseConstants {
     const val DEFAULT_DB_NAME: String = "${DB_PREFIX}_default"
 
     const val CACHE_LIMIT_KEY: String = "node_db_cache_limit"
-    const val DEFAULT_CACHE_LIMIT: Int = 3
+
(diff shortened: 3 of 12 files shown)


jamesarich:
Follow-up per adversarial review: the claim that no UI-facing eviction call site exists was wrong — `SettingsViewModel.setDbCacheLimit()` (Settings > Persistence, both Android and Desktop) calls `DatabaseManager.setCacheLimit()` directly with full ViewModel context available.

Added `DatabaseManager.cachedDeviceDbCount()` and a shared `CacheLimitPreference` composable that warns before a lowered limit would actually evict cached devices, showing how many would be affected. Only gates the case where lowering the limit would cause an eviction — raising or leaving it unchanged applies immediately, no prompt. Wired into both `PersistenceSection.kt` (Android) and `DesktopSettingsScreen.kt` (Desktop), since both call `setDbCacheLimit` independently.

Full baseline + screenshot validation green. Title/body updated — this now fixes #6186 fully rather than partially.

jamesarich:
@coderabbitai review

jamesarich:
@coderabbitai full review

jamesarich:
@coderabbitai full review

jamesarich:
@coderabbitai full review

jamesarich:
The GitHub App's CodeRabbit review kept failing generically on this PR across three separate manual triggers, so I ran the local CodeRabbit CLI directly against the branch instead. It found two real issues, both fixed:

1. \`cachedDeviceDbCount()\` ran the mutex-protected file listing without dispatching to \`dispatchers.io\`, unlike every other filesystem-touching method in this class (\`getOrOpenDatabase\`, \`drainWriters\`, etc. all wrap in \`withContext(dispatchers.io)\`). Fixed.
2. \`CacheLimitPreference\`'s selection handler had no guard against out-of-order async completion: picking two different lower limits in quick succession could let the first (slower) eviction check overwrite the dialog state set by the second (faster) one. Added a request-token guard so only the latest selection's result is applied.

Full baseline green.

PR #6743, opened 2026-08-16
fix(connections): open the region picker directly from the region card

### LINKED ISSUE #6586: [Bug]: Blank dialog appears when tapping "Set your region" on the Connection screen
### Contact Details

pablo.leandro.ac@gmail.com

### Checklist

### Affected app version

2.7.14 (29321034) google

### Affected Android version

Android 14 (UP1A.230905.011) - HyperOS 1.0.13.0.UNLMIXM

### Affected phone model

Xiaomi POCO X6 Pro 5G

### Affected node model

seeed studio - Wio Tracker L1 Pro

2.7.26.54e0d8d Beta

1. Flash or reset a node so that no LoRa region is configured.
2. Connect to the node using the Android app (BLE or Serial).
3. Wait until the "Set your region" card appears on the Connection screen.
4. Tap the "Set your region" card.
5. A loading indicator is displayed briefly.
6. Observe that an empty dialog is shown instead of the region selection menu.

### Actual behavior

The loading indicator is displayed briefly before the dialog opens. The first time it remains visible for about one or two seconds; subsequent attempts complete almost instantly.

Instead of displaying the region selection menu, an empty dialog containing only the "Close" button is shown.

The same issue has been reproduced on multiple Android devices:
- Xiaomi POCO X6 Pro 5G (HyperOS)
- Google Pixel running GrapheneOS
- Samsung Galaxy S21 FE

I first noticed this behavior while using a node running firmware 2.7.15, and later confirmed that it still occurs with firmware 2.7.26.

As a workaround, the region can still be configured successfully by going to **Settings → LoRa → Region**, where the list of available regions is displayed normally.

### Expected behavior

When tapping the "Set your region" card, the app should display the region selection menu, allowing the user to choose a LoRa region without navigating to Settings → LoRa → Region.

Image
Image
Image
Image

### Relevant log output

```shell

```

### Additional information

This issue has also been observed with multiple node models from different manufacturers. I have not yet verified every affected model, but it does not appear to be limited to the Seeed Studio Wio Tracker L1 Pro.

jamesarich:
Fixes #6586

Tapping the "Set your region" card on Connections showed a brief spinner and then an empty dialog with only a **Close** button, instead of opening the region picker. Settings → LoRa → Region was unaffected, which is the workaround users found.

The card did not navigate. It called `radioConfigViewModel.setResponseStateLoading(ConfigRoute.LORA)` and put up a `PacketResponseStateDialog`, navigating to `SettingsRoute.LoRa` only from that dialog's `onComplete`. Two things made that unreliable:

- `LORA` is a fan-out route (`hasReadFanOut = true`) — it issues `getChannel(0)` *and* `getConfig(LORA_CONFIG)` — but the generic `is ConfigRoute ->` branch leaves `Loading.total` at its default of `1`, unlike `CHANNELS` and `AdminRoute`, which call `setResponseStateTotal(...)`.
- `PacketResponseStateDialog` fires `onComplete()` from inside composition (`if (state.completed >= state.total) onComplete()`), so it only navigates if a composition happens to *observe* the completed `Loading` state.

Once the last request id is removed, `completePacketResponse` calls `clearPacketResponse()`, which resets `responseState` to `Empty`. If both admin responses land before a composition observes the completed `Loading` state, the dialog composes straight to `Empty` — a blank body with a Close button, and `onComplete` can never fire again. That is the reported screenshot.

Note this is why simply adding the missing `setResponseStateTotal(2)` is *not* a fix: `clearPacketResponse()` is unconditional for non-Admin routes, so `Empty` is always the terminal state and the navigation would still be racing it, just at `completed >= 2` instead of `>= 1`.

The dialog was never load-bearing here. `configComposable` already calls `viewModel.loadConfigRoute(routeInfo)` on entry, and `LoRaConfigScreen`/`RadioConfigScreenList` render from the connect-time snapshot while that refresh runs (the overlay is suppressed for local settings, and the dialog only appears for `Success`/`Error`). `settingsRadioConfigSession()` already names "Connections -> LoRa" as a supported direct entry. Worse, the Connections screen's `RadioConfigViewModel` was an unparameterised `koinViewModel()` — a different instance from the keyed one the Settings graph builds — so the fetch it was waiting on did not benefit the destination screen anyway.

Changes:

- `ConnectionsScreen`: the "Set your region" card navigates directly with `onConfigNavigate(SettingsRoute.LoRa)`.
- Removed the now-dead `isWaiting` / `PacketResponseStateDialog` block, the `radioConfigState` collection, and the `radioConfigViewModel` parameter (plus the matching argument in `ConnectionsNavigation`).

On the happy path the user-visible result is identical to what the old `onComplete` did when it won the race — the same destination, minus the dialog.

**Not covered by this PR**

- The second report on the thread (button missing entirely on a fresh 2.8.0 flash) is out of scope. This change touches only the card's `onClick`; its visibility gate — `uiState == ConnectionUiState.CONNECTED_WITH_NODE && regionUnset && sessionAuthorized && isPhysicalDevice` — is untouched, so that report is neither fixed nor made worse here. `sessionAuthorized` being false on a not-yet-authorised node is the likely cause and wants its own fix.
- `ChannelScreen`'s `ModemPresetInfo` tap has the same defect (`setResponseStateLoading(ConfigRoute.LORA)` behind the same dialog). Left alone deliberately: it shares `onComplete`/`getNavRouteFrom` with the channel-edit path, which has a legitimate multi-fetch progress at `total = maxChannels + 1`. Worth a follow-up.

**Testing**

No automated test added: this deletes a code path rather than adding logic, and the only assertion left to make (a click navigates) would need a `runComposeUiTest` over `ConnectionsScreen` with the full Koin graph and ~15 backing flows stubbed. Verified via the repo baseline (`spotlessApply spotlessCheck detekt kmpSmokeCompile assembleDebug test allTests`).

## Summary by CodeRabbit

* **New Features**
  * Region selection now opens LoRa settings directly for a faster configuration flow.
  * Modem preset selection now navigates directly to the LoRa settings screen.

* **Bug Fixes**
  * Removed unnecessary loading and response dialogs during radio configuration navigation.
  * Simplified connections navigation to provide a more consistent experience.

### DIFF
--- a/feature/connections/README.md
+++ b/feature/connections/README.md
@@ -112,7 +112,7 @@ Android and JVM subclasses override `requestBonding(entry)` and `requestPermissi
 // Registration (in androidApp / desktopApp nav graph)
 fun EntryProviderScope<NavKey>.connectionsGraph(backStack: NavBackStack<NavKey>) {
     // Registers ConnectionsRoute.Connections entry
-    // Injects ScannerViewModel + RadioConfigViewModel via Koin
+    // Injects ScannerViewModel via Koin
 }
 ```
 
@@ -129,7 +129,6 @@ feature:connections
   ├── core:domain, core:model, core:navigation
   ├── core:prefs, core:resources, core:service, core:ui
   ├── org.meshtastic:protobufs     (Maven artifact)
-  ├── feature:settings             (RadioConfigViewModel)
   └── usb-serial-android           (Android only)
 ```
 
@@ -151,7 +150,6 @@ graph TB
   :feature:connections -.-> :core:ui
   :feature:connections -.-> :core:ble
   :feature:connections -.-> :core:network
-  :feature:connections -.-> :feature:settings
   :feature:connections -.-> :core:testing
 
 classDef android-application fill:#CAFFBF,stroke:#000,stroke-width:2px,color:#000;
--- a/feature/connections/build.gradle.kts
+++ b/feature/connections/build.gradle.kts
@@ -37,7 +37,6 @@ kotlin {
             implementation(projects.core.ui)
             implementation(projects.core.ble)
             implementation(projects.core.network)
-            implementation(projects.feature.settings)
         }
 
         androidMain.dependencies { implementation(libs.usb.serial.android) }
--- a/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/navigation/ConnectionsNavigation.kt
+++ b/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/navigation/ConnectionsNavigation.kt
@@ -39,7 +39,6 @@ import org.meshtastic.core.ui.component.MeshtasticDialog
 import org.meshtastic.feature.connections.NO_DEVICE_SELECTED
 import org.meshtastic.feature.connections.ScannerViewModel
 import org.meshtastic.feature.connections.ui.ConnectionsScreen
-import org.meshtastic.feature.settings.radio.RadioConfigViewModel
 
 /** Navigation graph for for the top level ConnectionsScreen - [ConnectionsRoute.Connections]. */
 fun EntryProviderScope<NavKey>.connectionsGraph(backStack: NavBackStack<NavKey>) {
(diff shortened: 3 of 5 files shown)


jamesarich:
Follow-up per adversarial review: the ModemPresetInfo tap in ChannelScreen had the identical #6586 bug (progress dialog + `setResponseStateLoading(ConfigRoute.LORA)`), and the fix was already available via the existing `onNavigate` callback — no reason to leave it out of scope. Applied the same direct-navigation fix there.

Also removed the now-dead `feature:settings` dependency from `feature/connections/build.gradle.kts` (no longer referenced anywhere in the module) and regenerated the module's README dependency graph, which was still documenting the removed `RadioConfigViewModel` injection.

Full baseline (spotlessApply/spotlessCheck/detekt/assembleDebug/test/allTests) green. No UI/rendering change (behavior-only `onClick` swap), so no screenshot goldens needed regeneration.

jamesarich:
@coderabbitai review

jamesarich:
@coderabbitai full review

Issue #6685, opened 2026-08-13
[Bug]: Unable to connect to already paired node over BT.


cheaterenator:
### Contact Details

cheaterenator@gmail.com

### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [x] I actually read this list, and should be taken seriously.

### Affected app version

2.8.0

### Affected Android version

Android 12

### Affected phone model

POCO F2 PRO

### Affected node model

SEEED XIAO-S3

### Affected node firmware version

2.8.0

### Steps to reproduce the bug

1. Pair node with phone using app.
2. Turn it off/ disconnect/go out of range.
3. Try to connect to it again.
4. App keeps showing Not connected / with Stop connecting button screen.

### Actual behavior

The problem initially appeared in version 2.7.14 of the app, and exist in version 2.8.0 stable, and also in snapshot 2.8.1. Older app version (2.7.12) connects to node without a problem. In these newer versions, the only way to connect to the node is to remove the Bluetooth pairing, and the first connection right after a new pairing works properly. I also tried other firmware versions (2.7.25, .26 and 2.8.0). Manual Disconnect/Reconnect is not helping.

### Expected behavior

The connection with the previously paired node should be automatically restored when node is turned on / in range.

### Relevant log output

```shell

```

### Additional information

Android device is on latest official ROM (MIUI 13 / Android 12). Battery saving for app is turned off. Bluetooth scanning shows my node in range properly.

b8b8:
how long did you wait before it "didnt auto reconnect"? ive noticed sometimes it takes literally a minute or two.

jamesarich:
Traced this a bit further. This isn't the same bug as #3589 and isn't fixed by the CDM work in #6479/#6477 — those address Android 12+ blocking `startForegroundService()` from the background (a presence-detection gap), whereas here the app is foregrounded and actively retrying, but every reconnect attempt genuinely fails at the BLE layer.

[...]

Issue #3589, opened 2025-10-31
[Bug]: App sometimes refuses to reconnect to node over BLE after node reboot


jp-bennett:
### Contact Details

jp-bennett

### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

### Affected app version

2.7.5

### Affected Android version

Android 16

### Affected phone model

Pixel 9a and others

### Affected node model

Currently testing with a Thinknode-m5

### Affected node firmware version

2.7.13

### Steps to reproduce the bug

1. Connect to node over BLE in Android
2. Trigger reboot on node
3. Look at Android BLE list and see node is connected
4. Look in Meshtastic App, see node is unconnected

### Actual behavior

Android sees node as connected, Meshtastic app sees node as not connected.

### Expected behavior

Meshtastic app should recognize that node is connected via BLE and start want-config flow to re-establish API connection.

### Screenshots/Screen recordings

Will share screen recording privately

### Relevant log output

```shell
INFO  | ??:??:?? 11 [Screen] Done with boot screen
DEBUG | ??:??:?? 11 [Screen] Show standard frames
DEBUG | ??:??:?? 11 [Screen] Show 10 module frames
DEBUG | ??:??:?? 11 [Screen] Added modules.  numframes: 10
DEBUG | ??:??:?? 11 [Screen] Finished build frames. numframes: 10
DEBUG | ??:??:?? 11 [Screen] determineMode(): 
DEBUG | ??:??:?? 11 [Screen] refresh=FULL, reason=FLAGGED_COSMETIC, frameFlags=0x5
DEBUG | ??:??:?? 11 [Screen] Update E-Paper
DEBUG | ??:??:?? 12 [Screen] done
DEBUG | ??:??:?? 12 [Screen] Async full-refresh begins (drop frames)
DEBUG | ??:??:?? 12 [Screen] refresh=SKIPPED, reason=EXCEEDED_RATELIMIT_FAST, frameFlags=0x3
DEBUG | ??:??:?? 12 [GPS] Upgrade time to quality GPS
[...]

PR #6744, opened 2026-08-16
fix(ble): refresh a stale GATT cache during ordinary reconnects

jamesarich:
Addresses #6685 (needs hardware verification — see notes below)

## Problem

After a BLE-paired node goes out of range or powers off and then comes back, the app gets permanently stuck on "Not connected" / "Stop connecting". The only workaround users report is unpairing and re-pairing Bluetooth at the OS level.

The "stuck" UI is not a hang. `BleRadioTransport` runs `BleReconnectPolicy` with `maxFailures = Int.MAX_VALUE`, so it is an endless *silent-failure* retry loop with exponential backoff (5 s → 60 s cap). Every attempt genuinely fails; nothing ever escalates.

A stale Android GATT/service cache after a prolonged disconnection is a well-documented cause of exactly this symptom. Android keeps serving the service table it recorded before the absence, discovery replays that stale table, the Meshtastic profile setup fails, and the loop retries forever. An OS-level unpair is what clears that cache by hand.

## Fix

The repair machinery already existed in this codebase but was reachable only from the post-OTA path: `BleConnection.invalidateServiceCache()` refreshes the platform cache, and the transport then disconnects and reconnects to force rediscovery. This wires the same machinery into the general reconnect-failure path.

`GattCacheInvalidationGate` (new, `core/network`, ~40 lines of logic) decides when a failure streak has earned one refresh:

- **Threshold — 6 consecutive failures.** Deliberately far above `BleReconnectPolicy.DEFAULT_FAILURE_THRESHOLD` (3), which only decides when to tell higher layers a disconnect is more than a blip: three failures are reached about 47 s in, which is an entirely ordinary out-of-range or powered-off gap and no evidence of a stale cache. Six is the first count at which the backoff ladder has been saturated at its 60 s cap for two consecutive cycles — the ladder is exhausted and every further attempt is identical. The refresh therefore lands on the 7th attempt, **3 min 36 s** (7 × 3 s settle + 5+10+20+40+60+60 s backoff) into an unbroken failure streak, matching the prolonged absence #6685 describes ("go out of range/power off") rather than a sub-minute blip.
- **Post-invalidation reconnect delay — `BleReconnectPolicy.DEFAULT_SETTLE_DELAY` (3 s).** This is the same disconnect → reconnect cycle the reconnect loop performs, waiting on the same firmware-side GATT release, so it uses the same empirically validated delay. (The probe data recorded on that constant: 1.5 s fails 3–4 times out of 5; ≥ 5 s is reliable.)
- **At most once per streak.** A refresh that did not help must not repeat on every subsequent attempt — that would just add a disconnect/reconnect round trip to each retry.
- **The allowance is consumed only on a real refresh.** On Android the invalidation is a reflection hop into `BluetoothGatt`; if it misses, it is a free no-op. Burning the allowance on a miss would disable the recovery for the rest of the streak.
- **The end of a streak re-arms it**, so a later streak earns its own refresh.

The post-OTA and stale-cache triggers now share one `refreshGattCacheAndReconnect()` helper, but **only the stale-cache trigger spends the streak's allowance** — a post-OTA refresh is scheduled by the firmware-update flow, not by any failure streak, and charging it to a streak would disable stale-cache recovery for a streak that had not even started (the gate is only re-armed by a stable or intentional disconnect, which cannot happen mid-streak). The post-OTA one-shot flag is read into a `val` before the `||` so short-circuit evaluation can never skip consuming it.

## Notes on the re-arm (the subtle part)

Re-arming is an explicit `onFailureStreakEnded()` call from the disconnect path, mirroring the two outcomes that reset the policy's own counter: a stable session and an intentional disconnect.

It deliberately does **not** infer the reset by watching for `consecutiveFailures == 0`. The gate is consulted only on an attempt that reached a connected link, and the counter reset lands *after* that attempt returns — so a radio that disappears a second time fails its next attempt long before any zero could be observed. That would leave the gate consumed forever and silently reintroduce this bug the second time a device goes away. There is a regression test for exactly this (`a second failure streak after the previous one ended earns its own cache refresh`).

## Limitations / what still needs verifying

- **Hardware verification is still required.** This was developed and tested against fakes only; I have no radio to reproduce #6685 against. Please exercise it with a real node: pair, take the node out of range or power it off for several minutes, bring it back, and confirm recovery without unpairing.
- **Scope limit — this only fires on an attempt that already reached `Connected`.** The platform cache refresh needs a live GATT connection handle, so a connect that never establishes a link at all is untouched. **If the field failure in #6685 is connect-time (e.g. `status 133` before ever reaching Connected), this fix will not fire.** That is the single most useful thing to instrument when reproducing: check whether failing attempts reach `Connected` and then fail downstream, or never connect at all. If it is the latter, the root cause is elsewhere and this change is at best partial.
- **The threshold does not eliminate the false-fire class.** The gate is consulted after `connectAndAwait` succeeds, so a radio returning from *any* prolonged absence (left at home overnight — failure count in the dozens) still pays one teardown + reconnect it did not need. What makes that acceptable is the cost: one 3 s settle round trip with a proven-reliable delay, at most once per streak. The earlier revision of this PR fired at 3 failures and reconnected after 500 ms, which turned ordinary reconnects into failures plus another 40–60 s of backoff — see the follow-up commit and comment below. The threshold is a constructor parameter, so it stays cheap to retune.
- Not to be confused with #3589 / #6479 / #6477 (Companion Device Manager). Those address Android 12+ blocking background service restarts — a different failure mode from this foregrounded, actively-retrying case.

## Testing

Full repo baseline, all green:

```
./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile
```

(`kmpSmokeCompile` added per AGENTS.md since this touches `commonMain` in KMP modules.)

New tests — 9 unit tests on the gate in isolation, plus 5 transport-level tests driving the real reconnect loop in virtual time:

- `a stale service table recovers on the attempt that follows the cache refresh` — the real recovery case: connects succeed but discovery keeps replaying a table without the Meshtastic service, and the fake only makes the service reappear after a *successful* `invalidateServiceCache()`, so the refresh is what lets the next attempt succeed
- `a stale service table never recovers when the platform cannot refresh the cache` — the negative control that makes the above non-vacuous (invalidation reports failure → the table is never repaired and the loop keeps retrying)
- `a post-OTA cache refresh does not consume the stale-cache allowance` — verified to fail if the conditional consume is reverted
- `stale GATT cache is refreshed after the reconnect failure streak reaches the threshold`
- `a reconnect failure streak below the threshold never refreshes the GATT cache` — the discriminator; it fails if the threshold is ever tuned back down into blip territory

The gate's own tests pin the *intent* rather than the literal: the default threshold must be strictly greater than the policy's transient-disconnect threshold, and the elapsed time before it can fire — derived from the production settle delay and backoff ladder — must be at least 3 minutes.

The pre-existing post-OTA test (`post-OTA cache invalidation flag is consumed during connect and triggers reconnect`) still passes; it is the regression surface of extracting the shared helper.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bluetooth reconnection reliability when stale GATT service caches cause repeated connection failures.
  * Automatically refreshes the Bluetooth service cache after repeated failed reconnect attempts.
  * Preserves retry behavior when cache refreshes are unsuccessful.
  * Coordinates cache recovery with post-update refreshes and reconnect timing.

* **Documentation**
  * Clarified when to invalidate and refresh Bluetooth service caches and when to reconnect afterward.

* **Tests**
  * Added comprehensive coverage for cache recovery, retry behavior, failure thresholds, and recovery scenarios.

### DIFF
--- a/core/ble/src/commonMain/kotlin/org/meshtastic/core/ble/BleConnection.kt
+++ b/core/ble/src/commonMain/kotlin/org/meshtastic/core/ble/BleConnection.kt
@@ -91,9 +91,16 @@ interface BleConnection {
     fun requestBalancedConnectionPriority(): Boolean = false
 
     /**
-     * Clears the platform's cached GATT service table for the connected peripheral. Necessary when a device reboots
-     * into a different GATT profile (e.g., ESP32 OTA loader) on the same BLE MAC. Returns `true` if the cache was
-     * invalidated. Default implementation returns `false` for platforms without a service cache.
+     * Clears the platform's cached GATT service table for the connected peripheral, so the next discovery re-reads the
+     * device instead of replaying the cache. Necessary when a device reboots into a different GATT profile (e.g., ESP32
+     * OTA loader) on the same BLE MAC, and as recovery when a bonded device that was out of range for a long time
+     * reconnects against a cache the platform never refreshed.
+     *
+     * Requires a live connection: the implementation needs the platform's connection handle, so callers must invoke
+     * this while connected and then reconnect to pick up the fresh service table.
+     *
+     * Returns `true` if the cache was invalidated. Default implementation returns `false` for platforms without a
+     * service cache.
      */
     fun invalidateServiceCache(): Boolean = false
 }
--- a/core/network/src/commonMain/kotlin/org/meshtastic/core/network/radio/BleRadioTransport.kt
+++ b/core/network/src/commonMain/kotlin/org/meshtastic/core/network/radio/BleRadioTransport.kt
@@ -133,8 +133,14 @@ private val PRIORITY_DOWNGRADE_DELAY = 30.seconds
 /**
  * Settle delay after disconnecting to let the BLE stack release GATT resources before reconnecting post cache
  * invalidation.
+ *
+ * Reuses [BleReconnectPolicy.DEFAULT_SETTLE_DELAY] (3 s) rather than a shorter bespoke value: this is the same
+ * disconnect → reconnect cycle the reconnect loop performs, and the same firmware-side GATT release has to complete
+ * first. The probe data recorded on that constant is explicit that 1.5 s fails 3–4 times out of 5 and that ≥ 5 s is
+ * needed for reliable reconnection, so a 500 ms pause here reliably turned a cache refresh into a failed attempt plus a
+ * fresh round of backoff — the opposite of the recovery it exists to provide.
  */
-private val POST_INVALIDATION_RECONNECT_DELAY = 500.milliseconds
+private val POST_INVALIDATION_RECONNECT_DELAY = BleReconnectPolicy.DEFAULT_SETTLE_DELAY
 
 /**
  * A [RadioTransport] implementation for BLE devices using the common BLE abstractions (which are powered by Kable).
@@ -396,9 +402,16 @@ class BleRadioTransport(
         val consecutiveFailures = reconnectPolicy.consecutiveFailures
         val staleCacheSuspected = gattCacheInvalidationGate.shouldInvalidateOnAttempt(consecutiveFailures)
         if (postOtaRequested || staleCacheSuspected) {
-            val reason =
-                if (postOtaRequested) "post-OTA reboot" else "$consecutiveFailures consecutive reconnect failures"
-            refreshGattCacheAndReconnect(device, reason)
+            val triggers = buildList {
+                if (postOtaRequested) add("post-OTA reboot")
+                if (staleCacheSuspected) add("$consecutiveFailures consecutive reconnect failures")
+            }
+            val reason = triggers.joinToString(" + ")
+            // Only the stale-cache trigger spends the streak's one refresh. A post-OTA refresh is scheduled by the
+            // firmware-update flow, not by any failure streak, so charging it to the streak would disable stale-cache
+            // recovery for a streak that had not even started — and the gate can only be re-armed by a stable or
+            // intentional disconnect, which by definition cannot happen while a streak is running.
+            refreshGattCacheAndReconnect(device, reason, consumeStreakAllowance = staleCacheSuspected)
         }
 
         val gattConnectedAt = nowMillis
@@ -506,16 +519,23 @@ class BleRadioTransport(
      * disconnected and the current link is used as-is.
      *
      * @param reason short diagnostic describing which trigger asked for the refresh
+     * @param consumeStreakAllowance true only when the refresh was triggered by a suspected stale cache, so that a
+     *   post-OTA refresh never spends the failure streak's single allowance
      * @throws RadioNotConnectedException when the post-refresh reconnect does not reach Connected, so the caller
      *   returns a retryable failure to [BleReconnectPolicy]
      */
-    private suspend fun refreshGattCacheAndReconnect(device: BleDevice, reason: String) {
+    private suspend fun refreshGattCacheAndReconnect(
+        device: BleDevice,
+        reason: String,
+        consumeStreakAllowance: Boolean,
+    ) {
         val invalidated = bleConnection.invalidateServiceCache()
         Logger.d { "[${address.anonymize()}] GATT cache invalidation ($reason): $invalidated" }
         if (!invalidated) return
 
-        // Consume the streak's single allowance only now that the platform reported a real refresh.
-        gattCacheInvalidationGate.onCacheInvalidated()
+        // Consume the streak's single allowance only for a stale-cache refresh, and only now that the platform
+        // reported a real refresh.
+        if (consumeStreakAllowance) gattCacheInvalidationGate.onCacheInvalidated()
 
         Logger.i { "[${address.anonymize()}] Reconnecting after GATT cache refresh to force service rediscovery" }
         bleConnection.disconnect()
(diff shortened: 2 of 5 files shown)


jamesarich:
Follow-up commit 46a2df428 addresses two blocking findings from review of the first commit, plus the test gap behind them. Net effect: this PR goes from "a fix that could make ordinary reconnects worse" to one that only intervenes for genuinely prolonged failures.

### 1. The gate fired on essentially every normal reconnect (regression)

The threshold was `BleReconnectPolicy`'s transient-disconnect threshold (3), which only decides when to tell higher layers a disconnect is more than a blip. Three consecutive failures are reached about **47 s** into an entirely ordinary out-of-range / powered-off gap — no evidence of a stale cache at all. The gate then tore down a link that had just connected and reconnected after `POST_INVALIDATION_RECONNECT_DELAY = 500.milliseconds`, a delay the probe data recorded on this repo's own `DEFAULT_SETTLE_DELAY` says fails 3–4 times out of 5 (≥ 5 s needed for reliable reconnection). So an otherwise-successful reconnect was sabotaged into a failure plus another 40–60 s of backoff, repeatedly, for any radio that regularly leaves BLE range.

- `DEFAULT_FAILURE_THRESHOLD` 3 → **6**. Six is the first count at which the backoff ladder has been saturated at its 60 s cap for two consecutive cycles — the ladder is exhausted and every further attempt is identical. The refresh now lands on the seventh attempt, **3 min 36 s** (7 × 3 s settle + 5+10+20+40+60+60 s backoff) into an unbroken failure streak: the minutes-long absence #6685 actually describes ("go out of range/power off"), not a sub-minute blip.
- `POST_INVALIDATION_RECONNECT_DELAY` 500 ms → `BleReconnectPolicy.DEFAULT_SETTLE_DELAY` (3 s). It is the same disconnect → reconnect cycle the reconnect loop performs, waiting on the same firmware-side GATT release, so it should use the same empirically validated delay.

Worth stating plainly rather than papering over: **the higher threshold does not eliminate the false-fire class.** The gate is consulted after `connectAndAwait` succeeds, so a radio returning from any prolonged absence (left at home overnight — failure count in the dozens) still pays one teardown + reconnect. What makes that acceptable is the cost: one 3 s settle round trip with a proven-reliable delay, once per streak, instead of a likely-to-fail 500 ms reconnect that dumped the user into another 40–60 s of backoff. The threshold scopes the intervention; the delay makes even a false fire cheap.

### 2. A post-OTA refresh burned the stale-cache one-shot allowance

`onCacheInvalidated()` was called unconditionally, but the post-OTA refresh is scheduled by the firmware-update flow, not by any failure streak. If the post-OTA connection then entered an unstable streak, stale-cache recovery for *that* streak was already disabled — the gate is only re-armed by a stable or intentional disconnect, which by definition cannot happen mid-streak. The trigger reason is now threaded through to the consumption point, and only the stale-cache trigger consumes the allowance (post-OTA never does).

### 3. The tests demonstrated cache wiping, not recovery

All three original tests built the streak with `FakeBleConnection.failNextN`, which fails at connect time, so they only asserted "after N failures the cache gets wiped" — encoding the mechanism rather than testing recovery. Now:

- **`a stale service table recovers on the attempt that follows the cache refresh`** — connects succeed but service discovery keeps replaying a table without the Meshtastic service (`missingServices`), and the fake only makes the service reappear after a *successful* `invalidateServiceCache()`. The streak reaches the threshold, the cache is refreshed, and the next attempt connects specifically because of it; the loop then settles (asserted by `connectAndAwaitCalls` not moving across a further 5 minutes of virtual time).
- **`a stale service table never recovers when the platform cannot refresh the cache`** — negative control with `invalidateServiceCacheResult = false`: the table is never repaired and the loop keeps retrying. This is what makes the recovery test non-vacuous — it proves the invalidation is the cause, not the retry loop eventually getting lucky.
- **`a post-OTA cache refresh does not consume the stale-cache allowance`** — covers finding 2; I verified it fails if the conditional consume is reverted.
- **`a reconnect failure streak below the threshold never refreshes the GATT cache`** — now sits at five failures, so it fails if the threshold is ever tuned back down into blip territory.
- The gate unit test's "default threshold matches the policy's transient threshold" assertion is inverted into the real invariant (strictly greater), plus a test that derives elapsed time from the production settle delay and backoff ladder and requires ≥ 3 minutes before the gate can fire. The existing post-OTA test still passes unchanged.

The threshold is still reachable in tests (virtual time), so the gate stays fully testable.

Local validation: `spotlessApply spotlessCheck detekt assembleDebug test allTests` — all green. The change is confined to `core/network` (+ the earlier `core/ble` KDoc), touches no Compose UI, so no screenshot goldens are involved.

jamesarich:
@coderabbitai review

jamesarich:
@coderabbitai full review

jamesarich:
Addressed both CodeRabbit findings:

1. **Test doesn't distinguish the fixed 500ms vs 3s delay**: the post-OTA test's single 20s advance comfortably fit either value. Split it into a checkpoint that lands after a regressed 500ms delay would already have fired but before the real 3s delay does, asserting the reconnect hasn't happened yet. Verified this actually discriminates by temporarily reintroducing the old 500ms constant — the new assertion fails as expected (`expected: <1> but was: <2>`), then passes again once reverted.
2. **Fake-store end-state assertion**: the stale-service recovery test asserted against `missingServices`, a store mutated by the test's own callback based on a call count that's already independently verified. Removed that assertion — the existing `invalidateServiceCacheCalls`/`connectAndAwaitCalls` checks already cover the production-observable behavior without testing the fake against itself.

Full baseline green. Didn't attempt the review's broader suggestion to add a combined "successful invalidation + failed post-refresh reconnect" scenario test — the fake's `failNextN` is a simple sequential counter that can't selectively fail one specific later call while letting an earlier one succeed, and forcing that in would need a less clean fault-injection mechanism than what's here now. Flagging as a possible follow-up rather than bolting it on.

jamesarich:
recheck

Issue #6583, opened 2026-08-04
[Bug]: Tak Integration : can't send Markers or chat messages in ATAK through Meshtastic


Guylby:
### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [x] I actually read this list, and should be taken seriously.

### Affected app version

2.8.0 snapshot 29321761 universal debug

### Affected Android version

Android 13 and 9

### Affected phone model

Sony XQ-AD51 and XQ-BT52

### Affected node model

Seeed T1000-E

### Affected node firmware version

2.7.15

### Steps to reproduce the bug

Hi,

I've been unable to send or receive any marker shared or any chat message (through all ATAK means: broadcast, all chat, direct message to contact)
I can send messages from the meshtastic app to the other meshtastic app without problems (shortfast tak private channel, direct message). 
Each node show on the other phone in the contact list as connected and a marker is placed on the map with the Short Name + Full name of the node.

On the config side:
In the meshtastic app: in Settings>Tak Server:
Local tak server is ON both sides
Mesh to CoT converter is ON both sides
On both phones the Meshtastic tak server is connected (green status) and setup through the data package created by the meshtastic app

Both nodes are setup with TAK profiles.

I've run the TAK Mesh Test (Debug) on both sides with 0 errors (all 41 tests show all PLI, markers, contacts and chat messages on the other device, both sides)

The ATAK meshtastic plugin is disabled (if enabled it does not work either) and the app is restarted. 

The full setup was fully working with Meshtastic 2.7.13 + ATAK Meshtastic plugin 1.1.42

### Actual behavior

Nodes show on the map and in the contact list, but when sending markers or chat messages, nothing happens

### Relevant log output

```shell

```

kd2var:
I wanted to add some testing feedback from the Mesh-to-CoT work. The node/contact portion of the Mesh-to-CoT functionality has been working well in my testing with the new TAK server implementation.

[...]

PR #6746, opened 2026-08-16
fix(takserver): route TAK self-test through the real v1/v2 dispatch path

jamesarich:
## Summary

Relates to #6583 — fixes the self-test's blind spot to firmware-version-gated dispatch; does not itself resolve the reported ATAK marker/chat loss.

The debug "TAK Mesh Test" self-test (Settings → TAK Server, debug builds only) called `TakSdkCompressor.compressCoT()` directly and always sent fixtures on `ATAK_PLUGIN_V2`, hardcoding the v2 (zstd/full-CoT) pipeline. It never went through `TAKMeshIntegration.useTakV2()`, the firmware-version gate (`Capabilities(fw).supportsTakV2`, requires firmware >= 2.8.0) that real ATAK traffic is routed through. On a radio running firmware < 2.8.0, real traffic silently falls back to the much more limited legacy v1 `TAKPacket` schema (PLI + chat only, everything else dropped) — but the self-test kept reporting 41/41 regardless of which path a connected radio would actually take. That false-confidence gap is what this PR closes.

### Changes

- **`TAKMeshIntegration`**: added `sendCoTToMeshForTest(cotMessage, forceV2)`, an `internal` test-only seam that calls the same production `sendCoTToMeshV1`/`sendCoTToMeshV2` functions with the protocol forced explicitly instead of read from the connected radio's firmware. This is a deliberate design choice — `useTakV2()` reads live radio state, which the self-test needs to bypass in order to exercise *both* paths deterministically in one run regardless of which firmware happens to be connected. It's the only new call site of `sendCoTToMeshV1`/`V2` outside the real dispatch flow, and it's gated the same way the rest of the card is (`buildConfig.isDebug`).
- **`TakSendOutcome.Dropped`** gained a `schemaLimited: Boolean` field so callers can tell a real MTU oversize drop apart from an intentional, permanent schema-coverage gap (e.g. v1's `TAKPacket` having no field for a marker/shape type at all). Previously both were reported as an opaque "dropped".
- **`TakMeshTestRunner`** now runs every fixture through both the v2 and v1 dispatch paths (`TakProtocol.V2` then `TakProtocol.V1`), calling the real `TAKMeshIntegration` via `sendCoTToMeshForTest` rather than the SDK compressor directly. A dropped v1 result is labeled `expectedDrop` only when `schemaLimited` is true — an oversize drop is a real problem and is surfaced as a failure even on v1, not lumped in with the known "v1 doesn't support markers/shapes" limitation.
- **UI** (`TAKConfigItemList.kt`): the results card now shows a separate section per protocol (v2 / v1) with its own pass / expected-drop / failed counts, so a low v1 pass rate reads as the documented legacy-schema limitation rather than a self-test regression.
- Added regression tests: `TakMeshTestRunnerTest` (5 tests) verifies both protocols run for every fixture, that a v1 pass sends on the legacy port and a v2 pass sends on the v2 port *regardless of the connected radio's reported firmware*, and that v1's expected drops are labeled correctly. `TAKMeshIntegrationTest` gained 2 tests directly exercising the `schemaLimited` discrimination: an unsupported CoT type is schema-limited, but an oversize-yet-representable PLI is not (guards against the oversize-drop-mislabeled-as-expected class of bug).

### Operator-facing note

The self-test now sends each fixture twice (once per protocol) instead of once, roughly doubling total airtime for a debug user who taps "Run" with a radio connected — paced by the existing 5s inter-send delay on successful sends only (dropped fixtures never reach the radio, so they don't incur the delay).

## Test plan

- [x] `spotlessApply spotlessCheck detekt assembleDebug test allTests` — full repo-mandated baseline, green (see below)
- [x] `TakMeshTestRunnerTest`: 5/5 passing on both `jvmTest` and `testAndroidHostTest` targets
- [x] `TAKMeshIntegrationTest`: 14/14 passing (12 existing + 2 new) on both targets
- [ ] Manual verification against a real radio on firmware < 2.8.0 (not done here — no such radio available in this environment)


jamesarich:
Follow-up commit ef6f753: regenerated the 4 stale screenshot goldens (ScreenshotTakTestCardResults/Running, Light + Dark). The TAK test card was intentionally restructured into two sections (V2 and V1 results, with new headers and dividers), but the reference PNGs were never updated, which is why CI's screenshot validation failed. Only those 4 PNGs changed; validateDebugScreenshotTest plus spotless/detekt/assembleDebug/test/allTests all pass locally, so CI should now be green.

jamesarich:
@coderabbitai review

jamesarich:
@coderabbitai full review

jamesarich:
@coderabbitai full review

jamesarich:
@coderabbitai full review

PR #5434, opened 2026-05-12
feat: TAK v2 protocol integration with zstd compression and full CoT type support

thebentern:
## Summary

Upgrades TAK (Team Awareness Kit) integration from the legacy v1 protocol (port 72, PLI + GeoChat only) to the new TAK v2 protocol (port 78, `ATAK_PLUGIN_V2`) with zstd dictionary compression and support for all CoT payload types. Enables rich tactical data—markers, routes, drawn shapes, emergencies, tasks, and ranging—over LoRa mesh between ATAK/iTAK clients.

**Spec**: `specs/005-tak-v2-protocol/` (spec, plan, tasks, research, data model, wire protocol, quickstart)

### TAK v2 Protocol (`core:takserver` KMP module)
- **28 CoT type mappings** (23 CoT types + 4 HOW types + fallback) via `TakV2TypeMapper`
- **Zstd dictionary compression** (`TakV2Compressor`) with separate dictionaries for aircraft vs non-aircraft types
- **CoT detail stripping** removes 16 bloat XML elements before mesh transmission to fit ~225B usable MTU
- **Bidirectional bridge** (`TAKMeshIntegration`) between local TAK server and mesh network
- **Legacy fallback**: Auto-detects firmware version; uses v1 (port 72) for firmware < 2.8.0, v2 (port 78) for >= 2.8.0
- **Offline message queue**: 50-message cap, 5-minute TTL, FIFO eviction, replay on client reconnect
- **Route data packages**: Generates KML for ATAK import when route CoT is received

### TAK Server Lifecycle
- Local TLS/mTLS server on port 8089 for ATAK/iTAK connections (`TAKServerJvm`)
- Partial wake lock in `MeshService` to prevent CPU throttling during TAK operations
- 10-second keepalive interval (below ATAK's 15-second stale threshold)
- Data package export (.zip with server.p12, client.p12, ca.pem)

### TAK Configuration UI (`feature:settings`)
- Compose Multiplatform settings screen (`TAKConfigItemList`) for server enable/disable, team/role selection, data package export, and diagnostic test runner
- `ACCESS_LOCAL_NETWORK` permission handling for Android 17+ (API 37)

### Testing (89+ tests across 12 test classes)
- `TAKMeshIntegrationTest` — 10 tests: lifecycle, inbound mesh forwarding, firmware gating, GeoChat enrichment
- `TAKServerManagerTest` — 9 tests: broadcast routing, offline queue, broadcastRawXml
[...]

PR #6759, opened 2026-08-18
fix(tak): duplicate ATAK contacts on v1 GeoChat + surface v1-fallback notice

jamesarich:
Two fixes for #6583, addressing the two distinct problems reported in that thread.

## Duplicate ATAK contacts on legacy GeoChat (real regression)

`TAKMeshIntegration.convertV1ToCoT()` built a v1 GeoChat's sender UID from the raw `device_callsign` wire string — but that string is actually a smuggled `"<senderUid>|<messageId>"` pair (see `TAKPacketConversion.createChatPacket`). Since `messageId` is random per message, every incoming chat from the same device got a different `senderUid`, so ATAK treated each one as a brand-new contact with an empty history (@Guylby's report).

The correct parsing already existed — `TAKPacketConversion.toCoTMessage()`, using `TakConversionHelpers.parseDeviceCallsign()` — but was dead code, only referenced from a test. A refactor in #5434 swapped its call site for the new (buggy) private `convertV1ToCoT()` and never wired the correct implementation back in. Fixed by pointing `handleV1Packet()` at the existing correct implementation and deleting the duplicate. Verified team/role mapping is unaffected: `TakConversionHelpers.teamToColorName()`/`roleToName()` are thin wrappers around the exact same extension functions `convertV1ToCoT()` called directly.

## Silent v1 fallback (UX gap)

@jamesarich's diagnosis in the thread: real ATAK traffic gates on the connected node's firmware (`Capabilities.supportsTakV2`, firmware ≥ 2.8.0) and silently falls back to a legacy path (PLI + one chat type only — no markers, POIs) on older firmware, with zero user-facing signal, only a log line. Since 2.8.0 is currently nightly-only, this affects most real users today.

Added a warning row in TAK Server settings (`TakV1FallbackNotice`), shown when Mesh-to-CoT is enabled but the connected node doesn't support v2, explaining what still works (location, chat) vs. what needs a firmware update (markers, other event types).

## Testing

- `:core:takserver:allTests` (253 tests), `:feature:settings:allTests` — pass
- `detekt`, `spotlessCheck` (both modules) — pass
- `assembleDebug` — full clean build succeeds

## Summary by CodeRabbit

- **New Features**
  - Added a warning in TAK server settings when connected firmware supports only legacy TAK integration.
  - The notice explains available functionality and the firmware version needed for full integration.
  - Added preview support for both full-integration and legacy-fallback states.

- **Bug Fixes**
  - Improved TAK v1 packet conversion by using the standard conversion process, helping maintain consistent message handling.

### DIFF
--- a/.skills/compose-ui/strings-index.txt
+++ b/.skills/compose-ui/strings-index.txt
@@ -1730,6 +1730,7 @@ tak_server_test_result_unknown_error
 tak_server_test_results_v2
 tak_server_test_run
 tak_server_test_running
+tak_server_v1_fallback_notice
 tak_team
 tak_team_blue
 tak_team_brown
--- a/core/resources/src/commonMain/composeResources/values/strings.xml
+++ b/core/resources/src/commonMain/composeResources/values/strings.xml
@@ -1781,6 +1781,7 @@
     <string name="tak_server_test_results_v2">%1$d passed, %2$d expected drop, %3$d failed of %4$d/%5$d</string>
     <string name="tak_server_test_run">Run</string>
     <string name="tak_server_test_running">Running: %1$s</string>
+    <string name="tak_server_v1_fallback_notice">Connected node firmware doesn't support full TAK integration — only location and chat messages will bridge to ATAK. Markers and other event types need firmware 2.8.0 or newer.</string>
     <string name="tak_team">Team Color</string>
     <string name="tak_team_blue">Blue</string>
     <string name="tak_team_brown">Brown</string>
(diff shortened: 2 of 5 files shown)


jamesarich:
@coderabbitai Addressed the first pre-merge finding:

**"Fallback notice can incorrectly say a node doesn't support TAK v2 when firmware is unavailable"** — real bug, fixed. `Capabilities(null).supportsTakV2` evaluates to `false` (not "unknown") since `forceEnableAll` defaults to a hardcoded `isDebug = false` constant in `commonMain`, so the notice was showing even with no connected node at all. Changed to `myNodeInfo?.firmwareVersion?.let { Capabilities(it).supportsTakV2 } ?: true` — defaults to "assume supported" (no warning) when firmware is genuinely unknown, only reports `false` when we actually have firmware confirming it's sub-2.8.0.

The second finding ("visibility widening" on `TAKPacketConversion.toCoTMessage()`) isn't applicable — `object TAKPacketConversion` has no visibility modifier, so it was already `public` by default; nothing widened, just a new same-module caller for an already-public, previously-unused function.

Left the "add a Compose UI test for the notice" suggestion (nice-to-have, not a correctness issue) for a follow-up rather than this PR.

PR #6834, opened 2026-08-23
fix(map): parse KMZ archives in the F-Droid map overlay renderer

### LINKED ISSUE #6833: [Bug]: Crash after adding KMZ layer on Maps
### Contact Details

thecoolestname36

### Checklist

### Affected app version

Since v2.8.0 (29321705)

### Affected Android version

Android 16

### Affected phone model

Samsung Galaxy A36 5G

### Affected node model

Heltec V4

2.7.15.567b8ea

Open the app, navigate to the maps tab, click the layers button, click "add layer", pick your KMZ file from the file picker, notice app crash after selection.

### Actual behavior

The app crashes when selecting a KMZ file.

### Expected behavior

The app does not crash and the KMZ file imports a new layer as normal.

_No response_

### Relevant log output

```shell

```

### Additional information

_No response_

jamesarich:
Fixes #6833.

`.kml` and `.kmz` both resolve to `LayerType.KML` with no extension-level way to tell them apart, and `FdroidMapOverlayRenderer` fed a KMZ's raw zip bytes straight to osmbonuspack's `parseKMLStream()`, which expects KML XML. The SAX parse just failed and returned `false` with no exception thrown, so `reconcile()`'s `parse(...) ?: continue` silently dropped the layer — no crash, no log, nothing on screen.

PR #6811 (merged 2026-08-21) separately fixed a real crash on the `google` flavor (`NoSuchMethodError` from an xmlutil/maps-utils version mismatch). That fix doesn't touch `fdroid` at all, and a collaborator confirmed on #6833 that the latest `fdroid` snapshot still doesn't crash but the KMZ layer "does not load/display" — this PR is that remaining gap.

### 🐛 Bug Fixes
- The `google` flavor already sniffs the zip magic bytes (`PK`) before choosing `KmlParser` vs `KmzParser`; that sniff was flavor-agnostic, so it's now a shared `InputStream.isKmzArchive()` extension in `MapLayer.kt` (the doc-designated flavor-neutral file) instead of a private duplicate.
- osmbonuspack (the `fdroid` KML library) has no stream-native KMZ parser — only `KmlDocument.parseKMZFile(File)`, which needs random-access zip seeking to resolve embedded images. `FdroidMapOverlayRenderer` now spools a detected KMZ stream to a temp file in `cacheDir` (deleted immediately after parsing) and calls that, mirroring how the `google` flavor treats KMZ uniformly regardless of whether the layer's source is a local file or a network fetch.
- `parse()`'s previously-silent `ok == false` path (malformed KML/KMZ/GeoJSON) now logs via Kermit instead of vanishing.

### Testing Performed
- `./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests` — all green, both `google` and `fdroid` flavors.
- No new test added (no local fixture KMZ available in this pass); manual verification would be importing a `.kmz` on an `fdroid` build and confirming the overlay renders.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map overlay support for KML and KMZ files.
  * KMZ archives are now identified and processed correctly, while standard KML files continue to load normally.
  * Improved handling of larger map files during import.
  * Malformed or unreadable documents now fail gracefully with clearer error reporting.

### DIFF
--- a/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/FdroidMapOverlayRenderer.kt
+++ b/androidApp/src/fdroid/kotlin/org/meshtastic/app/map/FdroidMapOverlayRenderer.kt
@@ -19,6 +19,7 @@ package org.meshtastic.app.map
 import android.graphics.Color
 import androidx.core.graphics.toColorInt
 import co.touchlab.kermit.Logger
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import org.meshtastic.app.map.cluster.RadiusMarkerClusterer
@@ -37,9 +38,16 @@ import org.osmdroid.views.overlay.Polyline
 import java.io.BufferedInputStream
 import java.io.File
 import java.io.InputStream
+import java.io.OutputStream
 import kotlin.math.roundToInt
 
 private const val TAG = "MapOverlayRenderer"
+
+/**
+ * Cap on a spooled KMZ's size: `openStream` can be an unbounded network fetch, and osmbonuspack's KMZ path needs the
+ * whole archive on disk before it can parse it. 50 MB comfortably covers a legitimate overlay plus embedded icons.
+ */
+private const val MAX_KMZ_BYTES = 50L * 1024 * 1024
 
 // simplestyle-spec fallbacks, mirroring the Google flavor's applySimpleStyleSpec(); tune here.
 private const val DEFAULT_GEOJSON_FILL_OPACITY = 0.35f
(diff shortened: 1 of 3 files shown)


jamesarich:
## Fixes Applied Successfully

Fixed 2 file(s) based on 4 CodeRabbit feedback item(s), plus one finding from an adversarial review pass (the oversized-KMZ log level/message).

**Files modified:**
- `androidApp/src/main/kotlin/org/meshtastic/app/map/MapLayer.kt` — `isKmzArchive()` now loops `read()` until the magic bytes are filled or EOF (short reads could misclassify a KMZ as bare KML)
- `androidApp/src/fdroid/kotlin/org/meshtastic/app/map/FdroidMapOverlayRenderer.kt` — `CancellationException` rethrown before the generic catch; spooled KMZ temp copy capped at 50 MB with partial-file cleanup; malformed/oversized input logs at warning, not error

**Commit:** `df81f06a5`

Also verified against osmbonuspack 6.9.0 bytecode that `parseKMZFile` decodes embedded icons eagerly and closes the `ZipFile` before returning, so deleting the temp file immediately after parsing is safe.

Baseline (`spotlessCheck detekt assembleDebug test allTests`) green after the fixes.

PR #6839, opened 2026-08-24
fix(mqtt): tolerate object-typed payload in MQTT JSON messages

jamesarich:
## Why

Crashlytics `31a6a0931012d08caf0c22f8505abad9` is the third-highest-volume issue on 2.8.1: 186 users / 3,763 events (and ~3.5% of Datadog RUM sessions carry the same signature). `MqttJsonPayload` declares `payload` as `String?`, but the firmware's own JSON MQTT output publishes `payload` as a nested JSON object for position, telemetry, nodeinfo, and map-report messages (only text messages use a string). Every JSON-enabled gateway therefore threw `JsonDecodingException` on most uplinks, and the message was dropped instead of proxied.

The decode at `MQTTRepositoryImpl.processMessage` is a pure validation gate: its result is discarded and the raw JSON text is forwarded verbatim, so the old strictness bought nothing and silently broke a documented firmware feature (JSON `sendposition` downlink commands, which carry object payloads, were dropped by the phone in client-proxy mode).

## What

`payload` now uses a private `JsonTransformingSerializer` that passes string primitives through unchanged and coerces objects, arrays, and non-string primitives to their compact JSON text. Null/absent stay null; genuinely malformed JSON still fails as before. No consumer change needed: the field keeps its `String?` type, and the forwarded raw text is byte-identical to today.

Behavior note for reviewers: messages that previously failed decode are now forwarded to the radio. On current firmware these are uplinks the node already receives elsewhere (it logs and drops invalid envelopes); the user-visible win is that JSON downlink commands work again. Opted-in (`json_enabled`) gateways will see somewhat more BLE proxy traffic.

## Testing Performed

- 8 new unit tests in `core/model` commonTest covering the string/object/array/primitive/null/absent/garbage matrix, executed on both `jvmTest` and `testAndroidHostTest` (JUnit XML checked for masked failures).
- Full local baseline: `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` all green in a clean worktree.
- Two independent adversarial reviews (correctness/regression and conventions/constitution) both returned APPROVE; the correctness pass verified the serializer semantics against kotlinx-serialization 1.11.0 sources and traced the forwarding path through firmware `MQTT.cpp` on both current and pre-#10152 firmware.

## Constitution Check

All seven principles evaluated: I KMP purity (commonMain, kotlinx-serialization only); II zero lint (spotless + detekt clean); III CMP UI (no UI change); IV privacy (no logging change, no proto edits); V design standards (N/A, no UI); VI documentation freshness (no user-facing docs affected); VII verify before push (full baseline run locally; CI to be confirmed on this PR after push).

### DIFF
--- a/core/model/src/commonMain/kotlin/org/meshtastic/core/model/MqttJsonPayload.kt
+++ b/core/model/src/commonMain/kotlin/org/meshtastic/core/model/MqttJsonPayload.kt
@@ -18,6 +18,10 @@ package org.meshtastic.core.model
 
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import kotlinx.serialization.builtins.serializer
+import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonPrimitive
+import kotlinx.serialization.json.JsonTransformingSerializer
 
 @Serializable
 data class MqttJsonPayload(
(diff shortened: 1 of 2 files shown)


jamesarich:
@coderabbitai full review

PR #6844, opened 2026-08-24
fix(navigation): never let the active backstack empty under NavDisplay

jamesarich:
## Why

Crashlytics `113670c3bd77018f67e6f4030246cc49` (fatal, 36 users / 42 events in the 2.8.1 window, firing since 2.7.14; Datadog `404a8928`): `IllegalArgumentException: NavDisplay backstack cannot be empty`. Root cause, verified against the nav3 1.1.1 sources: the two docs screens registered `NavigationBackHandler` pops with no `isBackEnabled` gate. Entry-level handlers register after NavDisplay's own handler and win LIFO priority, and the outgoing entry's handler stays enabled for the whole 350ms exit crossfade, so a second back press during the animation pops the tab root and leaves the stack empty; a tab-switch variant drains the now-inactive tab's stack and crashes on the next switch to it. The docs handlers were the only ungated pop handlers in the codebase (the ~56 toolbar `removeLastOrNull` callbacks are all `dropUnlessResumed`-wrapped and lifecycle-capped by nav3's decorator).

## What

Two layers, both needed:

- Root cause: `popUnlessRoot()` added to `NavBackStackExt`; the docs handlers now gate with `isBackEnabled = backStack.size > 1` and pop via `popUnlessRoot()` (the gate can be stale within a frame, hence both), and the docs toolbar back pops are guarded the same way.
- Defense for the latent class: `MultiBackstack.activeBackStack` self-heals an empty stack to its tab root, and `MeshtasticNavDisplay` re-adds the remembered first-composition root if a displayed stack empties. Both heals log a greppable `Backstack self-healed to root` non-fatal (via throwable-less `Logger.e`) so any future regression of this class is visible in Crashlytics instead of silently snapping to the root.

Reviewer note: the heals write to snapshot state during composition deliberately; a `SideEffect` would run after NavDisplay's `require` already threw. Single-tap back behavior is unchanged everywhere (docs screens always sit at depth 2 or more in healthy states). Public contracts of `MeshtasticNavDisplay` and `MultiBackstack` are unchanged; `popUnlessRoot()` is a new additive public extension in core/navigation (relevant to the SDK cutover inventory).

Follow-up (out of scope here): a konsist/lint rule requiring `isBackEnabled` on every `NavigationBackHandler`, which would close this defect class permanently.

## Testing Performed

- 9 new commonTest cases across `MultiBackstackTest` (root retention, rapid double-back, stale-handler drain heal, empty-restored-stack heal, tab-switch heal) and `NavBackStackExtTest` (`popUnlessRoot` matrix incl. rapid double-pop); suites now 13 and 15 tests, all green on fresh JUnit XML (cache replay ruled out).
- Full local baseline: `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` green (2,677 tests), clean tree.
- Two independent adversarial reviews returned APPROVE; the correctness pass verified the crash mechanism against the extracted nav3 1.1.1 / navigationevent 1.1.1 sources (the `require` at NavDisplay.kt:360, LIFO dispatcher precedence, and the library's own stale-enabled race comment).

## Constitution Check

All seven principles evaluated: I KMP purity (all commonMain, no framework bleed); II zero lint (spotless + detekt clean); III CMP UI (`MeshtasticNavDisplay`/`NavigationBackHandler` remain the mandated surface, contracts unchanged); IV privacy (heal logs carry class simple names only); V design standards (no visual change); VI documentation freshness (no doc-page content affected; `skip-docs-check` applied: crash fix to back-press plumbing with no documented user-visible behavior change); VII verify before push (full baseline run locally; CI confirmed on this PR after push).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved back navigation to preserve the root screen and prevent empty navigation stacks.
  - Added recovery when navigation state is unexpectedly empty or restored without a valid root.
  - Improved handling of rapid repeated back presses and screen exit transitions.
  - Updated documentation navigation to safely stop at its root screen.

- **Tests**
  - Added coverage for root preservation, repeated back presses, empty-stack recovery, and saved-state restoration.

### DIFF
--- a/core/navigation/src/commonMain/kotlin/org/meshtastic/core/navigation/MultiBackstack.kt
+++ b/core/navigation/src/commonMain/kotlin/org/meshtastic/core/navigation/MultiBackstack.kt
@@ -28,6 +28,7 @@ import androidx.compose.runtime.setValue
 import androidx.navigation3.runtime.NavBackStack
 import androidx.navigation3.runtime.NavKey
 import androidx.navigation3.runtime.rememberNavBackStack
+import co.touchlab.kermit.Logger
 
 /** Manages independent backstacks for multiple tabs. */
 class MultiBackstack(val startTab: NavKey, private val currentTabState: MutableState<NavKey>) {
@@ -37,7 +38,15 @@ class MultiBackstack(val startTab: NavKey, private val currentTabState: MutableS
         private set
 
     val activeBackStack: NavBackStack<NavKey>
-        get() = backStacks[currentTabRoute] ?: error("Stack for $currentTabRoute not found")
+        get() {
+            val stack = backStacks[currentTabRoute] ?: error("Stack for $currentTabRoute not found")
+            // A drained stack crashes NavDisplay ("backstack cannot be empty"), so restore the tab root.
+            if (stack.isEmpty()) {
+                Logger.e { "Backstack self-healed to root (MultiBackstack): tab=${currentTabRoute::class.simpleName}" }
+                stack.add(currentTabRoute)
+            }
+            return stack
+        }
 
     /** Switches to a new top-level tab route. */
     fun navigateTopLevel(route: NavKey) {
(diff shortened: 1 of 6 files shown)


jamesarich:
@coderabbitai full review

PR #6846, opened 2026-08-24
fix(map): survive the WebView provider update race in Site Planner

jamesarich:
## Why

Crashlytics `7b41bd386280b17278b8f6d0450dd0dc` (fatal, 30 users / 30 events, new in the 2.8.1 line, one event per user, never a loop): `Resources$NotFoundException: failed to redirect ResourcesImpl` thrown inside `WebView.<init>` at `SitePlannerRunner`. This is the well-known AOSP race where the system WebView provider package updates while the app is running; the first WebView constructed afterwards throws. Site Planner (#6136) is our only WebView, so the crash arrived with it.

## What

- WebView construction in the Site Planner's headless runner is wrapped in a recovery seam: a classifier (`isWebViewProviderUpdateException`) recognizes exactly the two shapes this race produces (`Resources.NotFoundException`, and `AndroidRuntimeException` mentioning WebView, e.g. `MissingWebViewPackageException`); both catch arms route through it so tests and production cannot drift. The try wraps only the bare `WebView(context)` constructor, so no partially configured instance can be orphaned.
- On a classified failure the attach retries once on a later main-looper dispatch (250ms); a second failure fails soft: a toast (`site_planner_webview_updating`, new string resource, sorted) plus dialog dismissal, matching the file's existing soft-fail style. Everything else rethrows unchanged.
- The WebView now lives in a container that owns the retry `Handler`: release cancels pending retries and bridge posts, detaches the child, then destroys it. Happy-path setup is byte-identical, moved verbatim into a configure function.

## Testing Performed

- 4 new Robolectric tests over the production classifier (both recoverable shapes recognized; unrelated `AndroidRuntimeException` and `IllegalStateException` propagate). The retry path itself requires a real WebView provider and is not unit-testable; the seam was kept minimal. Test pins `@Config(application = Application::class)` per repo precedent to avoid the real-Application scope leak (#6644).
- Full local baseline: `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` green (1,789 tests), clean tree.
- Two independent adversarial reviews (correctness APPROVE, conventions APPROVE); review hardening applied: constructor-only try scope, detach-before-destroy teardown, classifier-routed catches.

## Constitution Check

All seven principles evaluated: I KMP purity (Android-only file in androidApp; android.* imports appropriate there); II zero lint (clean; catches are type-specific, delay is a named constant); III CMP UI (string surfaced via Compose resource accessor); IV privacy (no data logged); V design standards (soft-fail matches existing pattern in the same flow); VI documentation freshness (`skip-docs-check` applied: crash fix, no documented behavior change); VII verify before push (full baseline run locally; CI confirmed on this PR after push).

### DIFF
--- a/.skills/compose-ui/strings-index.txt
+++ b/.skills/compose-ui/strings-index.txt
@@ -1670,6 +1670,7 @@ site_planner_tx_power_watts
 site_planner_use_current_location
 site_planner_use_map_center
 site_planner_use_node_location
+site_planner_webview_updating
 skip
 slot
 smart_position
--- a/androidApp/src/main/kotlin/org/meshtastic/app/map/SitePlannerRunner.kt
+++ b/androidApp/src/main/kotlin/org/meshtastic/app/map/SitePlannerRunner.kt
@@ -17,13 +17,17 @@
 package org.meshtastic.app.map
 
 import android.annotation.SuppressLint
+import android.content.Context
+import android.content.res.Resources
 import android.os.Handler
 import android.os.Looper
+import android.util.AndroidRuntimeException
 import android.webkit.JavascriptInterface
 import android.webkit.WebResourceError
 import android.webkit.WebResourceRequest
 import android.webkit.WebView
 import android.webkit.WebViewClient
+import android.widget.FrameLayout
 import android.widget.Toast
 import androidx.compose.foundation.layout.Arrangement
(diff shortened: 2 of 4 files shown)


jamesarich:
@coderabbitai full review

PR #6847, opened 2026-08-24
fix(metrics): eliminate the Vico canvas restore underflow crash

jamesarich:
## Why

Crashlytics `7744c73d302c0af1676f31f80802d59f` (fatal, 106 users / 177 events on 2.8.1): `IllegalStateException: Underflow in restore - more restores than saves` thrown inside Vico's own `BaseCartesianLayer.draw`. The upstream routes are closed: we ship Vico 3.3.0, which is the latest release (no post-3.3.0 drawing fixes on master, no matching upstream issue), and version correlation pins the defect to the 3.3.0 line (2.7.9 on 3.2.x was crash-free; firstSeen is 2.8.0's 3.3.0-next.1). Reading the complete 3.3.0 draw path shows every save/restore pair statically balanced, and the field stacks contain no app frames, so the imbalance is a runtime condition in Vico's animation-frame layer machinery that our stock-component configuration cannot avoid. A defensive guard at our single chart host is the remaining sound option.

## What

- `ChartDrawGuard` (feature/node, expect/actual for android/jvm/ios): snapshots the native canvas save count before the chart draws, catches only `IllegalStateException` whose message contains "Underflow in restore" (which also covers the `restoreToCount` variant), rethrows everything else, and rebalances via `restoreToCount(snapshot)` in a `finally`. The rebalance is load-bearing: our charts use fading edges, and Vico pushes a non-finally `saveLayer` on the node canvas for them, so a naive catch would leave the compose frame's save stack corrupted.
- Applied as `Modifier.chartRestoreUnderflowGuard()` at the single `CartesianChartHost` call site (`BaseMetricChart`), through which every metric chart funnels.
- Each swallow records a Crashlytics non-fatal (`Logger.e`) deliberately, so the next release can confirm the fatal population converts to guard firings and reveal when a Vico upgrade silences it; downgrade the severity once field-verified.
- Documented limitation: a swallowed frame can leave Vico's cached offscreen layer canvas corrupted until the canvas resizes or the screen recreates (frozen or partially clipped chart in that rare case), which the node-canvas rebalance cannot reach. Degraded visuals beat the fatal.

Follow-ups (out of scope here): file the field facts upstream with patrykandpatrick/vico, and consider a key()-reset that rebuilds the chart host after a swallow to also refresh the cached layer canvas.

## Testing Performed

- 6 jvmTest cases against a real Skia canvas: leaked-save rebalance, underflow swallow, rethrow of unrelated ISE and non-ISE, balanced no-op, and stack-driven-below-snapshot repair.
- Full local baseline: `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` green, clean tree.
- Two independent adversarial reviews returned APPROVE; the correctness pass verified the exception message is hardcoded English in the Android SDK Canvas source, that `restoreToCount` cannot throw from the finally on either backend, and that no graphics layer sits between the guard and Vico's canvas.

## Constitution Check

All seven principles evaluated: I KMP purity (expect/actual split; commonMain uses compose + kermit only); II zero lint (clean, no suppressions or baseline edits); III CMP UI (multiplatform accessors: nativeCanvas on Android, skiaCanvas on skiko); IV privacy (log carries the exception only); V design standards (no visual change on healthy frames); VI documentation freshness (`skip-docs-check` applied: crash fix, no documented behavior change); VII verify before push (full baseline run locally; CI confirmed on this PR after push).

### DIFF
--- a/feature/node/src/androidMain/kotlin/org/meshtastic/feature/node/metrics/ChartDrawGuard.android.kt
+++ b/feature/node/src/androidMain/kotlin/org/meshtastic/feature/node/metrics/ChartDrawGuard.android.kt
@@ -1,0 +1,26 @@
+/*
+ * Copyright (c) 2026 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+package org.meshtastic.feature.node.metrics
+
+import androidx.compose.ui.graphics.Canvas
+import androidx.compose.ui.graphics.nativeCanvas
+
+internal actual fun Canvas.platformSaveCount(): Int = nativeCanvas.saveCount
+
+internal actual fun Canvas.platformRestoreToCount(count: Int) {
+    nativeCanvas.restoreToCount(count)
+}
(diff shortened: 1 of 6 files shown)


jamesarich:
@coderabbitai full review

PR #6848, opened 2026-08-24
fix(connection): extend BLE handshake deadlines while config progress flows

jamesarich:
## Why

Crashlytics `e61fc83f4f8155d35b6c0271a1871010` is the highest-volume non-fatal on 2.8.1: 2,159 users / 9,430 events of "handshake stall" reports (the title is frozen from 2.7-era wording; the bucket groups every throwable-less `Logger.e` from the stall guard). Sampled 2.8.1 events split into two populations:

- ~50% are BLE Stage 1/2 stalls that show REAL config-stream progress right up to the deadline, firing at exactly the 30s/60s budget on large meshes (one sampled mesh: 471 nodes). The budgets were fixed totals that never extended on progress (`onHandshakeProgress()` was a deliberate no-op on BLE), and each recovery restart re-drains the nodeDB from scratch into the same wall, looping to "recovery exhausted": these users cannot connect at all. That is the bug this PR fixes.
- ~50% are fast-transport total-silence stalls: a genuinely unresponsive device or session, where the restart is the recovery mechanism working. Environmental, not addressed here.

(The sibling "forcing reconnect" variant vanished after 2.8.0 because #6470 downgraded the fast watchdog to Warn, not because the underlying stall stopped.)

## What

- On BLE, `onHandshakeProgress()` now re-arms the current stage's guard with its full budget, converting 30s/60s into inactivity windows. Only the connected device's own session-admitted config-stream frames re-arm (my_info, config, module_config, channel, node_info, metadata, fileInfo); ambient mesh traffic cannot. A silent link fires the guard exactly as before. TCP/USB fast path unchanged.
- A per-stage total cap bounds the new behavior: a cap job on the same virtual clock de-arms the spec at 10x budget, so even pathological trickle-forever firmware hits a hard ceiling (330s Stage 1 / 660s Stage 2) instead of connecting forever. The stage spec is generation-guarded so a progress signal racing the Stage 1 to Stage 2 transition cannot resurrect a stale spec.
- Stall diagnostics enriched with `progressSignals` (count since stage arm) and `sinceConnectMs`: states, counters, and timings only, no PII. Crashlytics grouping is stack-based via the synthesized exception, so the enriched message keeps the existing bucket, letting the next release discriminate silent stalls (0 signals) from any residual false positives.
- Severity deliberately stays at `Logger.e` for one release so the fix's field effect is verifiable; recommend the #6470-style Warn downgrade afterward, keeping "recovery exhausted" as the actionable non-fatal. Also flagged for later: fast-transport Stage 2 arms at 12s of inter-stage silence (12% of sampled events), worth revisiting if it persists post-fix.

## Testing Performed

- 3 new virtual-time tests in `MeshConnectionManagerImplTest`: Stage 1 trickle (20s intervals past the 30s budget; the gaps also prove the 12s fast watchdog never engages on BLE), Stage 2 trickle (45s intervals past 60s), and total-cap (continuous 20s trickle for 400s virtual asserts exactly one restart). Both trickle tests then verify a single restart after full-budget silence. The old test pinning the BLE no-op was replaced, not ignored; the surviving silence-only stall tests remain valid.
- Full local baseline: `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` green (6,700 tests), clean tree.
- Two independent adversarial reviews; the correctness pass exhaustively verified that only session-admitted config-stream frames re-arm the guard, that restart hygiene leaves no stale spec or leaked timers, and required the total cap this PR includes.

## Constitution Check

All seven principles evaluated: I KMP purity (commonMain, coroutines + atomicfu, virtual-clock timers); II zero lint (clean); III CMP UI (no UI change); IV privacy (diagnostics are counters and timings only); V design standards (N/A); VI documentation freshness (no user-facing docs affected); VII verify before push (full baseline run locally; CI confirmed on this PR after push).

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConnectionManagerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshConnectionManagerImpl.kt
@@ -128,6 +128,18 @@ class MeshConnectionManagerImpl(
     private val handshakeCompleteLatch = atomic(false)
     private val locationRejectionLogged = atomic(false)
 
+    /** Armed stage guard: stage, budget, and a generation so a stale re-arm racing a stage change always loses. */
+    private data class StageGuardSpec(val stage: Int, val timeout: Duration, val generation: Long)
+
+    private val armedStageGuard = atomic<StageGuardSpec?>(null)
+    private val stageGuardGeneration = atomic(0L)
+
+    /** Clears [armedStageGuard] once a stage has run for 10x its budget, ending re-arms from trickle progress. */
+    private val stageCapJob = atomic<Job?>(null)
+
+    /** Progress signals seen since the current stage was armed; 0 in a stall report means total silence. */
+    private val stageProgressSignals = atomic(0)
+
     /**
      * Consecutive handshake-recovery failure count for [runSiblingHandshakeRecovery].
      *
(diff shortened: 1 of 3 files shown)


jamesarich:
@coderabbitai full review

Issue #6840, opened 2026-08-24
[Bug]: incorrect display of units of measurement


BrennXavier:
### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [ ] I actually read this list, and should be taken seriously.

### Affected app version

2.8.1 (29321949)

### Affected Android version

13

### Affected phone model

Realme GT

### Affected node model

Any

### Affected node firmware version

2.7.26

### Steps to reproduce the bug

1. Be a user from any country that uses metric units of measurement (meters, degrees Celsius, etc.).
2. Set your phone's language to English (US).
3. Receive imperial units of measurement (Fahrenheit, feet, etc.) in all node information. Any node settings that indicate metric units are ignored; imperial units will be received everywhere.

<img width="1080" height="492" alt="Image" src="https://github.com/user-attachments/assets/235a67ec-177d-4d28-a9cc-cbf9c1d04014" />
<img width="1080" height="784" alt="Image" src="https://github.com/user-attachments/assets/28de3234-bafd-404d-ae80-41bc8cd17c6a" />
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/90c2d7d5-6ddd-4801-84c9-9fffe24cc926" />
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/bff4ccbd-96d8-41fa-9ecb-92cefe9886e8" />

### Actual behavior

Please make the setting to be able to see normal metric units of measurement.

<img width="1080" height="2136" alt="Image" src="https://github.com/user-attachments/assets/cafabcfd-e38b-4814-b3a8-5ec4c17c9484" />

### Relevant log output

```shell

```

gztproject:
This is another concrete example of why I think OS locale is a good default, but perhaps too restrictive as the only allowed source of truth for units.

[...]

PR #6854, opened 2026-08-24
feat(units): follow the OS locale for units and number formatting

jamesarich:
## Why

[#6840](https://github.com/meshtastic/Meshtastic-Android/issues/6840) reports device-side metric settings being ignored: an en-US-locale user sees imperial everywhere regardless of the radio's `DisplayConfig.DisplayUnits`. Two defects sat underneath it.

**Units came from two different places.** Temperature already followed the phone locale (#6775) and maps read locale through a bridge, but the formatters and state fields were still *typed* on the radio's proto enum, and node-detail distance carried it end to end. `DisplayConfig.DisplayUnits` configures the device's own screen; it must never drive app formatting. Since every producer already sourced from the locale, the proto enum survived purely as a carrier type — one new call site away from silently reintroducing device-config sourcing.

**Numbers ignored the locale entirely.** A fixed dot separator and no grouping, so a German user read "1.6 km" and "3.85 V". Android's i18n guidance is that a number shown to a user should read the way that user writes numbers.

**Units also went stale.** A ViewModel outlives the configuration change a locale switch triggers, so units snapshotted at construction kept showing the old system until the screen was rebuilt.

## What

**Locale is the only source of units.** Every display-path formatter, state field and composable parameter moved from `Config.DisplayConfig.DisplayUnits` to `core/common`'s `MeasurementSystem`, and the `DistanceUnit.getFromLocale()` bridge is gone in favour of `getSystemMeasurementSystem()`. The radio settings screen that *configures* the device's own display units is untouched. A konsist rule (`MeasurementSystemSourceTest`) fails the build if the proto enum reappears outside that screen, with a staleness check so the allowlist can't rot silently.

**Numbers follow the locale.** A new `formatDecimalLocalized` expect/actual backs `NumberFormatter.format`, the chokepoint for nearly every user-facing number, so telemetry, metrics screens, site planner and discovery reports all localise without touching their call sites. Rounding is pinned HALF_UP because `java.text` defaults to HALF_EVEN and localising a value must not also re-round it.

**Machine-readable output stays invariant.** `NumberFormatter.formatInvariant` keeps the fixed-dot behaviour for output another program parses. This is not theoretical: the first revision of this PR localised GPX coordinates (`lat="52,5200000"` — a file no importer accepts) and a CSV column. Both, plus the CoT payloads read by other ATAK clients, now use the invariant path, pinned by tests that run under a German locale because the whole suite otherwise runs under a pinned en-US where the two formatters are indistinguishable.

**Units re-read on change.** `LocaleUnitsProvider` owns the read once and exposes it as flows; `LocaleChangeNotifier` wraps `ACTION_LOCALE_CHANGED`, the documented signal for Android 14 regional preferences, shared so one platform listener serves every screen. The node list, the map and `CommonGetNodeDetailsUseCase` all consume it — the last of which was still going stale in the first revision, on the most temperature-dense screen in the app.

**Test JVMs are pinned to en-US** in one place, because locale-aware output would otherwise fail every test pinning a formatted string on a contributor's de-DE machine. Locale behaviour is asserted by tests that set the locale themselves.

### Deliberately not included

ICU `MeasureFormat` (translated unit symbols) and `usage("road")` (CLDR unit choice) were both implemented and then removed:

- **ICU is unreachable from this code.** `commonTest` runs against the Android target's stub `android.jar` where every `android.icu` call throws, and it can't use Robolectric because it also compiles for iOS. Symbols were fixed English before this PR, so nothing regressed — only the translation improvement was given up. Removing it also collapsed `formatMeasure` to a plain common function with no platform seam.
- **`usage("road")` degraded precision.** CLDR road preferences impose road *rounding* — nearest 10 m below 300 m, nearest 50 m above — so a node 87 m away read "90 m". Its guard is never false in practice (every caller derives its system from the same function), so it governed all distance rendering on API 33+, including the GNSS-accuracy and position-precision labels, which aren't road distances. It also returned kilometres on LR/MM locales where the app reports imperial.

## Testing Performed

- Full local baseline: `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` plus `:desktopApp:test`, green.
- Five parallel adversarial reviews (correctness, KMP idioms, tests, architecture, and a fact-check of every API-level claim against Android/ICU/CLDR docs). They found the GPX corruption, the stale node-detail units, a negative-tie rounding disagreement between the two formatters (`-6.25 dB` rendered `-6.3` on screen and `-6.2` in CoT — firmware reports SNR in quarter-dB steps, so those ties are routine), a missing NaN guard, and two false claims in my own KDoc. All fixed; the fact-check confirmed the API levels, `ACTION_LOCALE_CHANGED` semantics and `RECEIVER_NOT_EXPORTED` delivery.
- New tests: locale separators and grouping; both formatters agreeing on a binary-exact negative tie (the first version of this assertion used `1.005`, which isn't a representable tie and passed regardless of rounding mode); GPX and CoT invariance under a German locale; `LocaleUnitsProvider` priming and de-duplication; real metric-to-imperial flips in the node-list and map ViewModels, replacing an assertion that only counted subscribers.

Refs #6840

https://claude.ai/code/session_012mU7P5KvLFZFNoHL4ExrBN

## Summary by CodeRabbit

* **New Features**
  * Measurement and temperature units now follow system locale preferences and update automatically when changed.
  * Added localized formatting for distances, speeds, weights, and other measurements.
  * Added support for importing KMZ map overlays with a 50 MB size limit.

* **Bug Fixes**
  * Exported GPX, CSV, and telemetry values now consistently use locale-independent decimal formatting.
  * Improved handling of malformed, oversized, or interrupted KMZ imports.


jamesarich:
@coderabbitai full review

PR #6871, opened 2026-08-25
Use Scan-Only Probes After Prolonged Bonded BLE Reconnect Failures

jeremiah-k:
## Overview

When a bonded radio stays away — out of range or powered off — the reconnect loop first scans for a fresh advertisement, then falls back to the bonded handle and gives Android up to fifteen seconds to establish the GATT connection. Once the retry ladder reaches its sixty-second cap, that same expensive sequence can repeat indefinitely while the radio is simply absent.

This changes prolonged bonded-device reconnects to use scan-only probes after six consecutive failures. Most long-streak attempts avoid opening another GATT connection entirely: a missed probe becomes the same retryable failure the reconnect policy already handles, while a fresh advertisement flows directly into the normal connection path.

Scan visibility is not equivalent to connectability on Android, though. On Android 8–11, system Location state can suppress BLE scan results; on Android 12+, `BLUETOOTH_SCAN` can become unavailable independently of the bonded connection path. A permanently armed scan-only gate could therefore strand a radio that bonded `autoConnect` would still recover.

To preserve that recovery path, every fifth long-streak attempt deliberately falls through to normal bonded discovery and `autoConnect`. A stable connection resets the existing reconnect-policy failure counter and disarms probing automatically for the next absence.

### Scan-only reconnect probing

* Add `ScanOnlyProbeGate`, activating after six consecutive reconnect failures.
* Use scan-only probes for most attempts once the reconnect ladder has saturated.
* Restrict probe-only behavior to bonded devices; non-bonded discovery remains unchanged.
* Use a fresh advertisement directly when a probe succeeds.
* Fall through to the normal bonded discovery/`autoConnect` path on every fifth long-streak attempt so unavailable scanning cannot permanently block reconnection.
* Derive all probe behavior from the existing `BleReconnectPolicy` failure counter rather than maintaining a second mutable streak.
* Automatically disarm probing when a stable connection resets that counter.

### BLE discovery

* Extract bonded lookup and bounded scanning into internal `BleDeviceLocator`.
* Keep the extracted discovery methods internal rather than widening the API during the refactor.
* Preserve the existing bonded flow: one bounded address-filtered scan followed by the bonded-handle fallback when the scan misses.
* Preserve the existing non-bonded flow of three bounded scan attempts with retry delay.
* Keep coroutine cancellation cooperative while treating ordinary scanner failures as missed scans.
* Keep affected diagnostics and connection exceptions free of raw BLE addresses.

### Diagnostics

* Log the initial transition into probe mode at info severity and continuing probes at debug.
* Log periodic bonded self-heal attempts separately at debug.
* Leave the existing prolonged-failure GATT cache recovery path intact once a connection can actually be established.

## Testing

* Added exact boundary coverage proving all attempts through the threshold use the bonded fallback and the first armed probe miss does not increment `connectAndAwait`.
* Added coverage proving prolonged scan misses periodically return to the bonded `autoConnect` path.
* Added coverage using distinct bonded and freshly advertised device instances, proving recovery connects with the fresh advertisement rather than the stale bonded handle.
* Added coverage proving a stable reconnect-policy outcome resets the failure streak and disarms probing.
* Added coverage proving non-bonded devices retain the full three-scan discovery sequence even after the bonded probe threshold would have armed.
* Added gate coverage for threshold boundaries, periodic bonded fallback, stable reset, invalid thresholds, and the prolonged-failure timing implied by the production retry constants.
* CI passes across the affected targets.

## Scope

* The change applies only to prolonged reconnect failures for bonded BLE radios.
* Normal initial connections and short out-of-range interruptions retain the existing bonded fallback behavior.
* Non-bonded BLE discovery, reconnect backoff, USB, serial, and TCP transports are unchanged.
* No protocol, persistence, or firmware changes are introduced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Bluetooth reconnection reliability for bonded devices.
  * Added scan-only probing after repeated connection failures, avoiding unnecessary connection attempts while the device is not advertising.
  * Automatically resumes normal connections when the device becomes discoverable again.

* **Tests**
  * Added coverage for reconnection thresholds, scan-only probing, retry behavior, and invalid configuration values.

### DIFF
--- a/core/network/src/commonMain/kotlin/org/meshtastic/core/network/radio/BleDeviceLocator.kt
+++ b/core/network/src/commonMain/kotlin/org/meshtastic/core/network/radio/BleDeviceLocator.kt
@@ -1,0 +1,131 @@
+/*
+ * Copyright (c) 2026 Meshtastic LLC
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ */
+@file:Suppress("TooGenericExceptionCaught")
+
+package org.meshtastic.core.network.radio
+
+import co.touchlab.kermit.Logger
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.withTimeoutOrNull
+import org.meshtastic.core.ble.BleDevice
+import org.meshtastic.core.ble.BleScanner
+import org.meshtastic.core.ble.BluetoothRepository
+import org.meshtastic.core.ble.MeshtasticBleConstants.SERVICE_UUID
+import org.meshtastic.core.model.RadioNotConnectedException
+import org.meshtastic.core.model.util.anonymize
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+
+private const val SCAN_RETRY_COUNT = 3
+private val SCAN_RETRY_DELAY = 1.seconds
+
+/**
+ * Bounded scan duration used by both discovery paths in [BleDeviceLocator.findDevice]:
+ * - Bonded devices get one address-filtered scan before falling back to the bonded handle.
+ * - Non-bonded retries each use this duration.
+ *
+ * Keeping the bonded path to one scanner registration is important on Android, which throttles applications that start
+ * BLE scans too frequently. A single 5s window still covers multiple advertising intervals for typical power-save slots
+ * (~1–2s each), resolves immediately when the target advertises, and avoids consuming two scan starts per reconnect. If
+ * the scan misses, [BleDeviceLocator.findDevice] falls back to the bonded handle and `attemptConnection` keeps that
+ * patient `autoConnect` path bounded through `CONNECTION_TIMEOUT`.
+ */
+internal val SCAN_TIMEOUT = 5.seconds
+
+/**
+ * Locates the BLE device for one transport address: bonded-handle lookup with a single fresh-advertisement scan, or
+ * retried bounded scans for non-bonded addresses.
+ *
+ * Extracted from [BleRadioTransport] (unchanged behavior) so the transport class stays within detekt's LargeClass
+ * budget while sibling reconnect fixes continue to grow it.
+ */
+internal class BleDeviceLocator(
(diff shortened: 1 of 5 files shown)


jeremiah-k:
~~CI note: the current `shard-feature` failure is unrelated to this PR. It is the scheduler-dependent `NodeListViewModelTest` locale-units race reproduced independently across these branches and fixed separately in #6870.~~ (Now resolved)

jamesarich (commented):
The threshold reasoning is solid and I like that the tests derive their budgets from the production constants instead of hardcoded ms. The `BleDeviceLocator` extraction reads clean too.

One thing I want sorted before this goes in though.

`scanForFreshDevice` swallows every non-cancellation exception to null, so once the gate arms at 6 the probe can't tell "radio isn't advertising" from "we can't scan at all". Null becomes a `RadioNotConnectedException`, that bumps `consecutiveFailures`, and the gate stays armed - there's no way back to the bonded connect except a successful scan.

Two cases where that bites and the bonded autoConnect would still have worked:

- our min is 26, and on 26-30 scan results are gated on system Location being on. Location off and the scan is always empty.
- on 31+ `BLUETOOTH_SCAN` is a separate runtime perm from `BLUETOOTH_CONNECT`. Revoke just scan in settings and `scanner.scan` throws SecurityException, which we catch and drop.

Both of those reconnect fine today via the bonded fallback. And the only log that would tell them apart is `Logger.v`, so the report we get is going to be "it stopped reconnecting" with nothing useful in the logs.

Worth comparing against `GattCacheInvalidationGate` - that one goes out of its way to never end up permanently consumed, and says so in the comment on `onFailureStreakEnded`. This gate can end up permanently armed and there's no equivalent escape.

Either let every 5th probe or so fall through to the full `findDevice()` as a self heal, or have `scanForFreshDevice` return a scanner error distinctly from a scan miss and skip probing (at warn) when it's an error. Either works for me.

Nits while you're in there:

`ScanOnlyProbeGate` has no mutable state, so the "not thread-safe by design" line copied over from `GattCacheInvalidationGate` isn't true here.

CodeRabbit's inline about the raw address in the exception is stale - you already anonymized it, that one can be resolved.

Its premerge warning on the early-failure test is fair though. `connectAndAwaitCalls >= 5` can't fail on the regression it names - that's the class C thing in our review checklist. Pin the exact count, then assert the next missed probe doesn't move it.

jeremiah-k:
@jamesarich Good catch on the scan-only trap. I went with the periodic bonded fallback rather than trying to distinguish scanner errors from misses, since the Android 8–11 Location-off case can look like an ordinary empty scan.

After the six-failure threshold, four attempts stay scan-only and every fifth long-streak attempt falls through to the normal bonded discovery/`autoConnect` path. That gives us the cheap absence probes without making scan availability a permanent requirement for an already-bonded device.

I also tightened the tests around the actual regression boundary: exact connect count through the threshold, the first missed probe leaving that count unchanged, periodic bonded self-heal, a distinct fresh-advertisement instance being used on recovery, stable-policy reset, and the non-bonded three-scan guard. `BleDeviceLocator`'s moved methods are explicitly internal now, and I removed the stale thread-safety note from the stateless gate.

The raw-address CodeRabbit finding was already fixed in the current branch as you noted, so there was nothing further to change there.

PR #6872, opened 2026-08-25
Demote Repeated Offline Heartbeat Rejection Logs to Debug

jeremiah-k:
## Overview

While a radio is unavailable, each periodic heartbeat handoff is rejected by the transport. That expected offline state could produce two warning lines for the same heartbeat: the transport's generic admission warning followed by the heartbeat sender's rejection warning. With heartbeats continuing during an extended disconnect, those duplicate warnings create substantial log noise without adding new diagnostic information.

This makes `HeartbeatSender` the owner of generic heartbeat rejection severity. The first two consecutive rejected handoffs remain warnings, while later consecutive rejections are demoted to debug. Any accepted heartbeat resets the streak, so a new failure after connectivity returns is visible at warn severity again.

BLE and Android serial transports route heartbeat sends through a quiet unavailable-admission path. That suppresses only the duplicate generic unavailable warning for heartbeat traffic. Ordinary application sends continue to warn, and cause-specific BLE write-backlog pressure remains independently visible at warning severity on every rejected admission.

### Heartbeat rejection severity

* Track consecutive generic heartbeat rejection logs in `HeartbeatRejectionLogPolicy`.
* Keep the first two consecutive rejections at warn severity.
* Demote subsequent consecutive rejections to debug until a handoff succeeds.
* Reset the streak immediately after any accepted heartbeat.
* Keep rejection logging outside `nonceMutex` so logger work cannot extend the heartbeat critical section.
* Keep the policy independent from Kermit's process-wide logger configuration.

### Transport routing

* Route BLE heartbeat sends through normal admission without emitting the duplicate generic transport-unavailable warning.
* Apply the same behavior to Android serial heartbeats.
* Preserve warning behavior for ordinary application sends.
* Preserve the dedicated BLE backlog warning on every full-queue rejection, independently of the generic heartbeat rejection streak.
* Keep `BLE_MAX_PENDING_WRITES` private rather than widening a production implementation detail for tests.

### Testability and diagnostics

* Use per-instance internal logging seams to verify routing without modifying Kermit's process-wide writers.
* Keep the testing seam off the public API.
* Anonymize the touched BLE and serial transport identifiers in the affected diagnostics.
* Test BLE queue pressure behaviorally rather than exposing its private capacity constant.

## Testing

* Added sender-level coverage proving the exact warn/warn/debug rejection sequence.
* Added coverage proving a successful heartbeat resets the streak so the next rejection warns again.
* Added BLE coverage proving offline heartbeat rejection suppresses the duplicate generic admission warning while ordinary sends retain it.
* Added Android serial coverage for the same heartbeat-versus-ordinary-send distinction.
* Added repeated BLE backlog coverage proving every rejected admission from a full queue retains its dedicated transport warning, even though the generic heartbeat rejection line may be demoted.
* CI passes across the affected targets.

## Scope

* Heartbeat cadence, nonce handling, encoding, and transport admission decisions are unchanged.
* Accepted-heartbeat behavior and BLE post-heartbeat drain scheduling are unchanged.
* Cause-specific transport failures remain visible at warning severity.
* No protocol, persistence, or firmware changes are introduced.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced duplicate offline connection warnings during heartbeat activity.
  * Heartbeat rejection messages now become less prominent after repeated failures and reset after a successful send.
  * Protected serial connection details in diagnostic logs.
  * Preserved clear warnings for ordinary send failures and full BLE write backlogs.

* **Tests**
  * Added coverage for heartbeat logging, offline sends, unavailable BLE characteristics, and write-backlog limits.

### DIFF
--- a/core/network/src/androidHostTest/kotlin/org/meshtastic/core/network/radio/SerialRadioTransportTest.kt
+++ b/core/network/src/androidHostTest/kotlin/org/meshtastic/core/network/radio/SerialRadioTransportTest.kt
@@ -111,6 +111,26 @@ class SerialRadioTransportTest {
     }
 
     @Test
+    fun `offline heartbeat owns rejection logging while an ordinary send keeps the serial warning`() = runTest {
+        val address = "serial-device"
+        val transportRejections = mutableListOf<String>()
+        val transport = createTransport(address, backgroundScope)
+        transport.sendRejectionLogger = { transportRejections += it }
+
+        transport.keepAlive()
+        testScheduler.runCurrent()
+        assertTrue(
+            transportRejections.isEmpty(),
+            "HeartbeatSender must be the sole owner of expected offline rejection severity",
+        )
+
+        assertFalse(transport.handleSendToRadio(byteArrayOf(1)))
+        assertEquals(1, transportRejections.size)
+        assertTrue("Serial connection not available" in transportRejections.single())
+        transport.close()
+    }
+
+    @Test
     fun `send is rejected promptly while connection is still opening`() = runBlocking<Unit> {
         val address = "serial-device"
         val connectEntered = CountDownLatch(1)
--- a/core/network/src/androidMain/kotlin/org/meshtastic/core/network/radio/SerialRadioTransport.kt
+++ b/core/network/src/androidMain/kotlin/org/meshtastic/core/network/radio/SerialRadioTransport.kt
@@ -32,6 +32,7 @@ import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeoutOrNull
 import org.meshtastic.core.common.util.handledLaunch
 import org.meshtastic.core.common.util.nowMillis
+import org.meshtastic.core.model.util.anonymize
 import org.meshtastic.core.network.repository.SerialConnection
 import org.meshtastic.core.network.repository.SerialConnectionListener
 import org.meshtastic.core.network.transport.HeartbeatSender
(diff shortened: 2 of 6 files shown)


jeremiah-k:
~~CI note: the current `shard-feature` failure is unrelated to this PR. It is the scheduler-dependent `NodeListViewModelTest` locale-units race reproduced independently across these branches and fixed separately in #6870.~~ (Now resolved)

jamesarich (commented):
Good trim. Heartbeat is every 30s and keeps ticking while we're disconnected, so an out of range radio was churning out a couple hundred warn lines an hour across the two sites. Two per streak with a reset on the first accepted handoff is the right shape, and keeping the logging outside `nonceMutex` is the right call.

Couple of things:

CodeRabbit's right about `BLE_MAX_PENDING_WRITES` - I'd rather not widen it to `internal` just so a test can read it. `repeat(4)` in the test is fine, or drive the backlog behaviorally.

`HeartbeatRejectionLogPolicy` lumps backlog-full rejections into the same streak as the offline ones. So a radio that reconnects and then runs into write backlog gets demoted to debug on the third rejection, even though backlog pressure is the more interesting of the two. Not a blocker, but the streak might want to be per-cause.

Three constructors to inject the sink feels heavy, though it does keep the seam off the public API - I'm fine leaving it.

Rest looks good to me. Also your CI note is stale, everything's green now. And I checked this against #6871 - they merge clean against each other, no rebase needed either way.

jeremiah-k:
@jamesarich Fixed the visibility widening — `BLE_MAX_PENDING_WRITES` is private again, and the test fills the four-slot queue behaviorally instead of reaching into the production constant.

On the per-cause point, I kept the `HeartbeatRejectionLogPolicy` focused on the generic heartbeat rejection line rather than expanding the handoff result/API. BLE backlog pressure already has its own cause-specific transport warning, which fires independently on every full-queue rejection even after the generic heartbeat line has been demoted. I strengthened that test to make three consecutive backlog rejections and assert that all three retain the dedicated diagnostic.

I left the internal sink constructors as-is per your note. The stale CI note is marked resolved; the updated summary reflects the green run.

PR #6894, opened 2026-08-26
fix(ci): retry snapshot asset uploads instead of losing the release

jamesarich:
## What broke

[Run 33015261548](https://github.com/meshtastic/Meshtastic-Android/actions/runs/33015261548) failed `publish-snapshot` with:

```
HTTP 400: 400 Bad Request (https://uploads.github.com/repos/meshtastic/Meshtastic-Android/releases/377444082/assets?label=&name=Meshtastic_Desktop-snapshot-aarch64-29322037.AppImage)
```

Every other job in the run was green. The `snapshot` release is currently **gone entirely** — not stale, absent.

## Why one 400 removes the release

`publish-snapshot` deletes the previous snapshot release *and its tag*, then recreates both by handing all 15 assets (~1.5 GB) to a single `gh release create`. A 400 on any one of them aborts the whole command; `gh` then deletes the release it was staging. The old release and tag are already gone by that point, so main is left with no snapshot until the next push. Confirmed — release `377444082` now 404s and no orphaned draft remains.

The asset itself was fine: a 119 MB AppImage, normal size. The run 7 hours earlier was green with byte-comparable artifacts (351,280,800 vs 351,331,258 for `desktop-app-Linux-ARM64`), and nothing in #6890 touches the publish path. This is a transient `uploads.github.com` failure that the job had no way to absorb.

## The change

- Create the release with **no assets**, then upload each file separately with up to 3 attempts and a backing-off sleep. A flake now costs one retry of one file rather than the entire release. `--clobber` keeps a retry idempotent.
- Raise `timeout-minutes` from 10 to 25. The failing run was already 3m25s into uploading when it died, on top of ~2m of checkout and artifact download — there was no room for even one retry.

**Trade-off, stated up front:** `create` no longer publishes only once every asset has landed, so the release is briefly visible while assets are still arriving. For a rolling prerelease that is replaced on every push to main, a briefly-incomplete release beats an absent one. Happy to switch to staging a draft and publishing it last if reviewers would rather keep strict atomicity.

## Verification

**This job cannot be exercised by any PR check** — `publish-snapshot` runs only on push to `main`, so nothing here executes the changed job before merge. What does stand behind it:

- `actionlint` clean, which shellchecks the `run:` blocks.
- The upload loop run locally against a fake `gh`, covering both paths:
  - an asset that 400s twice then succeeds → recovers on attempt 3, loop exits 0;
  - an asset that always 400s → warns 3×, emits `::error::`, exits 1 (the job still goes red rather than silently publishing an incomplete release).
- Filenames containing spaces (`Meshtastic Desktop-2.8.2-*.dmg`) and `continue 2` semantics under `bash -e` both verified in those runs.

The real test is the first push to main after merge — that push recreates the snapshot, and the release should come back with all 15 assets.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **Chores**
  * Improved snapshot release publishing reliability.
  * Extended the publishing timeout to support longer-running releases.
  * Added individual asset uploads with automatic retries and clearer failure handling.

### DIFF
--- a/.github/workflows/main-check.yml
+++ b/.github/workflows/main-check.yml
@@ -53,7 +53,10 @@ jobs:
     # APK build itself failed, the artifact download below fails and this job goes red.
     if: github.repository == 'meshtastic/Meshtastic-Android' && !cancelled()
     runs-on: ubuntu-24.04-arm
-    timeout-minutes: 10
+    # ~1.5 GB of assets, uploaded with up to 3 attempts each (see "Upload snapshot assets").
+    # The 2026-08-26 failure was already 3m25s into uploading when it died, on top of ~2m of
+    # checkout and artifact download -- 10 minutes left no room for a single retry.
+    timeout-minutes: 25
     permissions:
       contents: write
     env:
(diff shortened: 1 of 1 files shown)


jamesarich:
## Fixes Applied Successfully

Fixed 1 file based on 1 CodeRabbit feedback item.

**Files modified:**
- `.github/workflows/main-check.yml`

**Commit:** `6185ed7eb`

### What changed and why

The finding was valid, and checking it turned up something the review didn't mention: uploading per file is **sequential**, where the `gh release create upload/*` it replaced uploaded **concurrently**. So the step was already slower than the code it replaced, and a single stalled upload had nothing bounding it — on its own it could have spent the whole job budget and had the job cancelled mid-release, which is the exact half-published outcome this step exists to prevent.

- `timeout 8m` per upload attempt — far beyond what a ~120 MB asset needs, so it fires only on a genuine stall. This is what makes a larger job cap a backstop rather than a blank cheque.
- Backoff 15s → 10s per attempt, cutting worst-case sleep from 11m15s to 7m30s.
- Job cap 25 → 35 minutes, so even "every asset flakes twice" completes.

One correction to the finding's framing, for the record: all 15 assets cannot exhaust their attempts, because the first exhaustion `exit 1`s the step. The 45-upload / 11m15s figure is the worst case that still *succeeds* — every asset failing twice, then landing. The arithmetic was right; the scenario is narrower than stated.

### Verification

`publish-snapshot` runs only on push to `main`, so no PR check exercises the changed job. Standing behind it:

- `actionlint` clean (it shellchecks `run:` blocks).
- The loop run locally against a fake `gh` covering three paths:
  - asset 400s twice then succeeds → recovers on attempt 3, exits 0;
  - asset **stalls** → killed by `timeout` and retried, succeeding on attempt 2 (6.2s elapsed instead of the 30s stall, confirming the bound fires);
  - asset always 400s → warns 3×, emits `::error::`, exits 1 so the job still goes red rather than publishing an incomplete release.

The Gradle baseline was not run: this commit touches only workflow YAML, no Kotlin or Gradle files.

The latest autofix changes are on the `fix/ci-snapshot-appimage-upload` branch.

PR #6919, opened 2026-08-27
ci(flatpak): generate sources inside each arch's offline build

jamesarich:
`verify-flatpak.yml` spent more time queueing than building. On run [33092180649](https://github.com/meshtastic/Meshtastic-Android/actions/runs/33092180649) — 40 min wall:

| phase | x86_64 |
| --- | --- |
| queue → `generate-sources` | 4m45s |
| gradle `captureFlatpakSources` | 9m01s |
| **queue → `build-flatpak`** | **13m49s** |
| flathub runtime install | 50s |
| flatpak-builder source download | 2m19s |
| offline gradle build in the sandbox | 5m23s |
| `build-bundle` export | 2m07s |

The aarch64 leg paid 8m30s for the same second queue wait. That is exactly the fan-in serialization `reusable-check.yml` documents avoiding at the top of its `jobs:` block.

So: one job per arch, generation included. Generation runs twice now, but in parallel, and the run drops to a single queue wait plus ~20 min of work — ~40 min → ~23 min, +9 runner-minutes.

Two consequences of the merge, both in the diff:

- Each leg captures `flatpak-sources.json` on the arch that consumes it. Today x86_64-resolved URLs feed the aarch64 build.
- The rsync excludes had to become unanchored. `/build/` and `*/build/` only match at the transfer root and one level down — harmless when the build job checked out a clean tree, but it now runs after Gradle, so `core/*/build/` etc. would land in the flatpak source dir.

Also skips the bundle export + upload on `pull_request`: the offline build succeeding is the verification, and `build-bundle` is ~2 min per arch for an artifact nobody installs from a PR. `workflow_dispatch` still produces it.

Deliberately **not** done here: caching the isolated Gradle home (capture only sees `ExternalResourceReadBuildOperation` on cache misses — a warm cache empties the manifest), caching `~/.local/share/flatpak` (worth 50s), and caching `.flatpak-builder/downloads` (GBs against the 10 GB Actions quota, and re-resolving the mirror is the thing under test).

This workflow stays outside the required-check set — `Check Workflow Status` does not read it. Folding it into `pull-request.yml` would make it blocking; that is a separate decision.

Verification: `actionlint` 1.7.12 clean; a `workflow_dispatch` run on this branch to follow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Flatpak verification reliability across architectures.
  * Added fallback handling when the preferred Java runtime setup fails.
  * Improved generated-source reporting and build-directory synchronization.
  * Prevented unnecessary bundle exports and artifact uploads for pull requests.

### DIFF
--- a/.github/workflows/verify-flatpak.yml
+++ b/.github/workflows/verify-flatpak.yml
@@ -29,9 +29,17 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  generate-sources:
-    runs-on: ubuntu-24.04
-    timeout-minutes: 20
+  # One job per arch: splitting generation from the build cost a second runner-queue wait
+  # (8-14 min measured, against ~9 min of generation) — the fan-in serialization
+  # reusable-check.yml documents avoiding. Each leg now captures sources on its own arch.
+  verify-flatpak:
+    name: Verify Flatpak Offline Build (${{ matrix.arch }})
+    runs-on: ${{ matrix.arch == 'aarch64' && 'ubuntu-24.04-arm' || 'ubuntu-24.04' }}
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        arch: [x86_64, aarch64]
+      fail-fast: false
     steps:
       - uses: actions/checkout@v7.0.1
         with:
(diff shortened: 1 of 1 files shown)


jamesarich:
Both legs green on [run 33097130393](https://github.com/meshtastic/Meshtastic-Android/actions/runs/33097130393) - export/upload correctly skipped on `pull_request`, JBR resolved on `ubuntu-24.04-arm` without hitting the Foojay fallback, and the unanchored rsync excludes held.

Work per leg (queue excluded):

| | generate | flatpak build | job total |
| --- | --- | --- | --- |
| aarch64 | 7m00s | 7m43s | **16m00s** |
| x86_64 | 9m35s | 10m43s | **21m18s** |

Against [33092180649](https://github.com/meshtastic/Meshtastic-Android/actions/runs/33092180649)'s 9m41s + 11m52s across two serialized jobs, so the merged leg costs about what the old tail leg did - the saving is the queue wait that no longer sits between them.

Wall clock this run was 60 min, worse than the 40 min baseline, entirely from a 32-38 min queue wait against 4m45s on the baseline run. That is pool congestion, not the change: under this run's congestion the old two-job shape would have paid it twice, for ~85 min.

Issue #6931, opened 2026-08-28
[ALIGNMENT]: Mesh Beacon config editor (design#140)


garthvh:
Tracks the Android side of [design#140](https://github.com/meshtastic/design/issues/140) — Mesh Beacon config editor alignment.

Align the Mesh Beacon module config editor (or build it to this spec if not yet implemented) with the required client behavior in design#140:

- Region shown read-only from the radio's LoRa config; block beacon setup until a region is set
- Preset pickers filtered to the region's legal set (advertised region→preset map, conservative set otherwise); keep a stored preset visible
- Offered channel required, chosen from the radio's channel list, defaulting to the primary channel
- Offered preset not editable — always the radio's configured modem preset
- Broadcast interval as a list of fixed intervals with a one-hour minimum, using the standard interval control
- One Broadcast On list of identical channel + preset rows; one row saves as the single broadcast config, added rows save as broadcast targets — no mode switch
- Selecting a channel preselects the radio's current modem preset
- Send-as-node not exposed; stored value preserved on save
- Plain help text; say "beacon", not MESH_BEACON_APP
- Suppress invitations whose offered channel matches an already-configured channel (same name and key) — see design#140 behavior 10 and meshtastic/firmware#11645

See design#140 for the full decision and rationale.

PR #6963, opened 2026-08-30
feat(settings): align the Mesh Beacon config editor with design#140

jamesarich:
### Why

The iOS Mesh Beacon editor was reworked to remove foot-guns, and [design#140](https://github.com/meshtastic/design/issues/140) makes that behavior the cross-platform standard. Our editor still let users pick a region and an offered preset by hand (describing the offered channel wrong), offered unfiltered preset lists, allowed a beacon with no joinable channel, and took a raw-seconds interval.

Part of #6931 (behaviors 1-5, 7, 9).

### What

- Region is never chosen: shown read-only from the radio's LoRa config, stamped into the offer and every broadcast target on save; a radio with no region blocks beacon setup with a clear message. An unmapped region code falls back to its enum name rather than rendering blank.
- Target preset pickers are filtered to the region's legal set from the firmware-advertised region-to-preset map, falling back to a conservative widely-supported set; a stored out-of-list preset stays visible so an existing config never renders blank.
- The offered channel is required, picked from the radio's own channel list (default primary), replacing the free-text name/PSK fields; only name and PSK are carried into `broadcast_offer_channel`.
- The offered preset is no longer editable: the radio's configured modem preset is written on save.
- Broadcast interval is a fixed-interval dropdown (one-hour minimum) using the standard control; a stored nonstandard value stays visible.
- Target rows pick a channel instead of typing an index, and picking a channel preselects the radio's current preset when the row has none.
- Save-time normalization and pickers live in a testable `MeshBeaconConfigPolicy`.

**Custom LoRa params (`use_preset=false`):** firmware (develop, `MeshBeaconModule.cpp`) transmits the stored `broadcast_offer_*` fields verbatim, with no validation against the live config, so stamping the stale `modem_preset` would advertise a preset the mesh does not actually run. The editor now gates broadcasting under custom params: the toggle can be turned off but not on, the broadcast sections show stored values read-only when broadcast is already on, listening stays fully available, and save preserves all stored broadcast fields verbatim (no stamping) so other clients and the CLI are never stomped. The gate keys off the stored flag, so toggling off in-session is recoverable.

**Placeholder channel slots:** pickers filter empty secondary slots using the exact liveness test the firmware itself applies to beacon channels (name or PSK non-empty; the raw PSK, since the decoded accessor remaps the cleartext sentinel), while keeping true slot indices for `channel_index`. Previously placeholders rendered as several identical selectable preset-name rows. A stored reference to a filtered-out slot shows as a disabled "Channel N" fallback item, and an empty channel list shows an explicit placeholder row. This deliberately diverges from Apple, which lists every slot 0-7.

### Notes for design#140 / firmware

- Behavior 4 has no carve-out for `use_preset=false`; Apple's shipped editor stamps the stale preset unconditionally, so the reference build carries the same defect. Worth a design#140 follow-up.
- Firmware's `reconfigureForBeaconTX` never touches `use_preset` or bandwidth/SF/CR, so a `broadcast_targets` preset switch is a silent PHY no-op on a custom-params radio. Worth a firmware issue.

### Testing Performed

- 28 unit tests in `MeshBeaconConfigPolicyTest` (preset filtering and fallback, interval fallback, channel-index matching, save-time stamping and offer narrowing, preset preselection, broadcast gate matrix, verbatim preservation under custom params, placeholder filtering) plus `runComposeUiTest` coverage in `MeshBeaconConfigUiTest` (region-unset gate, required-channel save path, stored-on/stored-off gate states, empty-channel-list placeholder).
- Full local baseline green after each commit: `spotlessApply spotlessCheck detekt assembleDebug test allTests`.


jamesarich:
@coderabbitai full review

PR #6964, opened 2026-08-30
feat(discovery): suppress beacon invitations for channels the radio already has

jamesarich:
### Why

A beacon inviting you to a mesh you are already on is not an invitation ([design#140](https://github.com/meshtastic/design/issues/140) behavior 10, from meshtastic/firmware#11645). The firmware cannot know an offer is redundant; the client can. iOS shipped this in 2.7.20 (Meshtastic-Apple#2371's sibling, commit f6b2e293).

Part of #6931 (behavior 10).

### What

- New pure matcher in `core/model`: an offer matches a configured channel when the name and PSK are both equal (same name with a different key is a different mesh and still shows). Placeholder secondary slots are excluded from matching.
- `DiscoveryViewModel` filters reactively by combining the stored offers with the live channel set: joining a mesh hides its standing invitations from the Discovery invitations section and the beacon scan-target list, and deleting the channel later makes the still-stored invitation reappear. Nothing is deleted from `MeshBeaconRepository`.
- `MeshDataHandlerImpl` no longer posts the first-sight notification for an already-joined offer (the offer is still stored).

### Testing Performed

- Matcher edge cases in `MeshBeaconOfferTest` (empty-name primary, default PSK, same-name/different-key, placeholder slots).
- Reactive filtering in new `DiscoveryViewModelTest` cases (join hides, delete reappears, unrelated channel changes leave offers alone, re-broadcast dedup unaffected).
- Notification suppression covered in `MeshDataHandlerTest`.
- Full local baseline green: `spotlessApply spotlessCheck detekt assembleDebug test allTests`.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshDataHandlerImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/MeshDataHandlerImpl.kt
@@ -42,6 +42,7 @@ import org.meshtastic.core.model.source
 import org.meshtastic.core.model.textMentionsNode
 import org.meshtastic.core.model.util.MeshDataMapper
 import org.meshtastic.core.model.util.decodeOrNull
+import org.meshtastic.core.model.util.isAlreadyJoined
 import org.meshtastic.core.model.util.isValidCodePoint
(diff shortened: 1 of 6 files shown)


jamesarich:
@coderabbitai full review

PR #6978, opened 2026-08-31
fix(map): resolve the persisted basemap before the map first renders

jamesarich:
Testers of the 2026-08-29 snapshot hit an error dialog over a dead map the moment the MapLibre Mesh Map opened (reported on Discord against the desktop snapshot on a Clockwork uConsole, arch64):

> Source '__MAPLIBRE_COMPOSE_source_0' was not added: its style is no longer loaded. Any layer referencing it will fail to attach.

The basemap preferences (`MapPrefs.mapStyle`, `MapTileProviderPrefs.selectedCustomTileProviderId`) are eager `StateFlow`s that start at a hardcoded default and take on the persisted value only after their first disk read. A map composed before that read resolves opens on the default basemap and swaps to the persisted one moments later. That swap is not cosmetic: `MaplibreMap` reacts to a changed `baseStyle` by unloading the live style **synchronously** (maplibre-compose 0.15.0, `MlnFfiMapSession.setBaseStyle`), and map content still attaching its sources to that style throws the error above rather than no-opping.

Fast hardware masks the race, which is why development never saw it: the preference read settles milliseconds after open, long before the style's network fetch completes, so the swap lands while nothing is attached yet. On slow storage the read can land exactly in the attach window — on every open.

### 🐛 Bug Fixes

- `rememberBasemapSelection` now returns null until both basemap preferences have been read from disk once, so the map only ever opens with one style. All five call sites (main map, inline node map, traceroute map, discovery map, node track map) skip composing the map for that first moment.
- Added `MapPrefs.awaitMapStyle()`, following the same await-the-first-disk-load convention as `awaitCameraPosition()` and `awaitHiddenLayerUrls()`; `MapTileProviderPrefs.awaitSelectedCustomTileProviderId()` already existed and is now used.

### Remaining upstream

maplibre-compose 0.15.0's `Source.attach` hard-throws on an unloaded binding that its own `mutate()` documents as "normal for a frame during a style swap", so a manual basemap switch landing in the previous style's attach window can still reach the same check. That will be reported separately against maplibre/maplibre-compose; this PR removes the systematic trigger on our side.

## Testing Performed

- New unit test: `MapPrefsImplTest` — `map style await sees the persisted value rather than the eager default`.
- Full local baseline green on the rebased branch: `./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved map startup behavior by waiting for saved map and tile preferences to load before rendering.
  * Prevented temporary blank, incorrect, or unstable basemap states during initial map display.
  * Ensured the saved map style is restored reliably, including after cold starts.

* **Tests**
  * Added coverage verifying persisted map styles are returned correctly.

### DIFF
--- a/core/prefs/src/commonMain/kotlin/org/meshtastic/core/prefs/map/MapPrefsImpl.kt
+++ b/core/prefs/src/commonMain/kotlin/org/meshtastic/core/prefs/map/MapPrefsImpl.kt
@@ -48,6 +48,8 @@ class MapPrefsImpl(private val dataStore: MapDataStore, dispatchers: CoroutineDi
         scope.launch { dataStore.edit { it[KEY_MAP_STYLE_PREF] = style } }
     }
 
+    override suspend fun awaitMapStyle(): Int = dataStore.data.map { it[KEY_MAP_STYLE_PREF] ?: 0 }.first()
+
     override val showOnlyFavorites: StateFlow<Boolean> =
         dataStore.data.map { it[KEY_SHOW_ONLY_FAVORITES_PREF] ?: false }.stateIn(scope, SharingStarted.Eagerly, false)
 
--- a/core/prefs/src/commonTest/kotlin/org/meshtastic/core/prefs/map/MapPrefsImplTest.kt
+++ b/core/prefs/src/commonTest/kotlin/org/meshtastic/core/prefs/map/MapPrefsImplTest.kt
@@ -79,6 +79,13 @@ class MapPrefsImplTest {
     }
 
     @Test
+    fun `map style await sees the persisted value rather than the eager default`() = testScope.runTest {
+        prefs.setMapStyle(3)
+
+        assertEquals(3, prefs.awaitMapStyle())
+    }
+
+    @Test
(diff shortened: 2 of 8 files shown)


jamesarich:
## Fixes Applied Successfully

Fixed 1 file based on 1 CodeRabbit feedback item (🟠 Synchronize readiness with the values used for the initial render).

**Files modified:**
- `feature/map-maplibre/src/commonMain/kotlin/org/meshtastic/feature/map/maplibre/component/BasemapSelection.kt`

**Commit:** `505a03514`

`rememberBasemapSelection` now seeds the initial selection from the awaited disk reads themselves: each live StateFlow is subscribed with `drop(1)` (discarding only the subscription replay, which can predate the first disk read), the awaited reads fill the still-null initial slots, and every later emission is a genuine update that wins. The gate can no longer open on a default the renderer would immediately swap away from. Full local baseline green (`spotlessApply spotlessCheck detekt assembleDebug test allTests`).

The latest autofix changes are on the `fix/maplibre-style-reload-source` branch.

Issue #7001, opened 2026-09-01
[Bug]: v2.8.2-closed.1 (29322131) map tab crash the FDroid app


ccchdro:
### Contact Details

ccchdro@gmail.com

### Checklist

- [x] I am able to reproduce the bug with the latest version.

- [x] I have updated to the latest *Alpha* firmware, and am able to reproduce the bug. Many issues are fixed quickly in alpha before the general beta release.

- [x] I made sure that there are no existing **OPEN or CLOSED issues** which I could contribute my information to.

- [x] I have taken the time to fill in all the required details. I understand that the bug report will be dismissed otherwise.

- [x] This issue contains only one bug.

- [x] I have read and understood the **Contribution Guidelines**.

- [x] I agree to follow this project's Code of Conduct

- [x] I actually read this list, and should be taken seriously.

### Affected app version

v2.8.2-closed.1 (29322131) fdroid

### Affected Android version

Android 11

### Affected phone model

Samsung

### Affected node model

Heltec T190

### Affected node firmware version

2.7.26.54e0d8d Beta

### Steps to reproduce the bug

It seems that the changes made to the map do not work well on all phones, on some the application crashes.

https://github.com/user-attachments/assets/592092bb-0f37-458b-9b97-5ee7322b4324

### Relevant log output

```shell

```

b8b8:
The only time ive seen this, clearing the app storage solved it. Ive had to do this maybe twice a year.

Can you try that?

ccchdro:
> The only time ive seen this, clearing the app storage solved it. Ive had to do this maybe twice a year.
> 
> Can you try that?


New installation deleted all the cash still behaves the same, 2.8.2 (29322015) works without any problems I think it has to do with the changes to the map recently made.

b8b8:
try 29322132

https://github.com/meshtastic/Meshtastic-Android/releases/download/snapshot/androidApp-fdroid-universal-debug-29322132.apk

ccchdro:
> try 29322132
> 
> https://github.com/meshtastic/Meshtastic-Android/releases/download/snapshot/androidApp-fdroid-universal-debug-29322132.apk

after installation and wizard the entire application crashes
https://github.com/user-attachments/assets/2f5e9312-1c05-40b1-af82-d7cce50309b8

jamesarich:
Root caused - the armeabi-v7a split ships no MapLibre native library at all, so this is not phone-specific.

[...]

PR #658, opened 2023-07-14
Honor radio's option to use Fahrenheit.

jp-bennett:
Closes #599

PR #7003, opened 2026-09-01
fix(map): pin the maplibre native ffi snapshot so 32-bit F-Droid gets a map engine

jamesarich:
## Why

`#6901` moved the F-Droid flavor from OSMdroid to maplibre-compose. maplibre-compose 0.15.0's native stack — `maplibre-native-ffi 0.202608.3` — publishes **arm64-v8a and x86_64 only**, but `androidApp/build.gradle.kts` still lists `armeabi-v7a` in both `abiFilters` and the ABI `splits`. The 32-bit split builds and installs happily with no map engine in it, and opening the map tab dies:

```
FATAL EXCEPTION: main
java.lang.UnsatisfiedLinkError: dlopen failed: library "libjniMaplibreNativeC.so" not found
  at org.maplibre.nativeffi.internal.javacpp.MaplibreNativeC.<clinit>(MaplibreNativeC.java:10)
  at org.maplibre.compose.mlnffi.EnsureMlnFfiConfigured(EnsureMlnFfiConfigured.android.kt:12)
  at org.meshtastic.feature.map.maplibre.MeshMapKt.MeshMap(MeshMap.kt:139)
  at org.meshtastic.feature.map.maplibre.MapLibreMapViewProvider.MapView(MapLibreMapViewProvider.kt:150)
```

Every armeabi-v7a install of 29322131 and later has no map, by construction. Shipped artifacts from `v2.8.2-closed.1`:

| release APK | size | MapLibre `.so` |
| --- | --- | --- |
| `androidApp-fdroid-arm64-v8a-release.apk` | 32.5 MB | present |
| `androidApp-fdroid-armeabi-v7a-release.apk` | 18.2 MB | **absent** |

Relates to #7001.

## 🐛 Bug Fixes

- Pin the `org.maplibre.nativeffi` group to `0.1.0-SNAPSHOT`, the rolling publication that carries armeabi-v7a, so the 32-bit split ships a map engine again.
- Set `cacheChangingModulesFor(0, "seconds")` so a snapshot pin is not served stale from the dependency cache.

## Not for merge in this form

A snapshot pin is not shippable. Upstream merged Android ARM32 on 2026-08-24 — maplibre-native-ffi [#658](https://github.com/maplibre/maplibre-native-ffi/pull/658) (ARMv7 OpenGL), [#659](https://github.com/maplibre/maplibre-native-ffi/pull/659) (Kotlin/Native `androidNativeArm32`), [#660](https://github.com/maplibre/maplibre-native-ffi/pull/660) (ARMv7 Vulkan) — but the newest release, `0.202608.3` (2026-08-20), predates them, and maplibre-compose `main` still pins that version.

The shipping fix is a two-step upstream wait:

1. maplibre-native-ffi cuts a release containing #658/#659/#660.
2. maplibre-compose bumps its `maplibre-nativeFfi` pin to it and publishes.

Then this becomes a one-line `maplibre-compose` bump in `gradle/libs.versions.toml` and this block gets deleted.

This PR exists to prove the fix works and to produce a 32-bit APK affected users can test.

## Testing Performed

Verified against the artifacts, not just the build:

- `:androidApp:assembleFdroidDebug` — the armeabi-v7a split grows 60 MB → 70 MB and now contains `lib/armeabi-v7a/libmaplibre-native-c.so` (8.8 MB) and `lib/armeabi-v7a/libjniMaplibreNativeC.so` (908 KB).
- Both are genuine 32-bit binaries: `ELF 32-bit LSB shared object, ARM, EABI5 version 1`, `Tag_CPU_arch: v7`, NDK r28c, built for Android 24.
- The snapshot FFI is drop-in compatible with maplibre-compose 0.15.0 — the arm64 build from this branch installs and renders on a Pixel 6a: `maplibre-compose: Rendered the first map frame with OPENGL`. This also clears the concern that #660's Vulkan handle ABI break would affect our OpenGL path. It does not.
- Reproduced the original failure first, to confirm the diagnosis: stripping the two MapLibre `.so` files from the arm64 APK (leaving exactly the lib set the v7a split ships), re-signing and installing gives the identical `UnsatisfiedLinkError` above.
- `spotlessApply spotlessCheck detekt` green.

Not verified: the ARM32 code path actually executing. armeabi-v7a emulator system images stop at API 25 and `minSdk` is 26, so no emulator can run it, and the only physical device here is arm64-only. That rests on upstream's CI until a 32-bit phone runs the artifact from this PR.

### DIFF
--- a/androidApp/build.gradle.kts
+++ b/androidApp/build.gradle.kts
@@ -343,3 +343,19 @@ dependencies {
     // plugin merges the generated rules into src/<variant>/generated/baselineProfiles at build time.
     baselineProfile(projects.baselineprofile)
 }
+
+// DO NOT MERGE AS-IS — pinned to an unreleased snapshot.
+//
+// #6901 moved F-Droid from OSMdroid to maplibre-compose, whose 0.15.0 native stack
+// (maplibre-native-ffi 0.202608.3) publishes arm64-v8a and x86_64 only. Our abiFilters above
+// still list armeabi-v7a, so the 32-bit split installs with no map engine and the map tab dies
+// with UnsatisfiedLinkError on libjniMaplibreNativeC.so — meshtastic/Meshtastic-Android#7001.
+//
+// Upstream merged ARM32 on 2026-08-24 (maplibre-native-ffi#658, #659, #660) but has not cut a
+// release carrying it. The 0.1.0-SNAPSHOT publication does carry it, which is what this pin is
+// for. Replace with a maplibre-compose version whose ffi pin is >= the first release containing
+// those PRs, then delete this block.
+configurations.configureEach {
+    resolutionStrategy.eachDependency { if (requested.group == "org.maplibre.nativeffi") useVersion("0.1.0-SNAPSHOT") }
+    resolutionStrategy.cacheChangingModulesFor(0, "seconds")
+}


jamesarich:
Closing - the pin cannot ship in any form, and the check that replaces it is #7013.

- `0.1.0-SNAPSHOT` was re-published overnight: yesterday `20260825.201433-1`, today `20260902.064417-1`. Same version string, different engine, and `rb-check` assembles once so it would never notice.
- A timestamped pin does not work either: the timestamps are per artifact (`-android` `064417-1`, `-runtime-opengl-android` `064601-1`), and Central prunes snapshots after 90 days, so a merged pin 404s for any rebuild of that tag after late November.
- Renovate cannot see `resolutionStrategy.eachDependency`, so the force would silently override the `maplibre-compose` bump it does deliver.

The v7a test build from this PR's CI run stays valid for the artifact's 7-day retention: https://github.com/meshtastic/Meshtastic-Android/actions/runs/33562083023/artifacts/9821973919

PR #7005, opened 2026-09-02
fix(map): show a message instead of crashing when a device has no map engine, and fail CI on the gap

jamesarich:
## Why

MapLibre Compose reaches its renderer through a JNI shim, and an app can be built for an architecture that shim was never published for. `maplibre-native-ffi 0.202608.3` publishes arm64-v8a and x86_64 only, so the F-Droid `armeabi-v7a` split carries no engine — 18.2 MB against the arm64 split's 32.5 MB, with CI green — and the first frame of a map takes the whole app down:

```
FATAL EXCEPTION: main
java.lang.UnsatisfiedLinkError: dlopen failed: library "libjniMaplibreNativeC.so" not found
  at org.maplibre.compose.mlnffi.EnsureMlnFfiConfigured(EnsureMlnFfiConfigured.android.kt:12)
  at org.meshtastic.feature.map.maplibre.MapLibreMapViewProvider.MapView(MapLibreMapViewProvider.kt:150)
```

Losing the map on those devices is unavoidable until upstream ships a 32-bit build. Losing every other screen is not, and neither is shipping the gap without knowing. Two commits:

1. **Runtime** — ask whether the engine can load before composing a map, and say so plainly when it cannot.
2. **Build time** — fail CI when an ABI split ships fewer native libraries than its siblings, with the current gap recorded as a known exception that fails the *other* way once it closes.

Relates to #7001.

## 🐛 Bug Fixes

- `isMapLibreRuntimeAvailable()` — an `expect`/`actual` probe. Android loads the JNI shim inside `runCatching`; `UnsatisfiedLinkError` is an `Error`, so a `catch (e: Exception)` would have missed it and kept the crash.
- Guarded at all three places that compose `MaplibreMap`: `MapLibreMapViewProvider` (so the toolbar and zoom controls do not float over an empty screen driving a map that is not there), plus `MeshMap` and `SecondaryMapSurface`, which reach it directly.
- `MapEngineUnavailable` — a deliberately actionless screen. There is nothing a user can do about their phone's architecture, so it states what happened, states the rest of the app is fine, and stops.

## 🧹 Chores

- `scripts/lib/abi-parity.sh` — the classifier. The reference set is the union of native libs across a flavor's split APKs; every split must carry all of it unless a known-gap line says otherwise.
- `scripts/verify-abi-parity.sh` — runs in `android-check` right after the debug assemble, where every PR already builds the splits.
- `scripts/verify-abi-parity-selftest.sh` — fixture self-test in `lint-check`, mirroring `verify-rb-selftest.sh`.
- Two strings, sorted with `scripts/sort-strings.py`; `strings-index.txt` regenerated.

## Notes for review

- **Known gaps are checked both ways.** The two recorded lines are the MapLibre gap on `fdroid/armeabi-v7a`. An entry must be *absent* from its split, and the check *fails once the library turns up*, so an entry cannot outlive its reason. Renovate tracks the 8 `org.maplibre.compose` catalog entries and will bump `maplibre-compose` when a release with the 32-bit build lands (upstream merged it 2026-08-24: maplibre-native-ffi [#658](https://github.com/maplibre/maplibre-native-ffi/pull/658), [#659](https://github.com/maplibre/maplibre-native-ffi/pull/659), [#660](https://github.com/maplibre/maplibre-native-ffi/pull/660); no release yet). Without this, that bump lands green and nothing marks #7001 closed. With it, the bump goes red with *"the gap has closed, delete those lines"* — and deleting them is the proof the v7a APK really has an engine.
- **The probe loads the shim, not the 13 MB engine.** `libjniMaplibreNativeC.so`'s `DT_NEEDED` lists `libmaplibre-native-c.so`, so one load answers for both. `System.loadLibrary` is process-wide and idempotent — MapLibre's own load later finds it already done.
- **The shim name is hardcoded in two places** (the probe and the known-gap list), and that is the maintenance edge. An upstream rename fails the probe closed — fallback screen on every device — with no JVM-side test to catch it. The parity check's *stale* direction is the backstop: a renamed shim shows up as a new lib in the arm64 split and a stale entry for the old name.
- **Off Android nothing changes** — every published desktop and iOS artifact carries its own engine, so those actuals return `true` unconditionally.
- **Why not pin the upstream snapshot** (#7003, closed): it was re-published overnight with different bytes under the same version, timestamps are per artifact so a fixed pin cannot address the group, and Central prunes snapshots after 90 days. `rb-check` assembles once and would never have noticed.

## Testing Performed

**Runtime guard**
- `MapLibreRuntimeTest` — `probeNativeRuntime` returns `false` for a thrown `UnsatisfiedLinkError` (an `Error`, which is the whole point) and `true` for a load that returns; `SecondaryMapSurface` under a `LocalMapLibreRuntimeProbe` that says no composes the fallback and its content **zero** times. Mutation-checked: deleting the guard fails the test with `No ComposeMapHost is installed` — MapLibre being reached. `MeshMap` and the provider carry the identical one-line guard but need a seven-repository view model and Koin to compose, so they are not harnessed.
- `MapEngineUnavailableTest` — both strings displayed, and `onAllNodes(hasClickAction())` is empty.
- `ScreenshotMapEngineUnavailable` added to `MapScreenshotTests`, light and dark references committed; `:screenshot-tests:validateDebugScreenshotTest` green.
- Probe passes (engine present): the arm64 build installs on a Pixel 6a and renders as before — `maplibre-compose: Rendered the first map frame with OPENGL`, no fallback in the view tree.
- Probe fails (engine absent): stripping the two MapLibre libs from that same APK — exactly the lib set the v7a split ships — re-signing and installing shows "Map unavailable on this device" instead of the FATAL, with Nodes, Messages and Settings still working.
- Baseline green: `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile`. detekt's `LongMethod` caught the first draft of the guard at 63 lines against a 60 limit, hence the single-line early return.

**Parity check**, run against three real APK sets:

| input | result |
| --- | --- |
| shipped `v2.8.2-closed.1` fdroid splits + google splits | ✅ google clean; ✅ fdroid "splits match apart from 2 recorded known gap(s)" |
| the #7003 snapshot build, where v7a **has** the engine | ❌ "known-gap entries whose library is now present" — the self-healing signal |
| shipped splits with the two lines deleted | ❌ "native libraries missing from a split: armeabi-v7a/libjniMaplibreNativeC.so armeabi-v7a/libmaplibre-native-c.so" |

- Self-test: 22 checks green — 15 on the classifier, including wrong-ABI and wrong-flavor entries that must *not* excuse a gap, and 7 that run `verify-abi-parity.sh` itself against fixture output trees (universal APK ignored, unrecorded gap, the checked-in gap held and then closed, a lone split, an empty directory, two failing flavors summing in the exit status).
- `shellcheck -x` over `scripts/` (the lint-check invocation) and `actionlint` on both workflows: clean.

Not verified: execution on real 32-bit ARM hardware. armeabi-v7a emulator images stop at API 25 and `minSdk` is 26, and the only device here is arm64-only — the probe-fail path was exercised by removing the libraries rather than by running on a v7a phone.

## Summary by CodeRabbit

* **New Features**
  * Added a clear fallback screen when maps are unavailable on a device, with an explanation that other app features remain usable.
  * Added localized messaging and visual coverage for the unavailable-map experience.

* **Bug Fixes**
  * Prevented map screens from crashing when the required native map renderer is unavailable.
  * Added validation to detect missing native libraries across Android ABI splits during builds and pull-request checks.


jamesarich:
@coderabbitai review

PR #7015, opened 2026-09-02
fix(connections): scope region warnings to the active connection

jeremiah-k:
## Overview

The Connections screen could make region and configuration state actionable before it was known to belong to the active transport session. Repository node state can briefly outlive a device switch or reconnect, so a non-null cached node is not enough to establish that configuration guidance describes the current connection.

`ConnectionStatus` also carried a `MUST_SET_REGION` value that mixed configuration health into the connection lifecycle. Once region guidance is owned by the connected-state configuration card, that status has no distinct presentation role. Keep connection status lifecycle-only and use `regionUnset` together with active-session node readiness for the actionable region card instead.

Configuration warnings additionally used `sessionAuthorized` as a proxy for write eligibility. That flag represents a successful lockdown authentication session rather than general configuration availability. Pre-2.8 firmware, newer builds that do not participate in runtime lockdown status, and lockdown-capable devices that explicitly report `DISABLED` do not require an authenticated lockdown session for normal configuration writes.

Lockdown command dispatch had a related edge case: the coordinator could enter `AwaitingResponse` even when no transport accepted the request. Managed mode is separate from lockdown and follows the existing Android convention of treating a device as managed only when `SecurityConfig.is_managed` is explicitly true.

This keeps connection lifecycle separate from configuration health, scopes actionable region/transmit guidance to the active connection, and makes lockdown pending state depend on actual transport admission.

### Active-session readiness and connection status

* Add `activeNodeInfoReady` by matching `ConnectionIdentity.nodeNum` against `ourNodeInfo.num`.
* Ignore cached or mismatched node information left over from an earlier connection when deciding whether configuration guidance is actionable.
* Withdraw node readiness when the active identity is cleared or replaced and restore it only after matching node information arrives.
* Keep `ConnectionStatus` limited to connection lifecycle states and remove `MUST_SET_REGION` from the enum.
* Keep `regionUnset` as a separate configuration-health input rather than rewriting the connection label.
* Stop rendering the old `You must set a region!` connection-status message; the actionable `Set your region` card is the single region-warning surface.
* Cover same-node reconnect ordering where active-node readiness can become true before the service reaches its post-handshake `CONNECTED` state without changing the connection lifecycle label.

### Lockdown and configuration eligibility

* Add `LockdownState.allowsConfigWrites` as the lockdown-state-only configuration-write eligibility signal.
* Treat no received lockdown status, explicit `DISABLED`, and authenticated `UNLOCKED` as not withholding normal configuration writes on lockdown grounds.
* Keep locked, provisioning, pending-response, authentication-failure, and backoff states non-actionable.
* Propagate direct transport admission from `PacketHandler` through `CommandSender` for lockdown requests.
* Enter `AwaitingResponse` only after a transport accepts the request.
* Preserve retryable lockdown state when dispatch is rejected.
* Do not stage or erase credentials for rejected requests.
* Fall back to the locked state when stored-passphrase auto-unlock cannot be dispatched.
* Arm the Lock Now acknowledgement transition only after the command has been admitted.

### Connections warning presentation

* Require connected-node content and `activeNodeInfoReady` before showing region or transmit-disabled cards.
* Match the existing Android managed-mode convention by suppressing local configuration actions only when `localConfig.security?.is_managed == true`.
* Treat missing `SecurityConfig` as not known-managed so first-run region guidance is not hidden while configuration is still arriving.
* Continue suppressing configuration actions for mock and replay devices.
* Remove the connection-warning dependency on raw `sessionAuthorized`.
* Keep the actionable `Set your region` card as the single connected-state region warning.

## Testing

Coverage includes:

* connection lifecycle remaining `CONNECTED` when the active device has an unset region;
* active-node readiness for matching and mismatched connection identities;
* readiness withdrawal and restoration when the active identity is cleared or replaced;
* same-node readiness before the final connected state without rewriting the connection lifecycle;
* lockdown configuration-write eligibility boundaries;
* direct transport admission and rejection for lockdown commands;
* propagation of admission results through the command-sender layer;
* manual and stored-passphrase request ordering;
* rejected unlock, disable, auto-unlock, and Lock Now requests;
* firmware responses replacing an admitted pending lockdown state;
* active-node readiness, managed-device, missing-node, and virtual-device warning policy;
* Compose rendering of the production region and transmit-disabled warning cards.

## Scope

* Connection lifecycle, active-session configuration readiness, lockdown request admission/state handling, and Connections warning presentation only.
* Radio configuration values and firmware lockdown/authentication rules are unchanged.
* Managed-mode firmware behavior is unchanged; the Connections screen follows the Android client's existing `SecurityConfig` convention.
* Transport selection, persistence schema, protocol format, and firmware behavior are unchanged.
* Existing translated `must_set_region` resources are left untouched; removing unused localization keys is outside this behavioral fix.
* No database or user-data migration is required.

## Summary by CodeRabbit

- **New Features**
  - Added an “awaiting response” lockdown state while passphrase requests are processed.
  - Lockdown controls now reflect firmware responses and reappear when authentication is required.
  - Added configuration-write eligibility handling based on device lockdown status.
  - Improved connection tracking to ignore stale node information from previous sessions.
  - Added persistent region warnings when the active device requires configuration.

- **Bug Fixes**
  - Prevented premature region and configuration warnings during handshakes, managed mode, lockdown, or virtual-device connections.
  - Improved connection status accuracy when reconnecting to the same device.

### DIFF
--- a/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/LockdownCoordinatorImpl.kt
+++ b/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/LockdownCoordinatorImpl.kt
@@ -127,14 +127,22 @@ class LockdownCoordinatorImpl(
                 }
             if (stored != null) {
                 Logger.i { "Lockdown: Auto-unlocking with stored passphrase" }
-                wasAutoAttempt = true
-                commandSender.sendLockdownPassphrase(
-                    stored.passphrase,
-                    stored.boots,
-                    stored.hours,
-                    stored.maxSessionSeconds,
-                )
-                return
+                // A fresh LOCKED status ends any previous auto-attempt outcome. Only an admitted replay arms failure
+                // handling for the new attempt.
+                wasAutoAttempt = false
+                val dispatched =
+                    commandSender.sendLockdownPassphrase(
+                        stored.passphrase,
+                        stored.boots,
+                        stored.hours,
+                        stored.maxSessionSeconds,
+                    )
+                if (dispatched) {
+                    wasAutoAttempt = true
+                    serviceRepository.setLockdownState(LockdownState.AwaitingResponse)
+                    return
+                }
+                Logger.w { "Lockdown: Auto-unlock command was not accepted by the active transport" }
             }
         }
         serviceRepository.setLockdownState(LockdownState.Locked(lockReason))
@@ -212,13 +220,21 @@ class LockdownCoordinatorImpl(
         hours: Int,
         maxSessionSeconds: Int,
         disable: Boolean,
-    ) {
+    ): Boolean {
+        if (!commandSender.sendLockdownPassphrase(passphrase, boots, hours, maxSessionSeconds, disable)) {
+            Logger.w { "Lockdown: Passphrase command was not accepted by the active transport" }
+            return false
+        }
+
         wasAutoAttempt = false
         wasLockNow = false
         if (disable) {
-            // Turning lockdown OFF: the device will reboot to DISABLED, so there is nothing to re-save. Drop any
-            // stored passphrase now so a later reconnect doesn't auto-unlock a device the user just disabled.
+            // Turning lockdown OFF: the device will reboot to DISABLED, so there is nothing to re-save. Clear the
+            // stored passphrase only after transport admission succeeds so a rejected request remains retryable.
             pendingPassphrase = null
+            pendingBoots = LockdownPassphraseStore.DEFAULT_BOOTS
+            pendingHours = 0
+            pendingMaxSessionSeconds = 0
             val deviceAddress = radioInterfaceService.getDeviceAddress()
             if (deviceAddress != null) {
                 try {
@@ -233,13 +249,18 @@ class LockdownCoordinatorImpl(
             pendingHours = hours
             pendingMaxSessionSeconds = maxSessionSeconds
         }
-        serviceRepository.setLockdownState(LockdownState.None)
-        commandSender.sendLockdownPassphrase(passphrase, boots, hours, maxSessionSeconds, disable)
+        serviceRepository.setLockdownState(LockdownState.AwaitingResponse)
+        return true
     }
 
-    override fun lockNow() {
-        wasLockNow = true
-        commandSender.sendLockNow()
+    override fun lockNow(): Boolean {
+        val dispatched = commandSender.sendLockNow()
+        if (dispatched) {
+            wasLockNow = true
+        } else {
+            Logger.w { "Lockdown: Lock Now command was not accepted by the active transport" }
+        }
+        return dispatched
     }
 
     private fun resetTransientState() {
--- a/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/ConnectionsScreen.kt
+++ b/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/ConnectionsScreen.kt
@@ -61,6 +61,7 @@ import org.meshtastic.core.model.DeviceType
 import org.meshtastic.core.model.FirmwareUpdateDestination
 import org.meshtastic.core.model.FirmwareUpdateNotice
 import org.meshtastic.core.model.InterfaceId
+import org.meshtastic.core.model.service.LockdownState
 import org.meshtastic.core.navigation.FirmwareRoute
 import org.meshtastic.core.navigation.Route
 import org.meshtastic.core.navigation.SettingsRoute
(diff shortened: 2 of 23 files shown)


feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/ConnectionsScreen.kt
```diff
@@ -341,18 +390,33 @@ fun ConnectionsScreen(
...
+            regionWarningLabel =
+            if (
+                shouldShowPersistentRegionWarning(
+                    connectionStatus = connectionStatus,
+                    lockdownState = lockdownState,
+                    isManaged = managedMode,
+                    isPhysicalDevice = isPhysicalDevice,
+                )
+            ) {
+                stringResource(Res.string.must_set_region)
```
jamesarich: `shouldShowPersistentRegionWarning` and `canShowConfigWarnings && regionUnset` are true under the same conditions, so "You must set a region!" renders inside the card and "Set your region" right under it (https://github.com/meshtastic/Meshtastic-Android/blob/fb188e4925375ddb00db8ebb0a8c930c44f7455d/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/ConnectionsScreen.kt#L580-L592). Drop the label - the card is the actionable one.

feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/ConnectionsScreen.kt
```diff
@@ -140,13 +143,57 @@ import org.meshtastic.feature.connections.ui.components.CurrentlyConnectedText
...
+ * authoritative config arrives.
+ */
+internal fun resolveManagedMode(firmwareVersion: String?, localConfig: LocalConfig): Boolean? {
+    localConfig.security?.let {
+        return it.is_managed
+    }
+
+    val version = firmwareVersion?.let(::DeviceVersion)?.takeIf { it.isValid }
+    val predatesSecurityConfig = version != null && version < SecurityConfigManagedModeVersion
+    return if (predatesSecurityConfig) localConfig.device?.is_managed else null
```
jamesarich: This fails closed and version-splits, every other site reads `security?.is_managed` fail-open - https://github.com/meshtastic/Meshtastic-Android/blob/971885339bcf7fe74bfd5342edf657bb73066852/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/channel/ChannelViewModel.kt#L52, https://github.com/meshtastic/Meshtastic-Android/blob/971885339bcf7fe74bfd5342edf657bb73066852/feature/settings/src/androidMain/kotlin/org/meshtastic/feature/settings/SettingsScreen.kt#L230. Unresolved metadata now hides a first-run warning. Either lift a shared helper or match the convention.

jamesarich (changes requested):
The stale-node race in the title is already closed by the handshake ordering - `Connected` only fires after Stage 2 and `connectionIdentity` is cleared on switch and at every generation boundary, so the `activeNodeInfoReady` gate is harmless but not the fix. The shipped bug is the `sessionAuthorized` gate - it is only ever set true at https://github.com/meshtastic/Meshtastic-Android/blob/971885339bcf7fe74bfd5342edf657bb73066852/core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/LockdownCoordinatorImpl.kt#L174, so both cards have been unreachable on every non-lockdown device since #5939 - https://github.com/meshtastic/Meshtastic-Android/blob/v2.8.1/feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/ConnectionsScreen.kt#L429. `allowsConfigWrites` is the right replacement. Lead with that in the title/body and link an issue - the race framing can go down to a sentence.

core/data/src/commonMain/kotlin/org/meshtastic/core/data/manager/LockdownCoordinatorImpl.kt
```diff
@@ -233,7 +234,7 @@ class LockdownCoordinatorImpl(
             pendingHours = hours
             pendingMaxSessionSeconds = maxSessionSeconds
         }
-        serviceRepository.setLockdownState(LockdownState.None)
+        serviceRepository.setLockdownState(LockdownState.AwaitingResponse)
         commandSender.sendLockdownPassphrase(passphrase, boots, hours, maxSessionSeconds, disable)
     }
```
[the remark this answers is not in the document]
jeremiah-k: @coderabbitai Addressed. Lockdown dispatch now reports whether a transport actually accepted the frame, and the coordinator enters `AwaitingResponse` only after successful admission.

Rejected manual or stored-passphrase unlocks retain a retryable lockdown state. Rejected manual/disable requests do not stage or erase credentials, and rejected Lock Now requests do not arm the acknowledgement transition.

Coverage exercises packet-handler admission, command-sender propagation, and coordinator admission/rejection ordering, including the state observed at the actual dispatch boundary.

feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/ConnectionsScreen.kt
jeremiah-k: Agreed. I dropped the persistent connected-card label implementation and removed that third commit from the rewritten history.

I simplified the remaining path further as well: `MUST_SET_REGION` is now removed from `ConnectionStatus` entirely. Region state no longer rewrites the connection lifecycle label; `regionUnset` plus `activeNodeInfoReady` and the configuration policy drive the actionable `Set your region` card directly. That leaves a single region-warning surface without a transition-specific workaround.

feature/connections/src/commonMain/kotlin/org/meshtastic/feature/connections/ui/ConnectionsScreen.kt
jeremiah-k: Agreed. I removed the version-split resolver and matched the existing client convention. Connections now treats the device as managed only when `localConfig.security?.is_managed == true`.

Missing `SecurityConfig` therefore fails open, so first-run region guidance is not hidden while configuration or metadata is still settling. The nullable/unresolved managed-mode policy and its associated resolver tests were removed accordingly.

jamesarich (approved):
Both threads addressed - label gone, `MUST_SET_REGION` out of the lifecycle enum, managed mode fail-open like the rest of the client. `AwaitingResponse` only has a lost-frame residual and disconnect resets it, same as the old `None`. One non-blocking thing - https://github.com/meshtastic/Meshtastic-Android/blob/53a129b20/feature/settings/src/commonMain/kotlin/org/meshtastic/feature/settings/radio/component/SecurityConfigScreen.kt#L222 still drops the new `Boolean`, so a rejected Lock Now is a silent no-op. Fine as a follow-up.

jamesarich (approved):

PR #7020, opened 2026-09-03
fix(takserver): skip compressed v1 TAK packets to stop duplicate contacts

### LINKED ISSUE #7019: Legacy v1 TAK path renders unishox2-compressed callsigns as literal text, creating a duplicate contact with an unreadable name
## Summary

With firmware ≤ 2.7.x (the current stable line), every ATAK PLI arriving over the mesh produces **two** contacts in ATAK: the correct one, and a duplicate whose `uid` and `callsign` are mojibake.

The cause is a mismatch between firmware and app:

1. **Firmware `AtakPluginModule`** unishox2-compresses `contact.callsign` / `contact.device_callsign` for LoRa transport and sets `is_compressed = true`. On receive it decompresses into a **copy** and sends that to the phone — but the original, still-compressed packet is *also* delivered. The phone therefore receives both.
2. **`handleV1Packet`** in this app reads the string fields verbatim and never checks `is_compressed`, so the compressed copy is turned into a CoT event whose `uid`/`callsign` are raw compressed bytes. Protobuf's string decoder replaces every non-UTF-8 byte with `U+FFFD`, and ATAK renders the result as a second, permanent contact.

The comment at `TAKPacketConversion.kt:36-37` states the flag exists *"but the firmware doesn't act on it"*. That assumption is incorrect — the firmware does act on it, and that is precisely why the compressed copy is never decoded.

## Environment

| | |
|---|---|
| App | 2.8.1 (29321949) google, release build |
| App source | `main` @ `9718853` (2026-09-02) — code paths below still current |
| Radio firmware | 2.7.15 (both nodes) |
| Clients | ATAK-CIV 5.7.0.10 and 5.7.0.14, connected to the app's local TAK server over mTLS |
| Mesh-to-CoT pref | **disabled** on both devices (not related to this bug) |
| ATAK Meshtastic plugin | **not installed** on either device |

## Observed behaviour

Captured by attaching a passive mTLS client to the app's local TAK server (port 8089) using the bundled `tak_certs/client.p12`. Coordinates and device UIDs redacted.

**Device A sees Device B's PLI twice:**

```xml
<event uid='ANDROID-<redacted-a>' type='a-f-G-U-C' time='...:25:01.966Z' ...>
  <point lat='53.2148xx' lon='63.6294xx' hae='126.0' .../>
  <detail><contact endpoint='*:-1:stcp' callsign='Ohio'/><uid Droid='Ohio'/>
  <__group role='Team Member' name='Green'/><status battery='100'/>
  <track course='277.0' speed='0.0'/></detail></event>

<event uid='&#xFFFD;Nu&#xFFFD;&#xFFFD;&#xFFFD;&#xFFFD;E`&#xFFFD;&#xFFFD;~&#xFFFD;sw&#xFFFD;&#xFFFD;' type='a-f-G-U-C' time='...:25:02.044Z' ...>
  <point lat='53.2148xx' lon='63.6294xx' hae='126.0' .../>
  <detail><contact endpoint='*:-1:stcp' callsign='&#xFFFD;v&#xFFFD;'/><uid Droid='&#xFFFD;v&#xFFFD;'/>
  <__group role='Team Member' name='Green'/><status battery='100'/>
  <track course='277.0' speed='0.0'/></detail></event>
```

The two events arrive **~75 ms apart** with **byte-identical** `lat`, `lon`, `hae`, `battery` and `course`. Only the string fields differ. Over a 10-minute capture the pairing was exactly 1:1 (5 clean / 5 corrupt, no drops), across ATAK restarts on both ends.

It is **symmetric** — each device shows a corrupt twin of the *other* device, because the duplicate is manufactured by the receiving radio.

### The corrupted bytes are unishox2 output, not random corruption

Raw bytes of the corrupt fields (hex, as delivered in the XML):

```
callsign: ef bf bd 76 ef bf bd                    -> U+FFFD 'v' U+FFFD
uid:      ef bf bd 4e 75 ef bf bd ef bf bd
          ef bf bd ef bf bd 45 60 ...             -> U+FFFD 'N' 'u' ... 'E' '`' ...
```

Two independent devices produce corrupt UIDs with an **identical 8-byte prefix**:

```
Device A sees:  <FFFD> N u <FFFD> <FFFD> <FFFD> <FFFD> E ` <FFFD> <FFFD> ~ <FFFD> s w <FFFD> <FFFD>
Device B sees:  <FFFD> N u <FFFD> <FFFD> <FFFD> <FFFD> E c <FFFD> A C c F + <TAB>
```

Both real UIDs begin with the literal `ANDROID-`. A shared plaintext prefix yielding a shared compressed prefix is the signature of the compression, and confirms these are compressed strings rather than damaged ones.

Note the corruption is **lossy and irreversible** by the time it reaches XML: each invalid byte has already been replaced with `U+FFFD`, so the original compressed bytes cannot be recovered downstream.

### Firmware side (≤ 2.7.x)

`src/modules/AtakPluginModule.cpp`, receive branch of `alterReceivedProtobuf`:

```cpp
auto decompressedCopy = packetPool.allocCopy(mp);
decompressedCopy->decoded.payload.size =
    pb_encode_to_bytes(decompressedCopy->decoded.payload.bytes, ..., &uncompressed);
service->sendToPhone(decompressedCopy);
```

The transmit branch modifies `mp` **in place**; the receive branch instead sends a copy and leaves the original untouched, so the compressed original is delivered to the phone as well. Present in 2.7.15 and unchanged through **2.7.26** (the only delta there is `strlen()` → `pb_string_length()`).

Fixed in 2.8.0 by rewriting the module to a passthrough on `ATAK_PLUGIN_V2`:

```cpp
ProcessMessage AtakPluginModule::handleReceived(const meshtastic_MeshPacket &mp)
{
    (void)mp; // Passthrough - no processing needed, apps handle compression/decompression
    return ProcessMessage::CONTINUE;
}
```

### App side (still present on `main`)

`TAKMeshIntegration.kt:442`:

```kotlin
private suspend fun handleV1Packet(payload: okio.ByteString) {
    val takPacket = TAKPacket.ADAPTER.decode(payload)
    val cotMessage = takPacket.toCoTMessage() ?: return
```

`TAKPacketConversion.kt:130`:

```kotlin
fun TAKPacket.toCoTMessage(): CoTMessage? {
    val rawDeviceCallsign = contact?.device_callsign ?: "UNKNOWN"
    val senderCallsign = contact?.callsign ?: "UNKNOWN"
```

`is_compressed` is never read, and the strings are never validated as UTF-8.

## Expected behaviour

A v1 `TAKPacket` with `is_compressed = true` should have its string fields unishox2-decompressed before conversion to CoT. At minimum, a packet whose strings are not valid UTF-8 should be dropped rather than surfaced to TAK clients as a contact.

## Suggested fix

In `handleV1Packet` (or in `TAKPacket.toCoTMessage()`), branch on the flag:

```kotlin
val takPacket = TAKPacket.ADAPTER.decode(payload)
val decoded = if (takPacket.is_compressed) takPacket.decompressStrings() else takPacket
val cotMessage = decoded.toCoTMessage() ?: return
```

decompressing `contact.callsign`, `contact.device_callsign`, and the GeoChat `message` / `to` / `to_callsign` fields with unishox2 (`USX_PSET_DFLT`, matching the firmware).

This is worth doing independently of the firmware fix, because the **official ATAK plugin also sets `is_compressed`** — so any sender using the documented flag is currently decoded incorrectly. The proto comment is explicit: *"Are the payloads strings compressed for LoRA transport?"*

A defensive UTF-8 validity check before emitting a contact would also prevent unreadable entries from reaching ATAK if any other sender gets this wrong.

## Impact

Cosmetic but persistent, and not self-clearing: the duplicate is re-broadcast with every PLI (~15 s here) with a fresh 10-minute stale, so it cannot be dismissed from the ATAK contact list. Positional data itself is unaffected — both copies carry correct coordinates.

## Note on upgrading

For anyone hitting this: upgrading only *one* radio to 2.8.0 makes it worse rather than better. A 2.8.0 radio no longer registers port 72, so a compressed packet from a 2.7.x peer passes through undecompressed, and since the app does not decompress either, the *clean* copy is lost and only the corrupt contact remains. Both radios need to move together.

texaskst:
## Problem

On firmware <= 2.7.x, every inbound ATAK PLI produces **two** contacts in the connected TAK client: the correct one, and a duplicate whose `uid` and `callsign` are unreadable.

The firmware's `AtakPluginModule` unishox2-compresses `contact.callsign` / `device_callsign` for LoRa transport and sets `is_compressed`. On receive it decompresses into a **copy** and sends that to the phone — but unlike the transmit branch, it does not rewrite `mp` in place, so the original compressed packet is delivered as well:

```cpp
// AtakPluginModule::alterReceivedProtobuf, receive branch
auto decompressedCopy = packetPool.allocCopy(mp);
decompressedCopy->decoded.payload.size = pb_encode_to_bytes(..., &uncompressed);
service->sendToPhone(decompressedCopy);   // original mp still goes to the phone too
```

`handleV1Packet` never checked the flag, so the compressed twin was converted using its raw string fields. Those bytes are not valid UTF-8 and decode to replacement characters, which the TAK client renders as a second contact.

## Evidence

Captured with a passive mTLS client on the app's local TAK server (port 8089), using the bundled `tak_certs/client.p12`. Coordinates and device UIDs redacted:

```xml
<event uid='ANDROID-<redacted>' type='a-f-G-U-C' time='...:25:01.966Z'>
  <point lat='53.2148xx' lon='63.6294xx' hae='126.0'/>
  <detail><contact callsign='Ohio'/><status battery='100'/>
  <track course='277.0' speed='0.0'/></detail></event>

<event uid='&#xFFFD;Nu&#xFFFD;&#xFFFD;&#xFFFD;&#xFFFD;E`&#xFFFD;&#xFFFD;~&#xFFFD;sw&#xFFFD;&#xFFFD;' type='a-f-G-U-C' time='...:25:02.044Z'>
  <point lat='53.2148xx' lon='63.6294xx' hae='126.0'/>
  <detail><contact callsign='&#xFFFD;v&#xFFFD;'/><status battery='100'/>
  <track course='277.0' speed='0.0'/></detail></event>
```

The pair arrives ~75 ms apart with byte-identical `lat`, `lon`, `hae`, `battery` and `course` — only the string fields differ. Over a 10-minute capture the pairing was exactly 1:1 (5 clean / 5 corrupt, no drops), across client restarts, and symmetric: each device shows a corrupt twin of the other, because the duplicate is produced by the receiving radio.

That the corrupt strings are compressed rather than damaged is confirmed independently of any source reading — two different devices yield corrupt UIDs sharing an identical 8-byte prefix, because both real UIDs begin with the literal `ANDROID-`:

```
Device A:  <FFFD> N u <FFFD> <FFFD> <FFFD> <FFFD> E ` <FFFD> <FFFD> ~ <FFFD> s w <FFFD> <FFFD>
Device B:  <FFFD> N u <FFFD> <FFFD> <FFFD> <FFFD> E c <FFFD> A C c F + <TAB>
```

Tested with app 2.8.1 (release), radios on firmware 2.7.15, ATAK-CIV 5.7.0.10 and 5.7.0.14. The mesh-to-CoT preference was disabled and the ATAK plugin was not installed, so neither contributes.

## Fix

Skip inbound v1 packets flagged `is_compressed`. The decompressed copy the radio already delivers carries identical position, team and telemetry, so nothing usable is lost — and this module has no unishox2 implementation to decode the compressed one with.

The `TAKPacketConversion` doc comment stated the firmware "doesn't act on" `is_compressed`. It does, and that assumption is why the flag was never handled — corrected here.

## Notes

Impact is cosmetic but not self-clearing: the duplicate is re-sent with every PLI and carries a fresh stale time, so it cannot be dismissed from the contact list. Position data itself is unaffected.

Firmware 2.8.0 removes the duplicate at source by rewriting the module to a passthrough on `ATAK_PLUGIN_V2`, but 2.7.x is the current stable line and this path is retained for it. Handling the flag is worth doing independently in any case, since the official ATAK plugin also sets `is_compressed` — any sender using the documented flag is currently decoded incorrectly.

A full unishox2 decode would be the richer fix, but it needs an implementation this module does not have; skipping is the minimal correct behaviour until then.

## Testing

- New regression test `inbound compressed V1 packet is not broadcast to TAK clients`; verified it fails without the guard and passes with it
- `:core:takserver:jvmTest` — 18 tests, 0 failures
- `spotlessApply` and `detekt` clean

Fixes #7019

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented compressed legacy packets from being forwarded to TAK clients when an equivalent decompressed packet is available.
  * Improved handling of compressed callsign and GeoChat data over LoRa transport.
  * Clarified supported packet formats and compression behavior for legacy TAK integrations.

* **Tests**
  * Added regression coverage to verify compressed inbound position packets are not broadcast.


core/takserver/src/commonMain/kotlin/org/meshtastic/core/takserver/TAKMeshIntegration.kt
```diff
@@ -434,14 +434,30 @@ class TAKMeshIntegration(
...
+            // Firmware <= 2.7.x unishox2-compresses contact.callsign / device_callsign for LoRa transport and sets
+            // is_compressed. On receive it decompresses into a *copy* and sends that to the phone, but the original
+            // compressed packet is delivered as well (AtakPluginModule::alterReceivedProtobuf allocates a copy rather
+            // than rewriting mp in place, as the transmit branch does). We therefore see every PLI twice.
+            //
+            // The compressed twin cannot be rendered here: this module has no unishox2 implementation, so its string
+            // fields decode as replacement characters and ATAK surfaces them as a second contact with an unreadable
+            // callsign, re-broadcast on every PLI and impossible to dismiss. Skip it and keep the decompressed copy,
+            // which carries identical position, team and telemetry.
+            if (takPacket.is_compressed) {
```
jamesarich: Gate this on `!useTakV2()` - the decompressed twin only exists when the local radio is 2.7. On 2.8 this is the only copy of a 2.7 peer's PLI (spec 005 US5, https://github.com/meshtastic/Meshtastic-Android/blob/971885339bcf7fe74bfd5342edf657bb73066852/specs/005-tak-v2-protocol/spec.md#L49-L53).

core/takserver/src/commonTest/kotlin/org/meshtastic/core/takserver/TAKMeshIntegrationTest.kt
```diff
@@ -271,6 +271,18 @@ class TAKMeshIntegrationTest {
         assertTrue(h.serverManager.broadcasts.first().type.startsWith("a-f-"))
     }
 
+    @Test
+    fun `inbound compressed V1 packet is not broadcast to TAK clients`() = runTest(UnconfinedTestDispatcher()) {
```
jamesarich: `FakeNodeRepository` defaults to `2.8.0.0` (https://github.com/meshtastic/Meshtastic-Android/blob/c572c2689e1e551ac509410e207cc6d92c1c7c67/core/takserver/src/commonTest/kotlin/org/meshtastic/core/takserver/TAKMeshIntegrationTest.kt#L83), so this asserts the drop on exactly the firmware where dropping is wrong. Pin it to `2.7.0.0` and add the inverse - 2.8 local + compressed v1 packet still broadcasts.

core/takserver/src/commonMain/kotlin/org/meshtastic/core/takserver/TAKMeshIntegration.kt
```diff
@@ -434,14 +434,30 @@ class TAKMeshIntegration(
...
-     * v1 receive path (firmware <= 2.7.x): decode bare protobuf [TAKPacket] (no compression) from port 72 (ATAK_PLUGIN)
-     * and convert to CoT for forwarding to attached TAK clients. Kept indefinitely so users on stable 2.7.x firmware
-     * retain PLI + GeoChat interop; new typed payloads (shapes, markers, routes, etc.) still require a v2-capable radio
-     * (firmware >= 2.8.0).
+     * v1 receive path (firmware <= 2.7.x): decode bare protobuf [TAKPacket] from port 72 (ATAK_PLUGIN) and convert to
+     * CoT for forwarding to attached TAK clients. Kept indefinitely so users on stable 2.7.x firmware retain PLI +
+     * GeoChat interop; new typed payloads (shapes, markers, routes, etc.) still require a v2-capable radio (firmware >=
+     * 2.8.0).
+     *
+     * Packets flagged `is_compressed` are skipped — see the comment inline for why the radio delivers them alongside a
```
jamesarich: Only true of 2.7 - worth saying so here.

jamesarich (changes requested):
The diagnosis holds for 2.7 - https://github.com/meshtastic/firmware/blob/54e0d8d0ab2ff56b3a9ce967e53f79e49af560fb/src/modules/AtakPluginModule.cpp#L199-L204 sends the decompressed copy and never rewrites `mp`, so https://github.com/meshtastic/firmware/blob/54e0d8d0ab2ff56b3a9ce967e53f79e49af560fb/src/mesh/MeshService.cpp#L112 forwards the compressed original too. 2.8 dropped unishox2 entirely and the module is a passthrough - https://github.com/meshtastic/firmware/blob/76e45c5b44ac0dce6b5ce490659cb959e759b909/src/modules/AtakPluginModule.cpp#L5-L11. A port 72 packet from a 2.7 peer reaches a 2.8 radio's phone once, compressed, and this `return` drops that contact instead of rendering a garbled uid with a correct position.

`docs/en/user/tak.md` L97 - "mixed-firmware meshes keep working" stops being true as written - the gate keeps it true.

core/takserver/src/commonMain/kotlin/org/meshtastic/core/takserver/TAKMeshIntegration.kt
texaskst: Done in f5c6a8a — the guard is now `takPacket.is_compressed && !useTakV2()`.

You are right that the twin is a property of the local radio, not of the packet. I had reasoned about the 2.8 passthrough while writing the issue and then failed to carry it into the code, so the first version regressed exactly the case US5 exists to protect.

One edge worth your call: `useTakV2()` returns false when `firmwareVersion` is still null, so during that window a compressed packet is dropped rather than rendered. That matches the conservative default the send path already uses, and it is transient, but if you would rather bias toward never losing an inbound contact I can restrict the skip to a positively-identified 2.7.x radio instead.

core/takserver/src/commonTest/kotlin/org/meshtastic/core/takserver/TAKMeshIntegrationTest.kt
texaskst: Good catch, and it was the sharpest one — the test passed for the wrong reason. With the default of `2.8.0.0` it asserted the drop on precisely the firmware where dropping loses the contact, so it would have locked in the regression.

Now two tests:

- `compressed V1 packet is dropped on a legacy radio` — pinned to `2.7.0.0`
- `compressed V1 packet is still broadcast on a V2 radio` — `2.8.0.0`, asserts the PLI still reaches TAK clients

I checked the second one is load-bearing rather than vacuous: reverting the guard to the unconditional `return` makes it fail, and it passes with the gate.

core/takserver/src/commonMain/kotlin/org/meshtastic/core/takserver/TAKMeshIntegration.kt
texaskst: Fixed. The KDoc now says the skip applies only when the local radio is 2.7.x, and the inline comment spells out that it is that firmware which decompresses into a copy while `MeshService` forwards the original.

I scoped the note in `TAKPacketConversion` the same way, since it had the same universal-sounding wording and would have gone stale for the same reason — it now says firmware <= 2.7.x acts on `is_compressed`, and that 2.8+ dropped unishox2 for the port 78 passthrough.

core/takserver/src/commonMain/kotlin/org/meshtastic/core/takserver/TAKMeshIntegration.kt
```diff
@@ -434,14 +434,30 @@ class TAKMeshIntegration(
...
+            // compressed packet is delivered as well (AtakPluginModule::alterReceivedProtobuf allocates a copy rather
+            // than rewriting mp in place, as the transmit branch does). We therefore see every PLI twice.
+            //
+            // The compressed twin cannot be rendered here: this module has no unishox2 implementation, so its string
+            // fields decode as replacement characters and ATAK surfaces them as a second contact with an unreadable
+            // callsign, re-broadcast on every PLI and impossible to dismiss. Skip it and keep the decompressed copy,
+            // which carries identical position, team and telemetry.
+            if (takPacket.is_compressed) {
+                Logger.d { "Skip compressed V1 packet; the radio also delivers a decompressed copy" }
+                return
```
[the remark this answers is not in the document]
texaskst: Addressed in f5c6a8a. The guard is now gated on `!useTakV2()`, so a compressed port-72 packet reaching a 2.8+ radio is still decoded and broadcast, and only the genuine 2.7.x duplicate is skipped. Covered by a new test on a 2.8 local radio.

Of the two options suggested, decoding unishox2 here is the better long-term fix but needs an implementation this module does not have, and a port would have to reproduce the firmware `USX_PSET_DFLT` tables exactly. Twin-correlation would mean buffering and matching packets on a timing window, which seemed like more machinery than the problem warrants. The firmware-version gate is deterministic and needs neither, since whether a twin exists is a property of the local radio.

core/takserver/src/commonMain/kotlin/org/meshtastic/core/takserver/TAKMeshIntegration.kt
jamesarich: As-is is fine. The window is the handshake before metadata lands, so it costs one PLI at most on a 2.8 radio, and it matches the send path's rule.

jamesarich (approved):
Gate, tests and doc scoping all match. :core:takserver:jvmTest green locally at d5d3f63.

PR #7072, opened 2026-09-08
fix(firmware): source the nightly channel from nightly.meshtastic.org

jamesarich:
The nightly firmware channel read its version pointer and its artifacts from the `firmware-nightly/` folder on `meshtastic.github.io`. That folder last published on 2026-09-06, while `nightly.meshtastic.org` is current as of 2026-09-08, so the in-app nightly channel was serving a build two days behind. `web-flasher` already treats `nightly.meshtastic.org` as the nightly source (`utils/firmwareUrl.ts`, `NIGHTLY_BASE`), so this brings Android onto the same host.

The two publishers sit on different firmware commits, so the pointer and the artifacts have to come from the same host. Reading the version from one and building artifact filenames against the other resolves files that do not exist.

### 🐛 Bug Fixes

- Point the nightly version pointer at `https://nightly.meshtastic.org/index.json` instead of the `raw.githubusercontent.com` copy of `firmware-nightly/index.json`.
- Point nightly artifact downloads (`.bin`, `.uf2`, `-ota.zip`, `.mt.json`) at the same host, which serves them flat at its root.

- Add `HttpClientDefaults.NIGHTLY_BASE_URL` so the hostname has one home, and derive the nightly index URL from it.
- Replace `FirmwareRelease.artifactFolder` (a path fragment) with `FirmwareRelease.artifactBaseUrl` (a full base URL), so the nightly channel can resolve against a different host rather than a different folder on the same one. This mirrors `web-flasher`'s `getFirmwareBaseUrl`.

### 🧹 Chores

- Refresh the KDoc that named `meshtastic.github.io` as the nightly source across `core:model`, `core:network`, `core:data`, `core:database` and `core:repository`.

### Scope

Stable and alpha are unchanged. They keep resolving against the versioned `firmware-<version>/` folders on `meshtastic.github.io`; the nightly host is flat and serves no versioned folders (`nightly.meshtastic.org/firmware-2.8.0/` returns 404).

### Verification

Checked against the live hosts before writing the change:

- `nightly.meshtastic.org` serves `index.json` plus every per-target artifact at its root, as `application/json` / `application/octet-stream`.
- `HEAD` and range requests both work there, which `FirmwareFileHandler.checkUrlExists` depends on.
- The nightly index 404s before the first nightly of a series is published, which is the existing "no nightly" path and is preserved.

One infrastructure note for reviewers: the new host sets a Cloudflare edge `cache-control: max-age=14400`, against the app's 1 hour `CACHE_EXPIRATION_TIME_MS`. A freshly published nightly can therefore take up to 4 hours to surface. That is unchanged in behaviour from a stale upstream folder and is not addressed here.

### Testing Performed

Added:

- `ApiServiceTest.nightly pointer is fetched from the nightly host root` pins the exact index URL.
- `ApiServiceTest.nightly pointer returns null when nothing is published` covers the 404 path, where the body is the host's HTML error page and the status is what distinguishes "gone" from "unreachable".
- `CommonFirmwareRetrieverTest.stable release keeps using the versioned folder on meshtastic github io` is a regression guard that stable is never looked up on the nightly host.

Updated the three existing nightly cases in `CommonFirmwareRetrieverTest` to the new base URL.

Baseline run locally, all green:

```
./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile
```

Locally that exercised the `jvm` variant of `CommonFirmwareRetrieverTest`; the `androidHostTest` subclass is covered by CI.

### Follow-up, not in this PR

`web-flasher` also serves stable and alpha from `release.meshtastic.org/<version>/` (R2-backed) rather than `raw.githubusercontent.com`. I confirmed that host serves the current stable's artifacts, so moving the release channels off `raw.githubusercontent.com` is available as a separate change.

### DIFF
--- a/androidApp/src/main/kotlin/org/meshtastic/app/di/NetworkModule.kt
+++ b/androidApp/src/main/kotlin/org/meshtastic/app/di/NetworkModule.kt
@@ -39,9 +39,13 @@ import io.ktor.client.plugins.HttpTimeout
 import io.ktor.client.plugins.contentnegotiation.ContentNegotiation
 import io.ktor.client.plugins.logging.LogLevel
 import io.ktor.client.plugins.logging.Logging
-import io.ktor.client.request.url
+import io.ktor.client.request.header
+import io.ktor.http.HttpHeaders
 import io.ktor.serialization.kotlinx.json.json
 import kotlinx.serialization.json.Json
+import okhttp3.Cache
+import okhttp3.ConnectionPool
+import okhttp3.Dispatcher
 import okio.Path.Companion.toOkioPath
 import org.koin.core.annotation.Module
 import org.koin.core.annotation.Single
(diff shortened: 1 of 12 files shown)


jamesarich:
Superseded by #7073. This branch was accidentally cut from a local in-progress branch, so it also carried an unrelated `feat(network)` commit. #7073 is the same change on a branch cut from `main`, with only the one commit.

PR #7077, opened 2026-09-08
fix(settings): fetch Mesh Beacon config from remote-admin nodes

### LINKED ISSUE #7060: [Bug]: Remote Admin Beacon info does not populate
### Checklist

### Affected app version

2.8.2 Snapshot 29322184

### Affected Android version

17

### Affected phone model

Pixel 9a

### Affected node model

Heltec Wireless Tracker V2, Seeed X1

2.8.1fdb6730 nightly (and all previous firmwares for the last week lets say)

1. Directly connect to a slave node and load some mesh beacon parameters
2. Disconnect from slave node and do all further steps over remote LoRa admin
3. Use master node on latest nightly firmware to remote into slave node running latest nightly firmware.
4. Navigate to Mesh Beacon section
5. Notice remote Mesh Beacon config section shows "set your radio's region before setting up a beacon"
6. Region is already set though
7. Back out and enter remote slave LoRa config section and let it load (proving region is set)
8. Back out of remote slave LoRa config section
9. Reenter remote slave Mesh Beacon config
10. Notice config sections renders as empty (everything turned off)
11. Physically connect to slave node and load Mesh Beacon config
12. Notice Mesh Beacon config is loaded with previously set settings

jamesarich:
Over remote admin the Mesh Beacon editor asked for a region and then rendered everything off. `MESH_BEACON` was marked `refreshable = false` on the premise that no `ModuleConfigType` existed for it, so no get was ever sent, and the `ModuleConfigResponse` merge dropped `mesh_beacon` anyway. `MESHBEACON_CONFIG` is in the pinned protos and 2.8 firmware answers it.

Fixes #7060

### 🐛 Bug Fixes
- `ModuleRoute.MESH_BEACON` requests `MESHBEACON_CONFIG`; the `refreshable` flag and its early return are gone (nothing else used it)
- `loadModuleRoute` also reads the LoRa config and the primary channel for this route, which the editor's region gate and offer picker need on a remote node, gated on `supportsMeshBeacon` so firmware without the module does not sit on a get it will never answer
- `mesh_beacon` is carried through both `ModuleConfigResponse` merges (read and optimistic save)

### 🧹 Chores
- `settings-validation.md` and the editor KDoc describe the real read path. Over remote admin the picker offers the primary channel only.

### Testing Performed
`RadioConfigViewModelTest`, each pins one production line:
- `MESH_BEACON route on a remote node reads the beacon module config plus LoRa config and channel 0`
- `MESH_BEACON route skips the module get on firmware without the module but still reads LoRa and channel 0`
- `MESH_BEACON remote read completes only after all three responses land` (also pins the loading total and the channel merge)
- `ModuleConfigResponse carrying mesh_beacon is merged and survives a later response without it`
- `setModuleConfig merges mesh_beacon into state before the radio answers`

Baseline green (`spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile`). Not verified on hardware yet: on the bench the remote module list hides Mesh Beacon until the app has the remote node's metadata, which is a separate gap.

## Summary by CodeRabbit

* **Improvements**
  * Mesh Beacon settings now load through the standard settings flow, providing more reliable configuration updates.
  * The editor now loads the required region, radio, and channel information before displaying available options.
  * Remote administration now limits the offer-channel picker to the primary channel when applicable.
  * Mesh Beacon changes appear immediately in the settings view while updates are being processed.
  * Settings remain synchronized as configuration responses arrive.

### DIFF
--- a/feature/settings/settings-validation.md
+++ b/feature/settings/settings-validation.md
@@ -379,10 +379,12 @@ configuration settings screen. Constraints are sourced from two layers:
 
 ### Mesh Beacon (`ModuleConfig.MeshBeaconConfig`)
 
-Gated behind the `supportsMeshBeacon` capability. There is no `AdminMessage.ModuleConfigType`
-value for this module, so the editor reads from the connect-time config sync (no per-module
-refresh) and writes via `AdminMessage.setModuleConfig`. Flag edits are read-modify-write so
-`FLAG_LEGACY_SPLIT` and any unknown bits survive.
+Gated behind the `supportsMeshBeacon` capability. The editor reads via
+`AdminMessage.getModuleConfigRequest(MESHBEACON_CONFIG)` plus the LoRa config and channel 0
+(the region gate and the offer-channel picker need both on a remote node) and writes via
+`AdminMessage.setModuleConfig`. Flag edits are read-modify-write so `FLAG_LEGACY_SPLIT` and any
+unknown bits survive. Over remote admin only the primary channel is read, so the picker offers
+the primary only there.
 
 | Field | Type | Validation | Notes |
 |-------|------|------------|-------|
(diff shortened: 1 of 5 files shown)


jamesarich:
@coderabbitai full review

PR #7079, opened 2026-09-08
fix(messaging): keep the composer IME session stable while the node list changes

### LINKED ISSUE #6950: [Bug]: While in a channel, if the message box is tapped on to show the keyboard the keyboard will appear and disappear randomly
### Contact Details

I will reply in this issue.

### Checklist

### Affected app version

2.8.1 (29321948) fdroid

### Affected Android version

Adnroid 10

### Affected phone model

Xiaomi Redmi 4X

### Affected node model

DIY node, ESP32-C3 + SX1276

2.8.0(develop branch)

1. Start the meshtastic app
2. Go into the a channel
3. Tap on the "Message" chatbox area
4. Virtual keyboard will constantly appear and re-appear

### Actual behavior

I go into a channel and I tap on the input field for a message. After that the virtual keyboard will appear and disappear randomly(possibly when the nodes list updates or something else updates). I could only see this happen when I am using the global channel with MQTT uplink and downlink because there are no local nodes I can talk with to try and replicate this issue.

### Expected behavior

Keyboard should appear properly since the input field is in focus.

https://github.com/user-attachments/assets/7fd948af-ec7c-4d38-b951-87cbdbed34c1

### Relevant log output

```shell

```

### Additional information

This is AnySoftKeyboard but it happens with the AOSP keyboard too. It seems to me that it happens when the nodes list updates but I am not certain about this. I tested it in the latest pre-release(29322086) and the same issue exists there as well.

jamesarich:
The composer's mention `OutputTransformation` was rebuilt whenever a node's name or id changed. `BasicTextField` keys its transformed state on that instance, so the decorator node restarted the IME session on every rebuild, and on Android a restart lands as a keyboard hide followed by a show. Busy channels and 2.8 firmware churn the node table constantly, so the keyboard dropped and came back while typing.

Fixes #6950

### 🐛 Bug Fixes
- `MessageInput` keeps one `OutputTransformation` for the field's life and reads the latest mention candidates through a `rememberUpdatedState` holder, so renames still render without touching the IME session

### Testing Performed
- `MessageInputImeSessionTest.renamingAMentionedNodeWhileFocusedKeepsTheImeSessionAndRendersTheNewName` (new): renames a mentioned node while the field is focused, asserts the transformation instance is unchanged and the new name renders; fails on the old `remember(mentionCandidates)` code

Baseline green (`spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile`). Not reproduced on hardware here; the second reporter's "only on 2.8 over BLE" is not explained by this change, though every IME-restart path found goes through this transformation identity.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated mention suggestions without restarting the keyboard input session.
  * Ensured renamed mention candidates are reflected immediately while removing outdated display text.

* **Tests**
  * Added coverage verifying mention updates preserve the active keyboard session and display the latest candidate information.

### DIFF
--- a/feature/messaging/src/commonMain/kotlin/org/meshtastic/feature/messaging/Message.kt
+++ b/feature/messaging/src/commonMain/kotlin/org/meshtastic/feature/messaging/Message.kt
@@ -55,6 +55,7 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.rememberUpdatedState
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.runtime.snapshotFlow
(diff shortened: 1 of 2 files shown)


jamesarich:
@coderabbitai full review

PR #7101, opened 2026-09-09
ci(flatpak): verify the offline build post-merge, not on every PR

jamesarich:
`verify-flatpak` is not a required check and never runs in the merge queue, but its path filter covers the whole of `gradle/libs.versions.toml`, so it spent two runner-heavy jobs on nearly every Renovate PR for a signal that blocks nothing. This scopes the pre-merge run to changes in the flatpak tooling itself and moves the dependency surface to post-merge, which also closes a gap rather than only cutting cost: `main` was never verified at all.

## Why

Across the last 100 `Verify Flatpak Offline Build` runs there were 24 failures, cross-checked against the Pull Request CI conclusion on the same head SHA:

| Category | Count |
| --- | --- |
| Also red in Pull Request CI on the same SHA (compile errors, unresolvable deps, cancellations) | 22 |
| Unique to verify-flatpak | 2 |
| Real regressions that Pull Request CI missed | 0 |

Both unique failures were the `Reclaim the isolated Gradle home` step rather than the verification: `rm -rf "$RUNNER_TEMP/flatpak-gradle-home"` hitting `Directory not empty` on `caches/9.7.1/transforms` because Gradle still held the files, in both cases *after* the offline build had already printed `BUILD SUCCESSFUL`.

## 🧹 Chores

- `pull_request` now fires only on `scripts/verify-flatpak/**` and the workflow itself, where a break has no other cover and every later run would be red.
- The dependency surface it captures (`gradle/libs.versions.toml`, `desktopApp/**`, `gradle/wrapper/**`, the root build scripts) moves to `push` on `main`, plus a nightly cron for drift no path filter can see, such as a Flathub runtime bump or an upstream artifact that moved.
- Pinned both export steps to `workflow_dispatch`. They were gated on the event not being a `pull_request`, which meant dispatch-only while dispatch was the only other trigger. Left alone they would have started bundling on every `main` push and every nightly, which is the opposite of the point.
- Added the `github.repository ==` fork guard that `scheduled-updates.yml` and `scheduled-baseline.yml` already carry, so a fork does not run the nightly.
- Keyed the concurrency group on `github.event_name`. A nightly and a `main` push share `github.ref`, and `cancel-in-progress` would otherwise let one kill the other.
- Corrected `.skills/testing-ci/SKILL.md`, which still described this as a `build-flatpak-src` job inside `reusable-check.yml` gated on `run_desktop_flatpak_src`. That job no longer exists.

## Trade-off

A PR that breaks the offline build through a dependency bump now reds `main` instead of the PR. Given zero such breaks in the last 100 runs, and that `main` previously had no coverage at all, that looks like the better side of the trade. Reverting is a trigger edit.

## Testing Performed

- `actionlint` 1.7.12, the version pinned in `pull-request.yml`, clean on the changed workflow and on the whole workflow directory.
- No Gradle run. The change touches only workflow YAML and a skill doc, and spotless targets `*.kt` and `*.gradle.kts` only, so the change-type matrix puts this in the docs-only class.
- This PR modifies `.github/workflows/verify-flatpak.yml`, so it exercises the new `pull_request` filter on itself.

## Summary by CodeRabbit

- **CI Improvements**
  - Flatpak verification now runs automatically for relevant pull requests, pushes to the main branch, and nightly scheduled checks.
  - Verification covers both source generation and fully offline Flatpak builds across supported Ubuntu architectures.
  - Flatpak bundle exports and artifact uploads are available for manually triggered runs.
  - Forked repositories are protected from unsupported scheduled verification runs.

- **Documentation**
  - Updated CI guidance to reflect the new Flatpak verification workflow and runner strategy.


jamesarich:
@coderabbitai full review

PR #7115, opened 2026-09-10
refactor: construct protos through Builder for buildersOnly protobufs

jamesarich:
`protobufs` is moving its Kotlin codegen to Wire's `buildersOnly` (meshtastic/protobufs#1074) because the generated all-args constructor encodes every field in its signature: adding one proto field moves it, and a consumer compiled against a different `protobufs` version fails at runtime with `NoSuchMethodError` while the wire format stays compatible and `buf breaking` stays silent. This app is the consumer that most needs it fixed, since it tracks an unreleased protobufs snapshot while the libraries pin releases. This rewrites the 2698 affected sites.

## Changes

🛠️ **Refactoring & Architecture**
- 278 files, 2698 sites, to `X.Builder().also { wb -> … }.build()` and `x.newBuilder().also { wb -> … }.build()`.
- Always `newBuilder()` for read-modify-write, never a rebuild from fields: it carries `unknownFields` across, and a rebuild silently drops forward-compatible data from newer firmware.
- Uses `.also { wb -> }`, **not** `.apply { }`. Inside `apply` the Builder's own property names shadow outer identifiers, so `apply { this.model = model ?: "" }` reads the builder's empty default instead of the outer value and silently writes the wrong thing. `also` takes the builder as a parameter and cannot shadow.

🧹 **Chores**
- Three detekt thresholds raised, each with the cause in a comment: `allowedComplexity` 15 → 35, `LongMethod` 60 → 70, `LargeClass` 600 → 1100. detekt counts each `also { }` lambda toward complexity, so a settings screen scores one point per row it writes without gaining a branch — `MQTTConfigScreen` reads 33 against a limit of 15. Nothing there got harder to follow. Measured both ways first: detekt is clean on `main` and reports 27 on this branch, so the migration causes all of them.
- `MagicNumber` was **not** loosened. Its seven reports were numbers that had been exempt as named arguments (`ignoreNamedArgument` defaults true) until the rewrite made them assignments. Two in `CommandSenderImpl` now use the `MILLIS_PER_SECOND` constant that file already declared; the rest are Compose preview sample data, so the rule gains the same `ignoreAnnotated` list `LongMethod` already carried, plus one in-place suppression on a preview helper that is not itself annotated.
- `takpacket-sdk` pin moved, because it re-exports the proto types (see Landing).
- Rebased onto `main` at #7135 (2026-09-11). One site landed on `main` in between, `Node.toSharedContact`, converted the same way; the first-wins key handling from #7118 kept its logic and only changed shape.

### Name collisions, resolved by import rather than by name

`org.meshtastic.core.model.Channel` is a hand-written wrapper, not the proto, and `Node`, `NodeState`, `DataPacket`, `Reaction`, `RadioConfigState` and friends are hand-written state classes whose `copy()` calls are untouched even where they sit in the same expression as a proto. `EnvironmentMetrics` and `PowerMetrics` are also `@Composable` function names; two test call sites were wrongly rewritten as `Builder()` and are restored. A name match is not a proto — it is the proto only where the file imports it from `org.meshtastic.proto` or spells it fully qualified.

## Testing performed

Full baseline, run with `--rerun-tasks` so nothing could be replayed: **1844 of 1844 tasks executed, 0 FROM-CACHE**.

`spotlessCheck detekt assembleDebug kmpSmokeCompile test allTests` — all green, **8167 tests across 874 suites, 0 failures**, results timestamped after the final commit.

One lesson worth recording: `jvmTestClasses` is not a sufficient oracle. It never compiles `androidMain`, `androidHostTest`, or the `google`/`fdroid` flavor test source sets, and this branch looked finished while still hiding 22 sites there — including an import alias (`Position as ProtoPosition`), an `x?.copy(...)` whose safe call stopped covering the chain, and a `node.copy(user = node.user.copy(…))` where only the inner receiver is a proto.

After the rebase only `assembleFdroidDebug` was re-run (green); the full baseline above predates it and has not been repeated.

No UI or behavioural change, so no screenshots: the builder rewrite is construction-shape only, and `ADAPTER.encode`/`decode` are untouched.

## Landing

Draft, and **red in CI by design**. The pins are `2.8.1-buildersonly-SNAPSHOT` and `takpacket-sdk 0.9.2-buildersonly-SNAPSHOT`, which exist only in a local `~/.m2`; both are placeholders for real tags.

This is the last of the five to land, because `takpacket-sdk` re-exports the proto types and this app consumes its published jar — the old TAK artifact would fail at runtime, not at compile time. Order: protobufs tags → `TAKPacket-SDK` releases → this bumps both pins.

Related: meshtastic/protobufs#1074, meshtastic/TAKPacket-SDK#141, meshtastic/meshtastic-sdk#125.


jamesarich:
Rebased 54 commits forward onto `main` and squashed to one commit, so the diff reads as construction-syntax-only. Baseline is green.

**Baseline:** `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` — all satisfied, **8,345 tests across 893 suites, 0 failures, 0 skipped**.

**Rebase conflicts, and how each was resolved:**

- `SwitchingChannelSetDataSource.kt` — `main` added the `reconcile` pass and `lastReconciled`. Kept `main`'s semantics and construct through the Builder; the branch's side was stale text, not a competing behaviour.
- `NodeTest.kt` — kept `main`'s two new `heardOnCurrentLora` tests, built `Position` through the Builder.
- `MigrationTest.kt` — `main` deleted it in #7162, so the deletion wins; the branch's work on it is dead.
- `gradle/libs.versions.toml` — both pins carry builders.

**~80 sites arrived from `main` un-converted** and only the private constructor caught them, across 9 files in `core/data`, `core/database`, `core/model`, `core/network`, `core/ui` and `desktopApp`. Worth noting two that would not have been obvious:

- `desktopApp/…/DesktopProtoDataStoreModule.kt` — four *production* DataStore default-value lambdas (`{ LocalConfig() }` → `{ LocalConfig.Builder().build() }`).
- Two `copy()` read-modify-writes now `newBuilder()`. One of them failed with a completely misleading error — Kotlin resolved `longFast.copy(region = …)` to the stdlib `Map.Entry.copy()` extension and complained about inferring `K`/`V`, which reads like a generics bug rather than a missing proto method.

**Correctness checks the compiler cannot do**, run over all 2,185 diff hunks:
- **0** `.apply { }` on a builder (an implicit receiver lets a one-word field name bind to the enclosing scope and silently set the wrong thing).
- **0** `copy()` sites converted to a fresh `Builder()` — all 368 became `newBuilder()`, so `unknownFields` from a peer survive. A fresh builder there would compile perfectly and silently drop them.
- No `decoded`/`encrypted` oneof pair assignment anywhere, so no oneof can be clobbered by assignment order.

**CI will be red until two things land, neither of them this diff.** `takpacket-sdk` is pinned to a local `0.9.2-buildersonly-SNAPSHOT` that exists only in `~/.m2`, so resolution fails until TAKPacket-SDK#141 merges and publishes `0.9.2-SNAPSHOT`. The `protobufs` pin wants the snapshot after meshtastic/protobufs#1098. This repo builds on JDK 25, so unlike the library repos it is not affected by the class-file-69 problem #1098 fixes — only by the coordinate.

PR #7120, opened 2026-09-10
refactor(prefs): keep each surface's filters in one state object

jamesarich:
`UiPrefs` exposed nine separate node filter flows and `MapPrefs` thirteen map ones, so every reader had to recombine them downstream. The typed `combine` overloads cap at five, which the map had already outgrown: it carried two intermediate combines, a `NodeFilters` box that existed only "so they fit one combine", and a hand duplicated initial value.

The store already holds these together. They are now handed back together.

No behaviour change. This is groundwork for adding filters in #7117, where a new one costs one property and one key instead of touching seven files.

## 🛠️ Refactoring & Architecture

- `nodeFilters: StateFlow<NodeFilterPrefs>` and `mapFilters: StateFlow<MapFilterPrefs>` replace the per toggle flows. Writes go through `updateNodeFilters` / `updateMapFilters`, applying a pure transform inside DataStore's own transaction, which is the lost update guard `updateHiddenLayerUrls` already used.
- Node list `combine` drops from five flows plus a nested combine to three. The map drops from eleven flows and two intermediate combines to a single `map`, and loses both the `NodeFilters` box and the duplicated initial value.
- `NodeFilterTextField` took twenty parameters and now takes the `NodeFilterToggles` it was already constructing internally. That retires three detekt baseline entries (`ComposableParamOrder`, `LambdaParameterEventTrailing`, `ParameterNaming`) rather than adding to them.

## 🐛 Bug Fixes

- `toggleRoleExcluded` decoded the stored role names into enum values, toggled, then re-encoded. `decodeExcludedRoles` deliberately drops names the current protobufs do not define, so every role toggle silently discarded any role excluded by a newer build, which is the exact data loss the "names rather than ordinals" comment exists to prevent. It now toggles by name on the stored set. Pre-existing, but the refactor is what puts the string set in hand.

## Testing Performed

- `BaseMapViewModelTest`: new `toggling a role keeps role names this build does not know` excludes an unknown role name, toggles ROUTER on and off, and asserts the unknown name survives both. The `updateMapFilters` mock now applies the transform so toggles are observable. Eleven `every { }` stubs collapse to one.
- `NodeListViewModelTest`: seven per toggle stubs collapse to one `nodeFilters` stub.
- `MapNodePolicyTest`, `MapFilterSheetTest`, `GetFilteredNodesUseCaseTest` updated for the new shapes.
- Baseline green: `spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile`.

## Summary by CodeRabbit

- **Enhancements**
  - Consolidated node-list and map filtering preferences into unified filter settings.
  - Preserved existing filtering behavior for favorites, visibility, connectivity, ignored nodes, MQTT, unheard nodes, and role exclusions.
  - Map role exclusions continue to preserve names not recognized by current device definitions.

- **Bug Fixes**
  - Improved consistency when updating multiple filter settings by applying changes atomically.

### DIFF
--- a/core/prefs/src/commonMain/kotlin/org/meshtastic/core/prefs/map/MapPrefsImpl.kt
+++ b/core/prefs/src/commonMain/kotlin/org/meshtastic/core/prefs/map/MapPrefsImpl.kt
@@ -16,6 +16,8 @@
  */
 package org.meshtastic.core.prefs.map
 
+import androidx.datastore.preferences.core.MutablePreferences
+import androidx.datastore.preferences.core.Preferences
 import androidx.datastore.preferences.core.booleanPreferencesKey
 import androidx.datastore.preferences.core.doublePreferencesKey
 import androidx.datastore.preferences.core.edit
@@ -34,6 +36,7 @@ import org.koin.core.annotation.Single
 import org.meshtastic.core.di.CoroutineDispatchers
 import org.meshtastic.core.prefs.di.MapDataStore
 import org.meshtastic.core.repository.MapCameraPosition
+import org.meshtastic.core.repository.MapFilterPrefs
 import org.meshtastic.core.repository.MapPrefs
 
 @Single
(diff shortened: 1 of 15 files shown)


jamesarich:
@coderabbitai full review

PR #7125, opened 2026-09-10
docs: defer prose style to the design standards

jamesarich:
Section 11 of the [Meshtastic Client Design Standards](https://github.com/meshtastic/design/blob/master/standards/meshtastic_design_standards_v1_5.md) landed in meshtastic/design#150 and is now the org-wide style guide for documentation prose. This repo has carried its own guide since August. Two guides saying the same things in slightly different words is how they drift, so this one steps back to what Section 11 does not cover.

11.1 says client repos that keep their own guide keep it in agreement with Section 11 rather than overriding it. This is that agreement.

### 🧹 Chores

- `docs/en/developer/documentation-style.md` points at Section 11 and drops everything it now covers: voice, plain language, terminology, page structure, instructions, cross-platform coverage, code and CLI examples, media, accessibility, when to use an admonition, units, translation, versioning and in-product text. 160 lines to 102.
- What stays is what Section 11 does not decide: this repo's frontmatter, `DocBundleLoader` registration, screenshot paths and link mechanics; the emoji admonition form; the spaced-arrow menu path; and eight prose rules (en-US spelling, requirement words, inclusive language, Oxford comma, spaced em dashes, number style, contractions, `above`/`below`).
- Surviving rules keep their IDs so old review comments still resolve. Retired IDs are not reused. Nothing else in the repo cites a rule ID, so there are no dangling references.
- `contributing.md` points prose at Section 11 first. `developer.md` gets a What's New entry.
- The `DocBundleLoader` entry for the page has its `charCount` updated to what `generateDocsBundle` computes and gains a `standards` keyword.

### Two rules change, neither as a sweep

**Headings are sentence case** (11.5, and a quick check in 11.16). The old HEAD-1 mandated Title Case because re-casing invalidates translation memory across 40+ locales. That cost is real, so the rule changes and the corpus does not: headings re-case as pages are touched. About 291 of 446 headings read Title Case today. Frontmatter `title` values and the literal `## Related Topics` heading are shared nav labels across 22+ pages, so those stay until they all change together, and the page says so.

**The hardware is a node** (11.4). This repo used *radio* for the hardware to disambiguate it from the phone, and 11.4 rules *radio* out, reserving it for the LoRa transceiver. Roughly 300 uses remain. New writing prefers *node*, and *radio* goes when a page is edited for another reason. *Phone* stays as the Android handset, which Section 11 has no word for.

Wi-Fi needed nothing: 11.4 adopted the dash, which is what the word list already said.

### Testing Performed

No code paths changed. Docs gates run locally, all green:

- `node scripts/validate-doc-links.js docs/en`
- `node scripts/check-doc-coverage.js .`
- `node scripts/check-doc-aliases.js .`
- `node scripts/check-doc-freshness.js docs/en --max-age-days=180`
- `./gradlew :feature:docs:spotlessCheck :feature:docs:detekt generateDocsBundle validateDocsBundle`
- Full pre-push baseline: `./gradlew spotlessCheck detekt assembleDebug test allTests`

The rewritten page was also checked against its own surviving rules and against the 11.16 quick checks.

## Summary by CodeRabbit

- **Documentation**
  - Updated the developer documentation style guide to align with Section 11 of the Meshtastic Client Design Standards.
  - Clarified remaining repository-specific writing rules, including admonitions and menu paths.
  - Added guidance for sentence-case headings, terminology updates, and quick checks for new pages.
  - Updated contributing guidance and documentation metadata.
  - Improved discoverability and accuracy of the style guide in the in-app documentation.

### DIFF
--- a/docs/en/developer.md
+++ b/docs/en/developer.md
@@ -30,6 +30,8 @@ Things that trip up first-time contributors — check these before requesting re
 Keep the last 5–8 entries and trim older ones from the bottom.
 -->
 
+**September 2026** — [Documentation Style](developer/documentation-style) — Section 11 of the Meshtastic design standards is now the style guide for `docs/en/`; this page keeps only the repository mechanics, the in-app renderer's admonition form, and the prose rules the standards leave open.
+
 **August 2026** — [Documentation Style](developer/documentation-style) — New page: the house style guide for `docs/en/` prose, with rule IDs, a project word list, and the reasoning behind each convention.
 
 **August 2026** — Map tile sources are one shared catalogue in `feature/map` (`MapTileCatalogue`, `RasterTileSpec`), so both flavors draw the same raster base maps and overlays from one definition.
--- a/docs/en/developer/contributing.md
+++ b/docs/en/developer/contributing.md
@@ -2,7 +2,7 @@
 title: Contributing
 parent: Developer Guide
 nav_order: 8
-last_updated: 2026-08-29
+last_updated: 2026-09-10
 description: Branch naming, commit style, PR workflow, and the verification gates a change must pass before merge.
 aliases:
   - contributing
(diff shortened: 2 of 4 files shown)


rcarteraz (approved):

rcarteraz:
@jamesarich I think my writing missed the prose rules you had. Do you think these should be incorporated into the standards?

jamesarich:
> @jamesarich I think my writing missed the prose rules you had. Do you think these should be incorporated into the standards?

Might not be a bad idea - it does have some decent guidance.

rcarteraz:
> > @jamesarich I think my writing missed the prose rules you had. Do you think these should be incorporated into the standards?
> 
> Might not be a bad idea - it does have some decent guidance.

Okay, I'll work on an update. I do like some of the things you have in there.

Issue #1079, opened 2024-06-08
2.3.11 Map Zoom doesn't Save


AlphaEcho7:
Map Zoom is resetting after closing the App

Edit:
Also the new sorting Feature and "include unknown" Not saving

On Android 12 & 14

PR #7138, opened 2026-09-11
fix(node): stop presenting MQTT-only nodes as unheard on the current LoRa

jamesarich:
Firmware [#11811](https://github.com/meshtastic/firmware/pull/11811) is the first build that actually populates `NodeInfo.heard_on_current_lora`, and reading it against a real mesh turns up one case the field does not mean what the UI claims of it.

The radio sets the mark only when a packet arrives over RF and not via MQTT. So a node that only ever reaches us through an uplink reports `false` for the entire life of the entry. On an MQTT-uplinked mesh that meant every one of those nodes wore the unreachable badge, was hidden by the unheard filter, and was offered for bulk removal — where deleting it achieves nothing, because the next uplinked packet brings it straight back.

That PR also changed the mechanism after review. The sweep this feature was built against is gone: bit 11 is now a sticky "heard over RF", bits 12..23 hold a fingerprint of the LoRa slot the node was heard on, and `heard_on_current_lora` is derived by comparing that against the slot the radio is committed to. Nothing is cleared when the config changes, so leaving a configuration and returning to it restores the previous answers by itself. Several of our comments and one user-facing string still describe the old behaviour.

## 🐛 Bug Fixes

- **MQTT-only nodes are no longer presented as unreachable.** New derived `Node.isUnheardOnCurrentLora` (`!heardOnCurrentLora && !viaMqtt`) replaces the raw flag at all three consumers: the `StatusAwareLastHeard` badge (`NodeItem`, `NodeItemCompact`), the hide-unheard filter (`GetFilteredNodesUseCase`), and the bulk-removal offer (`NodeListScreen`). One property so the three cannot drift apart.

## 🧹 Chores

- **String.** `node_not_heard_on_current_lora`: "Not heard since you changed settings" → "Not heard on your current LoRa settings". The old wording names a cause that no longer holds — under the derived scheme you can change settings and keep the marks, or change nothing and read unheard after a restore.
- **Docs.** `Capabilities`, `NodeEntity`, `Node`, `NodeItem` and the unheard preview all described the sweep.

## Notes

Shared contacts need no guard. Firmware's `addFromContact` marks a contact favourite to keep it off the eviction list, and the removal offer already exempts favourites. The two fallbacks that skip the favourite (role `CLIENT_BASE`, and a protected-cap refusal) call `stampContactHeardNow` instead, so a contact on a `CLIENT_BASE` can still be offered — narrow enough to leave alone.

`Capabilities.supportsHeardOnCurrentLora` is untouched and still gated at `UNRELEASED`. Nothing here ships behaviour until the firmware release carrying #11811 lands.

The proto doc comment is wrong in the same way and is fixed separately in protobufs [#1079](https://github.com/meshtastic/protobufs/pull/1079).

## Testing Performed

Added to `commonTest`:

- `NodeTest > isUnheardOnCurrentLora is true only for a node heard over RF on another config`
- `NodeTest > isUnheardOnCurrentLora excludes MQTT nodes the radio never heard over RF`
- `GetFilteredNodesUseCaseTest > the unheard filter drops a node the radio can no longer reach`
- `GetFilteredNodesUseCaseTest > the unheard filter keeps an MQTT node the radio never heard over RF`

Full baseline green: `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile`, BUILD SUCCESSFUL in 19m59s, 1856 tasks, no failures.

No screenshot changes: the unheard previews use a node with `viaMqtt = false`, so the badge still renders exactly as before.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved node status indicators to distinguish devices not heard over the current LoRa settings from nodes reachable through MQTT.
  - Corrected unheard markers, filters, and cleanup selections to retain MQTT devices, favorites, and the connected node while targeting only applicable nodes.

- **Documentation**
  - Clarified how current LoRa hearing status is determined and how it relates to node reachability.
  - Updated “unheard” messaging to reference the current LoRa settings.

### DIFF
--- a/core/database/src/commonMain/kotlin/org/meshtastic/core/database/entity/NodeEntity.kt
+++ b/core/database/src/commonMain/kotlin/org/meshtastic/core/database/entity/NodeEntity.kt
@@ -163,9 +163,11 @@ data class NodeEntity(
     /** True when this node signs its broadcasts via XEdDSA (NodeInfo.has_xeddsa_signed). */
     @ColumnInfo(name = "has_xeddsa_signed", defaultValue = "0") var signsPackets: Boolean = false,
     /**
-     * True when the radio has heard this node over RF since its current LoRa config took effect
-     * (NodeInfo.heard_on_current_lora). Defaults true so nodes stored before this column existed, and nodes from
-     * firmware that does not report it, are never shown as unheard.
+     * True when the radio has heard this node over RF on the LoRa configuration it is using now
+     * (NodeInfo.heard_on_current_lora). The radio derives this from the slot each node was heard on rather than
+     * clearing it on a config change, so returning to a configuration restores the previous answers. Defaults true so
+     * nodes stored before this column existed, and nodes from firmware that does not report it, are never shown as
+     * unheard.
      */
     @ColumnInfo(name = "heard_on_current_lora", defaultValue = "1") var heardOnCurrentLora: Boolean = true,
     /**
--- a/core/model/src/commonMain/kotlin/org/meshtastic/core/model/Capabilities.kt
+++ b/core/model/src/commonMain/kotlin/org/meshtastic/core/model/Capabilities.kt
@@ -98,9 +98,9 @@ data class Capabilities(val firmwareVersion: String?, internal val forceEnableAl
     val supportsMeshBeacon = atLeast(V2_8_0)
 
     /**
-     * Whether the node reports [NodeInfo.heard_on_current_lora] - whether it has heard each node since its current LoRa
-     * config took effect. Gated to [UNRELEASED] until the firmware side ships. Older firmware never sends the field,
-     * and a proto3 bool defaults to false, so an ungated read marks every node as unheard.
+     * Whether the node reports [NodeInfo.heard_on_current_lora] - whether it has heard each node over RF on the LoRa
+     * configuration it is using now. Gated to [UNRELEASED] until the firmware side ships. Older firmware never sends
+     * the field, and a proto3 bool defaults to false, so an ungated read marks every node as unheard.
      *
      * Deliberately outside [forceEnableAll]: every other capability being wrong in a debug build shows a UI the
      * firmware ignores, but this one being wrong persists false into the node DB and offers real nodes for removal.
--- a/core/model/src/commonMain/kotlin/org/meshtastic/core/model/Node.kt
+++ b/core/model/src/commonMain/kotlin/org/meshtastic/core/model/Node.kt
@@ -89,13 +89,19 @@ data class Node(
     val capabilities: Capabilities by lazy { Capabilities(metadata?.firmware_version) }
 
     /**
-     * True when this node is one the radio can no longer reach on its current LoRa configuration: the radio has heard
-     * it over RF at some point, but not on the configuration in force now.
+     * True when the radio does not report having heard this node over RF on the LoRa configuration in force now, and it
+     * is not an [viaMqtt] node.
      *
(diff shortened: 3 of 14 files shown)


jamesarich:
Both warnings addressed in `1cf6514`, and one pushed back on.

**Merge risk: inaccurate explanation on the removal offer.** Correct, and I'd missed it. I fixed the row marker string but not the banner above the offer: `nodes_unheard_banner` and `nodes_unheard_banner_one` both still read "not heard since you changed settings". Both now say "not heard on your current LoRa settings", matching the marker.

**Regression coverage for the changed consumers.** Added `NodeItemUnheardMarkerTest`. It drives the real `NodeItem` and `NodeItemCompact` rather than `StatusAwareLastHeard` directly, since the wiring is the part that can regress, and follows the existing `setNodeItem`/`setNodeItemCompact` shape from `NodeItemZeroMetricsTest`.

I checked it actually discriminates rather than assuming it. Reverting both rows to `isThisNode || thatNode.heardOnCurrentLora`:

```
nodeItem_withholdsTheMarkerFromAnMqttNode           FAILED
nodeItemCompact_withholdsTheMarkerFromAnMqttNode    FAILED
nodeItem_marksANodeHeardOverRfOnAnotherConfig       PASSED
nodeItemCompact_marksANodeHeardOverRfOnAnotherConfig PASSED
```

**Keeping `the unheard filter drops a node the radio can no longer reach`.** You're right that it doesn't discriminate this change - it passes against the old implementation too. I'm keeping it anyway: it pins the behaviour the change has to preserve. The fix narrows what the filter drops, and the obvious way to get that wrong is to narrow it too far. Without this case, a regression that stopped filtering anything would go uncaught, because the MQTT test only proves the exclusion. The pair is the coverage; either alone has a blind spot.

Full baseline green at `1cf6514`: `spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile`, BUILD SUCCESSFUL in 12m35s, no failures.

jamesarich:
@coderabbitai full review

jamesarich:
Both remaining points addressed in `69610df` and `5e8a3af`.

**Merge risk: the doc promised a prior-RF hear.** Right, and it was my error. `isUnheardOnCurrentLora`'s KDoc said the radio "has heard it over RF at some point, but not on the configuration in force now". `!heardOnCurrentLora && !viaMqtt` cannot establish that. `heard_on_current_lora` is a single bool and firmware does not expose its `HAS_RF_HEAR` bit separately, so a node heard under different settings and a node never heard at all are indistinguishable from here - both read true. The doc now says exactly that, and points callers needing the distinction at favourites, which is what firmware marks a node added as a shared contact.

I also renamed three tests that encoded the same overclaim in their names.

**Regression coverage for the bulk-removal offer.** Extracted the inline `unheardNodes` computation as `selectRemovableUnheardNodes` and covered each exemption separately, since this is the consumer that deletes:

```
offersANodeNotHeardOnTheConfigInForceNow      PASSED
neverOffersAnMqttNode                         PASSED
neverOffersAFavourite                         PASSED
neverOffersTheConnectedRadio                  PASSED
offersNothingUntilTheLocalNodeNumberIsKnown   PASSED
```

`neverOffersAFavourite` is the one that matters for contacts: firmware marks a shared contact favourite, so the offer already exempted them, and now that is pinned rather than incidental.

Full baseline green at `69610df`: BUILD SUCCESSFUL in 13m10s, no failures. Renames re-verified separately.

jamesarich:
@coderabbitai full review

PR #7165, opened 2026-09-15
docs: point design standards references at the index, and retire the prose rules it now carries

jamesarich:
## Why

Two consequences of the design repo's September standards work, both of which leave this repo pointing at something that no longer holds.

**The canonical design standards reference did not resolve.** The constitution (principle V), the spec template's UI checklist, and the design-standards skill all named the raw URL for `standards/meshtastic_design_standards_latest.md`. That file is a symlink, and GitHub serves a symlink as its target's *filename*, so raw and blob return 35 bytes with HTTP 200:

```shell
$ curl -sSfL https://raw.githubusercontent.com/meshtastic/design/refs/heads/master/standards/meshtastic_design_standards_latest.md
meshtastic_design_standards_v1_5.md
```

The constitution requires new screens to be reviewed against the standards before merge. Anything obeying that fetched no standards and got no error. [design#155](https://github.com/meshtastic/design/pull/155) added `standards/` as an index that always names the current version, for exactly this reason.

**Seven of our eight local prose rules now live upstream.** [design#154](https://github.com/meshtastic/design/pull/154) moved them into Section 11, where they read the same. `documentation-style.md` already says a local rule contradicting Section 11 is a bug in that page; a local rule duplicating it is the same bug one step earlier — two copies of a rule are two things to keep in agreement.

## Changes

🧹 **Chores (docs)**
- Constitution principle V, `spec-template.md`, and `.skills/design-standards/SKILL.md` point at `https://github.com/meshtastic/design/tree/master/standards`. The constitution and the skill also state why the symlink can't be fetched over HTTP.
- The skill claimed `meshtastic_design_standards_v1_4.md` "at time of writing" — two versions of drift, which is the staleness the index exists to prevent. Removed rather than re-pinned.
- `documentation-style.md` retires VOICE-2 → 11.2; LANG-3, LANG-4, LANG-6 and the Oxford comma half of LANG-7 → 11.3; LANG-1 → 11.4; LANG-9 → 11.5; LANG-8 → 11.12. IDs are not reused, and nothing else in the repo cites one. LANG-7 keeps the spaced em dash, the one rule Section 11 does not mention anywhere.
- Its standards link was a pinned `..._v1_5.md` blob; now the index.

Constitution 1.3.6 → **1.4.0**. MINOR rather than PATCH: principle V gains a prohibition (`MUST NOT` fetch the symlink over HTTP), and the versioning policy reserves PATCH for clarifications. Per the Amendment Procedure, `.skills/speckit/SKILL.md` follows to v1.4.0 in the same commit; `AGENTS.md` and `checklist-template.md` name no URL and the principle count is still 7, so neither changes.

**Nothing about how anyone writes changes** — the prose rules just have one home instead of two.

## Testing Performed

Docs-only; no Kotlin, no resources, and no page added or removed, so `DocBundleLoader.kt` is untouched. All four `docs-quality.yml` gates run locally and pass:

| Gate | Result |
|---|---|
| `validate-doc-links.js docs/en` | PASSED — all internal links and images valid |
| `check-doc-coverage.js .` | 11/11 required pages (100%) |
| `check-doc-aliases.js .` | PASSED — every frontmatter alias registered |
| `check-doc-freshness.js docs/en --max-age-days=180` | All 32 pages fresh |

The full Gradle baseline was not run: this diff touches no compiled source, and the in-app copy under `feature/docs/src/commonMain/composeResources/files/docs/` is gitignored and regenerated by `syncDocsToComposeResources`.

Checked that the new prose obeys the section it defers to, including the 11.2 ban on `now`/`simply` and 11.16.

## Summary by CodeRabbit

* **Documentation**
  * Updated design standards guidance to reference the current upstream standards index and clarify how to access the specification.
  * Revised the documentation style guide to reflect the latest prose requirements and standards structure.
  * Consolidated remaining prose guidance around spaced em dashes and documented which former rules are now covered by relevant standards sections.
  * Updated the documented constitution version from v1.3.6 to v1.4.0 while retaining its seven principles.


rcarteraz (approved):

jamesarich:
Both CodeRabbit findings were valid. Fixed in the force-push, and one of them went further than the review suggested.

**Record the prohibition as a requirement change** — correct, and it understated the problem. Principle V didn't just gain wording; it gained a `MUST NOT`. The versioning policy reserves PATCH for "clarifications, wording fixes, or non-semantic refinements" and MINOR for "materially expanded guidance", so the bump itself was wrong, not only the summary. Now **1.3.6 → 1.4.0**, with the report naming the new requirement and saying why it is MINOR.

**Synchronize all constitution consumers** — correct for `.skills/speckit/SKILL.md`, which declared v1.3.6 while the constitution declared a newer version. Amendment Procedure step 3 requires it in the same commit; it now reads v1.4.0. Checked the other consumers it named rather than assuming:

- `AGENTS.md` — no design-standards URL, and the principle count is unchanged at 7. Nothing to sync.
- `.specify/templates/checklist-template.md` — CHK005 references the design standards but names no URL. Nothing to sync.

Both are now recorded in the sync report as checked, so the next amendment doesn't have to re-derive it.

**Not doing: the raw URL in `specs/`.** Five tracked specs under `specs/` still link `meshtastic_design_standards_latest.md` (`005-tak-v2-protocol` spec and tasks, `20260513-075218-lockdown-mode`, `20260520-153449-node-list-context-menu`, `20260601-074653-air-quality-telemetry`). Those are point-in-time records of features that have shipped, describing what was true when they were written. Editing them to fix a link rewrites history rather than fixing anything a reader acts on — the live paths a contributor actually follows are the constitution, the spec template, and the skill, and all three now point at the index. `spec-template.md` is fixed, so no new spec inherits the bad URL.

For anyone finding this later: the URL returns HTTP 200 and 35 bytes (`meshtastic_design_standards_v1_5.md`), because GitHub serves a symlink as its target's filename. It fails silently, which is what made it worth chasing across four repos.

PR #7180, opened 2026-09-15
chore(deps): move Compose to the 1.13 alpha train for the lookahead crash

jamesarich:
`LookaheadDelegate has not been measured yet when measureResult is requested.` is
crashing the node list and the message list on 2.8.2-open. Datadog RUM puts it at
**8.8% of users on 2.8.2-open.2** (versionCode 29322268) against **0.010%** on 2.8.1
production — 261 of that build's 284 crashes. Strip that one issue out and the build
is actually cleaner than production, so it accounts for the entire regression.

Both screens are list-detail panes, so they render inside `ThreePaneScaffold`'s own
`LookaheadScope`. The `LazyLayoutCacheWindow` added in #7093 prefetches items outside
the normal pass; the main placement pass then reaches an item the lookahead pass never
measured, and `LookaheadDelegate` throws. `LazyListMeasuredItem.place` →
`MeasurePassDelegate.placeAt` → `LookaheadPassDelegate` with no `measureResult`.

compose-foundation **1.13.0-alpha02** fixes it upstream — *"only keep around cached
items when not in a lookahead pass"* (`isKeepAroundDuringLookaheadDisabled`, `I110e9`).
Taking the alpha means the lists **keep** the prefetch win from #7093 instead of giving
it up.

### 🧹 Chores

- `compose-multiplatform` 1.12.0 → 1.13.0-alpha01
- `compose-multiplatform-material3` 1.12.0-alpha03 → 1.13.0-alpha01
- `androidx-compose-bom-aligned` 1.12.1 → 1.13.0-alpha02

Three version choices worth reviewing rather than skimming:

- **The AndroidX force is alpha02, not a later alpha, on purpose.** alpha03 raises the
  AAR metadata key `minCompileMinorSdk` from 0 to 1, so it requires `compileSdk` 37.1+;
  alpha02 keeps 0 and still carries the fix. `chore/compile-sdk-37.2` moves compileSdk
  to 37.2 and unblocks alpha03+ separately.
- **compose-multiplatform moves too, not just AndroidX.** `AndroidCompose.kt`'s
  `resolutionStrategy` force covers the Android target only, and both crashing lists are
  `commonMain` — an Android-only bump would have fixed the crash on the one target we
  happened to measure it on, leaving desktop and iOS on the old keep-around behaviour.
- **material3 rides its own AndroidX line** but has to move in step, because its 1.13
  requires the two source changes below.

Two material3 1.13 migrations, both the same shape — a member that needed a receiver
scope became something reached directly. Neither changes behaviour.

- **`ExposedDropdownMenu`** was a member of `ExposedDropdownMenuBoxScope` and is now a
  top-level extension on it, so the four call sites need an explicit import. The member
  survives as `@Deprecated(DeprecationLevel.HIDDEN)` for binary compatibility. This
  cannot be staged ahead of the bump — on 1.12 there is no top-level symbol to import,
  so the version change and the imports have to land together.
- **`SearchBarScrollBehavior`**'s `fun Modifier.searchBarScrollBehavior()` became the
  property `searchBarScrollBehaviorModifier`, so `DocsSearchBar` now `.then()`s it —
  the same shape `SearchBar` uses internally.

### Verification

No tests were added or changed; this is a dependency move plus the API migrations it
forces. Full baseline green on `ec0aa16b6d`: `spotlessApply spotlessCheck detekt
assembleDebug test allTests kmpSmokeCompile` — 224 tests passed, 0 detekt violations,
spotless modified nothing.

The crash fix itself is **not** verified on a device. The mechanism is inferred from the
stack and from the upstream release note, and the correlation is strong (the two screens
that got a cache window in #7093 are exactly the two crashing views), but nobody has
reproduced the crash and watched it stop. The proof will be RUM on the next open build.

### Reviewer notes

- **This ships three alpha Compose artifacts on a release line.** That is the real risk
  here and the reason it is a draft. The alternative branch
  `fix/lookahead-delegate-crash` instead removes the cache windows — it fixes the crash
  with no alpha dependency, at the cost of the prefetch win. **The two are alternatives;
  only one should merge.**
- `ListItem`'s `headlineContent` became a trailing `content` lambda in 1.13. Six call
  sites in `feature:docs` and `feature:map` now emit deprecation warnings. Left alone
  here to keep this diff to the crash fix.
- `NodeDetailCompassLifecycleTest` failed once during verification, then passed 3/3 with
  `--rerun-tasks` here and 3/3 in a control worktree with no Compose change. It is a
  documented pre-existing flake (43–66% failure on CI), not a regression from this
  change — but it never reproduced, so that is non-reproduction rather than proof.


jamesarich:
CI findings on `ec0aa16b6d`, including a re-run of the failed jobs on the same commit.

| job | first run | re-run |
|---|---|---|
| test-shards (shard-feature) | fail | **pass** |
| Build Desktop Debug ×4 | fail | **fail** |
| screenshot-check | fail | **fail** |

### Blocker: maplibre-compose is not compatible with CMP 1.13.0-alpha01

`:desktopApp:proguardReleaseJars` fails on all four platforms with:

```
Warning: org.maplibre.compose.map.SnapshotGraphicsContextKt$withSnapshotGraphicsContext$2:
  can't find referenced method 'SkiaGraphicsContext(boolean,int,kotlin.jvm.internal.DefaultConstructorMarker)'
  in program class androidx.compose.ui.graphics.SkiaGraphicsContext
Warning: ... can't find referenced method 'void dispose()' in ... SkiaGraphicsContext
Warning: there were 2 unresolved references to program class members.
```

maplibre-compose calls `androidx.compose.ui.graphics.SkiaGraphicsContext`'s constructor and
`dispose()`; both are gone in Compose 1.13.0-alpha01. **ProGuard only reports it — the methods
genuinely no longer exist, so the desktop map would `NoSuchMethodError` at runtime regardless of
the shrinker.** This is a correctness problem, not a ProGuard config one.

`0.17.0` is the newest maplibre-compose published and it declares CMP `1.12.0`. There is no version
to bump to, so this cannot be resolved inside this repo.

### Not blockers

- **shard-feature was a flake.** `QuickChatViewModelTest > addQuickChatAction delegates to
  repository` failed with `IllegalStateException at TestMainDispatcherJvm.kt:45` / `MainDispatchers.kt:77`
  — entirely inside `kotlinx-coroutines-test`'s `Dispatchers.setMain`/`resetMain` machinery, which is
  global JVM state shared across test classes. No Compose code on that path. It passed 3/3 locally on
  this branch, 3/3 on a control branch, and passed on the CI re-run of this same commit.
- **The Skiko warning is noise.** `org.jetbrains.skiko:skiko:0.150.1 -> 0.152.0-alpha02` comes from
  `io.coil-kt.coil3:coil-core-jvm:3.6.2`, and Gradle resolves it upward. Confirmed by
  `dependencyInsight`; `material3.adaptive` is *not* among the requesters.
- **screenshot-check** failing is expected of a Compose version bump and would just need reference
  regeneration.

### Recommendation

Park this until maplibre-compose ships a release built against CMP 1.13, and take the crash fix
that has no alpha dependency instead. Keeping the `LazyLayoutCacheWindow` prefetch win is not worth
holding back a fix for a crash hitting 8.8% of users on 2.8.2-open.2, and the win can come back
once the ecosystem catches up.

Also unaddressed here: `ListItem`'s `headlineContent` became a trailing `content` lambda in 1.13 —
seven call sites across `feature:docs`, `feature:map` and `feature:node` currently emit deprecation
warnings.

jamesarich:
The 1.13 alpha train is not the way in. Pull Request CI went red on all four Build Desktop Debug jobs plus screenshot-check, so the dependency move costs more than the crash fix buys.

#7185 closes the same crash by dropping the two cache windows that trip it — two files, six lines, no dependency change. Closing in favour of that.

PR #7251, opened 2026-09-19
docs(navigation): describe the deep-link matchers that replaced the when block

jamesarich:
## Why

#7246 replaced the hand-rolled `when` dispatcher in `DeepLinkRouter` with Navigation 3's
`UriDeepLinkMatcher` + `withBackStack`, but the developer page still tells you to add a
`when` branch to `route()`. Anyone following the "Adding a Deep Link" steps today lands on
a structure that no longer exists.

It also changed behaviour that nothing documented: patterns are anchored, so a path the
matchers do not model returns null instead of falling back to the family root.
`/firmware/anything-else` used to open the firmware screen and no longer does.

## 🧹 Chores

* Describe the matcher pipeline in the Deep Link Router section, and note the anchored-pattern
  behaviour change.
* Point the add-a-route steps and the source-of-truth list at `DeepLinkRouter.matchers`
  instead of the deleted `route()` `when` block.
* Bump `last_updated`.

`topLevelPathSegments`, `settingsSubRoutes` and `nodeDetailSubRoutes` all survived the
rewrite, so the rest of the page stays as it was.

## Testing Performed

All four docs gates run locally against this branch:

* `check-doc-aliases.js .` — every frontmatter alias registered in `DocBundleLoader.kt`
* `check-doc-coverage.js .` — 11/11 required pages present

Docs-only; no code touched.

## Summary by CodeRabbit

* **Documentation**
  * Updated the Navigation 3 deep-linking guide to describe matcher-based URI routing and route field decoding.
  * Clarified that unsupported or unmodeled paths no longer open screens.
  * Revised instructions for adding deep links through matcher configuration and updated the page’s source-of-truth guidance.


jamesarich:

PR #7252, opened 2026-09-19
docs(settings): lockdown's SWD lock is not reversible

jamesarich:
## Why

#7171 rewrote the lockdown enable dialog because the old wording understated what enabling it does. The user page was never updated, so it still carries the retracted claim — and on the most destructive setting in the app, it now says the opposite of the dialog the user is agreeing to.

Page, before this PR:

> You can turn lockdown off again at any time with the passphrase, and a full device erase restores the hardware regardless.

`lockdown_enable_warning`, live in the app since #7171:

> Enabling lockdown locks the debug port (SWD) on hardware where the lockout takes effect. Turning lockdown off later decrypts your data, but the port stays locked. Reopening it takes a full chip erase with a debug probe, which destroys everything on the device.

A user who reads the docs first is told the hardware change is undoable. It is not.

## 🐛 Bug Fixes

* Replace the reversibility claim with what actually happens: turning lockdown off decrypts storage and reboots, the SWD lock stays, and reopening the port needs a debug probe and destroys the device's contents.
* Bump `last_updated`.

## Testing Performed

Wording checked line by line against `lockdown_enable_warning` and `lockdown_disable_message` in `core/resources`. Docs-only; no code touched.

Found by a full sweep of `docs/en` against current behaviour; more fixes to follow.

## Summary by CodeRabbit

- **Documentation**
  - Updated the Settings documentation with the latest revision date.
  - Clarified Lockdown Mode behavior: the SWD port remains locked after Lockdown Mode is disabled on supported hardware.
  - Documented that reopening the port requires a full chip erase using a debug probe, which destroys device data.


jamesarich:

PR #7254, opened 2026-09-19
docs(discovery): correct beacon targets, signal bands and the gating claims

jamesarich:
## Why

`discovery.md` had drifted further than any other page. Seven passages no longer matched the app, including one that describes a state the beacon editor can no longer produce and a set of signal thresholds that a sibling commit corrected in three neighbouring pages while skipping this one. The page ships inside the app through the Help & Docs bundle, so the drift is not just website content.

Found by a full sweep of `docs/en` against current behaviour. Every claim below was verified against the code before editing.

## 🐛 Bug Fixes

* **Broadcast targets.** The page said targets are optional extras and that with none the beacon goes out on the offered channel alone. Since #7010 the list is floored at one row and the first row *is* the beacon's own transmission — `MeshBeaconConfigPolicy.kt:161` seeds an empty list with a default row, and `:182` re-seeds when you remove the last one.
* **Signal colours.** Replaced the fixed −7 / −15 dB thresholds with the preset-relative bands the app actually uses: `LoraSignalIndicator.kt:178-184` classifies a margin against `ModemPreset.snrLimit`, green above the floor then 5.5 and 7.5 dB steps, and there is a fourth red band the page never named. The floor itself moves with the preset (`ChannelOption.kt:496`), so −7 dB is Good on Short Fast and Bad on Long Fast.
* **Managed-mode claim.** It was wrong in both directions. The Advanced section is local-session only (`RadioConfig.kt:99`), and Debug Panel stays enabled on a managed device by design (`RadioConfig.kt:217-222` and its comment).
* **Transmit over LoRa.** Dropped the claim that it is unavailable on a default-named channel; the only gate is connectivity (`NeighborInfoConfigItemList.kt:87`).
* **Neighbor Info interval.** The 21600 / 14400 figures appear nowhere in the app, and the field has no minimum or rejection message (`NeighborInfoConfigItemList.kt:73-80`). Attributed the floor to firmware.
* **Keep screen awake.** The "a scan with it off can under-count" rationale was superseded by #7159 — the scan holds a partial wake lock for its whole lifetime and posts a foreground notification, so it survives the screen going off.

## Testing Performed

Docs-only; no code touched. Two genuine code defects turned up in the same sweep and are being handled separately.


jamesarich:

PR #7257, opened 2026-09-19
docs(nodes): document the not-heard-on-current-LoRa cluster

jamesarich:
## Why

`nodes.md` documents none of what the `heard_on_current_lora` work added, and one of those surfaces deletes nodes. A user can meet a banner offering **Remove** with nothing in the docs explaining what it removes or whether it can be undone.

Found by a full sweep of `docs/en` against current behaviour; every claim verified against the code before editing.

## 🌟 New Features

* New **Nodes Not Heard on Your Current Settings** section covering the banner (`NodeListScreen.kt:145`, strings `nodes_unheard_banner` / `_keep` / `_remove`), what **Remove** deletes, and that favorites and your own node are exempt (`RemovableUnheardNodes.kt:35`).
* Filter table gains the ninth toggle, **Hide unheard nodes** (`node_filter_exclude_unheard`), including that it is off by default and inert on firmware that does not report the flag.

## 🐛 Bug Fixes

* Status table gains the orange unreachable row and the struck-through ignored row (`NodeItem.kt:464-467` and `:548`), and the "no away tier" note now says the unreachable state takes precedence over the green online tint, and that MQTT-only nodes are never shown as unreachable — both stated in the renderer's own KDoc.
* Hop-window filter labels corrected to the on-screen spellings: **All**, **1 Hour**, **8 Hours**, **24H**.

## Not changed on purpose

The **Only show direct nodes** row was already correct. The filter behind it was not, and #7256 fixes the code to match this page rather than the other way round.

## Testing Performed

Docs-only; no code touched.


jamesarich:

PR #7258, opened 2026-09-19
docs(measurement): NumberFormatter.format follows the locale

jamesarich:
## Why

`measurement.md` states the opposite of the code's own contract, and the mistake it invites is a real bug rather than a typo.

The page says `NumberFormatter` "always uses `.` as the decimal separator" and justifies it with mesh data shared between nodes needing to be consistent. Since #6854 that is backwards: `format` follows the OS locale, delegating to `java.text.NumberFormat` on JVM and Android — the very `DecimalFormat` the page says is not used. The locale-independent path moved to a separate function, `formatInvariant`, which the page never mentions.

So a contributor writing exactly the case the page describes — a value another client parses — reads this page, calls `format`, and ships `1,5` on every comma-separator locale. `NumberFormatter.kt:42-46` spells out that hazard in its own KDoc.

Found by a full sweep of `docs/en` against current behaviour; every claim verified against the code.

## 🐛 Bug Fixes

* Rewrote the `NumberFormatter` section around the two halves: `format` for anything a person reads, `formatInvariant` for anything a machine parses, with the reason each exists. Notes the `—` placeholder and the half-away-from-zero rounding, which matters because firmware reports SNR in quarter-dB steps.
* Corrected two signatures in the `MetricFormatter` block — `snr(Float?)` and `rssi(Int?)` are nullable and render `—` — and added three functions the block omitted: `degreeSymbol`, `percent(Int)` and `weight`.
* `weight` converts kg to lb under `isImperial`, so the conversion count is four, not three, and the conversion table gains a row.
* Added `MeasureFormatting.kt` to the utilities table. Wind speed, rainfall and weight do not go through `NumberFormatter` at all; they pair a fixed English unit symbol with a locale-formatted number, and that file is where the symbols live.
* Replaced the two design-decision rows that asserted locale independence and pure arithmetic, both now true only of `formatInvariant`.

## Testing Performed

Docs-only; no code touched.


jamesarich:

PR #7259, opened 2026-09-19
docs(settings): the App Settings block has more than two entries

jamesarich:
## Why

The **App Settings** section claimed the Settings screen holds "two easy-to-miss entries" that configure the app. It holds about fourteen, in three named groups plus four standalone rows.

The concrete cost: `units-and-locale.md` tells readers to open the **Units** entry, and this page — the one that documents the settings screen — never mentions it exists. The same goes for the node database export, the packet exports and the full-timestamp toggle.

Found by a full sweep of `docs/en` against current behaviour. Every entry below was read from `SettingsScreen.kt:263-320` and its section composables, with labels taken from the string resources rather than guessed.

## 🐛 Bug Fixes

* Replaced the two-entry claim with the real inventory, grouped the way the screen groups it: Privacy (analytics, phone location, homoglyph), Appearance (Units, Theme, Language, full message timestamps), Persistence (cache limit, three exports), then Node Layout, Wi-Fi Provisioning, Message Filter and System AI.
* Noted that **Permissions** sits above the block and **App Info** below, and that System AI appears only where app functions are available.
* Linked **Units** to `units-and-locale`, closing the loop that page already assumed.

## Testing Performed

Docs-only; no code touched.

Note for reviewers: four committed screenshots under `docs/assets/screenshots/` (`settings_appearance`, `settings_persistence`, `settings_notifications`, `settings_app_info`) are referenced by no page at all, which suggests this section was always meant to be fuller. Wiring them in is left for a follow-up.


jamesarich:

PR #7260, opened 2026-09-19
docs(settings): fix the triple-click default and the enum lists

jamesarich:
## Why

Five claims on the radio settings page no longer match the app, one of them a default stated backwards.

Stacked on #7252 (the lockdown correction) because it touches the same file. Retarget to `main` once that merges.

Found by a full sweep of `docs/en` against current behaviour; every claim verified against the code or the pinned protobuf artifact.

## 🐛 Bug Fixes

* **Triple Click Ad Hoc Ping was documented as disabled by default; it ships enabled.** The switch binds the inverted proto field — `DeviceConfigScreen.kt:253` `checked = !formState.value.disable_triple_click` — and `disable_triple_click` has no non-zero default, so the control renders on.
* **OLED type** lists four values; the enum has six. Corrected to the six constants, and noted the picker shows raw names (`DisplayConfigItemList.kt:198` maps `entries` to `it.name`).
* **Compass orientation** lists four; the enum has eight, the extra four being `_INVERTED` variants.
* **Managed Mode** was described as restricting non-admin channel changes. It disables the entire Configuration list — `RadioConfig.kt:86` folds `isManaged` into the `enabled` flag for every row, and the in-app summary says the user cannot access any device settings.
* **Region and presets** gained two behaviours the page never mentioned: the preset list is filtered to what the region legally permits (`LoRaConfigItemList.kt:98`), and changing region repairs a now-illegal preset instead of leaving it (`:194` `presetForRegionChange`). The region list is itself firmware-filtered, so the newer regions appear only on 2.8 and later.

## Testing Performed

Enum values read from the pinned `protobufs-jvm` 2.8.0.96 artifact, not from the proto source. Docs-only; no code touched.


jamesarich:

PR #7264, opened 2026-09-19
feat(coverage): compute RF coverage locally with kp1812 (ITU-R P.1812)

jamesarich:
Computes RF coverage in-process with [`org.meshtastic:kp1812`](https://github.com/meshtastic/kp1812), a pure-Kotlin Multiplatform implementation of ITU-R P.1812, instead of driving the hosted Site Planner through a hidden WebView. `SitePlannerRunner` is 319 lines that load site.meshtastic.org, wait up to 45 s for a JavaScript bridge, and need the network; on desktop it cannot run at all, so today the desktop app opens a browser and asks the user to export a `.geojson` and re-import it by hand. With this, coverage is a local computation over terrain the app already has, offline once the terrain is cached.

**Draft, and depends on an unpublished library.** kp1812 has not had a release yet; the branch resolves it from `mavenLocal` behind the repo's existing `-PuseMavenLocal` flag, so CI will fail at dependency resolution until kp1812's first release lands on Maven Central. The kp1812 repo is private in the org with Actions disabled for now. Predictions will also not match the hosted planner pixel for pixel: P.1812 is a different model from SPLAT!'s ITM, and that is expected rather than a defect.

## 🌟 New Features

- `feature:coverage`, a KMP module: `LocalCoverage.sweep()` computes received signal strength on a polar lattice (one terrain profile per bearing, every ring a prefix of that walk) and resamples onto a grid; `CoverageContours` merges cells into disjoint dBm bands exported as GeoJSON `MultiPolygon` iso-bands with simplestyle fill, the same shape the hosted planner exports.
- Terrain from Mapterhorn's XYZ endpoint over ktor, prefetched for the whole disc, decoded once into a process-wide bounded cache, and written through to the map's existing `TerrainTileStore` so it survives a restart. Zoom fits the sampled area so a wide or high-latitude disc cannot thrash the cache.
- Desktop: `DesktopSitePlannerSlot` runs the local engine and lands the result through `MapLayersManager.addGeoJsonLayer`, the path the F-Droid flavour already uses for the WebView's GeoJSON, so it shows on the map, appears in the layers list as `LayerType.COVERAGE`, and survives the sheet closing. The sheet's palette, min/max dBm and overlay transparency now drive the render; the six palettes are the matplotlib colormaps the hosted planner uses, and at the default six bands plasma and viridis emit their anchors exactly.

## 🛠️ Refactoring & Architecture

- The model, the `ElevationSource` seam and the GeoJSON export live in `commonMain`; only the java.awt demo renderer stays in `jvmMain`. Android takes kp1812's `jvm` artifact, the same choice as kzstd. iOS is compile-verified through `kmpSmokeCompile`, not executed.
- `ElevationSource` is a one-method suspend interface: the app backs it with Mapterhorn tiles, tests back it with a lambda, so the suite needs no network, WebView or terrain download.
- A failed terrain request is a failed estimate, not sea level: only 404 is treated as absence; anything else throws after one retry.

## 🧹 Chores

- `kp1812` added to the version catalog (`0.1.0`, unpublished, see above).

## Testing Performed

- `feature/coverage` tests, behavioural rather than conformance since kp1812 validates itself against the ITU reference: signal decays with distance over flat ground; a 400 m ridge shadows what is behind it; more transmit power reaches at least as far; `reachable` agrees with receiver sensitivity; every bearing contributes equally to the sweep; the geodesy round-trips; the GeoJSON is structurally valid (closed rings, numeric pairs); palette anchors and band spreading; the terrain zoom guard.
- Measured on the desktop demo (Seattle, 25 km radius, 10 cores): terrain cold 1.2 s, second launch 0.24 s from the disk cache, sweep 310 ms, repeat estimate 125 ms.
- Full baseline (`spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile` with `-PuseMavenLocal`) run locally before this push; result noted in the first comment.


jamesarich:
Baseline on this head (`9d7e0542d`), run locally with `-PuseMavenLocal` so `org.meshtastic:kp1812:0.1.0` resolves from `~/.m2`:

```
./gradlew spotlessApply spotlessCheck detekt assembleDebug test allTests kmpSmokeCompile --continue
BUILD SUCCESSFUL in 4m 16s (171 executed, 4 from cache, 1736 up-to-date after an earlier run of the same tree)
```

Spotless made no changes. CI on this PR is expected to fail at dependency resolution until kp1812's first release is on Maven Central; that is the one thing standing between draft and ready.

